PR #11, opened 2015-10-03
update modal logic

athurman:
- removes jquery logic from Modals, as jquery manipulating the DOM caused issues with React.

Now, the modal shows/hides solely based on changing the class names via state.


src/components/Modal/Modal.js
```diff
@@ -12,36 +10,36 @@ class Modal extends Component {
...
+        'ui',
+        {small: this.props.small},
+        'modal',
+        'transition',
+        'visible',
+        'active',
+        );
+    } else {
+      classes = classNames('sd-modal', this.props.className, 'ui', {small: this.props.small}, 'modal');
+    }
```
levithomason: Don't add Semantic classes on Stardust components, extend the Stardust component to expose props instead.

Keep DRY, let's not repeat `classNames()` and the list of classes.  Use the if conditional inside a single call to classNames to toggle classes on/off.

test/src/components/Modal-test.js
```diff
@@ -30,12 +25,12 @@ describe('Modal', () => {
   it('should show modal', () => {
     var renderedModal = render(<Modal />).first();
     renderedModal.showModal();
-    renderedModal.elm.modal.called.should.equal(true);
+    renderedModal.state.isShown.should.equal(true);
   });
```
levithomason, line 26: This test is for `showModal` testing that `it('sets isShown true)`.  The description should be updated as it says that it is testing that the the modal is shown.

test/src/components/Modal-test.js
```diff
@@ -30,12 +25,12 @@ describe('Modal', () => {
...
   it('should hide modal', () => {
     var renderedModal = render(<Modal />).first();
     renderedModal.hideModal();
-    renderedModal.elm.modal.called.should.equal(true);
+    renderedModal.state.isShown.should.equal(false);
```
levithomason, line 30: Same here, this is a `hideModal()` test.  When testing methods, you can give them their own describe block and set of it() blocks for what they do.

src/components/ConfirmationModal/ConfirmationModal.js
```diff
@@ -53,7 +53,7 @@ class ConfirmationModal extends Component {
 
   render() {
     return (
-      <Modal actionRequired ref='modal'>
+      <Modal ref='modal' small>
```
levithomason: Pending sizes enum PR.

PR #14, opened 2015-10-03
Docs link

levithomason:
Doc site should auto generate and publish to `gh-pages` as part of the wercker deploy step.  We'll need to merge a PR before that happens.  This PR just adds the link, but the docs won't actually be there, yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@
 npm i stardust -S
 ```
 
+## Learn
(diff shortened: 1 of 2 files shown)


athurman:
:o: :octopus:

eanplatter:
:fallen_leaf:

athurman:
:factory:

PR #17, opened 2015-10-04
Watch

levithomason:
Makes it so that components are rescanned and doc info is created on change.  Also, reloads when html changes.

Webpack still handles reloading when JS changes.  This just makes docgen regenerate doc info when component syntax changes.


eanplatter:
:fallen_leaf:

PR #19, opened 2015-10-04
Update/button

eanplatter:
The button wasn't able to take in typically used classes like fluid or change colors. 

Now you can do it like so:

``` jsx
  <Button 
    color='teal'
    fluid
    loading  
  />
```

### DIFF
--- a/docs/app/Component/ComponentList.js
+++ b/docs/app/Component/ComponentList.js
@@ -46,7 +46,9 @@ class ComponentList extends Component {
           <h3>Props:</h3>
           <ComponentProps props={val.props} />
 
-          <Button onClick={this.toggleShowDocgenJSON}>JSON</Button>
+          <Button onClick={this.toggleShowDocgenJSON} large teal>
+            JSON
+          </Button>
 
           {this.state.showDocgenJSON && docgenJSON}
         </div>
--- a/docs/app/Component/ComponentList.js
+++ b/docs/app/Component/ComponentList.js
@@ -46,7 +46,7 @@ class ComponentList extends Component {
           <h3>Props:</h3>
           <ComponentProps props={val.props} />
 
-          <Button onClick={this.toggleShowDocgenJSON} fluid size={'large'} color={'teal'}>
+          <Button onClick={this.toggleShowDocgenJSON} size={'large'} color={'red'}>
             JSON
           </Button>
 
--- a/docs/app/Component/ComponentList.js
+++ b/docs/app/Component/ComponentList.js
@@ -46,7 +46,7 @@ class ComponentList extends Component {
           <h3>Props:</h3>
           <ComponentProps props={val.props} />
 
-          <Button onClick={this.toggleShowDocgenJSON} large teal>
+          <Button onClick={this.toggleShowDocgenJSON} fluid size={'large'} color={'teal'}>
             JSON
           </Button>
 
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -23,7 +23,7 @@ class Button extends Component {
   render() {
     let classes = classNames('sd-button', 'ui', 'button', this.props.color, this.props);
     return (
-      <button className={classes}>
+      <button className={classes} {...this.props}>
         {this.props.children}
       </button>
     );
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -7,11 +7,13 @@ class Button extends Component {
     basic: PropTypes.bool,
     children: PropTypes.string,
     color: PropTypes.string,
+    disabled: PropTypes.bool,
     emphasis: PropTypes.string,
     fluid: PropTypes.bool,
     inverted: PropTypes.bool,
     loading: PropTypes.bool,
     onClick: PropTypes.func,
+    size: PropTypes.string,
     type: React.PropTypes.oneOf(['button', 'submit', 'reset']),
   };
 
@@ -21,7 +23,21 @@ class Button extends Component {
   };
 
   render() {
-    let classes = classNames('sd-button', 'ui', 'button', this.props.color, this.props);
+    let classes = classNames(
+      'sd-button',
+      'ui',
+      'button',
+      this.props.color,
+      this.props.size,
+      {animated: this.props.animated},
+      {basic: this.props.basic},
+      {children: this.props.children},
+      {disabled: this.props.disabled},
+      {emphasis: this.props.emphasis},
+      {fluid: this.props.fluid},
+      {inverted: this.props.inverted},
+      {loading: this.props.loading}
+    );
     return (
       <button className={classes} {...this.props}>
         {this.props.children}
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,10 +1,18 @@
 import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
 
 class Button extends Component {
   static propTypes = {
+    animated: PropTypes.bool,
+    basic: PropTypes.bool,
     children: PropTypes.string,
+    color: PropTypes.string,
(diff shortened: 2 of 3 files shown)


src/components/Button/Button.js
```diff
@@ -1,10 +1,18 @@
...
 class Button extends Component {
   static propTypes = {
+    animated: PropTypes.bool,
+    basic: PropTypes.bool,
     children: PropTypes.string,
+    color: PropTypes.string,
+    emphasis: PropTypes.string,
+    fluid: PropTypes.bool,
+    inverted: PropTypes.bool,
+    loading: PropTypes.bool,
```
eanplatter, line 14: There are more variations of buttons than this, but I haven't been using them. I thought we could just add them as necessary.

docs/app/Component/ComponentList.js
```diff
@@ -46,7 +46,9 @@ class ComponentList extends Component {
           <h3>Props:</h3>
           <ComponentProps props={val.props} />
 
-          <Button onClick={this.toggleShowDocgenJSON}>JSON</Button>
+          <Button onClick={this.toggleShowDocgenJSON} large teal>
+            JSON
+          </Button>
```
eanplatter, line 51: Updated the doc button to use updated changes.

test/src/components/Button-test.js
```diff
@@ -26,4 +26,9 @@ describe('Button', () => {
     React.addons.TestUtils.Simulate.click(node);
     customHandleClick.called.should.be.true;
   });
+  it('should take in a color attribute', () => {
+    let coloredButton = render(<Button color={'teal'}>Submit</Button>).findTag('button');
+    coloredButton.props.className.should.contain('teal');
+    coloredButton.getDOMNode().getAttribute('class').should.contain('teal');
+  });
```
eanplatter, line 33: Just wanted to test that the colors attribute was getting added in correctly.

src/components/Button/Button.js
```diff
@@ -13,8 +21,9 @@ class Button extends Component {
   };
 
   render() {
+    let classes = classNames('sd-button', 'ui', 'button', this.props.color, this.props);
```
eanplatter: Using the `classnames` library rather than statically setting them.

src/components/Button/Button.js
```diff
@@ -1,10 +1,18 @@
...
+    basic: PropTypes.bool,
     children: PropTypes.string,
+    color: PropTypes.string,
+    emphasis: PropTypes.string,
+    fluid: PropTypes.bool,
+    inverted: PropTypes.bool,
+    loading: PropTypes.bool,
     onClick: PropTypes.func,
-    type: React.PropTypes.oneOf(['button', 'submit']),
+    type: React.PropTypes.oneOf(['button', 'submit', 'reset']),
```
eanplatter, line 17: I don't think anyone uses reset button types very often, but there are only three types of buttons, so why not.

athurman:
:ocean: :octopus:

docs/app/Component/ComponentList.js
levithomason, line 51: color='teal'
size='large'

src/components/Button/Button.js
levithomason: Let's not pass in all props as classes.  We should map the support we want from the outside.

test/src/components/Button-test.js
levithomason, line 33: perfect!

src/components/Button/Button.js
```diff
@@ -1,10 +1,20 @@
...
+    color: PropTypes.string,
+    disabled: PropTypes.bool,
+    emphasis: PropTypes.string,
+    fluid: PropTypes.bool,
+    inverted: PropTypes.bool,
+    loading: PropTypes.bool,
     onClick: PropTypes.func,
-    type: React.PropTypes.oneOf(['button', 'submit']),
+    size: PropTypes.string,
+    type: React.PropTypes.oneOf(['button', 'submit', 'reset']),
```
eanplatter, line 17: Add some of the more common usage of buttons

test/src/components/Button-test.js
```diff
@@ -26,4 +26,19 @@ describe('Button', () => {
...
+  it('should take in boolean attributes', () => {
+    let disabledButton = render(<Button disabled>Submit</Button>).findTag('button');
+    disabledButton.props.className.should.contain('disabled');
+    disabledButton.getDOMNode().getAttribute('class').should.contain('disabled');
+  });
```
eanplatter, line 38: This test checks if the boolean values are getting added to the classes properly

test/src/components/Button-test.js
```diff
@@ -26,4 +26,19 @@ describe('Button', () => {
...
+  it('should ignore unknown attributes', () => {
+    let unknownAttrButton = render(<Button fakeValue>Submit</Button>).findTag('button');
+    unknownAttrButton.props.className.should.not.contain('fakeValue');
+    unknownAttrButton.getDOMNode().getAttribute('class').should.not.contain('fakeValue');
+  });
```
eanplatter, line 43: This test is checking that unknownvalues aren't being added to the button's classes

athurman:
this PR looks good, but before we merge, do we want to have size and color utils fully built out? just curious.

src/components/Button/Button.js
```diff
@@ -13,8 +23,23 @@ class Button extends Component {
...
+      {children: this.props.children},
+      {disabled: this.props.disabled},
+      {emphasis: this.props.emphasis},
+      {fluid: this.props.fluid},
+      {inverted: this.props.inverted},
+      {loading: this.props.loading}
+    );
     return (
-      <button className='sd-button ui blue button' {...this.props}>
+      <button className={classes} {...this.props}>
```
levithomason, line 42: Don't spread props since there are many props that are not applicable here, like color and size.

src/components/Button/Button.js
```diff
@@ -1,10 +1,20 @@
...
     children: PropTypes.string,
+    color: PropTypes.string,
+    disabled: PropTypes.bool,
+    emphasis: PropTypes.string,
+    fluid: PropTypes.bool,
+    inverted: PropTypes.bool,
+    loading: PropTypes.bool,
     onClick: PropTypes.func,
-    type: React.PropTypes.oneOf(['button', 'submit']),
+    size: PropTypes.string,
```
levithomason, line 16: Add todo to use the coming enum validation on size.

src/components/Button/Button.js
```diff
@@ -1,10 +1,20 @@
 import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
 
 class Button extends Component {
   static propTypes = {
+    animated: PropTypes.bool,
+    basic: PropTypes.bool,
     children: PropTypes.string,
+    color: PropTypes.string,
```
levithomason, line 9: Add todo to use the coming enum validation on color.

test/src/components/Button-test.js
levithomason, line 33: Instead of `findTag()`, use `findClass('teal')` this will find the element and do the assertion.  It also removes the dependency of this test on the structure of the button.  Right now, the test will fail if the className is not found on the first on the root element of the button (props.className).  With `findClass()` it will work for any structure and only fail if the class name is missing.

levithomason, line 33: No need to test `class` attribute of the DOM node if we test `className` contains `teal`.  It is enough to test our own code and prove `className` is set.  It is up to React's internals and tests to turn the `className` into DOM `class` attr on the DOM node.

test/src/components/Button-test.js
levithomason, line 38: Same comments

test/src/components/Button-test.js
levithomason, line 43: Same comments.

PR #20, opened 2015-10-05
Update stardust components

athurman:
- general updates to stardust components, based on the certain use case for them within builgin out UIs

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import React, {Children, Component, PropTypes} from 'react';
+import classNames from 'classnames';
 import TableColumn from './TableColumn';
 import TableHeader from './TableHeader';
 import TableRow from './TableRow';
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.js
```diff
@@ -59,8 +62,19 @@ class Table extends Component {
   }
 
   render() {
+    let classes = classNames(
+      'sd-table',
+      'ui',
+      {'very basic': this.props.basic},
```
levithomason, line 68: Since the `very` class is used for `very padded` `very basic` and `very compact` I think we should expose it as a separate prop.  This way it is fully composable with the other uses.

src/components/Table/Table.js
```diff
@@ -59,8 +62,19 @@ class Table extends Component {
...
+      'selectable',
+      'fixed',
+      'compact',
+      'small',
```
levithomason, line 72: We need to make all options configurable from the outside.

levithomason, line 72: Note, the small class should be part of the `size` prop.  This is in our guide line doc, almost finished.  I'll push it up soon.

PR #21, opened 2015-10-05
Component guidelines

levithomason:
# Deferred

> see comment below

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,9 @@ npm i stardust -S
 ```
 
 ## Learn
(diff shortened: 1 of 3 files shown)


athurman:
:+1: :100: :1234:

eanplatter:
Looking good! :fallen_leaf:

levithomason:
# Deferred

Implement tests from #33 instead of a readme

levithomason:
Both the guideline doc and the tests were implemented in #33.

PR #22, opened 2015-10-05
Add stardust size util

athurman:
- Adds utility for defining values of size property, so to render matching semantic ui class names associated with that size.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,9 +1,13 @@
 import React, {Component, PropTypes} from 'react';
+import _ from 'lodash';
+import SIZES from '../../utils/SIZES';
+import classNames from 'classnames';
 
 class Button extends Component {
   static propTypes = {
     children: PropTypes.string,
     onClick: PropTypes.func,
+    size: PropTypes.oneOf(_.values(SIZES)),
     type: React.PropTypes.oneOf(['button', 'submit']),
   };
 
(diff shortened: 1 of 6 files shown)


src/components/Button/Button.js
```diff
@@ -1,9 +1,13 @@
...
+import _ from 'lodash';
+import SIZES from '../../utils/SIZES';
+import classNames from 'classnames';
 
 class Button extends Component {
   static propTypes = {
     children: PropTypes.string,
     onClick: PropTypes.func,
+    size: PropTypes.oneOf(_.values(SIZES)),
     type: React.PropTypes.oneOf(['button', 'submit']),
```
athurman, line 11: creates an array of values from the Size utility

src/components/Button/Button.js
```diff
@@ -13,8 +17,9 @@ class Button extends Component {
   };
 
   render() {
+    let classes = classNames('sd-button', 'ui', this.props.size, 'blue', 'button');
     return (
```
athurman, line 21: place chosen size within the class names.

src/components/Table/Table.js
```diff
@@ -18,7 +21,13 @@ class Table extends Component {
...
       }
     },
+    compact: PropTypes.bool,
     data: PropTypes.array.isRequired,
+    fixed: PropTypes.bool,
+    selectable: PropTypes.bool,
+    size: PropTypes.oneOf(_.values(SIZES)),
+    striped: PropTypes.bool,
+    very: PropTypes.bool,
   };
```
athurman, line 32: break out all table styles into bool props, or strings, based on the class name.

src/utils/SIZES.js
```diff
@@ -0,0 +1,12 @@
...
+  mini: 'mini',
+  tiny: 'tiny',
+  small: 'small',
+  medium: 'medium',
+  large: 'large',
+  big: 'big',
+  huge: 'huge',
+  massive: 'massive',
+};
+
```
athurman, line 11: all possible semantic ui sizes available.

PR #31, opened 2015-10-07
add tests for index.js

levithomason:
`index.js` is now tested to ensure it contains exactly all the components from `src/components`.

### DIFF
--- a/src/components/Confirm/Confirm.js
+++ b/src/components/Confirm/Confirm.js
@@ -1,7 +1,10 @@
 import React, {Component, PropTypes} from 'react';
 import Promise from 'bluebird';
 
-import {Modal, ModalContent, ModalFooter, ModalHeader} from 'stardust';
+import Modal from 'src/components/Modal/Modal';
+import ModalContent from 'src/components/Modal/ModalContent';
+import ModalFooter from 'src/components/Modal/ModalFooter';
+import ModalHeader from 'src/components/Modal/ModalHeader';
 
 class Confirm extends Component {
   static propTypes = {
--- a/test/src/index-test.js
+++ b/test/src/index-test.js
@@ -1,0 +1,21 @@
+import _ from 'lodash';
+import path from 'path';
+import stardust from 'index';
+
+let componentPaths = require.context('src/components/', true, /[^Style\.js$]\.js$/i);
+let componentNames = _.map(componentPaths.keys(), key => {
+  return path.basename(key).replace('.js', '');
+});
+
+describe('index.js', () => {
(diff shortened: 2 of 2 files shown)


src/components/Confirm/Confirm.js
```diff
@@ -1,7 +1,10 @@
 import React, {Component, PropTypes} from 'react';
 import Promise from 'bluebird';
 
-import {Modal, ModalContent, ModalFooter, ModalHeader} from 'stardust';
+import Modal from 'src/components/Modal/Modal';
+import ModalContent from 'src/components/Modal/ModalContent';
+import ModalFooter from 'src/components/Modal/ModalFooter';
+import ModalHeader from 'src/components/Modal/ModalHeader';
```
levithomason, line 7: Importing from `stardust` was causing a failure probably due to some circular import issue.  When in the `src` directory, let's only import exact paths.

The issue was, `stardust.Confirm` was `undefined` when it should have been a `function`.

test/src/index-test.js
```diff
@@ -0,0 +1,21 @@
...
+  return path.basename(key).replace('.js', '');
+});
+
+describe('index.js', () => {
+  it('has a property for every component', () => {
+    _.each(componentNames, component => {
+      expect(component in stardust);
+      expect(stardust[component]).to.be.a('function');
+    });
+  });
```
levithomason, line 16: Here we test that all components in the components dir are keys on `index.js` and that their values are functions.

test/src/index-test.js
```diff
@@ -0,0 +1,21 @@
...
+
+  it('only has properties which correspond to components', () => {
+    _.each(stardust, component => expect(component in componentNames));
+  });
```
levithomason, line 20: We also test that there are no extra keys by testing that each stardust key has a component in the components dir.

test/src/index-test.js
athurman, line 16: awesome tests. :100:

athurman:
:ocean: :octopus:

src/components/Confirm/Confirm.js
eanplatter, line 7: That's disappointing! Owel.

eanplatter:
:fallen_leaf:

src/components/Confirm/Confirm.js
levithomason, line 7: @eanplatter keep in mind, this is only in the `/src` dir, when building stardust components that are comprised of other stardust components.  You can still import `stardust` in `/docs` and `tests`.

PR #35, opened 2015-10-07
Classes & Props: Input

levithomason:


src/components/Input/Input.js
```diff
@@ -1,64 +1,34 @@
 import React, {Component, PropTypes} from 'react';
-import Field from '../Field/Field';
-import $ from 'jquery';
 import classNames from 'classnames';
 
-class Input extends Component {
+export default class Input extends Component {
```
levithomason, line 5: export inline

src/components/Input/Input.js
```diff
@@ -1,64 +1,34 @@
...
   };
 
-  componentWillMount() {
-    this.elm = $(React.findDOMNode(this.refs[this.props.ref]));
-    if (this.props.dataContent) {
-      this.elm.popup({
-        on: 'focus',
-      });
-    }
-  }
```
levithomason, line 33: There is no more built in popup functionality, this should be accomplished with a popup from the outside.

src/components/Input/Input.js
```diff
@@ -1,64 +1,34 @@
...
-
-  handleChange = (e) => {
-    this.props.onChange(e);
-  };
```
levithomason, line 37: Removed duplicate click handler as it only passes the event to the prop.  This will happen without this extra handler.

src/components/Input/Input.js
```diff
@@ -1,64 +1,34 @@
...
       'sd-input',
       'ui',
-      {small: this.props.small},
-      {transparent: this.props.transparent},
-      {icon: !!this.props.iconClass},
+      this.props.className,
       'input'
     );
+    let inputProps = _.clone(this.props);
+    delete inputProps.className;
```
levithomason: We only want to spread non class props on the input, the classes are spread on the containing div.

src/components/Input/Input.js
```diff
@@ -1,64 +1,35 @@
...
       'sd-input',
       'ui',
-      {small: this.props.small},
-      {transparent: this.props.transparent},
+      this.props.className,
       {icon: !!this.props.iconClass},
       'input'
     );
+    let inputProps = _.clone(this.props);
+    delete inputProps.className;
```
levithomason, line 28: We only want to spread non class props on the input, the classes are spread on the containing div.

src/components/Input/Input.js
```diff
@@ -1,64 +1,35 @@
...
-      <Field label={this.props.label}>
-        <div className={classes}>
-          <input
-            {...this.props}
-            onChange={this.handleChange}
-            data-content={this.props.dataContent}
-          />
-          {this.props.iconClass && icon}
-        </div>
-      </Field>
```
levithomason, line 59: Field has been removed per our conclusion to remove magic and be more explicit.

test/src/components/Input-test.js
```diff
@@ -22,8 +22,7 @@ describe('Input', () => {
     render(<Input iconClass='linkedin' />).findTag('i').should.be.ok;
   });
 
-  it('shows a popup on focus if data-content is given', () => {
-    let renderedInput = render(<Input dataContent='Sample Popup'/>).first();
-    renderedInput.elm.popup.called.should.equal(true);
+  it('has an icon class if iconClass given', () => {
+    render(<Input iconClass='foo' />).findClass('ui icon input');
```
levithomason, line 26: Added test for the tiny magic of adding the icon class if iconClass prop is given.  I removed the functionality by mistake but didn't realize it until doing annotations.  Now, a test will fail if it is missing/broken.

athurman:
Looks good. :octopus:

src/components/Input/Input.js
```diff
@@ -1,64 +1,36 @@
...
+      <div className={classes}>
+        <input {...inputProps} />
```
eanplatter, line 31: I like this a lot more than trying to be magical.

eanplatter:
:fallen_leaf:

src/components/Input/Input.js
levithomason, line 31: +1

PR #39, opened 2015-10-08
Component APIs and Doc Structure

levithomason:


docs/app/Component/ComponentDescription.js
```diff
@@ -1,20 +1,67 @@
 import React, {Component, PropTypes} from 'react';
+import SEMANTIC_TYPES from 'docs/app/utils/SemanticTypes';
+import STARDUST_TYPES from 'docs/app/utils/StardustTypes';
```
levithomason, line 3: Defined types for components (elements, collections, etc).  Non semantic ui components are Stardust Types.  So far, they are all bundled under one category `addons`.  I imagine we'll iterate on this.

PR #44, opened 2015-10-09
Table update

levithomason:
Updates the Table component to inherit classes and spread props.  Also, uses it in the PropType list on the doc site.  This means the table is also responsive whereas it was not previously.

### DIFF
--- a/docs/app/Component/ComponentDescription.js
+++ b/docs/app/Component/ComponentDescription.js
@@ -59,9 +59,7 @@ export default class ComponentDescription extends Component {
             </a>
           </code>
         </div>
-        <p className='sub header'>
-          {this.isSemanticComponent && semanticDocsLink}
-        </p>
+        {this.isSemanticComponent && semanticDocsLink}
       </Segment>
     );
   }
--- a/docs/app/Component/ComponentDescription.js
+++ b/docs/app/Component/ComponentDescription.js
@@ -1,8 +1,9 @@
 import React, {Component, PropTypes} from 'react';
 import SEMANTIC_TYPES from 'docs/app/utils/SemanticTypes';
 import STARDUST_TYPES from 'docs/app/utils/StardustTypes';
+import {Segment} from 'stardust';
 
-class ComponentDescription extends Component {
+export default class ComponentDescription extends Component {
   static propTypes = {
     /**
      * String describing the Stardust component.
@@ -48,7 +49,7 @@ class ComponentDescription extends Component {
       </a>
     );
     return (
-      <div>
+      <Segment className='basic vertical'>
         <div className='ui large header'>
           {_.capitalize(this.props.name)}
           <code className='sub header' style={{float: 'right'}}>
@@ -58,10 +59,10 @@ class ComponentDescription extends Component {
             </a>
           </code>
         </div>
-        {this.isSemanticComponent && semanticDocsLink}
-      </div>
+        <p className='sub header'>
+          {this.isSemanticComponent && semanticDocsLink}
+        </p>
+      </Segment>
     );
   }
 }
-
-export default ComponentDescription;
--- a/docs/app/Component/ComponentList.js
+++ b/docs/app/Component/ComponentList.js
@@ -17,7 +17,7 @@ export default class ComponentList extends Component {
       .map(name => {
         let doc = new ComponentDoc(name);
         return (
-          <Segment key={doc.name} id={doc.name} className='vertical'>
+          <Segment key={doc.name} id={doc.name}>
             <ComponentDescription
               path={doc.path}
               name={doc.name}
--- a/docs/app/Component/ComponentProps.js
+++ b/docs/app/Component/ComponentProps.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, {Component, PropTypes} from 'react';
-import {Segment} from 'stardust';
+import {Segment, Table, TableColumn} from 'stardust';
 
 /**
  * Displays a table of a Component's PropTypes.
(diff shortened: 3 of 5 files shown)


docs/app/Component/ComponentDescription.js
```diff
@@ -1,8 +1,9 @@
 import React, {Component, PropTypes} from 'react';
 import SEMANTIC_TYPES from 'docs/app/utils/SemanticTypes';
 import STARDUST_TYPES from 'docs/app/utils/StardustTypes';
+import {Segment} from 'stardust';
 
-class ComponentDescription extends Component {
+export default class ComponentDescription extends Component {
```
levithomason, line 6: Updating old export syntax to our new standard.

docs/app/Component/ComponentDescription.js
```diff
@@ -58,10 +59,10 @@ class ComponentDescription extends Component {
             </a>
           </code>
         </div>
-        {this.isSemanticComponent && semanticDocsLink}
-      </div>
+        <p className='sub header'>
+          {this.isSemanticComponent && semanticDocsLink}
+        </p>
+      </Segment>
```
levithomason, line 63: The `ComponentDescription` is the top section of every component doc section.  It is now a basic segment so that it has a bottom margin.  This is needed to keep whitespace between the header and prop types table.

docs/app/Component/ComponentList.js
```diff
@@ -17,7 +17,7 @@ export default class ComponentList extends Component {
       .map(name => {
         let doc = new ComponentDoc(name);
         return (
-          <Segment key={doc.name} id={doc.name} className='vertical'>
+          <Segment key={doc.name} id={doc.name}>
```
levithomason, line 20: Removed the vertical class so each component segment is the default style.  Gives depth separation between component doc sections:

image

docs/app/Component/ComponentProps.js
```diff
@@ -14,13 +14,27 @@ export default class ComponentProps extends Component {
     props: PropTypes.object
   };
 
+  nameRenderer(item) {
+    let required = item.required && <span className='ui mini red circular label'>required</span>;
+    return <div>{item.name} {required}</div>;
+  }
```
levithomason, line 20: The name column also includes a `required` label, if the `prop` is required.

docs/app/Component/ComponentProps.js
```diff
@@ -14,13 +14,27 @@ export default class ComponentProps extends Component {
...
+  defaultValueRenderer(item) {
+    let defaultValue = _.get(item, 'defaultValue.value');
+    let defaultIsComputed = <span className='ui mini gray circular label'>computed</span>;
+
+    return (
+      <div>
+        {defaultValue} {_.get(item, 'defaultValue.computed') && defaultIsComputed}
+      </div>
+    );
+  }
```
levithomason, line 31: The Default Value column includes a `computed` label if the prop value is computed from a function.

docs/app/Component/ComponentProps.js
```diff
@@ -29,42 +43,28 @@ export default class ComponentProps extends Component {
...
-        </div>
-      );
+      return {
+        name,
+        type,
+        value,
+        required,
+        defaultValue,
+        description,
+      };
```
levithomason, line 56: We now return an array of objects describing each property.  The table will use this array of PropType objects creating a row for every object.

docs/app/Component/ComponentProps.js
```diff
@@ -29,42 +43,28 @@ export default class ComponentProps extends Component {
...
-          <b style={{flex: '4'}}>default</b>
-          <b style={{flex: '8'}}>description</b>
-        </div>
-        {content}
+        <Table data={content} className='very basic'>
+          <TableColumn dataKey='name' cellRenderer={this.nameRenderer} />
+          <TableColumn dataKey='type' />
+          <TableColumn dataKey='defaultValue' cellRenderer={this.defaultValueRenderer} />
+          <TableColumn dataKey='description' />
+        </Table>
```
levithomason, line 67: Here is the new PropTypes table, each column uses some key from each object in the array of PropTypes.  Columns with renderer functions (defined above) are combining multiple properties (like `name` and `required`) into a single column.

src/collections/Table/Table.js
```diff
@@ -65,16 +65,11 @@ class Table extends Component {
     let classes = classNames(
       'sd-table',
       'ui',
-      {'very basic': this.props.basic},
-      'selectable',
-      'fixed',
-      'compact',
-      'small',
-      {striped: this.props.striped},
+      this.props.className,
```
levithomason, line 68: Table now inherits classes.

src/collections/Table/Table.js
```diff
@@ -65,16 +65,11 @@ class Table extends Component {
...
       'table',
     );
     return (
-      <table className={classes}>
+      <table {...this.props} className={classes}>
```
levithomason, line 72: Table also now spreads props.

docs/app/Component/ComponentDescription.js
```diff
@@ -59,9 +60,7 @@ class ComponentDescription extends Component {
           </code>
         </div>
         {this.isSemanticComponent && semanticDocsLink}
-      </div>
+      </Segment>
```
levithomason, line 63: The `ComponentDescription` is the top section of every component doc section.  It is now a basic segment so that it has a bottom whitespace.  This is needed to keep whitespace between the header and prop types table segment.

docs/app/Component/ComponentProps.js
eanplatter, line 56: +1

athurman:
:octopus:

eanplatter:
WOO :fallen_leaf:

PR #45, opened 2015-10-09
Feature/button examples

kyleturco:
Added component examples for the button types in - http://semantic-ui.com/elements/button.html. Doesn't include button groups or conditionals.

### DIFF
--- a/docs/app/Examples/Button/Animated.js
+++ b/docs/app/Examples/Button/Animated.js
@@ -1,0 +1,15 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='animated'>
+        <div className='visible content'>Next</div>
+        <div className='hidden content'>
+          <i className='right arrow icon'></i>
+        </div>
+      </Button>
+    );
+  }
+}
--- a/docs/app/Examples/Button/Animated.js
+++ b/docs/app/Examples/Button/Animated.js
@@ -7,7 +7,7 @@ export default class extends Component {
       <Button className='animated'>
         <div className='visible content'>Next</div>
         <div className='hidden content'>
-          <i className='right arrow icon'></i>
+          <i className='right arrow icon' />
         </div>
       </Button>
     );
--- a/docs/app/Examples/Button/Basic.js
+++ b/docs/app/Examples/Button/Basic.js
@@ -5,20 +5,20 @@ export default class extends Component {
   render() {
     return (
       <div>
-        <Button className = 'basic'>Standard</Button>
-        <Button className = 'red basic'>Red</Button>
-        <Button className = 'orange basic'>Orange</Button>
-        <Button className = 'yellow basic'>Yellow</Button>
-        <Button className = 'olive basic'>Olive</Button>
-        <Button className = 'green basic'>Green</Button>
-        <Button className = 'teal basic'>Teal</Button>
-        <Button className = 'blue basic'>Blue</Button>
-        <Button className = 'violet basic'>Violet</Button>
-        <Button className = 'purple basic'>Purple</Button>
-        <Button className = 'pink basic'>Pink</Button>
-        <Button className = 'brown basic'>Brown</Button>
-        <Button className = 'grey basic'>Grey</Button>
-        <Button className = 'black basic'>Black</Button>
+        <Button className='basic'>Standard</Button>
+        <Button className='red basic'>Red</Button>
+        <Button className='orange basic'>Orange</Button>
+        <Button className='yellow basic'>Yellow</Button>
+        <Button className='olive basic'>Olive</Button>
+        <Button className='green basic'>Green</Button>
+        <Button className='teal basic'>Teal</Button>
+        <Button className='blue basic'>Blue</Button>
+        <Button className='violet basic'>Violet</Button>
+        <Button className='purple basic'>Purple</Button>
+        <Button className='pink basic'>Pink</Button>
+        <Button className='brown basic'>Brown</Button>
+        <Button className='grey basic'>Grey</Button>
+        <Button className='black basic'>Black</Button>
       </div>
     );
   }
--- a/docs/app/Examples/Button/Basic.js
+++ b/docs/app/Examples/Button/Basic.js
@@ -1,0 +1,25 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <Button className = 'basic'>Standard</Button>
+        <Button className = 'red basic'>Red</Button>
+        <Button className = 'orange basic'>Orange</Button>
+        <Button className = 'yellow basic'>Yellow</Button>
+        <Button className = 'olive basic'>Olive</Button>
+        <Button className = 'green basic'>Green</Button>
+        <Button className = 'teal basic'>Teal</Button>
+        <Button className = 'blue basic'>Blue</Button>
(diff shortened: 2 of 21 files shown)


docs/app/Examples/Button/Basic.js
```diff
@@ -0,0 +1,25 @@
...
+        <Button className = 'olive basic'>Olive</Button>
+        <Button className = 'green basic'>Green</Button>
+        <Button className = 'teal basic'>Teal</Button>
+        <Button className = 'blue basic'>Blue</Button>
+        <Button className = 'violet basic'>Violet</Button>
+        <Button className = 'purple basic'>Purple</Button>
+        <Button className = 'pink basic'>Pink</Button>
+        <Button className = 'brown basic'>Brown</Button>
+        <Button className = 'grey basic'>Grey</Button>
+        <Button className = 'black basic'>Black</Button>
```
levithomason: Hm, I would think the linter should have caught this.  There are extra spaces around the `=` operator here.

docs/app/Examples/Button/Disabled.js
```diff
@@ -0,0 +1,10 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='disabled' />
```
levithomason: Might be useful to have the content of the Button say "Disabled" in this case.

docs/app/Examples/Button/Emphasis.js
```diff
@@ -0,0 +1,13 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <Button className='primary' />
+        <Button className='secondary' />
```
levithomason: Here also might be useful to have some Button content.  Either a contrived example as in the [Semantic Docs](http://semantic-ui.com/elements/button.html#emphasis) or simply "Primary" and "Secondary".

docs/app/Examples/Button/Inverted.js
```diff
@@ -0,0 +1,43 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <div className='ui inverted segment'>
```
levithomason: Let's use Stardust components where ever possible, there is a component that will do this.

``` jsx
<Segment className='inverted'>
```

docs/app/Examples/Button/Labeled.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='labeled'>
+        <div className='ui button'>
+          <i className='heart icon'></i> Like
+        </div>
```
levithomason: Sticking with Stardust components, let use a nested Button here instead of a `div`.

We'll need to make sure the nested button is a `div` for this to work.  This looks like more evidence that we should default the `Button` to render as a `div` and only render as a button in select cases (ie `type='submit'`, etc).

Referencing #46 here as it deals with the type of element rendered by the Button.  Let's add another `// TODO: ...` referencing issue 46 in the code as well.

docs/app/Examples/Button/VerticallyAttached.js
```diff
@@ -0,0 +1,16 @@
...
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <Button className='top attached'>Top</Button>
+        <div className='ui attached segment'>
+          <p>Content goes here</p>
+        </div>
```
levithomason: Let's use a Stardust `<Segment>` here instead of a `div`.

docs/app/Examples/Button/Animated.js
```diff
@@ -0,0 +1,15 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='animated'>
+        <div className='visible content'>Next</div>
+        <div className='hidden content'>
+          <i className='right arrow icon'></i>
```
levithomason: Also strange the linter is not catching this, I'll look into it.  Elements with no body should be self closing:

``` jsx
<i className='right arrow icon' />
```

Instead of:

``` jsx
<i className='right arrow icon'></i>
```

There are a few instances of this in the PR, I'll only comment on this one occurrence.

levithomason: You can see the style guide here: https://github.com/TechnologyAdvice/javascript/tree/master/react#tags

levithomason:
Looking fantastic, first round of comments done for me.

docs/app/Examples/Button/Animated.js
eanplatter: I've noticed the linter half-assing on this as well. I only catch it because of my IDE.

docs/app/Examples/Button/Basic.js
eanplatter: Do we want to map over these? Or are they examples in the sense that we want to show the markup?

docs/app/Examples/Button/Icon.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='icon'>
+        <i className='world icon'></i>
```
eanplatter: This should also be self closing: `<i className='world icon' />`

eanplatter:
Overall it looks pretty great man!  Levi's comments on using stardust components whenever possible and then there are a lot of instances where `<i></i>`tags can be self closing; I didn't comment all of them.

Also I mentioned using map to handle lists of JSX elements, you might _not_ want to do that if the code is there specifically for someone to see, but if not something like this would work:

``` javascript
let colors = ['red', 'green', 'blue'];

let buttons = colors.map((item, index) => {
  return (
    <Button key={index} className={item}>
      {item}
    </Button>
  );
});
```

Then in the `return` of your `render` function, something like this:

``` jsx
<div>
  {buttons}
</div>
```

which would render this once for each color in the `colors` array:

``` jsx
  <Button className='red'>
    red
  </Button>
```

That way you don't have to repeat yourself too much! But like I said, just depends on what you're trying to do. Sometimes you need to repeat yourself when showing an example. Sorry if you already knew this :P

docs/app/Examples/Button/Labeled.js
levithomason: Clarifying:

A Stardust `<Button />` renders a `div` with the classes `ui button`.  We should use the Stardust `<Button />` instead of the `<div className='ui button'>` in this example.

docs/app/Examples/Button/Labeled.js
```diff
@@ -0,0 +1,18 @@
...
+import {Button} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Button className='labeled'>
+        <Button>
+          <i className='heart icon' /> Like
+        </Button>
+        // TODO: See PR #46 - button needs to render as a div
```
levithomason: I think you mean "See issue #46" here.

kyleturco: Dangit, yes

levithomason:
Looks good.  Just had one comment regarding the todo referencing a GitHub issue.  Looks like this PR has conflicts and needs [rebased](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request).

Then, :ship:!

PR #47, opened 2015-10-10
Eslint Ignore

levithomason:
`/docs/app` was excluded from linting on precommit.   This PR lints the entire code base on `npm run lint` which is also run on the precommit hook.  I've also added an `.eslintignore` to exclude vendors, build artifacts, and test artifacts.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,7 @@
+_builds/*
+_projects/*
+_steps/*
+bower_components/*
 node_modules/*
 docs/build/*
+testBundle.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+node_modules/*
+docs/build/*
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   ],
   "scripts": {
     "test": "karma start karma.conf.js",
-    "lint": "eslint src gulp test ./*.js",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   ],
   "scripts": {
     "test": "karma start karma.conf.js",
-    "lint": "eslint src gulp test ./*.js",
+    "lint": "eslint .",
```
levithomason, line 11: Linting `.` is recursive by default.

.eslintignore
```diff
@@ -0,0 +1,7 @@
+_builds/*
+_projects/*
+_steps/*
```
levithomason, line 3: Wercker artifacts.

.eslintignore
```diff
@@ -0,0 +1,7 @@
...
+bower_components/*
+node_modules/*
```
levithomason, line 5: Vendors.

.eslintignore
```diff
@@ -0,0 +1,7 @@
...
+docs/build/*
+testBundle.js
```
levithomason, line 7: Build and test artifacts.

eanplatter:
Seems legit :fallen_leaf:

athurman:
:octopus:

PR #52, opened 2015-10-13
Feature/example section

levithomason:
This PR restructures `/docs` for extensibility and parity with Semantic UI docs.  A few components were added to allow this.  Refer to the Checkbox examples to see how to use them.


docs/app/Components/ComponentDoc/ComponentDescription.js
```diff
@@ -59,7 +59,8 @@ export default class ComponentDescription extends Component {
             </a>
           </code>
         </div>
-        {this.isSemanticComponent && semanticDocsLink}
+        <p>{this.props.description}</p>
+        <p>{this.isSemanticComponent && semanticDocsLink}</p>
```
levithomason, line 63: Components now show their docBlock text as their description in the doc site.  So, this:

``` jsx
/**
 * It's a component
 */
export default class Component {...}
```

Would show up in the docs as:

> #### Component
> 
> It's a component.

docs/app/Components/ComponentDoc/ComponentDoc.js
```diff
@@ -0,0 +1,31 @@
...
+          parent={doc.parent}
+          type={doc.type}
+          description={doc.docBlock.description}
+        />
+        <ComponentProps props={doc.props} />
+        <ComponentExamples name={doc.name} />
+      </Segment>
+    );
+  }
+}
```
levithomason, line 31: This component is what generates a doc section for a given component.  Every component in stardust is looped and passed through this component to generate its doc.

docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -0,0 +1,65 @@
...
+  static propTypes = {
+    children: PropTypes.node,
+    description: PropTypes.string,
+    examplePath: PropTypes.string,
+    title: PropTypes.string,
+  };
+
+  state = {showCode: false};
+  fileContents = require(`!raw!docs/app/Examples/${this.props.examplePath}`);
+  component = req(`./${this.props.examplePath}.js`);
```
levithomason: We read the `examplePath` as plain text using the [`raw-loader`](https://github.com/webpack/raw-loader).  This is used to display the code in the example.

Then, we `require()` the actual component at `examplePath`.  This is used to render the example.

This is how we both show the code and rendered example using the path to the example file.

docs/app/Components/Sidebar/Sidebar.js
```diff
@@ -8,12 +7,10 @@ export default class DocsMenu extends Component {
...
 
   render() {
-    let menuItems = _(stardust)
-      .keys()
+    let menuItems = Object.keys(stardust)
       .sort()
       .filter(item => new RegExp(this.state.menuQuery, 'i').test(item))
-      .map(item => <MenuItem key={item} name={item} href={`#${item}`} />)
-      .value();
+      .map(item => <MenuItem key={item} name={item} href={`#${item}`} />);
```
levithomason, line 13: Just removed lodash here as it was using only native methods.

docs/app/Examples/elements/Button/Variations/Colored.js
```diff
@@ -0,0 +1,11 @@
+import React, {Component} from 'react';
+
+export default class extends Component {
+  render() {
+    return (
+      <div className='ui warning message'>
+        TODO
```
levithomason, line 7: I noticed this example was missing, so I added a filler "TODO" example so we don't forget it.

docs/app/Examples/modules/Checkbox/Types/Types.js
```diff
@@ -0,0 +1,36 @@
...
+          examplePath='modules/Checkbox/Types/Slider'
+        />
+
+        <ComponentExample
+          title='Radio'
+          description='A checkbox can be formatted as a radio element. This means it is an exclusive option.'
+          examplePath='modules/Checkbox/Types/Radio'
+        />
+
+        <ComponentExample examplePath='modules/Checkbox/Types/RadioGroup' />
```
levithomason, line 32: Notice examples do not have to have titles and descriptions.  By listing multiple ComponentExamples under another with a title/description you can get a nested set of examples:

image

src/modules/Checkbox/Checkbox.js
```diff
@@ -70,6 +58,15 @@ export default class Checkbox extends Component {
...
 
   render() {
+    let type = this.props.type;
+
+    if (!type) {
+      type = 'checkbox';
+      if (_.includes(this.props.className, 'radio')) {
+        type = 'radio';
+      }
+    }
```
levithomason, line 61: Make sure Checkboxes get the proper default `type` attribute.  If no `type` was supplied, we infer it from the `className`.

kyleturco:
🐎

PR #53, opened 2015-10-14
Conformance Tests

levithomason:
The below document is linked in the README.md.  I included it here for ease of access.

---
# Stardust Component Guidelines

Every component in Stardust must conform to these guidelines.  They ensure consistency and optimal development experience for Stardust users.

This document will be minimally maintained.  See [`\test`](https://github.com/TechnologyAdvice/stardust/tree/master/test) for the current conformance tests.
## Extends React.Component

**Valid**

``` jsx
import React, {Component} from 'react';

export default class MyComponent extends Component {...}
```

**Invalid**

``` jsx
import React, {Component} from 'react';

export default class MyComponent {...}
```

> This is a new class, does not extend React.Component.
## Constructor name matches component name

Give `MyComponent.js` is a component attached to `stardust.MyComponent`:

**Valid**

``` jsx
export default class MyComponent extends Component {...}
```

**Invalid**

``` jsx
export default class extends Component {...}
```

> This is an anonymous class, actually named `_deafult`.

``` jsx
export default class YourComponent extends Component {...}
```

> This class has the wrong name, it should be .
## Has the `sd-<component>` element as its first child (no wrapper elements)

**Valid**

``` jsx
export default class Input extends Component {
  render() {
    return (
      <Input />
    );
  }
}
```

**Invalid**

``` jsx
export default class Input extends Component {
  render() {
    return (
      <Field
        <Input />
      </Field>
    );
  }
}
```

> Never wrap the component with other components, whether DOM or Composite.
## Has props.className after `sd-<component>`

**Valid**

``` jsx
<Field className='inherit-this' />
// => <div className='sd-field inherit-this field>...
```

**Invalid**

``` jsx
<Field className='inherit-this' />
// => <div className='sd-field field>...
```

> className was not inherited

``` jsx
<Field className='inherit-this' />
// => <div className='inherit-this sd-field field>...
```

> className was inherited before sd-field

``` jsx
<Field className='inherit-this' />
// => <div className='inherit-this sd-field field>...
```

> className was not inherited before sd-field
## Has `sd-<component>` as the first class

**Valid**

``` jsx
<Divider />
// => <div className='sd-divider ui divider' /> 
```

**Invalid**

``` jsx
<Divider />
// => <div className='ui divider' /> 
```

> Missing `sd-divider` className.

``` jsx
<Divider />
// => <div className='ui sd-divider divider' /> 
```

> `sd-divider` className does not come first.
## Has `ui` class immediately after `sd-<component>`

Only for components that utilize the `ui` class (i.e `ui grid` but not `column`).

**Valid**

``` jsx
<Grid />
// => <div className='sd-grid ui grid' /> 
```

**Invalid**

``` jsx
<Grid />
// => <div className='sd-grid grid ui' /> 
```

> `grid` is immediately after `sd-grid`.
## Has props.className immediately after `ui`

Only for components that utilize the `ui` class (i.e `ui form` but not `field`).

**Valid**

``` jsx
<Form />
// => <div className='sd-form ui form' /> 
```

**Invalid**

``` jsx
<Form className='loading' />
// => <div className='sd-form loading ui form' /> 
```

> Inherited `loading` className comes before `ui`.

``` jsx
<Form className='loading' />
// => <div className='sd-form ui form loading' /> 
```

> Inherited `loading` className comes after `form`.
## Spreads props

Allows access to the underlying element of concern.

**Valid**

``` jsx
<Input onFocus={this.handleFocus} />
// => <input type='text' className='sd-input ui input' onFocus={this.handleFocus} /> 
```

``` jsx
<Input data-my-plugin='do-magic' />
// => <input type='text' className='sd-input ui input' data-my-plugin='do-magic' /> 
```

**Invalid**

``` jsx
<Input onFocus={this.handleFocus} />
// => <input type='text' className='sd-input ui input' /> 
```

> `onFocus` prop was lost.

``` jsx
<Input data-my-plugin='do-magic' />
// => <input type='text' className='sd-input ui input' /> 
```

> `data-my-plugin` prop was lost.

---

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,10 @@ npm i stardust -S
 ```
 
 ## Learn
-Checkout the **[Docs](https://technologyadvice.github.io/stardust/)**.
+Checkout the [Documentation](https://technologyadvice.github.io/stardust/).
+
+Review our [Component Guidelines]
+(https://github.com/TechnologyAdvice/stardust/blob/master/docs/app/Component Guidelines.md).
 
 ## Develop
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,6 @@ gulp help                 # list gulp commands
 ### Deploying
 
 ```
-npm version <version>
-npm run deploy
+npm version <version>     # bump the new version
+npm run deploy            # pushes tags and publishes package
 ```
--- a/docs/app/Component Guidelines.md
+++ b/docs/app/Component Guidelines.md
@@ -1,0 +1,199 @@
+# Stardust Component Guidelines
+
+Every component in Stardust must conform to these guidelines.  They ensure consistency and optimal development experience for Stardust users.
+
+This document will be minimally maintained.  See [`\test`](https://github.com/TechnologyAdvice/stardust/tree/master/test) for the current conformance tests.
+
+## Extends React.Component
+
+**Valid**
+
+```jsx
+import React, {Component} from 'react';
+
+export default class MyComponent extends Component {...}
+```
+
+**Invalid**
+
+```jsx
+import React, {Component} from 'react';
+
+export default class MyComponent {...}
+```
+>This is a new class, does not extend React.Component.
+
+## Constructor name matches component name
+
+Give `MyComponent.js` is a component attached to `stardust.MyComponent`:
+
+**Valid**
+
+```jsx
+export default class MyComponent extends Component {...}
+```
+
+**Invalid**
+
+```jsx
+export default class extends Component {...}
+```
+>This is an anonymous class, actually named `_deafult`.
+
+```jsx
+export default class YourComponent extends Component {...}
+```
+>This class has the wrong name, it should be .
+
+## Has the `sd-<component>` element as its first child (no wrapper elements)
+
+**Valid**
+
+```jsx
+export default class Input extends Component {
+  render() {
+    return (
+      <Input />
+    );
+  }
+}
+```
+
+**Invalid**
+
+```jsx
+export default class Input extends Component {
+  render() {
+    return (
+      <Field
+        <Input />
+      </Field>
+    );
+  }
+}
+```
+>Never wrap the component with other components, whether DOM or Composite.
+
+## Has props.className after `sd-<component>`
+
+**Valid**
+
+```jsx
+<Field className='inherit-this' />
+// => <div className='sd-field inherit-this field>...
+```
+
+**Invalid**
+
+```jsx
+<Field className='inherit-this' />
+// => <div className='sd-field field>...
+```
+>className was not inherited
+
+```jsx
+<Field className='inherit-this' />
+// => <div className='inherit-this sd-field field>...
+```
+>className was inherited before sd-field
+
+```jsx
+<Field className='inherit-this' />
+// => <div className='inherit-this sd-field field>...
+```
+>className was not inherited before sd-field
+
+## Has `sd-<component>` as the first class
+
+**Valid**
+
+```jsx
+<Divider />
+// => <div className='sd-divider ui divider' /> 
+```
+
+**Invalid**
+
+```jsx
+<Divider />
+// => <div className='ui divider' /> 
+```
+>Missing `sd-divider` className.
+
+```jsx
+<Divider />
+// => <div className='ui sd-divider divider' /> 
+```
+>`sd-divider` className does not come first.
+
+## Has `ui` class immediately after `sd-<component>`
+Only for components that utilize the `ui` class (i.e `ui grid` but not `column`).
+
+**Valid**
+
+```jsx
+<Grid />
+// => <div className='sd-grid ui grid' /> 
+```
+
+**Invalid**
+
+```jsx
+<Grid />
+// => <div className='sd-grid grid ui' /> 
+```
+>`grid` is immediately after `sd-grid`.
+
+## Has props.className immediately after `ui`
+
+Only for components that utilize the `ui` class (i.e `ui form` but not `field`).
+
+**Valid**
+
+```jsx
+<Form />
+// => <div className='sd-form ui form' /> 
+```
+
+**Invalid**
+
+```jsx
+<Form className='loading' />
+// => <div className='sd-form loading ui form' /> 
+```
+>Inherited `loading` className comes before `ui`.
+
+```jsx
+<Form className='loading' />
+// => <div className='sd-form ui form loading' /> 
+```
+>Inherited `loading` className comes after `form`.
+
+## Spreads props
+Allows access to the underlying element of concern.
+
+**Valid**
+
+```jsx
+<Input onFocus={this.handleFocus} />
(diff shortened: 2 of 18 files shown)


athurman:
Nice explanation!! Reads super easy!!

src/addons/Textarea/Textarea.js
```diff
@@ -1,40 +1,19 @@
...
-    onChange: PropTypes.func,
-    ref: PropTypes.string,
-    rows: PropTypes.oneOfType([
-      PropTypes.string,
-      PropTypes.number
-    ]),
-    value: PropTypes.string,
-  };
-
-  static defaultProps = {
```
athurman, line 18: were these unnecessary default props, and that is why they were removed?

src/collections/Form/Form.js
```diff
@@ -21,13 +22,14 @@ class Form extends Component {
...
         });
       }
+
       loopThruChildren(props.children);
 
       if (formButtons.length === 0 || !_.include(formButtons, 'submit')) {
         return new Error('`Form` must have `Button` child with type of "submit".');
       }
     },
-    onSubmit: PropTypes.func,
```
athurman, line 30: Do we no longer need this prop?

athurman:
Just a few general questions, but everything looks good!! :octopus:

src/addons/Textarea/Textarea.js
levithomason, line 18: Yep, none of these are required now since we inherit props and classes.  We don't need to define every prop and it's type, since any prop and type you add will be passed down.  We just need to define the ones we need to use.  Like `className` and `label`.

Same for `defaultProps`, this library will be generic and assume no defaults.  If you'd like 3 rows, simply add `rows={3}` to the component. :+1:

athurman, line 18: ah, makes total sense! sounds good.

src/collections/Form/Form.js
levithomason, line 30: Nope.  Notice all it did was handle this: `onSubmit={this.props.onSubmit}`.  That means when you submitted the form `this.props.onSubmit` was called with the default params.  Now, that is handled by spreading props as well.  If you add an `onSubmit` prop, it will be added to the component automatically and called automatically in default React style.

Since there a massive amount of props someone might pass down, we are opting to not validate all of them.  Devs should just know, the `onSubmit` prop is a function.  Along with all the other event listeners.  These are regular html concepts.

levithomason:
Cool!  I think I covered them all now. Let me know if there are any other questions/comments/doubts.

README.md
```diff
@@ -20,7 +20,10 @@ npm i stardust -S
 ```
 
 ## Learn
-Checkout the **[Docs](https://technologyadvice.github.io/stardust/)**.
+Checkout the [Documentation](https://technologyadvice.github.io/stardust/).
+
+Review our [Component Guidelines]
+(https://github.com/TechnologyAdvice/stardust/blob/master/docs/app/Component Guidelines.md).
```
levithomason, line 26: Added a link to the guideline doc in the readme.

docs/app/Component Guidelines.md
```diff
@@ -0,0 +1,199 @@
+# Stardust Component Guidelines
```
levithomason, line 1: This entire doc is the identical to the pull request description.

src/addons/Confirm/Confirm.js
```diff
@@ -1,14 +1,16 @@
...
 import Promise from 'bluebird';
+import classNames from 'classnames';
 
 import Modal from 'src/modules/Modal/Modal';
 import ModalContent from 'src/modules/Modal/ModalContent';
 import ModalFooter from 'src/modules/Modal/ModalFooter';
 import ModalHeader from 'src/modules/Modal/ModalHeader';
 
-class Confirm extends Component {
+export default class Confirm extends Component {
```
levithomason, line 10: We inline export our components

src/addons/Confirm/Confirm.js
```diff
@@ -52,8 +54,12 @@ class Confirm extends Component {
   };
 
   render() {
+    let classes = classNames(
+      'sd-confirm',
+      this.props.className,
+    );
```
levithomason, line 60: We use the `classnames` module to add the `sd-*` class, inherit `this.props.className` and add Semantic UI classes.

src/addons/Confirm/Confirm.js
```diff
@@ -52,8 +54,12 @@ class Confirm extends Component {
...
     return (
-      <Modal ref='modal' className='small'>
+      <Modal {...this.props} className={classes} ref='modal'>
```
levithomason, line 62: We [spread](https://facebook.github.io/react/docs/jsx-spread.html) props and use our merged classes for `className`.

Spreading props is the same as mixing objects.  All the keys/values on `this.props` are now on `Modal`.  Specifically listing prop values will override the spread values.  In this case, even if you pass in `className`, it will be overridden by `classes` (which includes `className`).

src/addons/Textarea/Textarea.js
```diff
@@ -1,40 +1,19 @@
...
+      this.props.className
+    );
     return (
-      <Field label={this.props.label}>
-        <textarea
-          {...this.props}
-          onChange={this.handleChange}
-          data-content={this.props.dataContent}
-        />
-      </Field>
```
levithomason, line 35: We do not wrap our components with extra components.  This dictates usage and limits reusability.  We instead go for single purpose generic components that are more reusable.

test/specs/Conformance-test.js
```diff
@@ -0,0 +1,102 @@
...
+
+const TestUtils = React.addons.TestUtils;
+
+let getSDClassName = componentName => `sd-${_.kebabCase(componentName)}`;
+
+/**
+ * This test ensures all Stardust components conform to our guidelines.
+ */
+describe('Conformance', () => {
+  _.each(stardust, (SDComponent, name) => {
```
levithomason, line 14: Here we loop over all components in the Stardust object and run our conformance tests against them.  Since we have both the component name (Stardust object keys) and the actual component (Stardust object values), we can generate the test names, class names, test the class, and render the class.

This means we can easily test many aspects of the component and make test messages very readable.

test/specs/addons/Textarea-test.js
```diff
@@ -10,10 +10,6 @@ describe('Textarea', () => {
     var renderedTextarea = render(<Textarea name='sample-post' />);
     expect(renderedTextarea.first().props.name).to.equal('sample-post');
   });
-  it('has 3 rows by default', () => {
-    var renderedTextarea = render(<Textarea name='sample' />);
-    expect(renderedTextarea.findTag('textarea').props.rows).to.equal('3');
-  });
```
levithomason, line 16: We don't set arbitrary styling defaults.  This will be left to the users of Stardust.  There will be future PRs to remove other styling and defaults as well as document any magic we are doing.

kyleturco:
Everything here looks great to me 🌞

PR #54, opened 2015-10-15
Feature/new example format

kyleturco:
Add buttons & checkboxes in our updated example format

### DIFF
--- a/docs/app/Examples/elements/Button/States/States.js
+++ b/docs/app/Examples/elements/Button/States/States.js
@@ -1,29 +1,27 @@
 import React, {Component} from 'react';
 import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample';
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection';
 
 export default class extends Component {
   render() {
     return (
-      <div>
+      <ExampleSection title='States'>
         <ComponentExample
           title='Active'
           description='A button can show it is currently the active user selection'
           examplePath='elements/Button/States/Active'
         />
-
         <ComponentExample
           title='Disabled'
           description='A button can show it is currently unable to be interacted with'
           examplePath='elements/Button/States/Disabled'
         />
-
         <ComponentExample
           title='Loading'
           description='A button can show a loading indicator'
           examplePath='elements/Button/States/Loading'
         />
-
-      </div>
+      </ExampleSection>
     );
   }
 }
--- a/docs/app/Examples/elements/Button/States/States.js
+++ b/docs/app/Examples/elements/Button/States/States.js
@@ -1,0 +1,29 @@
+import React, {Component} from 'react';
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <ComponentExample
+          title='Active'
+          description='A button can show it is currently the active user selection'
+          examplePath='elements/Button/States/Active'
+        />
+
+        <ComponentExample
+          title='Disabled'
+          description='A button can show it is currently unable to be interacted with'
+          examplePath='elements/Button/States/Disabled'
+        />
+
+        <ComponentExample
+          title='Loading'
+          description='A button can show a loading indicator'
+          examplePath='elements/Button/States/Loading'
+        />
+
+      </div>
+    );
+  }
+}
--- a/docs/app/Examples/modules/Checkbox/States/Checked.js
+++ b/docs/app/Examples/modules/Checkbox/States/Checked.js
@@ -1,12 +1,10 @@
 import React, {Component} from 'react';
-import {Form, Checkbox} from 'stardust';
+import {Checkbox} from 'stardust';
 
 export default class extends Component {
   render() {
     return (
-      <Form>
-        <Checkbox className='checked' checked={true} label='This checkbox comes prechecked' />
-      </Form>
+      <Checkbox className='checked' checked={true} label='This checkbox comes prechecked' />
     );
   }
 }
--- a/docs/app/Examples/modules/Checkbox/States/Checked.js
+++ b/docs/app/Examples/modules/Checkbox/States/Checked.js
@@ -1,0 +1,12 @@
+import React, {Component} from 'react';
+import {Form, Checkbox} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Form>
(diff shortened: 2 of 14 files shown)


kyleturco:
This PR feels a touch long, but it's generally basic stuff and made sense to group it all together

docs/app/Examples/elements/Button/States/States.js
```diff
@@ -0,0 +1,29 @@
...
+          examplePath='elements/Button/States/Disabled'
+        />
+
+        <ComponentExample
+          title='Loading'
+          description='A button can show a loading indicator'
+          examplePath='elements/Button/States/Loading'
+        />
+
+      </div>
```
levithomason: The extra line breaks between the components are probably fine, but this last one is only in a couple files and feels like it is breaking our `no-padded-blocks` rule.  That rule isn't working here because this is JSX and technically not a JS block.  I'd like to vote we remove them anyway, to keep consistent with the other files.

levithomason:
No worries on the size, I think it is OK.  In this case we aren't forging new heavy logic or precedence.  More so duplicating what has already been set as a standard.

docs/app/Examples/modules/Checkbox/States/Disabled.js
```diff
@@ -0,0 +1,17 @@
...
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <Form>
+          <Checkbox className='disabled' label='Disabled' />
+        </Form>
+        <Form>
+          <Checkbox className='toggle' disabled={true} label='Disabled' />
+        </Form>
```
levithomason: Use a single `Form` and wrap the `Checkbox` components in `Field` components to get the proper whitespace around them.  The root `div` will also not be required then as the `Form` will be the root.

image

vs

image

docs/app/Examples/modules/Checkbox/States/Checked.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+import {Form, Checkbox} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Form>
+        <Checkbox className='checked' checked={true} label='This checkbox comes prechecked' />
+      </Form>
```
levithomason: `Form` is unnecessary here.  Just the `Checkbox` as the only element.

docs/app/Examples/elements/Button/States/States.js
```diff
@@ -0,0 +1,29 @@
+import React, {Component} from 'react';
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <ComponentExample
```
levithomason, line 9: Questioning our approach now that I see this all laid out.  Instead of having a wrapper div here, wouldn't it make sense to use the `ExampleSection` in these files that bundle `ComponentExamples`?  So, here we'd have:

``` jsx
<ExampleSection title='States'>
  <ComponentExample
    title='Active'
    description='A button can show it is currently the active user selection'
    examplePath='elements/Button/States/Active'
  />
  <ComponentExample
    title='Disabled'
    description='A button can show it is currently unable to be interacted with'
    examplePath='elements/Button/States/Disabled'
  />
  <ComponentExample
    title='Loading'
    description='A button can show a loading indicator'
    examplePath='elements/Button/States/Loading'
  />
</ExampleSection>
```

Then, in `Button/ButtonExamples.js` instead of wrapping every import an `ExampleSection`, we could just have:

``` jsx
<div>
  <Types />
  <States />
  <Variations />
</div>
```

Thoughts?

docs/app/Examples/modules/Checkbox/States/ReadOnly.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+import {Form, Checkbox} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <Form>
+        <Checkbox className='read-only' label='This checkbox is read-only' />
+      </Form>
```
levithomason: `Form` not needed here.

docs/app/Examples/modules/Checkbox/Variations/Fitted.js
```diff
@@ -0,0 +1,26 @@
...
+import {Form, Checkbox, Segment} from 'stardust';
+
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <Segment className='compact'>
+          <Form>
+            <Checkbox className='fitted' />
+          </Form>
```
levithomason: `Form` no needed aroundt these checkboxes.

docs/app/Examples/modules/Checkbox/Variations/Variations.js
```diff
@@ -0,0 +1,16 @@
...
+export default class extends Component {
+  render() {
+    return (
+      <div>
+        <ComponentExample
+          title='Fitted'
+          description='A fitted checkbox does not leave padding for a label'
+          examplePath='modules/Checkbox/Variations/Fitted'
+        />
+      </div>
```
levithomason: Extra `div` not needed here, only when nesting multiple children.  More accurately, a component can only have a single root node with no siblings.

If we adopt the [above comment](#r42073397) regarding `ExampleSection` wrapping `ComponentExamples` this comment will no longer be valid.

levithomason:
While we're at it, I noticed the Radio example also has an extra wrapper div.  We should remove that:

image

Just the single `Form` as the root is OK.

docs/app/Examples/elements/Button/States/States.js
kyleturco: Went ahead and removed line breaks from all of the equivalent files. I was concerned with readability, but that was probably an overreaction.

docs/app/Examples/modules/Checkbox/States/Checked.js
kyleturco: Makes sense, didn't occur to me Checkbox could be standalone. You've noted it on a few others, and I'll double check the rest

docs/app/Examples/elements/Button/States/States.js
kyleturco, line 9: That seems pretty spiffy to me. I made the `div vs ExampleSection` update on all of the different sections, now updating the main template

kyleturco, line 9: Update - this works quite well

levithomason:
Beautiful :boom:

PR #59, opened 2015-10-16
fix warnings, add class names, necessary updates

kyleturco:
lots of small updates

### DIFF
--- a/docs/app/Examples/elements/Button/States/Loading.js
+++ b/docs/app/Examples/elements/Button/States/Loading.js
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import {Button} from 'stardust';
 
-export default class LoadingExapmle extends Component {
+export default class LoadingExample extends Component {
   render() {
     return (
       <div>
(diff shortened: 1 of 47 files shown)


docs/app/Examples/elements/Button/States/Loading.js
```diff
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import {Button} from 'stardust';
 
-export default class extends Component {
+export default class LoadingExapmle extends Component {
```
levithomason: "Exapmle" typo.

docs/app/Examples/elements/Button/Types/Labeled.js
```diff
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import {Button} from 'stardust';
 
-export default class extends Component {
+export default class LabelExample extends Component {
```
levithomason: "Labeled" example.

webpack.docs.js
```diff
@@ -60,6 +60,7 @@ module.exports = {
     },
   },
   plugins: [
-    new webpack.HotModuleReplacementPlugin()
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.NoErrorsPlugin(),
```
levithomason: This extra plugin did not solve the issue with the JSX parer stopping the build.  Let's pull it since it is doing nothing.

PR #65, opened 2015-10-19
Feature/add sd item

athurman:
2 New SD components for the [Semantic UI Item](http://semantic-ui.com/views/item.html) view:

> An item view presents large collections of site content for display
##### EDIT

PR also addresses GH Issue #68

### DIFF
--- a/src/utils/customPropTypes.js
+++ b/src/utils/customPropTypes.js
@@ -1,0 +1,16 @@
+import _ from 'lodash';
+
+let customPropTypes = {
+  mutuallyExclusive: (exclusives, props, propName, componentName) => {
+    _.forEach(exclusives, exclusiveProp => {
+      if (props[propName] && props[exclusiveProp]) {
+        throw new Error(`\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\`, ` +
+          `not both.`);
+      }
+
+      return null;
+    });
+  },
+};
+
+export default customPropTypes;
--- a/src/utils/customPropTypes.js
+++ b/src/utils/customPropTypes.js
@@ -1,15 +1,16 @@
 import _ from 'lodash';
 
 let customPropTypes = {
-  mutuallyExclusive: (exclusives, props, propName, componentName) => {
-    _.forEach(exclusives, exclusiveProp => {
-      if (props[propName] && props[exclusiveProp]) {
-        throw new Error(`\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\`, ` +
-          `not both.`);
-      }
-
-      return null;
-    });
+  mutuallyExclusive: exclusives => {
+    return (props, propName, componentName) => {
+      _.forEach(exclusives, exclusiveProp => {
+        if (props[propName] && props[exclusiveProp]) {
+          throw new Error(
+            `\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\` not both.`
+          );
+        }
+      });
+    };
   },
 };
 
--- a/src/views/Items/Item.js
+++ b/src/views/Items/Item.js
@@ -5,7 +5,11 @@ export default class Item extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
-    description: PropTypes.node,
+    description: (props, propName, componentName) => {
+      if (props.children && props.description) {
+        return new Error('`Item` should only have one of type `description` or `children`, not both.');
+      }
+    },
     extra: PropTypes.node,
     heading: PropTypes.node,
     image: PropTypes.node,
--- a/src/views/Items/Item.js
+++ b/src/views/Items/Item.js
@@ -1,0 +1,43 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+
+export default class Item extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    description: PropTypes.node,
+    extra: PropTypes.node,
+    heading: PropTypes.node,
+    image: PropTypes.node,
+    meta: PropTypes.node,
+  };
+
+  render() {
+    let heading = <div className='header'>{this.props.heading}</div>;
+    let meta = <div className='meta'>{this.props.meta}</div>;
+    let description = <div className='description'>{this.props.children || this.props.description}</div>;
+    let extra = <div className='extra'>{this.props.extra}</div>;
+    let content = (
+      <div className='middle aligned content'>
+        {this.props.heading && heading}
+        {this.props.meta && meta}
+        {this.props.children && description || this.props.description && description}
+        {this.props.extra && extra}
+      </div>
+    );
+    let hasContent = !!this.props.heading || !!this.props.meta || !!this.props.extra || !!this.props.children
+    || !!this.props.description;
+
+    let classes = classNames(
+      'sd-item',
+      this.props.className,
+      'item',
+    );
+    return (
+      <div {...this.props} className={classes}>
+        {this.props.image}
+        {hasContent && content}
+      </div>
+    );
+  }
+}
--- a/src/views/Items/Item.js
+++ b/src/views/Items/Item.js
@@ -1,17 +1,14 @@
 import React, {Component, PropTypes} from 'react';
 import classNames from 'classnames';
+import customPropTypes from '../../utils/customPropTypes';
 
 export default class Item extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
-    description: (props, propName, componentName) => {
-      if (props.children && props.description) {
-        return new Error('`Item` should only have one of type `description` or `children`, not both.');
-      }
-    },
+    description: customPropTypes.mutuallyExclusive.bind(null, ['children']),
     extra: PropTypes.node,
-    heading: PropTypes.node,
+    heading: PropTypes.node.isRequired,
     image: PropTypes.node,
     meta: PropTypes.node,
   };
--- a/src/views/Items/Item.js
+++ b/src/views/Items/Item.js
@@ -6,20 +6,26 @@ export default class Item extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
+    contentClass: PropTypes.string,
     description: customPropTypes.mutuallyExclusive(['children']),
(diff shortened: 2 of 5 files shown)


src/views/Items/Item.js
```diff
@@ -0,0 +1,43 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+
+export default class Item extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    description: PropTypes.node,
```
athurman: Need to add utility for adding warning if children exists.

src/views/Items/Item.js
```diff
@@ -0,0 +1,43 @@
...
+    extra: PropTypes.node,
+    heading: PropTypes.node,
+    image: PropTypes.node,
+    meta: PropTypes.node,
+  };
+
+  render() {
+    let heading = <div className='header'>{this.props.heading}</div>;
+    let meta = <div className='meta'>{this.props.meta}</div>;
+    let description = <div className='description'>{this.props.children || this.props.description}</div>;
```
athurman, line 24: By default, will render children before the default property if they both exist on the component.

src/views/Items/Items.js
```diff
@@ -0,0 +1,23 @@
...
+import classNames from 'classnames';
+
+export default class Items extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+  };
+
+  render() {
+    let classes = classNames(
```
athurman, line 11: Parent items div if more than one item needs to exist.

src/views/Items/Item.js
```diff
@@ -0,0 +1,43 @@
...
+export default class Item extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    description: PropTypes.node,
+    extra: PropTypes.node,
+    heading: PropTypes.node,
+    image: PropTypes.node,
+    meta: PropTypes.node,
+  };
```
athurman, line 15: Pass in all "pieces" of the component as props to make sure that all content matches semantic's markup.

src/views/Items/Item.js
```diff
@@ -0,0 +1,47 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+
+export default class Item extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    description: (props, propName, componentName) => {
+      if (props.children && props.description) {
```
athurman: prop validation that spits out an error if both `children` and `description` exists.

test/specs/views/Items/Item-test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {Item} from 'stardust';
+
+describe('Item', () => {
+  it('has children', () => {
+    let renderedItem = render(
+      <Item description='discription'>
```
athurman: added this test to confirm we receive a consoled error regarding description and children for an Item component.

src/views/Items/Item.js
levithomason: Nice.  That was easy :)  Let's pull this function out into a method on a props util so we can use it on all the components.  I'll also be adding a util soon to get spreadable props, I'll add it to your util after merge.

levithomason: Thinking about our other utils and how we can pull these together.  I think we should actually just this util as a standalone util in the `utils` directory.  We can make an index file later to organize them.  I've added 3 issues for util work that will cover the rest:
#68 CustomPropTypes util
#69 Spreadable props util
#70 Static meta info object

test/specs/views/Items/Item-test.js
levithomason: I've added an issue to extend conformance tests to handle this so we only have to write it once.  See #71.  We can safely remove this test for that issue.

For future reference, use a `find*()` method to assert the presence of something in the hierarchy.  Avoid writing tests that depend on component structure.  The find method will fail the test if it does not find exactly one occurrence.  So, no matches or more than one will fail.

``` jsx
let child = faker.hacker.phrase();
render(<Item description='foo'>{child}</Item>).findText(child);
```

Tests that depend on the structure of the component are more apt to fail for the wrong reason.  In this case, if the children were rendered somewhere else, this test would fail even though the component is still rendering children.

src/views/Items/Item.js
eanplatter: that is neat.

eanplatter:
:fallen_leaf:

src/utils/customPropTypes.js
```diff
@@ -0,0 +1,16 @@
...
+        throw new Error(`\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\`, ` +
+          `not both.`);
+      }
+
+      return null;
+    });
+  },
+};
+
+export default customPropTypes;
```
athurman, line 17: new utility to be used with customPropType validation and checking.

src/utils/customPropTypes.js
```diff
@@ -0,0 +1,16 @@
+import _ from 'lodash';
+
+let customPropTypes = {
+  mutuallyExclusive: (exclusives, props, propName, componentName) => {
+    _.forEach(exclusives, exclusiveProp => {
+      if (props[propName] && props[exclusiveProp]) {
+        throw new Error(`\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\`, ` +
+          `not both.`);
+      }
```
athurman: In order to compare the specific propType to the exclusive props passed in, I had to take advantage React's custom validation function found in [this article](https://facebook.github.io/react/docs/reusable-components.html) - so we have to bind the `exclusives` to this function in order to access and take advantage of the additional params `props` `propName` and `componentName`. 

This way, we have the current prop to compare the `exclusives` to.

src/views/Items/Item.js
```diff
@@ -0,0 +1,44 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+import customPropTypes from '../../utils/customPropTypes';
+
+export default class Item extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    description: customPropTypes.mutuallyExclusive.bind(null, ['children']),
```
athurman: Binding the exclusive props to the mutuallyExclusive utility function here.

src/utils/customPropTypes.js
```diff
@@ -0,0 +1,16 @@
+import _ from 'lodash';
+
+let customPropTypes = {
+  mutuallyExclusive: (exclusives, props, propName, componentName) => {
+    _.forEach(exclusives, exclusiveProp => {
+      if (props[propName] && props[exclusiveProp]) {
+        throw new Error(`\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\`, ` +
```
athurman: Had to `throw` error as opposed to `return` because the new Error needed to attach itself to the prop, not the utility property `mutuallyExclusive`

src/utils/customPropTypes.js
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash';
+
+let customPropTypes = {
+  mutuallyExclusive: exclusives => {
+    return (props, propName, componentName) => {
+      _.forEach(exclusives, exclusiveProp => {
+        if (props[propName] && props[exclusiveProp]) {
+          throw new Error(
+            `\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\` not both.`
+          );
```
eanplatter, line 10: I didn't think of throwing error messages with ticks. nice!

eanplatter:
:fallen_leaf:

src/views/Items/Item.js
```diff
@@ -0,0 +1,44 @@
...
+    let hasContent = !!this.props.heading || !!this.props.meta || !!this.props.extra || !!this.props.children
+    || !!this.props.description;
+
+    let classes = classNames(
+      'sd-item',
+      this.props.className,
+      'item',
+    );
+    return (
+      <div {...this.props} className={classes}>
```
levithomason: Component props used for Stardust purposes should be removed from `this.props` before spreading.  Only the user's props should be spread.  Since this component takes in:

```
children
className
description
extra
heading
image
meta
```

We should remove those props before spreading them onto the `Item`.  Otherwise, we'll be spreading components and other unintended things:

``` jsx
let props = _.clone(this.props);
delete props.children;
delete props.className;
delete props.description;
delete props.extra;
delete props.heading;
delete props.image;
delete props.meta;

// then

<div {...props} ...
```

Thinking about this, I believe we should always remove `static PropTypes` props, except for `static defaultProps` before spreading.  This way we only spread "extra" user defined props and defaultProps the user didn't specify.  Linking #69 spreadable props util here for this reference.

For now, add the clone/delete step before spreading.

src/views/Items/Item.js
```diff
@@ -0,0 +1,44 @@
...
+    meta: PropTypes.node,
+  };
+
+  render() {
+    let heading = <div className='header'>{this.props.heading}</div>;
+    let meta = <div className='meta'>{this.props.meta}</div>;
+    let description = <div className='description'>{this.props.children || this.props.description}</div>;
+    let extra = <div className='extra'>{this.props.extra}</div>;
+    let content = (
+      <div className='middle aligned content'>
```
levithomason: Stardust should not add default style classes that are not configurable.  Perhaps exposing a `contentClass` prop would make sense.  We can then use `defaultProps` to set it to this value but the user could supply their own classes.  Use `classNames()` to merge the content classes the same as we do for className, adding `sd-item-content` and `content` classes automatically.

**EDIT** fixed sd-\* class name typo.

test/specs/views/Items/Item-test.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import {Item} from 'stardust';
+import faker from 'faker';
+
+describe('Item', () => {
+  it('has children', () => {
```
levithomason: Small semantic request here.  This test isn't testing that `Item has children` but that `Item renders children`.

levithomason:
Good to merge after hitting the last to comments:
https://github.com/TechnologyAdvice/stardust/pull/65/files#r42529042
https://github.com/TechnologyAdvice/stardust/pull/65/files#r42529646

Also, can close #68.  Nice work on that, and bundling the `<Items />` component.

:+1: :zap:

Issue #61, opened 2015-10-16
Input does not handle labels


levithomason:
Though the Stardust Input API may be different than what was attempted below, we should be handling this functionality.

![image](https://cloud.githubusercontent.com/assets/5067638/10555624/c6a6c582-7426-11e5-9f50-c2f84e50d085.png)

levithomason:
Fixed in #67

PR #67, opened 2015-10-19
Feature/input examples

kyleturco:
Adding input examples to the doc site


docs/app/Examples/elements/Input/Variations/Size.js
```diff
@@ -0,0 +1,34 @@
...
+          <i className='search icon' />
+        </Input>
+        <br />
+        <Input className='massive icon' placeholder='Search...'>
+          <i className='search icon' />
+        </Input>
+      </div>
+    );
+  }
+}
```
kyleturco: Wasn't sure the best way to handle this example. I put all the size examples in one file with `br` between them. Alternate ideas would be create separate files.. eg.. `SizeMini.js`, `SizeBig.js` and then put them all in the Size example section. Thought it was 50/50 on adding lots of files vs putting all this in one. Open to suggestions

docs/app/Examples/elements/Input/Variations/Variations.js
```diff
@@ -0,0 +1,49 @@
...
+          examplePath='elements/Input/Variations/Inverted'
+        />
+        <ComponentExample
+          title='Fluid'
+          description='An input can take on the size of its container'
+          examplePath='elements/Input/Variations/Fluid'
+        />
+        <ComponentExample
+          title='Size'
+          description='An input can vary in sizee'
```
kyleturco: I'll fix the "sizee" typo after all the other comments are

docs/app/Examples/elements/Input/Variations/Labeled.js
```diff
@@ -0,0 +1,14 @@
+import React, {Component} from 'react';
+import {Input} from 'stardust';
+
+export default class Labeled extends Component {
+  render() {
+    return (
+      <Input className='labeled' placeholder='mysite.com'>
+        <div className='ui label'>
```
kyleturco: See Issue #61 - currently Input can't handle a label. This is causing it to render weird where the button can't go on the left side.

docs/app/Examples/elements/Input/Variations/Fluid.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+import {Input} from 'stardust';
+
+export default class Fluid extends Component {
+  render() {
+    return (
+      <Input className='fluid icon' placeholder='Search...'>
+        <i className='search icon' />
+      </Input>
```
kyleturco: Same issue as the "Labeled.js". Button is only displaying to the right of the input field

levithomason: fixed

PR #74, opened 2015-10-20
Feature/add sd image

athurman:
Add Stardust Image component that will be used in the `Item` component

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 // Addons
 import Confirm from 'src/addons/Confirm/Confirm';
+import Image from 'src/addons/Image/Image';
 import Textarea from 'src/addons/Textarea/Textarea';
 
 // Collections
@@ -38,6 +39,7 @@ import Items from 'src/views/Items/Items';
 export default {
   // Addons
   Confirm,
+  Image,
   Textarea,
 
   // Collections
--- a/index.js
+++ b/index.js
@@ -1,6 +1,5 @@
 // Addons
 import Confirm from 'src/addons/Confirm/Confirm';
-import Image from 'src/addons/Image/Image';
 import Textarea from 'src/addons/Textarea/Textarea';
 
 // Collections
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1,5 +1,6 @@
 // Addons
 import Confirm from 'src/addons/Confirm/Confirm';
+import Image from 'src/addons/Image/Image';
```
levithomason: Semantic UI Image is an `element`: http://semantic-ui.com/elements/image.html

The `addon` type is for components that Stardust adds to Semantic, like `Confirm`.  There is no `Confirm` component in Semantic UI so it is an `addon`.

src/addons/Image/Image.js
```diff
@@ -0,0 +1,24 @@
...
+  render() {
+    let classes = classNames(
+      'sd-image',
+      'ui',
+      this.props.className,
+      'image',
+    );
+    return (
+      <div {...this.props} className={classes}>
+        <img src={this.props.src} />
```
levithomason: The `img` element should at least also accept `alt` from props.

src/addons/Image/Image.js
```diff
@@ -0,0 +1,24 @@
...
+
+  render() {
+    let classes = classNames(
+      'sd-image',
+      'ui',
+      this.props.className,
+      'image',
+    );
+    return (
+      <div {...this.props} className={classes}>
```
levithomason: The `this.props` being spread here still contains `src`.  That should be deleted before spreading.  Make a clone of the props and delete the `src` key, then spread the clone:

``` jsx
const props = _.clone(this.props);
delete props.src;

<div {...props} ...
```

`alt` should be added to the delete list after adding it to the `img` as well, see next comment.

index.js
levithomason: done

src/addons/Image/Image.js
levithomason: done

PR #80, opened 2015-10-22
Checkbox plugin

### LINKED ISSUE #72: Conformance test for element and plugin
Semantic Modules each have a plugin. We expose this plugin at `this.element.<pluginName>` which is the component name.  The element is the jquery element.  We should test that both the element and the plugin are exposed.

Also, we should test that the plugin is called on `componentDidMount`, and that `this.element.off()` is called on `componentWillUnmount`.

To do this, we should certainly add the meta info in #70 so that we can determine if this is a Semantic component and if it is a module.

Before embarking on implementing this functionality in all components, consider a decorator that adds this boiler plate.

levithomason:
This PR sets precedence for how we deal with Semantic UI jQuery plugins.

Fixes #72

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,3 @@
 {
   "extends": "ta-webapp",
-  "env": {
-    "node": true
-  }
 }
--- a/docs/app/Components/ComponentDoc/ComponentExample.js
+++ b/docs/app/Components/ComponentDoc/ComponentExample.js
@@ -42,12 +42,12 @@ export default class ComponentExample extends Component {
       </Column>
     );
 
-    let linkIconStyle = {
+    const linkIconStyle = {
       display: this.state.showLink ? 'inline-block' : 'none',
       marginLeft: '0.25em',
     };
 
-    let children = <Column>{this.props.children}</Column>;
+    const children = <Column>{this.props.children}</Column>;
 
     return (
       <Grid className='one column' style={{marginBottom: '4em'}} id={this.anchor}>
--- a/docs/app/Components/ComponentDoc/ComponentExample.js
+++ b/docs/app/Components/ComponentDoc/ComponentExample.js
@@ -9,6 +9,7 @@ import exampleContext from 'docs/app/utils/ExampleContext';
  */
 export default class ComponentExample extends Component {
   static propTypes = {
+    children: PropTypes.node,
     description: PropTypes.string,
     examplePath: PropTypes.string.isRequired,
     title: PropTypes.string,
@@ -17,9 +18,19 @@ export default class ComponentExample extends Component {
   state = {showCode: false};
   fileContents = require(`!raw!docs/app/Examples/${this.props.examplePath}`);
   component = exampleContext(`./${this.props.examplePath}.js`);
+  // 'elements/Button/Types/Button' => #Button-Types-Button
+  anchor = this.props.examplePath.split('/').slice(1).join('-');
 
   toggleShowCode = () => {
     this.setState({showCode: !this.state.showCode});
+  };
+
+  handleMouseEnter = () => {
+    this.setState({showLink: true});
+  };
+
+  handleMouseLeave = () => {
+    this.setState({showLink: false});
   };
 
   render() {
@@ -31,13 +42,28 @@ export default class ComponentExample extends Component {
       </Column>
     );
 
+    let linkIconStyle = {
+      display: this.state.showLink ? 'inline-block' : 'none',
+      marginLeft: '0.25em',
+    };
+
+    let children = <Column>{this.props.children}</Column>;
+
     return (
-      <Grid className='one column' style={{marginBottom: '4em'}}>
+      <Grid className='one column' style={{marginBottom: '4em'}} id={this.anchor}>
         <Column>
           <Grid>
             <Column width={12}>
-              <h3 className='ui header' style={{marginBottom: 0}}>
+              <h3
+                className='ui header'
+                style={{marginBottom: 0}}
+                onMouseEnter={this.handleMouseEnter}
+                onMouseLeave={this.handleMouseLeave}
+              >
                 {this.props.title}
+                <a href={`#${this.anchor}`}>
+                  <i className='linkify icon' style={linkIconStyle} />
+                </a>
               </h3>
               <p>{this.props.description}</p>
             </Column>
@@ -49,7 +75,7 @@ export default class ComponentExample extends Component {
             </Column>
           </Grid>
         </Column>
-
+        {this.props.children && children}
         <Column>
           {createElement(this.component)}
         </Column>
--- a/docs/app/Examples/modules/Checkbox/States/CheckboxRemoteControlExample.js
+++ b/docs/app/Examples/modules/Checkbox/States/CheckboxRemoteControlExample.js
@@ -1,0 +1,26 @@
+import React, {Component} from 'react';
+import {Button, Checkbox} from 'stardust';
+
+export default class CheckboxRemoteControlExample extends Component {
+  state = {checked: false};
+
+  toggle = () => {
+    this.refs.checkbox.plugin('toggle');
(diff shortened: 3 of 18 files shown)


.eslintrc
```diff
@@ -1,6 +1,3 @@
 {
   "extends": "ta-webapp",
-  "env": {
-    "node": true
-  }
```
levithomason, line 5: Cruft, this is handled by the ta-webapp config now.

docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -49,7 +75,7 @@ export default class ComponentExample extends Component {
             </Column>
           </Grid>
         </Column>
-
+        {this.props.children && children}
```
levithomason, line 78: We now render children so we can add `<Message />`s to our examples.

docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -17,11 +18,21 @@ export default class ComponentExample extends Component {
   state = {showCode: false};
   fileContents = require(`!raw!docs/app/Examples/${this.props.examplePath}`);
   component = exampleContext(`./${this.props.examplePath}.js`);
+  // 'elements/Button/Types/Button' => #Button-Types-Button
+  anchor = this.props.examplePath.split('/').slice(1).join('-');
```
levithomason, line 22: Specific examples now have links on hover:

image

docs/app/Examples/modules/Checkbox/States/Checked.js
```diff
@@ -1,18 +1,8 @@
...
-    return (
-      <Checkbox label='This checkbox comes prechecked' checked={this.state.isChecked} onChange={this.handleChange} />
-    );
-  }
-}
+export default CheckboxCheckedExample => {
+  return (
+    <Checkbox defaultChecked label='This checkbox comes prechecked' />
+  );
+};
```
levithomason, line 8: To check a checkbox, just set the React `defaultChecked` prop.  This works because props are spread where they should be.

docs/app/Examples/modules/Checkbox/Variations/Fitted.js
```diff
@@ -6,13 +6,13 @@ export default class CheckboxFittedExample extends Component {
...
-          <Checkbox className='fitted' />
+          <Checkbox />
         </Segment>
         <Segment className='compact'>
-          <Checkbox className='fitted slider' />
+          <Checkbox className='slider' />
         </Segment>
         <Segment className='compact'>
-          <Checkbox className='fitted toggle' />
+          <Checkbox className='toggle' />
```
levithomason, line 15: The fitted class is automatically added.  Note the addition of the message explaining so on the example:

image

gulp/tasks/docs.js
```diff
@@ -39,7 +39,9 @@ gulp.task('generate-docs-json', cb => {
     paths.srcViews + '/**/*.js',
     '!' + paths.src + '/**/Style.js'
   ])
-    .pipe(g.plumber(err => {
+    // do not remove the function keyword
+    // we need 'this' scope here
+    .pipe(g.plumber(function(err) {
       g.util.log(err);
       this.emit('end');
```
levithomason, line 46: This was why watchers we dying on jsx parse error again.  `this.emit` was not the right `this`.  Arrow functions inherit `this` from the surrounding context, but we need a new scope here.

src/modules/Checkbox/Checkbox.js
```diff
@@ -76,12 +72,10 @@ export default class Checkbox extends Component {
       {fitted: !this.props.label},
       'checkbox'
     );
-
-    const checkboxProps = _.clone(this.props);
-    delete checkboxProps.className;
+    const props = getUnhandledProps(this);
```
eanplatter, line 75: I  hadn't seen this utility before! NEATO!

src/utils/Meta.js
```diff
@@ -15,7 +15,22 @@ const META = {
...
+  isCollection: ({_meta}) => _meta.type === META.type.collection,
+  isElement: ({_meta}) => _meta.type === META.type.element,
+  isView: ({_meta}) => _meta.type === META.type.view,
+  isModule: ({_meta}) => _meta.type === META.type.module,
+
+  // parent
+  isChild: ({_meta}) => !!_meta.parent,
 };
 
 export default META;
```
eanplatter, line 36: What exactly does META do again?

docs/app/Examples/modules/Checkbox/States/CheckboxRemoteControlExample.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+import {Button, Checkbox} from 'stardust';
+
+export default class CheckboxRemoteControlExample extends Component {
+  toggle = () => {
+    this.refs.checkbox.plugin('toggle');
+  };
```
eanplatter, line 7: Are we going to include these methods inside the stardust docs? Or just refer them to the semantic docs?

docs/app/Examples/modules/Checkbox/Variations/Variations.js
```diff
@@ -10,7 +11,15 @@ export default class CheckboxVariationsExamples extends Component {
...
           description='A fitted checkbox does not leave padding for a label'
           examplePath='modules/Checkbox/Variations/Fitted'
-        />
+        >
+          <Message>
+            The&nbsp;
+            <a href='http://semantic-ui.com/modules/checkbox.html#fitted' target='_blank'>
+              <code>fitted</code>
+            </a>
+            &nbsp;class is automatically applied if there is no <code>label</code> prop.
```
kyleturco, line 20: Is the `&nbsp;` normal to use here?

docs/app/Examples/modules/Checkbox/States/States.js
```diff
@@ -10,7 +11,17 @@ export default class CheckboxStatesExamples extends Component {
...
           examplePath='modules/Checkbox/States/Checked'
-        />
+        >
+          <Message>
+            Use
+            &nbsp;
+            <a href='https://facebook.github.io/react/docs/forms.html#default-value' target='_blank'>
+              <code>defaultChecked</code>
+            </a>
+            &nbsp;
```
kyleturco, line 21: Same thing with `&nbsp;` here - should that be used?

src/utils/Meta.js
levithomason, line 36: Defines meta data for components. What library is it from, what type is it, does it have a parent?  This let's us declaratively generate docs and tests for component specific things.

docs/app/Examples/modules/Checkbox/States/CheckboxRemoteControlExample.js
levithomason, line 7: Yup!

docs/app/Examples/modules/Checkbox/Variations/Variations.js
levithomason, line 20: Yes, React removes all whitespace regardless of how many blank lines (unlike html).  So, there is no space between words/components without this.

docs/app/Examples/modules/Checkbox/States/States.js
levithomason, line 21: See previous.

eanplatter:
:fallen_leaf:

src/modules/Checkbox/Checkbox.js
levithomason, line 75: This is a newer util.  Gets all props not defined in propTypes nor defaultProps.

docs/app/Examples/modules/Checkbox/Variations/Variations.js
kyleturco, line 20: Crazy! Good to know

PR #85, opened 2015-10-23
add getUnhandledProps util

### LINKED ISSUE #69: getSpreadableProps util
A util for getting spreadable props, to handle this use case:

``` jsx
let checkboxProps = _.clone(this.props);
delete checkboxProps.settings;  // all modules need to do this
delete checkboxProps.className; // this would be passed in

// then

<input {...checkboxProps} />
```

We need to remove `props.settings` from all Semantic UI Modules.  Also, many components need to exclude certain props from being spread as they belong elsewhere.  Like the checkbox example above, and this message example:

``` jsx
let messageProps = _.clone(this.props);
delete messageProps.icon;

// then

<div {...messageProps}>
```

The icon prop belongs as classes on the icon inside the message, then the rest of the props are spread.

This should be abstracted away with a util that understands what to do to get spreadable props.

levithomason:
Fixes #69 

This util gets all props for a component that are not defined in `propTypes` nor in `defaultProps`.  These extra props are considered 'unhandled` by the component at hand.

The immediate usage of this util will be for spreading props the user passed in that we aren't already explicitly handling.

### DIFF
--- a/test/utils/getUnhandledPops-test.js
+++ b/test/utils/getUnhandledPops-test.js
@@ -1,0 +1,33 @@
+import _ from 'lodash';
+import getUnhandledProps from 'src/utils/getUnhandledProps';
+
+// Helper class that takes in props and merges defaultProps
+class TestClass {
+  constructor(props) {
+    this.props = _.assign({}, this.constructor.defaultProps, props);
+    this.unhandledProps = getUnhandledProps(this);
+  }
+}
+
+describe.only('getUnhandledProps', () => {
+  it('removes props defined in defaultProps', () => {
+    TestClass.defaultProps = {imHandled: 'thanks'};
+    new TestClass()
+      .unhandledProps
+      .should.not.have.any.keys(_.keys(TestClass.defaultProps));
+  });
+  it('removes props defined in propTypes', () => {
+    TestClass.propTypes = {imHandled: 'thanks'};
+    new TestClass()
+      .unhandledProps
+      .should.not.have.any.keys(_.keys(TestClass.defaultProps));
+  });
+  it('leave props not in defaultProps || propTypes in tact', () => {
+    TestClass.defaultProps = {imHandled: 'thanks'};
+    TestClass.propTypes = {alsoHandled: 'got it'};
+    const props = {thisShould: 'still be here'};
+    new TestClass(props)
+      .unhandledProps
+      .should.eql(props);
+  });
+});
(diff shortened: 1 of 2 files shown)


eanplatter:
This is an awesome utility? Could it be extended to work with extra props on classes? I guess it would be hard to tell if the classes are being used or not. Either way :fallen_leaf:

test/utils/getUnhandledPops-test.js
```diff
@@ -0,0 +1,33 @@
...
+  it('removes props defined in propTypes', () => {
+    TestClass.propTypes = {imHandled: 'thanks'};
+    new TestClass()
+      .unhandledProps
+      .should.not.have.any.keys(_.keys(TestClass.defaultProps));
+  });
+  it('leave props not in defaultProps || propTypes in tact', () => {
+    TestClass.defaultProps = {imHandled: 'thanks'};
+    TestClass.propTypes = {alsoHandled: 'got it'};
+    const props = {thisShould: 'still be here'};
```
athurman, line 28: Just a quick question here as to why you chose to make this variable a `const` as opposed to a `let`? Is it solely because this variable is meant to be read only? Or another, more specific reason?

:smiley: Just curious

athurman:
:ocean: :octopus: Looks super handy

test/utils/getUnhandledPops-test.js
levithomason, line 28: Good question!  We've updated the linter.  If you make a variable whose value never changes, it will tell you to make it a const instead.  The only use case for a `let` now is for a variable whose value changes.  Example:

```
let hasCertainChild;
React.Children.forEach(child => {
  if (someCondition(child) {
    hasSomeChild = true;
  }
})
```

In this case, the var is initialized as `undefined` and then set to `true` if the condition is met.

Going a bit further, a const wouldn't work here for both of those reasons.  It cannot be initialized with an empty value nor can it's value change.

levithomason:
@eanplatter yes this util works with classes.  Classes are just functions with a new name.  The important thing is that the util looks for static properties `SomeClass.propTypes` and `SomeClass.defaultProps` to do it's magic.  It finds these via `this.constructor.<foo>`.

These are equivalent and all produce objects like the above:

``` jsx
// ES5
function SomeClass() {}

// ES6
class SomeClass {}
SomeClass.propTypes = {};
SomeClass.defaultProps = {};

// ES7
class SomeClass {
  static propTypes = {};
  static defaultProps = {};
}
```

Finally, whether or not the class `extends` anything is irrelevant.  Just so long as it has the right static props.

PR #100, opened 2015-10-27
Feature/list examples

kyleturco:


docs/app/Examples/elements/List/Types/ListListExample.js
```diff
@@ -0,0 +1,14 @@
...
+  render() {
+    return (
+      <List>
+        <ListItem description='Apples' />
+        <ListItem description='Pears' />
+        <ListItem description='Oranges' />
+      </List>
+    );
+  }
+}
```
kyleturco, line 14: Only did one example for the basic list - Semantic has a handful, would adding some variations be useful or necessary?

docs/app/Examples/elements/List/Content/ListDescriptionExample.js
```diff
@@ -0,0 +1,15 @@
...
+import {List, ListItem} from 'stardust';
+
+export default class ListDescriptionExample extends Component {
+  render() {
+    const mapIcon = <i className='map marker icon' />;
+
+    return (
+      <List>
+        <ListItem image={mapIcon} header='Chicago' description='This city is located in the state of Illinois' />
+        <ListItem image={mapIcon} header='Nashville' description='This city is located in the state of Tennessee' />
```
levithomason: Use the `icon` prop to pass in an icon.  The `image` prop is for passing in a constructed `<Image />`.  I'll add an issue for validation on this.

**EDIT**
I added #102 for this fix.  The reasoning for the different props (image/icon) is that we will support passing a string to the `image` prop and use it as `src` to construct an image inside the ListItem.  Technically, both of these props place the content in the same place for now.

**EDIT 2**
#102 lists a number of issues that affect my previous edit.  Let's make a separate PR that updates the ListItem, tests, and examples and leave this one as is.

docs/app/Examples/elements/List/Content/ListIconExample.js
```diff
@@ -0,0 +1,27 @@
...
+
+    return (
+      <List>
+        <ListItem
+          image={helpIcon}
+          header='Floated Icon'
+          description='This text will always have a left margin so it sits alongside the icon'
+        />
+        <ListItem
+          image={triangleIcon}
```
levithomason: Use icon props here, I won't comment on further icon/image props.

docs/app/Examples/elements/List/Content/ListIconExample.js
```diff
@@ -0,0 +1,27 @@
...
+          header='Icon Alignment'
+          description='Floated icons are by default top aligned'
+        />
+        <ListItem image={helpIcon}>
+          Inline Text
```
levithomason: Nice job including this example!  I was hoping to see it.

I think we should use React `children` terminology to describe this usage.  Also, I'd like to place it right at the top.  How about duplicating the very first example set, but with a description that says something like, "<code>ListItems</code> also render <code>children</code> as their description."?

docs/app/Examples/elements/List/Content/ListImageExample.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+import {List, ListItem} from 'stardust';
+
+export default class ListImageExample extends Component {
+  render() {
+    const helenAvatar = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/1' />;
```
levithomason: There is an `<Image />` component, let's use only Stardust components if we can.  Let's to do this for all the examples that use images.

docs/app/Examples/elements/List/Types/ListListExample.js
levithomason, line 14: I see what you mean.  It looks like they just show a mashup of various lists at the top, then dive into each specific capability.  I think we're good with the basics up top if we have links/icons/etc down below.

docs/app/Examples/elements/List/Types/ListOrderedExample.js
```diff
@@ -0,0 +1,21 @@
...
+    return (
+      <List className='ordered'>
+        <ListItem description='Apples' />
+          <div>
+            <List>
+              <ListItem description='Fuji' />
+              <ListItem description='Granny Smith' />
+              <ListItem description='Honeycrisp' />
+            </List>
+          </div>
```
levithomason: Let's remove the extra div here.  Turns out there is an issue with the `ListItem` component.  It currently renders `children || description` as the content.  I'll open up a separate issue for the `ListItem` as there are number of fixes it needs.

**EDIT**
I updated #102 to include this fix, we can leave this div for this PR.

docs/app/Examples/elements/List/Variations/ListAnimatedExample.js
```diff
@@ -0,0 +1,22 @@
...
+export default class ListAnimatedExample extends Component {
+  render() {
+    const avatar1 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/1' />;
+    const avatar2 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/2' />;
+    const avatar3 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/3' />;
+    return (
+      <div>
+        <Message className='info'>
+        Be sure content can fit on one line, otherwise text content will reflow when hovered.
+        </Message>
```
levithomason: This file is rendered on the doc site under the example.  To keep it clean, put the `<Message />` as a child of the `< ComponentExample />` in `ListVariationsExamples.js`.  We can also then remove the extra wrapping `div`.

The `<ComponentExample />` was extended to render children for this purpose.

docs/app/Examples/elements/List/Variations/ListDividedExample.js
```diff
@@ -0,0 +1,18 @@
+import React, {Component} from 'react';
+import {List, ListItem} from 'stardust';
+
+export default class ListDividedExample extends Component {
+  render() {
+    const avatar1 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/1' />;
+    const avatar2 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/2' />;
+    const avatar3 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/3' />;
```
levithomason: We have [Faker](https://github.com/marak/Faker.js/) installed for random data. Let's import it and use it for the image `src={faker.internet.avatar()}` for consistency and brevity.  As a vain bonus, we get random avatars on every render.

docs/app/Examples/elements/List/Variations/ListRelaxedExample.js
```diff
@@ -0,0 +1,17 @@
...
+  render() {
+    const avatar1 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/1' />;
+    const avatar2 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/2' />;
+    const avatar3 = <img className='ui avatar image' src='http://lorempixel.com/100/100/people/3' />;
+    return (
+      <List className='relaxed'>
+        <ListItem image={avatar1} header='Helen' />
+        <ListItem image={avatar2} header='Christian' />
+        <ListItem image={avatar3} header='Daniel' />
+      </List>
```
levithomason, line 15: Might be worth adding a `very relaxed` example as well.

docs/app/Examples/elements/List/Types/ListListExample.js
kyleturco, line 14: Sounds good  👍

docs/app/Examples/elements/List/Content/ListIconExample.js
kyleturco: I might need a little clarification what you mean here

docs/app/Examples/elements/List/Content/ListImageExample.js
kyleturco: Using the `<Image />` component makes them weirdly lopsided. I'm trying to figure it out but I thought I'd post it here if you have any ideas
screen shot 2015-10-30 at 4 51 03 pm

test/utils/React-util.js
```diff
@@ -19,7 +19,7 @@ import {
 /**
  * Render a ReactElement and use helpful test utility methods against it.
  * @param {ReactElement} reactElement The element to render.
- * @returns {RenderedTree}
+ * @returns {RenderedTree} The return
```
kyleturco: Not sure the deal here - for whatever reason the file threw a pre-commit error saying "Missing JSDoc return description". I added some text on the return line, and it seemed happy with that. Let me know if this needs an alternative solution.

levithomason: There is an eslint rule called [`valid-jsdoc`](https://github.com/eslint/eslint/blob/master/docs/rules/valid-jsdoc.md) that validates those `/**...*/` docblocks.  We don't want to require return descriptions, so I'll undo this change and add an override rule in the `test/.eslint` config to turn it off.

``` json
{
  "extends": "ta-webapp/tests",
  "globals": {
    "render": false
  },
  "rules": {
    "valid-jsdoc": [2, {
      "requireReturn": false,
      "requireReturnDescription": false,
    }]
  }
}
```

The only part added there ^ is the `"rules"` object.  By specifying rules there, we are overriding the rules we are extending in the `"extends"` section (our defaults).

levithomason:
Since I rebased this guy and updated it, would you check it out and make sure it still looks sane.  If so, it is good to merge!  :accept:

src/elements/Image/Image.js
```diff
@@ -24,9 +24,7 @@ export default class Image extends Component {
     );
 
     return (
-      <div className={classes}>
-        <img {...this.props} />
-      </div>
+      <img {...this.props} className={classes} />
```
athurman: since you are returning this one element, the parentheses are not necessary and you can move the image up to `line 26`.

.eslintrc
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": "ta-webapp",
+  "extends": "ta-webapp"
```
athurman: did this break something? As I thought we were leaving trailing commas in objects with ES6.

athurman:
1 question, 1 comment, overall good to merge!! :octopus:

.eslintrc
levithomason: In JSON (which an `.eslintrc` file _looks_ like), a trailing comma will break things.  I'm indifferent here since this looks like a JSON file but is apparently parsed with a YAML parser, so the trailing comma is valid.  I tried this and it is also valid:

```
{
  extends: ta-webapp
}
```

This is neither valid JSON nor JS but the YAML parser can handle it.  Let's leave trailing commas in for consistency moving forward and just leave this as is for now.

src/elements/Image/Image.js
levithomason: Nice catch, will update and merge.

PR #103, opened 2015-10-30
Feature/divider component

kyleturco:
Add a `<Divider />` component and examples for the divider. (Note: currently failing the Wercker test as a result of the Babel update)

### DIFF
--- a/docs/app/Examples/elements/Divider/DividerExamples.js
+++ b/docs/app/Examples/elements/Divider/DividerExamples.js
@@ -2,7 +2,7 @@ import React, {Component} from 'react';
 import DividerTypesExamples from './Types/DividerTypesExamples';
 import DividerVariationsExamples from './Variations/DividerVariationsExamples';
 
-export default class ButtonExamples extends Component {
+export default class DividerExamples extends Component {
   render() {
     return (
       <div>
--- a/docs/app/Examples/elements/Divider/DividerExamples.js
+++ b/docs/app/Examples/elements/Divider/DividerExamples.js
@@ -1,0 +1,14 @@
+import React, {Component} from 'react';
+import DividerTypesExamples from './Types/DividerTypesExamples';
+import DividerVariationsExamples from './Variations/DividerVariationsExamples';
+
+export default class ButtonExamples extends Component {
+  render() {
+    return (
+      <div>
+        <DividerTypesExamples />
+        <DividerVariationsExamples />
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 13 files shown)


docs/app/Examples/elements/Divider/Types/DividerVerticalExample.js
```diff
@@ -0,0 +1,28 @@
...
+        <Divider className='vertical'>And</Divider>
+        <div className='column'>
+          <Segment className='basic'>
+            Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec odio.
+          </Segment>
+        </div>
+      </div>
+    );
+  }
+}
```
kyleturco, line 28: Does SD have a `<Column>` component? I used regular `<div>` and set the className to column in this example. We do need the columns though to get the proper formatting to show the vertical dividers

docs/app/Examples/elements/Divider/DividerExamples.js
```diff
@@ -0,0 +1,14 @@
+import React, {Component} from 'react';
+import DividerTypesExamples from './Types/DividerTypesExamples';
+import DividerVariationsExamples from './Variations/DividerVariationsExamples';
+
+export default class ButtonExamples extends Component {
```
levithomason: Class name >.<

kyleturco: Oh no, it never stops

docs/app/Examples/elements/Divider/Types/DividerVerticalExample.js
```diff
@@ -0,0 +1,28 @@
+import React, {Component} from 'react';
+import {Segment, Divider} from 'stardust';
+
+export default class DividerVerticalExample extends Component {
+  render() {
+    return (
+      <div className='ui three column relaxed grid'>
+        <div className='column'>
```
levithomason: We can use Stardust Grid and Columns here.

docs/app/Examples/elements/Divider/Types/DividerVerticalExample.js
levithomason, line 28: Oops, comments on this above before seeing your comment.  Yep, just add `Grid, Column` to the import from stardust.  Like other components, they include all the necessary base classes so you only need to add modifier classes (like `three column relaxed` to the `<Grid />`).

levithomason:
After updating the one class name and using `Grid` and `Column` components, :ship: !

docs/app/Examples/elements/Divider/Types/DividerVerticalExample.js
kyleturco, line 28: Roger that, both `Column` and `Grid` work 👍

PR #113, opened 2015-11-18
Segment Examples

### LINKED ISSUE #112: Segment examples
http://semantic-ui.com/elements/segment.html

Add examples to Stardust for Segment control.

### LINKED ISSUE #114: <Segments /> component
To wrap grouped `Segment` controls

ghost:
Adds Semantic UI `Segment` examples to the Stardust repo. Also creates a `Segments` component which works in concert with `Segment` to provide grouping behavior. `Segments` component is considered a child of `Segment` from a documentation purpose and so it does not appear in the list of components, but will appear in the examples.

screen shot 2015-11-24 at 2 44 56 pm

Resolves #112 and closes #114. Discovers #116 and #120.

### DIFF
--- a/docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
+++ b/docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
@@ -1,23 +1,23 @@
 import React, {Component} from 'react';
-import {Segment, Segments} from 'stardust';
+import {Segment} from 'stardust';
 
 export default class SegmentNestedSegmentsExample extends Component {
   render() {
     return (
-      <Segments>
+      <Segment.Segments>
         <Segment>Top</Segment>
-        <Segments>
+        <Segment.Segments>
           <Segment>Nested Top</Segment>
           <Segment>Nested Middle</Segment>
           <Segment>Nested Bottom</Segment>
-        </Segments>
-        <Segments className='horizontal'>
+        </Segment.Segments>
+        <Segment.Segments className='horizontal'>
           <Segment>Top</Segment>
           <Segment>Middle</Segment>
           <Segment>Bottom</Segment>
-        </Segments>
+        </Segment.Segments>
         <Segment>Bottom</Segment>
-      </Segments>
+      </Segment.Segments>
     );
   }
 }
--- a/docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
+++ b/docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
@@ -1,0 +1,23 @@
+import React, {Component} from 'react';
+import {Segment, Segments} from 'stardust';
+
+export default class SegmentNestedSegmentsExample extends Component {
+  render() {
+    return (
+      <Segments>
+        <Segment>Top</Segment>
+        <Segments>
+          <Segment>Nested Top</Segment>
+          <Segment>Nested Middle</Segment>
+          <Segment>Nested Bottom</Segment>
+        </Segments>
+        <Segments className='horizontal'>
+          <Segment>Top</Segment>
+          <Segment>Middle</Segment>
+          <Segment>Bottom</Segment>
+        </Segments>
+        <Segment>Bottom</Segment>
+      </Segments>
+    );
+  }
+}
--- a/docs/app/Examples/elements/Segment/SegmentExamples.js
+++ b/docs/app/Examples/elements/Segment/SegmentExamples.js
@@ -1,0 +1,16 @@
+import React, {Component} from 'react';
+import SegmentTypesExamples from './Types/SegmentTypesExamples';
+// import SegmentStatesExamples from './Types/SegmentStatesExamples';
+// import SegmentGroupsExamples from './Content/SegmentGroupsExamples';
+// import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
+
+export default class SegmentExamples extends Component {
+  render() {
+    return (
+      <div>
+        <SegmentTypesExamples />
+
+      </div>
+    );
+  }
+}
--- a/docs/app/Examples/elements/Segment/SegmentExamples.js
+++ b/docs/app/Examples/elements/Segment/SegmentExamples.js
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import SegmentTypesExamples from './Types/SegmentTypesExamples';
 import SegmentGroupsExamples from './Groups/SegmentGroupsExamples';
-// import SegmentStatesExamples from './Types/SegmentStatesExamples';
+import SegmentStatesExamples from './States/SegmentStatesExamples';
 // import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
 
 export default class SegmentExamples extends Component {
@@ -10,7 +10,7 @@ export default class SegmentExamples extends Component {
       <div>
         <SegmentTypesExamples />
         <SegmentGroupsExamples />
-
+        <SegmentStatesExamples />
       </div>
     );
   }
--- a/docs/app/Examples/elements/Segment/SegmentExamples.js
+++ b/docs/app/Examples/elements/Segment/SegmentExamples.js
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import SegmentTypesExamples from './Types/SegmentTypesExamples';
+import SegmentGroupsExamples from './Groups/SegmentGroupsExamples';
 // import SegmentStatesExamples from './Types/SegmentStatesExamples';
-// import SegmentGroupsExamples from './Content/SegmentGroupsExamples';
 // import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
 
 export default class SegmentExamples extends Component {
@@ -9,6 +9,7 @@ export default class SegmentExamples extends Component {
     return (
       <div>
         <SegmentTypesExamples />
+        <SegmentGroupsExamples />
 
       </div>
     );
--- a/docs/app/Examples/elements/Segment/SegmentExamples.js
+++ b/docs/app/Examples/elements/Segment/SegmentExamples.js
@@ -2,7 +2,7 @@ import React, {Component} from 'react';
 import SegmentTypesExamples from './Types/SegmentTypesExamples';
 import SegmentGroupsExamples from './Groups/SegmentGroupsExamples';
 import SegmentStatesExamples from './States/SegmentStatesExamples';
-// import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
+import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
 
 export default class SegmentExamples extends Component {
   render() {
@@ -11,6 +11,7 @@ export default class SegmentExamples extends Component {
         <SegmentTypesExamples />
         <SegmentGroupsExamples />
         <SegmentStatesExamples />
+        <SegmentVariationsExamples />
       </div>
     );
   }
--- a/docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
+++ b/docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
@@ -1,7 +1,7 @@
 import React, {Component} from 'react';
 import {Segment} from 'stardust';
 
-export default class SegmentVerticalExample extends Component {
+export default class SegmentVerticalSegmentExample extends Component {
   render() {
     return (
       <Segment>
--- a/docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
+++ b/docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
@@ -1,0 +1,14 @@
+import React, {Component} from 'react';
+import {Segment} from 'stardust';
+
+export default class SegmentVerticalExample extends Component {
+  render() {
+    return (
+      <Segment>
+        <Segment className='vertical'>Te eum doming eirmod, nominati pertinacia argumentum ad his.</Segment>
+        <Segment className='vertical'>Pellentesque habitant morbi tristique senectus.</Segment>
+        <Segment className='vertical'>Eu quo homero blandit intellegebat. Incorrupte consequuntur mei id.</Segment>
+      </Segment>
+    );
+  }
+}
--- a/docs/app/Examples/elements/Segment/Variations/SegmentCompactGroupExample.js
+++ b/docs/app/Examples/elements/Segment/Variations/SegmentCompactGroupExample.js
@@ -1,17 +1,17 @@
 import React, {Component} from 'react';
-import {Segment, Segments} from 'stardust';
+import {Segment} from 'stardust';
 
 export default class SegmentCompactGroupExample extends Component {
   render() {
     return (
-      <Segments className='compact'>
+      <Segment.Segments className='compact'>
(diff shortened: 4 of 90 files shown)


src/elements/Segment/Segments.js
```diff
@@ -0,0 +1,46 @@
...
+    className: PropTypes.string,
+  };
+
+  componentDidMount() {
+    const {children} = this.props;
+
+    invariant(
+      !_.any( Children.map(children, child => child.type !== Segment) ),
+      'May only contain children of type Segment.',
+    );
```
ghost: @levithomason this would be our first use of `fbjs` to throw a possible `invariant` violation in this repo. The intent is to help prevent misuse of the components. Still need to add support for nested groups. (Note: the second argument to `invariant` is optional and would shed some weight during production builds if omitted.)

What's weird though is when I tried to create a simple `invariant` for `Children(count) >= 2` - I ran into issues with conformance tests.

If we like the pattern we could also consider adding a runtime check for number of children, though that's something we may want to pair on given my failed attempts.

Alternatively we could try an approach more like this perhaps?

``` jsx
static propTypes = {
  children: PropTypes.oneOfType([
    PropTypes.oneOfType([
      PropTypes.instanceOf(Segment),
      PropTypes.instanceOf(this),
    ]),
    PropTypes.arrayOf(PropTypes.oneOfType([
      PropTypes.instanceOf(Segment),
      PropTypes.instanceOf(this),
    ])),
  ]).isRequired,
  className: PropTypes.string,
};
```

src/elements/Segment/Segments.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  static _meta = {
+    library: META.library.semanticUI,
+    name: 'Segments',
+    type: META.type.element,
+    parent: 'Segment'
+  };
```
ghost, line 23: Impressed the linter runs with a precommit hook. This was moved down to satisfy the `react/sort-comp` rule. Arguably we should update the linter rule so static assignments like this can be kept at the top for visual consistency between modules and to keep instance and static properties grouped at the top of the object, as done when creating class diagrams for example.

test/specs/elements/Segment/Segments-test.js
```diff
@@ -0,0 +1,45 @@
...
+      component.querySelectorAll('.sd-segment').length
+    ).to.equal(3);
+  });
+
+  it('only allows children of type Segment', () => {
+    expect(global.shallowRender(
+      <Segments>Some text</Segments>
+    )).to.throw;
+
+    // FIXME: Figure out why this is giving false positives
```
ghost: Sadly I was unable to get this to work. Let's figure out what to do with this depending on what we decide regarding the introduction of `invariant` calls from the `fbjs` lib.

docs/app/Examples/elements/Segment/SegmentExamples.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+import SegmentTypesExamples from './Types/SegmentTypesExamples';
+import SegmentGroupsExamples from './Groups/SegmentGroupsExamples';
+// import SegmentStatesExamples from './Types/SegmentStatesExamples';
+// import SegmentVariationsExamples from './Variations/SegmentVariationsExamples';
```
ghost: There's still a little work to be done here, but at this point I'd like to get the introduction of the `Segments` component nailed down before going too much further.

src/elements/Segment/Segments.js
```diff
@@ -0,0 +1,46 @@
...
+import META from 'src/utils/Meta';
+import Segment from './Segment';
+
+export default class Segments extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+  };
+
+  componentDidMount() {
```
ghost: Tried to have this live in `componentWillMount` but ran into trouble getting tests to pass conformance tests.

src/elements/Segment/Segments.js
ghost: In attempting to allow `Segments` to support nested `Segments` I tried the propTypes approach, as well as the following without luck:

```
invariant(
  !_.any(Children.map(children, child => {
    return _.includes( ['Segment', 'Segments'], _.get(child, 'type._meta.name') );
  })),
  'May only contain children of type Segment or Segments.',
);
```

As a result, I'm going to backpedal to `children: PropTypes.node` for now, and remove the `fbjs` and `invariant` calls. The net result is the component will accept any children until we're able to figure out a method which works.

src/utils/customPropTypes.js
```diff
@@ -1,11 +1,36 @@
...
+   * Verifies exclusivity of a given prop
+   * @param {string} exclusives property used for exclusivity check
+   * @throws {Error} if named prop not mutually exclusive
+   * @returns no-op
+   */
   mutuallyExclusive: exclusives => {
     return (props, propName, componentName) => {
       _.forEach(exclusives, exclusiveProp => {
         if (props[propName] && props[exclusiveProp]) {
-          throw new Error(
```
ghost, line 8: Per [Reusable Components](https://facebook.github.io/react/docs/reusable-components.html) doc we should not throw, but `return` errors from custom validators.

docs/app/Examples/elements/Segment/Variations/SegmentVariationsExamples.js
```diff
@@ -0,0 +1,98 @@
...
+          examplePath='elements/Segment/Variations/SegmentInvertedExample'
+        />
+        <ComponentExample
+          title='Attached'
+          description='A segment can be attached to other content on a page.'
+          examplePath='elements/Segment/Variations/SegmentAttachedExample'
+        >
+          <Message className='info'>
+            Attached segments are designed to be used with other <code>attached</code> variations like
+            the <a href='#Header-Variations-HeaderAttachedExample'>attached header</a> or <i>attached messages</i>.
```
levithomason, line 22: Nice, thanks for the anchor link here!

docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
```diff
@@ -0,0 +1,23 @@
...
+        <Segments className='horizontal'>
+          <Segment>Top</Segment>
+          <Segment>Middle</Segment>
+          <Segment>Bottom</Segment>
+        </Segments>
+        <Segment>Bottom</Segment>
+      </Segments>
+    );
+  }
+}
```
kyleturco, line 23: Posing a question here - would it make sense here for the text to be taken in as a child and have all these be self closing? As in - `<Segment caption="Nested Top" />`. From a readability standpoint, it feels a bit cluttered.

docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
```diff
@@ -0,0 +1,14 @@
+import React, {Component} from 'react';
+import {Segment} from 'stardust';
+
+export default class SegmentVerticalExample extends Component {
```
kyleturco: Tiny difference between file & class name

docs/app/Examples/elements/Segment/Variations/SegmentCompactGroupExample.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+import {Segment, Segments} from 'stardust';
+
+export default class SegmentCompactGroupsExample extends Component {
```
kyleturco: Another tiny difference between file & class name, just a bonus 's' 😉

docs/app/Examples/elements/Segment/Groups/SegmentNestedSegmentsExample.js
kyleturco, line 23: I can see where that could get pretty unwieldy looking at the examples below with longer lines of text.

ghost, line 23: Agreed on the clutter. I'm not sure how many IRL use cases we'll have were we'll be using a simple text body. Regardless I'm happy to make a prop for text-only tag body content.

kyleturco, line 23: Yeah that's what I was wondering. The docs examples definitely have situations that rarely happen in the wild, so this might not be a big issue.

src/collections/Table/Table.js
```diff
@@ -1,23 +1,13 @@
...
-      Children.forEach(props.children, (child) => {
-        if (child.type !== TableColumn) {
-          isValid = false;
-        }
-      });
-      if (!isValid) {
-        return new Error('`Table` must only have `TableColumn` children.');
-      }
-    },
+    children: childrenOfTypesOnly(['TableColumn']),
```
levithomason: Our previous conversation about this type check only working on children was misguided.  After finding [this React issue](https://github.com/facebook/react/issues/2979), it seems the reason `instanceOf` doesn't work is because it is strictly doing `x instanceof y`.  However, when creating a class that extends `React.Component` instanceof checks apparently aren't going to work for things like, `TableColumn`.

**TL;DR**
I think a more appropriate name for this util is something like `componentTypes()` or `componentNames()` and drop the `children` reference.  Since, this is in fact what we are checking and the util will work for all props, not just `children`.

src/elements/Segment/Segment.js
```diff
@@ -2,10 +2,22 @@ import React, {Component, PropTypes} from 'react';
...
+ */
 export default class Segment extends Component {
   static propTypes = {
+    /**
+     * Segment tag body content.
+     */
     children: PropTypes.node,
+    /**
+     * Class name for custom styling.
+     */
```
levithomason: I'm torn on this particular doc block.  Partly because I think it is common knowledge.  Mostly because it seems if we add this one to `className` then we should add the same to all `className` props for all components.  If we did this, would suggest doing it in `ComponentProps.js`, where the prop types doc section is generated.  Then all components can have the same default doc string added in one place.  We could do this for `className` `style` `children`, etc.  Then, save component level doc strings for component specific things that folks aren't expected to know, like `heading`.

Thoughts?

src/collections/Table/Table.js
jhabdas: What's strange is the issue is closed but things seem not to work as expected. I know FB is inching away from `createClass`, yet most of the docs suggest many of the prop validators don't function otherwise (i.e. under ES6) classes. Though in my work on React Native I hadn't done much more complex than, for example:

src/elements/Segment/Segments.js
```diff
@@ -0,0 +1,44 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+import META from 'src/utils/Meta';
+import {childrenOfTypesOnly} from '../../utils/customPropTypes';
```
levithomason: Our app allows importing from the project root.  Suggest refactoring this util import to `src/utils/customPropTypes`.

**EDIT**
You can see the Webpack config for this [here](https://github.com/TechnologyAdvice/stardust/blob/master/webpack.dev.babel.js#L53)

src/collections/Table/Table.js
jhabdas: How about `ofTypes` for brevity? We could go with the singular if we want to accept simple string literals in addition to arrays as args too.

src/elements/Segment/Segment.js
ghost: All for DRY. If we move the doc upwards would we want to have the ability for doc overrides when specificity for a component is desired?

src/elements/Segment/Segments.js
ghost: :+1:

src/elements/Segment/Segments.js
levithomason, line 23: +1 to linter updates for sanity.  Could you add this to the Pivotal story for our organization lint update?  You should have access very soon :)

levithomason, line 23: Done.

src/index.js
```diff
@@ -68,6 +69,7 @@ export default {
   List,
   ListItem,
   Segment,
+  Segments,
```
levithomason, line 73: As of [this `Statistic` PR conversation](https://github.com/TechnologyAdvice/stardust/pull/117/files#r45690951), we've begun adding the plural grouping class as a sub-component of the singular component.  Have a look there and let me know what you think.

src/utils/customPropTypes.js
levithomason, line 8: Nice catch, thanks!

src/utils/customPropTypes.js
```diff
@@ -1,11 +1,36 @@
...
+   */
+  childrenOfTypesOnly: (allowedTypes) => {
+    return (props, propName, componentName) => {
+      const {children} = props;
+      const disallowed = Children.map(children, child => {
+        return _.includes(allowedTypes, _.get(child, 'type._meta.name')) ? null : child;
+      });
+      if (disallowed && disallowed.length !== 0) {
+        return new Error(
+          `\`${componentName}\` should only have children of type \`${allowedTypes}\`.`);
```
levithomason: Looks like the closing paren should be on the next line aligned with the `return`, or the entire error on one line.

test/specs/elements/Segment/Segments-test.js
```diff
@@ -0,0 +1,34 @@
...
+      </Segments>
+    ).scryClass('sd-segment');
+
+    segmentOne
+      .textContent
+      .should.equal('Top');
+    segmentTwo
+      .textContent
+      .should.equal('Bottom');
+  });
```
levithomason, line 19: Nice test!

test/specs/elements/Segment/Segments-test.js
```diff
@@ -0,0 +1,34 @@
...
+        <Segment>Top</Segment>
+        <Segment>Middle</Segment>
+        <Segment>Bottom</Segment>
+      </Segments>
+    ).scryClass('sd-segments');
+
+    expect(
+      component.querySelectorAll('.sd-segment').length
+    ).to.equal(3);
+  });
```
levithomason, line 31: Since we are testing how many `.sd-segment` elements there are, we can forgo accessing them through the `.sd-segments` element and directly test the count:

``` jsx
it('renders expected number of children', () => {
  render(
    <Segments>
      <Segment>Top</Segment>
      <Segment>Middle</Segment>
      <Segment>Bottom</Segment>
    </Segments>
  )
    .scryClass('sd-segment')
    .should.have.a.lengthOf(3);
});
```

> Makes use of chai [length/lengthOf](http://chaijs.com/api/bdd/#length-section) assertion.

src/elements/Segment/Segment.js
levithomason: Agreed, something like (pseudo code): `doc  = componentLevelDocBlock || defaultDocBlock` for each prop type.

ghost: Will do before merging.

test/specs/elements/Segment/Segments-test.js
ghost, line 31: Worked great! Thanks for the hot tip!

src/index.js
ghost, line 73: I like it. Consider it done before merge.

ghost:
All review feedback incorporated. While I was in there I also refactored `Statistic` to use the new custom propType created with this PR. In the refactored code we now have three `import` statements dipping into `src/utils`. If this gets unwieldy as more utils are introduced in the future we may want to look at allowing destructuring on `src/utils`.

PR #121, opened 2015-11-25
Add Progress Component

eanplatter:


levithomason:
@eanplatter where are we at on this guy, look like it needs a push from your local?

eanplatter:
@levithomason this is up to date, I am not sure why it says those commits are 9 days old. It has the updated meta and tests.

test/specs/modules/Progress/Progress-test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import {Progress} from 'stardust';
+
+describe('Progress', () => {
+  it('should be able to receive children', () => {
+    render(
+      <Progress>
+        <div className='new-child' />
+      </Progress>
+    ).scryClass('new-child');
```
eanplatter: Testing that it picks up it's children.

test/specs/modules/Progress/Progress-test.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+  it('should create a div with the class of bar', () => {
+    render(<Progress />).scryClass('bar');
+  });
```
eanplatter, line 21: Testing that it has a child with class of bar

test/specs/modules/Progress/Progress-test.js
```diff
@@ -0,0 +1,18 @@
...
+  it('should create a div with the class of progress', () => {
+    render(<Progress />).scryClass('progress');
+  });
+});
```
eanplatter, line 28: Checking that it has child with class of progress.

test/mocks/SemanticjQuery-mock.js
```diff
@@ -22,6 +22,7 @@ const jQueryPlugins = {
   dropdown: sandbox.stub().returnsThis(),
   modal: sandbox.stub().returnsThis(),
   popup: sandbox.stub().returnsThis(),
+  progress: sandbox.stub().returnsThis(),
```
eanplatter, line 25: Studding out since Progress is a stardust module.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+    onError: PropTypes.func,
+    onSuccess: PropTypes.func,
+    onWarning: PropTypes.func,
+    percent: PropTypes.number,
+    precision: PropTypes.number,
+    random: PropTypes.bool,
+    showActivity: PropTypes.bool,
+    total: PropTypes.bool,
+    value: PropTypes.bool,
+  };
```
eanplatter, line 25: These prop values were taken from semantic ui's progress docs:http://semantic-ui.com/modules/progress.html#/settings

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+    this.element = $(this.refs.element);
+    this.element.progress({
+      onActive: this.props.onActive,
+      onChange: this.props.onChange,
+      onError: this.props.onError,
+      onSuccess: this.props.onSuccess,
+      onWarning: this.props.onWarning,
+      percent: this.props.percent,
+    });
+  }
```
eanplatter, line 45: These are also taken from the stardust docs, and represent possible callbacks and plugin properties exposed to the user: http://semantic-ui.com/modules/progress.html#/usage

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+
+  static _meta = {
+    library: META.library.stardust,
+    name: 'Progress',
+    type: META.type.module,
+  };
```
eanplatter, line 51: Meta info for the component.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+
+  plugin() {
+    return this.element.progress(...arguments);
+  }
```
eanplatter, line 55: Exposing the plugin property.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+
+  render() {
+    const classes = classNames(
+      'sd-progress',
+      'ui',
+      this.props.className,
+      'progress',
+    );
```
eanplatter, line 86: configuring the classes.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,65 @@
...
+
+    return (
+      <div {...this.props} className={classes}>
+        <div className='bar'>
+          <div className='progress'/>
+        </div>
+        {this.props.children}
+      </div>
```
eanplatter, line 92: The markup as shown in semantic ui's documentation: http://semantic-ui.com/modules/progress.html#/definition

.gitignore
```diff
@@ -12,3 +12,4 @@ bower_components/
 testBundle.js
 docs/build
 docgenInfo.json
+.idea/
```
levithomason: Use [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for the `.idea/`directory.  This way not all projects have to have this entry, also not all developers will use a JetBrains product so this ignore only applies to a subset of users.

levithomason: Also, add `.DS_Store` to your global ignore.  Then remove it from this PR.

src/modules/Progress/Progress.js
levithomason, line 45: All the settings accepted as props should be passed to the plugin here.  Otherwise, we are defining props that are never used.  For instance:

``` jsx
<Progress total={1} value={0.2} />
```

This should do `this.element.progress({total: 1, value: 0.2})` on mount.  Right now, the component accepts the props, validates them, then does nothing with them.  So with the above example nothing will happen on mount.

If we instead pass all prop settings to the plugin, then it would mount and animate to 20% complete.

src/modules/Progress/Progress.js
levithomason, line 92: Looks like if there are children here they should be wrapped in a label `div`:

``` jsx
<div className='lablel'>{this.props.children}</div>
```

Make sure when there are no children there is not an empty label div present.

http://semantic-ui.com/modules/progress.html#label

levithomason, line 92: When the `attached` class is present, there should only be a `bar` div without a `progress` child div:

``` jsx
<div className='ui segment'>
  <div className='ui top attached progress'>
    <div className='bar'></div>
  </div>
</div>
```

http://semantic-ui.com/modules/progress.html#attached

test/specs/modules/Progress/Progress-test.js
levithomason: `scry` methods return arrays but do not throw if there are 0 results.  So, this test has no assertion it will always pass.  You can change it to `findClass` which will throw an error if it finds 0, or more than 1.  Or, we've added this `assertText` which does a simple assertion that text is present (text is still a child):

``` jsx
render(<Progress>child</Progress>)
  .assertText('child');
```

test/specs/modules/Progress/Progress-test.js
levithomason, line 28: same testing notes

test/specs/modules/Progress/Progress-test.js
```diff
@@ -0,0 +1,20 @@
...
+    ).assertText('Child');
+  });
+
+  it('should create a div with the class of bar', () => {
+    render(<Progress />).findClass('bar');
+  });
+
+  // it('should create a div with the class of progress', () => {
+  //   render(<Progress />).firstClass('progress');
+  // });
```
eanplatter: I was having some trouble understanding the React-utils. This test was failing because it found two `progress` classes on `<Progress />`. I'd be interested in pairing on it.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+
+  plugin() {
+    return this.element.progress(...arguments);
+  }
+
+  renderAttachedBar() {
+    return (
+      <div className='bar' />
+    );
+  }
```
eanplatter: The markup for an attached progress bar.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+      <div>
+        <div className='bar'>
+          <div className='progress'/>
+        </div>
+        <div className='label'>
+          {this.props.children}
+        </div>
+      </div>
+    );
+  }
```
eanplatter: The markup for a normal progress bar.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+      content = ::this.renderAttachedBar();
+    }
+
+    return (
+      <div {...this.props} className={classes}>
+        {content}
+      </div>
+    );
+  }
+}
```
eanplatter, line 95: I felt like this was kind of messy, it was like, iteration number three though, so I decided to not waste anymore time and just try to get some feedback. Essentially just want to check if the `attached` class exists, and if so render the proper progress bar.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+      onSuccess: this.props.onSuccess,
+      onWarning: this.props.onWarning,
+      percent: this.props.percent,
+      precision: this.props.precision,
+      random: this.props.random,
+      showActivity: this.props.showActivity,
+      total: this.props.total,
+      value: this.props.value,
+    });
+  }
```
eanplatter, line 45: Moved all props to the `this.element.progress` method.

test/specs/modules/Progress/Progress-test.js
ghost: Me too! I'm still getting adjusted to the test utils and am interested in a pair here.

src/modules/Progress/Progress.js
ghost, line 25: Any of these `props` required? If so, can we add the `.isRequired` to help with prop management/usage?

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+    precision: PropTypes.number,
+    random: PropTypes.bool,
+    showActivity: PropTypes.bool,
+    total: PropTypes.bool,
+    value: PropTypes.bool,
+  };
+
+  componentDidMount() {
+    this.element = $(this.refs.element);
+    this.element.progress({
```
ghost, line 29: To cut down on SLOC it'd be nice to spread our progress props of interest here, e.g.

`this.el.progress({...propsOfInterest})`

Also, it might be better for UX do a `document.createDocumentFragment` in `componentWillMount` and then inject the DOM as soon as the root node of the component becomes available. But I haven't done much in the way of this kind of change so that's only conjecture at this point.

package.json
```diff
@@ -83,7 +83,7 @@
     "sinon-chai": "^2.8.0",
     "through2": "^2.0.0",
     "watch": "^0.16.0",
-    "webpack": "^1.12.2",
-    "webpack-dev-server": "^1.12.0"
+    "webpack": "1.12.1",
+    "webpack-dev-server": "1.10.1"
```
jhabdas: fyi - https://github.com/webpack/webpack-dev-server/issues/326

ghost, line 87: fyi - https://github.com/webpack/webpack-dev-server/issues/334

src/modules/Progress/Progress.js
eanplatter, line 25: Technically none are, you could possibly make the `percent` prop required, but it's not necessary.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+
+  render() {
+    const classes = classNames(
+      'sd-progress',
+      'ui',
+      this.props.className,
+      'progress',
+    );
+
+    let content = ::this.renderStandardBar();
```
ghost: Though this app is currently set-up for `stage: 0` (strawman) I'm leery of using anything too experimental as it is more likely to cause breaking changes as the features hit a standards track. IIRC most of this app - if not everything used so far - is available in `stage: 1`, and I'd feel more comfortable sticking with >= 1 for now to mitigate possible growing pains as the specs settle.

levithomason: +1

src/modules/Progress/Progress.js
levithomason, line 29: -1 to manual DOM manipulation in React, it will freak out if you touch the DOM.  In this case, the `progress` plugin doesn't manipulate the DOM, only it's style.  If we need to wrangle a jQuery plugin for DOM manipulation then use [portals](https://github.com/ryanflorence/react-training/blob/gh-pages/lessons/05-wrapping-dom-libs.md).

**EDIT**
Also, to create `propsOfInterest` we'd need to write the same amount of code to create an object consisting of the props we want in order to spread it.  Prefer the current explicit list against a a new object that is used once.

src/modules/Progress/Progress.js
levithomason, line 45: Passing React children to a jQuery plugin are we ;)

levithomason, line 45: ...and html classes.

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,97 @@
...
+import classNames from 'classnames';
+import $ from 'jquery';
+import META from 'src/utils/Meta';
+
+export default class Progress extends Component {
+  static propTypes = {
+    autoSuccess: PropTypes.bool,
+    children: PropTypes.node,
+    className: PropTypes.string,
+    label: PropTypes.string,
```
levithomason: `label` only accepts two values, this should be `PropType.oneOf('percent', 'ratio')`.  Unless this prop is referring to a progress element label, and not the plugin property `label`.  In that case, there is a name conflict and we should fix it.

http://semantic-ui.com/modules/progress.html#/settings

> `label` Can be set to either to display progress as **percent** or **ratio**. Matches up to corresponding text template with the same name.

src/modules/Progress/Progress.js
levithomason: We should downgrade the stage plugin to account for this.

eanplatter: +1 especially for open source stuff.

src/modules/Progress/Progress.js
levithomason: Single line of JSX, `return <div...`

src/modules/Progress/Progress.js
levithomason, line 95: In keeping with our style thus far (and removing the need for the bind operator) you can define the elements and conditions at the top of the render method.  Then, use the conditions in the `return` to render the correct parts:

``` jsx
  render() {
    const isAttached = _.contains(this.props.className, 'attached');
    const attachedBar = <div className='bar' />;
    const label = (
      <div className='label'>
        {this.props.children}
      </div>
    );
    const standardBar = (
      <div>
        <div className='bar'>
          <div className='progress' />
        </div>
        {this.props.children && label}
      </div>
    );

    const classes = classNames(
      'sd-progress',
      'ui',
      this.props.className,
      'progress',
    );

    return (
      <div {...this.props} className={classes}>
        {isAttached ? attachedBar : standardBar}
      </div>
    );
  }
```

**EDIT**
Added support for conditional `label` div, if children are present.

test/specs/modules/Progress/Progress-test.js
levithomason: Hero me up folks, I'll clear this up for both of you.

src/modules/Progress/Progress.js
levithomason, line 95: I actually like how you broke those methods out though, then we aren't creating elements every render that we aren't using.  We should really think about that moving forward.

ghost:
Please :bomb: that `.DS_Store` file and consider adding to `.gitignore`. If you're using ST2/3 there's a plug-in called `Gitignore` to make generating ignore files quick and easy.

levithomason:
after removing DS_Store :accept:

src/modules/Progress/Progress.js
ghost, line 29: @eanplatter could you give this a try (as an example), it's highly expressive, doesn't require `_`, reduces typing and helps with minification (though as mentioned gzip will pretty much nullify the savings gained in the minified source over the wire):

```
componentDidMount() {
    let settings = [
      'autoSuccess',
      'children',
      'className',
      'label',
      'limitValues',
      'onActive',
      'onChange',
      'onError',
      'onSuccess',
      'onWarning',
      'percent',
      'precision',
      'random',
      'showActivity',
      'total',
      'value',
    ];
    settings = settings.map(setting => {
      const key = setting;
      return {[key]: this.props[key]};
    });

    this.element = $(this.refs.element);
    this.element.progress({...settings});
  }
```

src/modules/Progress/Progress.js
```diff
@@ -0,0 +1,96 @@
...
+    return this.element.progress(...arguments);
+  }
+
+  renderAttachedBar = () => {
+    return (
+      <div className='bar' />
+    );
+  };
+
+  renderStandardBar = () => {
```
ghost, line 63: I've asked this before but I'm still a little fuzzy as to why we're not prefixing instance methods like this with `_` (e.g. `_renderStandardBar`) to denote they are not apart of the api fingerprint for this component. /cc @levithomason

src/modules/Progress/Progress.js
levithomason, line 29: This will return an array of objects with one key/value pair for each prop:

``` jsx
[
  {autoSuccess: <this.prop.autoSuccess>},
  {label: <this.prop.label>},
  ...etc
]
```

> pseudo code

eanplatter, line 29: Should we omit `children` and `classNames` from the settings as well?

src/modules/Progress/Progress.js
levithomason, line 63: Per [Methods](https://github.com/TechnologyAdvice/javascript/blob/master/react/README.md#methods) section of our fork of the Airbnb styleguide:

> Do not use underscore prefix for internal methods of a react component.

Keep in mind these are not classes that users instantiate and use their methods as with a traditional library.  JSX handle the class instantiation and interaction is done by `props` mostly.

src/modules/Progress/Progress.js
ghost, line 29: Exactly what I wanted. Couldn't take it the full mile without the example, but we should be able to then iterate over the elements in the array and pass them into the `progress` method.

src/modules/Progress/Progress.js
levithomason, line 63: More in depth here: https://github.com/airbnb/javascript/issues/490

src/modules/Progress/Progress.js
levithomason, line 29: assuming the same settings array, we can create the object like this:

``` jsx
  let settingsObj = {};
  settings.forEach(key => settingsObj[key] = this.props[key]);

  this.element.progress(settingsObj);
```

eanplatter, line 29: Would this work?

``` javasript
  componentDidMount() {
    const whitelist = [
      'autoSuccess',
      'children',
      'className',
      'label',
      'limitValues',
      'onActive',
      'onChange',
      'onError',
      'onSuccess',
      'onWarning',
      'percent',
      'precision',
      'random',
      'showActivity',
      'total',
      'value',
    ];

    const settings = {};

    whitelist.map(key => {
      Object.assign(settings, {[key]: this.props[key]});
    });

    this.element = $(this.refs.element);
    this.element.progress({...settings});
  }
```

jhabdas: Even better. With the key destructuring I was doing the linter was barking without the additional assignment.

ghost:
:leaves:

PR #128, opened 2015-12-11
TextField addon

ghost:
In order to properly highlight an invalid control Stardust requires us to add the `error` class to a `Field` wrapping an `Input`. To enable automatic highlighting for required fields, ease the burden of form development and improve accessibility I've created a new composite `TextField` addon - borrowing heavily from MUI's [TextField](https://github.com/callemall/material-ui/blob/master/src/text-field.jsx#L318) and leveraging our existing components.

Before I start polishing, add tests, and integrate with form behaviors (in the backing topic branch) I'd like to get some initial impressions on the approach. As a result, you can expect to see some related hard-coding on `className`s for now. Alternative to merging into the backing branch, it might be wise for us to work this into `master` and then work it back into my topic branch if we like the approach.

Example (top is the `TextField` component and bottom is a plain jane `Input` with wrapping `Field`):

screen shot 2015-12-10 at 6 05 45 pm

See the `BehaviorDoc` component for example usage.

### DIFF
--- a/docs/app/Components/BehaviorDoc/BehaviorDoc.js
+++ b/docs/app/Components/BehaviorDoc/BehaviorDoc.js
@@ -14,6 +14,7 @@ import {
   Input,
   Message,
   Segment,
+  TextField,
 } from 'stardust';
 
 const DROPDOWN_GENDER_OPTIONS = [
@@ -140,13 +141,11 @@ export default class BehaviorDoc extends Component {
                   <Header.H3>About Me</Header.H3>
                 </legend>
                 <Fields evenlyDivided>
-                  <Field label='Name'>
-                    <Input
-                      ref='name'
-                      name='name'
-                      placeholder='First Name'
-                    />
-                  </Field>
+                  <TextField
+                    labelText='Name'
+                    name='name'
+                    placeholderText='First Name'
+                  />
                   <Field label='Gender'>
                     <Dropdown
                       ref='gender'
--- a/src/addons/TextField/TextField.js
+++ b/src/addons/TextField/TextField.js
@@ -1,0 +1,103 @@
+import React, {Component, PropTypes} from 'react';
+import classNames from 'classnames';
+import META from 'src/utils/Meta';
+
+import Field from 'src/collections/Form/Field';
+import Input from 'src/elements/Input/Input';
+
+export default class TextField extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    disabled: PropTypes.bool,
+    id: PropTypes.string,
+    labelText: PropTypes.string,
+    name: PropTypes.string.isRequired,
+    onBlur: PropTypes.func,
+    onEnterKeyDown: PropTypes.func,
+    onFocus: PropTypes.func,
+    onKeyDown: PropTypes.func,
+    placeholderText: PropTypes.string,
+    ref: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.string,
+    ]),
+    type: PropTypes.string,
+  };
+
+  defaultProps = {
+    ref: 'input',
+  };
+
+  handleInputBlur(e) {
+    this.setState({isFocused: false});
+    if (this.props.onBlur) {
+      this.props.onBlur(e);
+    }
+  }
+
+  handleInputFocus(e) {
+    const {disabled, onFocus} = this.props;
+    if (disabled) return;
+    this.setState({isFocused: true});
+    if (onFocus) onFocus(e);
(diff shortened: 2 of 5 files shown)


levithomason:
I think we should keep the `<Field />` and `<Input />` as separate components.

**Input Variations**
There many valid input [variations](http://semantic-ui.com/elements/input.html#variations) that would be very difficult to support through a single component api like `<TextField />`.  Inputs can have left/right attached buttons, drop downs, labels, and icons.  The plain `<Input />` component api is a bit bloated to handle these as it is.

**Plain Inputs vs Form Inputs**
Semantic UI supports inputs in many contexts.  There are two blocks of markup used in these cases.  A plain input (usable anywhere), and an form field input (used in forms).  The markup is slightly different in these cases, though the states are supported in all of them:

Plain Input in error state
image

Form Field Input in error state
image

If we bake validation into a component that includes the `<Field />` it will exclude validation and states for the plain input component.

**Inputs in other components**
The plain input from above (not wrapped in a field) is used inside of other components like [Menu Input](http://semantic-ui.com/collections/menu.html#input) and [Dropdown Input](http://semantic-ui.com/modules/dropdown.html#input).  These should also support validation and states (error, etc).  The markup in this case needs to be the plain input.  The field wrapper only works in a form.  If we bake validation into a coupled `<Field />` and `<Input />` then we it would exclude input validation and states in these use cases.  While validation in a menu input is likely rare, validation in a dropdown input is a likely need.  More so, a consistent validation experience is preferred.

---

### Conclusion & Suggestions

Based on the variations, use cases, and composability above, I think our goals are:
- form controls should support validation no matter where they are used
- the api surface area should be kept to a minimum
- create composable components rather than coupled abstractions

Two possibilities come to mind right off hand:
**Registering**

The field is the odd man out here.  Perhaps we could have hooks for form controls to register themselves with a Field (even the Form).  Then they can share validation information and stay decoupled.  The Form and Field could ask it's children to register and if accepted, bubble up validation and states.

**Validator**
Another option is to completely decouple the validation from the components all together.  We could have a validator utility, perhaps a decorator, that knows how to validate controls.  It could be used in controls, fields, and forms.  It would handle validating and propagating the validation and control states across controls/fields/forms.

Those are my initial thoughts.

ghost:
> I think we should keep the `<Field />` and `<Input />` as separate components.

Agreed. This is a convenience wrapper around form input controls. It allows children, giving us the ability to support existing `Input` [variants](http://semantic-ui.com/elements/input.html#variations) and accessible labels (those which are clickable and will, as a result, select the corresponding form control) using, in this case, _implicit label association_ as opposed to `htmlFor` for labels and therefore providing better a11y for older browsers that do not support _explicit label association_. Please see [H44 of WCAG 2.0](http://www.w3.org/TR/WCAG20-TECHS/H44.html) for more info on assessable labels.

> If we bake validation into a component that includes the `<Field />` it will exclude validation and states for the plain input component.

In the general use case (i.e. the 80/20 rule) _most or all_ inputs will be associated with a form, whether through use of the HTML5 [`form` attribute](http://devdocs.io/html/element/input#attr-form) (useful for multi-form documents and elements which float outside the `form` body) or not. Because Semantic UI is somewhat _loose_ when it comes to actual HTML semantics, having a convenience wrapper for forms will help encourage use of semantic markup and avoid the kinds of "how do I make this work" pitfalls we've seen building forms in other projects.

However, as suggested above, plain inputs would still be supported - giving us complete coverage over existing and currently possible use cases.

> Based on the variations, use cases, and composability above, I think our goals are:
> - form controls should support validation no matter where they are used
> - the api surface area should be kept to a minimum
> - create composable components rather than coupled abstractions

Dynamic control registration is the intent in the broader design. This should be done using the `form` attribute for better semantics as applicable, and, as a result, allows us to serialize `input` elements associated with a form regardless of their location in the DOM (i.e. they'll serialize as part of a form even if they're not within the form body).

The API surface produced here will reduce the amount of JSX written (and therefore the opportunity for implementation entropy). But the need for this component (or a more intelligent `Input` component) is required so we're able to dynamically add the `error` message to the surrounding field (the only way to highlight fields in error within a form given Semantic's current CSS specificity rules for inputs inside a form).

As for the point on _composable_ versus _coupled_, I feel this point may be moot given we're not losing our ability to compose with existing components and because we _will_ need to create [a behavioral abstraction](https://github.com/TechnologyAdvice/stardust/pull/122/files#diff-9a6b03300c6ab42f2f0750dd66944072R33) to allow our form validation to be easier to grok and maintain.

PR #132, opened 2015-12-18
Upgrade to babel6

levithomason:


jhabdas:
:shipit:

.babelrc
```diff
@@ -1,3 +1,10 @@
 {
-  "stage": 0
+  "presets": [
+    "es2015",
+    "react",
+    "stage-0"
+  ],
+  "plugins": [
+    "add-module-exports"
+  ]
```
levithomason: Upgrade to babel 6 presets.  Since we have karma running from node, we need common js interop.  This plugin adds the module exports to each file to achieve that.

docs/app/index.html
```diff
@@ -12,11 +12,6 @@
   <script src="//cdnjs.cloudflare.com/ajax/libs/Faker/3.0.1/faker.min.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.6/semantic.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/bluebird/2.9.33/bluebird.min.js"></script>
-
-  <!-- build:development-->
-  <script src="/webpack-dev-server.js"></script>
-  <!-- endbuild -->
-
```
levithomason, line 19: This was added to the webpack entry config.  It is no longer needed here.

karma.conf.babel.js
```diff
@@ -0,0 +1,80 @@
+import paths from './paths';
```
levithomason, line 1: Karma [doesn't support](https://github.com/karma-runner/karma/issues/1597) a babel register hook as gulp and webpack do, yet.  We write our karma babel file in future ES.  Then, karma runs `karma.conf.js` which uses the [babel register hook](https://babeljs.io/docs/usage/require/) to compile this file at run time.

dvdzkwsk:
Are we cool to mergerino?

karma.conf.js
```diff
@@ -1,83 +1,2 @@
...
-    webpackServer: {
-      progress: false,
-      stats: statConfig,
-      debug: true,
-      noInfo: true,
-      quiet: false
-    }
-  });
-};
+module.exports = require('./karma.conf.babel');
```
levithomason: We simply import our `karma.conf.babel.js` file after invoking the babel register hook.  So we can write our karma conf in future ES.

package.json
```diff
@@ -31,12 +31,23 @@
...
+  "devDependencies": {
+    "babel-cli": "^6.3.17",
+    "babel-core": "^6.3.2",
+    "babel-eslint": "^4.1.6",
+    "babel-loader": "^6.2.0",
+    "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-preset-es2015": "^6.1.18",
+    "babel-preset-react": "^6.1.18",
+    "babel-preset-stage-0": "^6.1.18",
+    "chai": "^3.3.0",
```
levithomason: Pulled runtime deps into `dependencies`, all others are `devDependencies`.

src/index.js
```diff
@@ -42,7 +42,7 @@ import Item from 'src/views/Items/Item';
 import Items from 'src/views/Items/Items';
 import Statistic from 'src/views/Statistic/Statistic';
 
-export default {
+const stardust = {
```
levithomason, line 45: This style export was not valid ES module syntax.  Babel has dropped support for it (common js interop).  Though we technically have the plugin to support it, we only have it because of karma's need to use the babel register hook.  Updated to proper ES module syntax.

karma.conf.babel.js
ghost, line 1: Looks like we're [using `babel-node`](https://github.com/TechnologyAdvice/tap-web-app/pull/9/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R17) to kick off the test runner. As a result, shouldn't our presets in the `.babelrc` handle the transpilation for us?

package.json
```diff
@@ -31,12 +31,23 @@
...
+  "devDependencies": {
+    "babel-cli": "^6.3.17",
+    "babel-core": "^6.3.2",
+    "babel-eslint": "^4.1.6",
+    "babel-loader": "^6.2.0",
+    "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-preset-es2015": "^6.1.18",
+    "babel-preset-react": "^6.1.18",
+    "babel-preset-stage-1": "^6.1.18",
+    "chai": "^3.3.0",
```
levithomason, line 51: Pulled runtime deps into dependencies, all others are devDependencies.  Also, bumped us down to stage 1 for a bit more stability.

.babelrc
```diff
@@ -1,3 +1,10 @@
 {
-  "stage": 0
+  "presets": [
+    "es2015",
+    "react",
+    "stage-1"
+  ],
+  "plugins": [
+    "add-module-exports"
+  ]
```
levithomason, line 9: Upgrade to babel 6 presets. Since we have karma running from node, we need common js interop. This plugin adds the module exports to each file to achieve that.

karma.conf.babel.js
levithomason, line 1: The current test script just runs karma directly: `"test": "karma start karma.conf.js",`.  The private template repo linked there is not implemented here yet.

karma.conf.babel.js
```diff
@@ -0,0 +1,80 @@
...
+      // webpack watches dependencies
+      devtool: 'inline-source-map',
+      eslint: {
+        formatter: friendlyFormatter,
+      },
+      module: {
+        loaders: [
+          {
+            test: /\.js$/,
+            loaders: ['babel', 'eslint'],
```
ghost, line 46: Thoughts on moving the linter into [a `preLoader`](http://devdocs.io/webpack/loaders) to avoid a heavier compilation at the linting stage?

karma.conf.babel.js
```diff
@@ -0,0 +1,80 @@
...
+        // require: '',
+        reporter: 'html',   // change Karma's debug.html to mocha web reporter
+        ui: 'bdd'
+      }
+    },
+    webpack: {
+      // karma watches the test entry points
+      // (you don't need to specify the entry option)
+      // webpack watches dependencies
+      devtool: 'inline-source-map',
```
ghost, line 38: Not sure if it's a Canary thing or a Webpack thing, but `inline-source-map` has never worked for me. I've always had to revert to `source-map` for proper handling of Source Maps in my browser.

karma.conf.babel.js
ghost, line 1: :+1:

karma.conf.babel.js
levithomason, line 46: I'm indifferent, I don't see any pragmatic gain from doing so.

karma.conf.babel.js
levithomason, line 38: +1 Let's keep this PR about the babel upgrade and add issues for anything else.  The template updates later should also cover this.

karma.conf.babel.js
ghost, line 1: Just noticed we have [`babel-core`](https://github.com/TechnologyAdvice/stardust/pull/132/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R43) as a dependency so we should still be able to `babel-node` it from `package.json` if desired.

webpack.dev.babel.js
```diff
@@ -9,7 +9,8 @@ import webpack from 'webpack';
 export default {
   entry: {
     app: [
-      'webpack/hot/only-dev-server',
+      'webpack-dev-server/client?http://localhost:8080',
```
ghost, line 12: Mind changing the loopback to `0.0.0.0` instead of localhost? Using `localhost` here will cause the dev server to bomb when run inside a docker container.

src/index.js
ghost, line 45: Seems like a few of the changes in this PR relate back to not being able to transpile the Karma config, which I believe calling the already available `babel-node` CLI command (made available by `babel-core` should resolve.

webpack.dev.babel.js
levithomason, line 12: Want to get this PR merged as it is blocking @davezuko.  Will fix errors in this PR and push improvements to another PR.

ghost:
:shipit:

src/index.js
levithomason, line 45: :+1: removed the `karma.conf.js` register hook and added scripts for running karma with babel-node.  It looks like babel-node requires the filename be named `karma.conf.babel.js`, so I left that as is.

karma.conf.babel.js
levithomason, line 1: Done.

dvdzkwsk:
:snowman:

PR #134, opened 2015-12-20
Improve README Documentation

ghost:
Because we're no longer using Wercker for builds, and the CI badge subsequently removed, this adds the CI badge back using the new build tooling and improves the documentation.

Also adds [EditorConfig](http://editorconfig.org/) to help with code consistency (e.g. prevent editors from automatically trimming trailing whitespace in Markdown documents).


README.md
```diff
@@ -1,16 +1,15 @@
...
+[React] component library based on [Semantic UI].
 
-```
-npm i stardust -S
-```
+    npm i stardust -S
 
 ## Learn
-Check out the [Documentation](https://technologyadvice.github.io/stardust/).
+Check out the [Doc Site] for usage documentation.
```
dvdzkwsk: Wow I had no idea you could do that (include the links at the bottom). Sweet :+1:

dvdzkwsk:
:snowman:

README.md
ghost: Another method is to do something like `[Doc Site][1]` and then use `[1]: http://link.com`, but I find it's a little harder to manage for docs with many links. I find this is a nice way to introduce [footnotes](https://github.com/jhabdas/habd.as/blame/master/_posts/2015-05-03-simple-websites-with-jekyll-and-docker.md#L213) as well.

dvdzkwsk: Awesome, and that does incentivize footnotes ^^

levithomason: Second on the doc site link!  I've only seen that for relative links.

README.md
```diff
@@ -1,17 +1,16 @@
-# Stardust  [![Code Climate](https://img.shields.io/codeclimate/github/TechnologyAdvice/stardust.svg?style=flat-square)](https://codeclimate.com/github/TechnologyAdvice/stardust) [![Gemnasium](https://img.shields.io/gemnasium/TechnologyAdvice/stardust.svg?style=flat-square)](https://gemnasium.com/TechnologyAdvice/stardust)
+# Stardust  [![Code Climate](https://img.shields.io/codeclimate/github/TechnologyAdvice/stardust.svg?style=flat-square)](https://codeclimate.com/github/TechnologyAdvice/stardust)
+[![Gemnasium](https://img.shields.io/gemnasium/TechnologyAdvice/stardust.svg?style=flat-square)](https://gemnasium.com/TechnologyAdvice/stardust)
+[![Circle CI](https://circleci.com/gh/TechnologyAdvice/stardust/tree/master.svg?style=svg)](https://circleci.com/gh/TechnologyAdvice/stardust/tree/master)
```
levithomason: Use the [shields.io](http://shields.io/) badge for the circle build as well.  Also, let's put them all on the same line as the codeclimate badge.

image

Or move them all below, just consistent.

README.md
```diff
@@ -48,22 +47,32 @@ var TryStardust = React.createClass({
...
+## Installation
+Start by cloning this repo and installing dependencies:
+
+    git clone git@github.com:TechnologyAdvice/stardust.git stardust && cd $_
+    npm install               # or simply npm i
+
+## Usage
+Once dependencies are installed, start the application with:
+
+    npm start
```
levithomason, line 56: Let's avoid "Installation" and "Usage" headings in this area.  The instructions here do not relate to installing or using Stardust, but to development.  Something like develop/contributing/etc is better suited.

**Edit**
Note there is already a **Usage** section under **Getting Started**

README.md
```diff
@@ -48,22 +47,32 @@ var TryStardust = React.createClass({
 module.exports = TryStardust;
 ```
 
-## Develop
+## Installation
+Start by cloning this repo and installing dependencies:
+
+    git clone git@github.com:TechnologyAdvice/stardust.git stardust && cd $_
```
levithomason: Suggest we simplify this a bit.

```
git clone git@github.com:TechnologyAdvice/stardust.git
cd stardust
npm install
```

README.md
ghost: http://shields.io/ is down again. I'll update and move them back up-top once service is restored.

README.md
ghost, line 56: I prefer to use a standard README document outline I see in most libraries so it's easier for users to grok our documentation. But I'll change this back so we can move forward.

README.md
ghost: That's three lines instead of two. I like to teach people while they're reading and this does that. Is this really worth an update in your eyes @levithomason?

levithomason: Agreed, not worth update if we aren't in agreement.  I brought it up as the README.ms is the first thing many folks will see of Stardust and I believe it sets the tone of the project.  I'll leave my peace here and move one :)

The parts of it I found strange were:
- The clone directory argument is unnecessary, it duplicates the repo name which is the default.  Why show how to duplicate arguments?
- There is a `&&` on line one, but the next command is on it's own line.  Why join some commands and not others?
- The combination of extra arguments, lesser known commands `$_`, and joining some commands and not others `&&`, and comment under this made it difficult to scan.  I had to scan it more than once.
- This README section of a well known project is typically pretty standard and I think consistency is good.

OK, with my thoughts out ^ I'm good moving on.

README.md
levithomason, line 56: We have an install and usage section.  I mean, there are now 2 **Install** and **Usage** sections with conflicting information.  We should specify which one is for which.  I think "using" Stardust ought to mean, developing React applications with it.  In which case, you should `npm i` and add the babel compiler.

### Install and Use Stardust in your project:

image

### Install and Use (Clone and Develop) Stardust locally:

image

levithomason:
:ship: with consideration to comments

PR #137, opened 2015-12-21
Update and fix doc site deploy

levithomason:
Resolves https://github.com/TechnologyAdvice/stardust/issues/129 (added by @jhabdas).


.gitignore
```diff
@@ -3,12 +3,8 @@
...
-_builds/
-_projects/
-_steps/
 npm-debug.log*
 node_modules/
 bower_components/
-testBundle.js
 docs/build
-docgenInfo.json
+docs/app/docgenInfo.json
```
levithomason, line 10: Cleaned up ignored Wercker cruft.  Explicitly referenced the `docInfo.json` so it is cleaned up on `npm run clean`.  This is the Stardust component info used to create the doc site.

circle.yml
```diff
@@ -1,12 +1,9 @@
 machine:
   node:
     version: 4.2.1
-  ruby:
-    version: 2.2.3
```
levithomason, line 5: No more Ruby needed on our CI since we get our changelog by [github changelog api](https://github.com/levithomason/github-changelog-api/) now. :D

circle.yml
```diff
@@ -1,12 +1,9 @@
...
 
 dependencies:
   pre:
-    - gem install github_changelog_generator
```
levithomason, line 9: No more waiting for this heavy gem install, thanks to the API.

scripts/circle_deploy.sh
```diff
@@ -10,26 +10,18 @@ git config --global user.name "tadeploy"
...
-#
-gulp docs
-
-#
 # generate changelog
 #
 echo "...generating changelog"
-github_changelog_generator $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+curl -X POST github-changelog-api.herokuapp.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME \
+ | json contents > CHANGELOG.md
```
levithomason, line 17: We post to the changelog api, grab the `contents` key from the json response and write it to the changelog.  Tested this locally and it works great.

circle.yml
ghost, line 9: +1 for improved speed. Once this app is Dockerized we'll be able to cache dependencies for breezy deployments, and eliminate the external build dependency.

scripts/circle_deploy.sh
```diff
@@ -64,3 +56,21 @@ else
...
+git add .
+git add -f docs/build
+
+if [[ -n $(git status --porcelain) ]]; then
+  echo "...starting push, gh-pages is dirty after build"
+  git commit -n -m "deploy gh-pages by $CIRCLE_USERNAME [ci skip]"
+  git push origin gh-pages
+else
+  echo "...skipping push, gh-pages is clean after build"
+fi
```
levithomason, line 76: Pulled the gh-pages bit into it's own block here. Haven't tested this yet (first deploy will).  It is pretty straight forward.
1. Push the current branch to gh-pages (deploy is master only right now)
2. Checkout gh-pages
3. Build docs, add everything, force add the ignored `docs/build`
   
   > This is ok because we're on the gh-pages branch here, master won't start tracking it.  Removes need for an integration branch.
4. If we have local changes, commit and push them

scripts/circle_deploy.sh
```diff
@@ -10,26 +10,18 @@ git config --global user.name "tadeploy"
...
-# build
-#
-gulp docs
-
-#
 # generate changelog
 #
 echo "...generating changelog"
-github_changelog_generator $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
+curl -X POST github-changelog-api.herokuapp.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME \
```
ghost, line 16: I was going to ask if the default machines came with `curl` installed (as opposed to just `wget`, for many vanilla *nix boxes), then I saw the build pass. ^^

scripts/circle_deploy.sh
```diff
@@ -64,3 +56,21 @@ else
...
+echo "...deploying gh-pages"
+git push -f $CIRCLE_BRANCH:gh-pages
+git checkout gh-pages
+gulp docs
+git add .
+git add -f docs/build
+
+if [[ -n $(git status --porcelain) ]]; then
+  echo "...starting push, gh-pages is dirty after build"
+  git commit -n -m "deploy gh-pages by $CIRCLE_USERNAME [ci skip]"
```
ghost: Rather than `[ci skip]` here it feels more straightforward to add a branch build specification to `circle.yml` and `ignore` the `gh-pages` branch: https://circleci.com/docs/configuration#branches

scripts/circle_deploy.sh
levithomason, line 16: I had seen it used somewhere.  Had to double check this one.  Found it in their [docs](https://circleci.com/docs/configuration) under configuration as a valid way to install phantom.js.

Keep in mind the build passing here is only testing our components.  The easiest way to test this deploy script, is to deploy it :)

scripts/circle_deploy.sh
```diff
@@ -64,3 +56,21 @@ else
...
+  git push origin gh-pages
```
ghost, line 73: If we don't `-f`orce here we may run into trouble with the deployment script during rebuilds, no?

circle.yml
ghost, line 5: Yesh!

circle.yml
dvdzkwsk, line 9: :+1:

ghost:
:ship: with suggestions considered

dvdzkwsk:
:ghost: along w/ @jhabdas' notes

scripts/circle_deploy.sh
levithomason, line 73: The checked out gh-pages here is forced pushed from master on ln 64.  So, the local copy is the latests remote copy.  A rebuild should force push master to gh-pages then checkout gh-pages again.  We'll see how it goes.

scripts/circle_deploy.sh
levithomason: :+1: updating

scripts/circle_deploy.sh
levithomason, line 73: Added -f on gh-pages push for good measure, since I had to update the script for the ignored branch.

PR #140, opened 2015-12-28
fix(imports): remove resolve.root config in webpack

dvdzkwsk:
…bution issues

By using `resolve.root` and `modulesDirectories` overrides with webpack, consuming packages _must_ use webpack and _must_ have a very specific setup in order for Stardust's imports to work correctly. By using all relative paths, Stardust is able to be consumed by any bundler/environment.

### DIFF
--- a/karma.conf.babel.js
+++ b/karma.conf.babel.js
@@ -54,10 +54,8 @@ export default (config) => {
         ]
       },
       resolve: {
-        root: paths.root,
         modulesDirectories: [
           'node_modules',
-          '.',
         ],
         alias: {
           // When these key names are require()'d, the value will be supplied instead
--- a/src/addons/Confirm/Confirm.js
+++ b/src/addons/Confirm/Confirm.js
@@ -2,11 +2,11 @@ import React, {Component, PropTypes} from 'react';
 import Promise from 'bluebird';
 import classNames from 'classnames';
 
-import META from 'src/utils/Meta.js';
-import Modal from 'src/modules/Modal/Modal';
-import ModalContent from 'src/modules/Modal/ModalContent';
-import ModalFooter from 'src/modules/Modal/ModalFooter';
-import ModalHeader from 'src/modules/Modal/ModalHeader';
+import META from '../../utils/Meta.js';
+import Modal from '../../modules/Modal/Modal';
+import ModalContent from '../../modules/Modal/ModalContent';
+import ModalFooter from '../../modules/Modal/ModalFooter';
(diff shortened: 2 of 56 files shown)


eanplatter:
That makes sense to me, but I'm not sure if @levithomason would have any other feedback concerning this. I _believe_ `resolve.root` was added in just for convenience tho.

dvdzkwsk:
Developer convenience makes sense, but the package ultimately has to be able to be safely imported in other environments. I came across this while importing Stardust into another application without the required (and undocumented) configuration, and while it would be easy to add I wanted to put this out in the open.

dvdzkwsk:
That said, I'm OK waiting on @levithomason for this, but it's a pretty critical fix that should be considered sooner rather than later.

eanplatter:
Oh yeah, I just mentioned that I think convenience was the primary motivating factor, so I don't think it's a big deal to do away with. It's easy to change either way, so :fallen_leaf:

ghost:
To make this work we'll need to circle through the components in the `docs` directory as well.

levithomason:
-1 to sweeping removal of root require.  The `test` and `docs` are never required by consumers.  Managing those paths is rather a pain without it.  Example, `Sidebar.js` doc site component:

``` jsx
import META from 'src/utils/Meta';
// vs
import META from '../../../../src/utils/Meta';
```

^ That is a loss of readability and maintainability with no gain.

**Short term**
This PR does hit the right issue, the imports in `src` _should_ use relative requires.  This makes more sense in most cases in the `src` directory anyhow since most imports are from somewhere else in `src` directory.  It also solves the  original problem.

**Long term**
We should have a dist build.

---

**EDIT**
Let's update this PR to only change to the `src` dir imports.

levithomason:
Side note, PR titles end up in the [CHANGELOG](https://github.com/TechnologyAdvice/stardust/blob/master/CHANGELOG.md).  Let's write them in consideration of others reading them as changelog bullets.

dvdzkwsk:
Agreed about the inconvenience in test and docs, but it's definitely a necessity for the source code. Personally I don't see how this package could be reasonably compiled to a distribution bundle due to its dependencies on lodash/jquery/semantic/etc., so it's vital that the source code be easy to import.

dvdzkwsk:
Also as a side note, the only reason I removed the aliases from the test bundle is that I wanted to ensure that the source code worked without relying on webpack's custom resolve utilities.

levithomason:
Cool, we should merge this as soon as possible.  It's borked ATM.

**EDIT**
Merging :)

PR #141, opened 2015-12-28
fix(customPropTypes): use named exports rather than default object

dvdzkwsk:
This fixes a Babel 6 issue where components destructuring this utility would break since keys on default exports are no longer treated as named exports.

For example:

``` js
import {ofComponentTypes} from 'src/utils/customPropTypes'
```

### DIFF
--- a/src/utils/customPropTypes.js
+++ b/src/utils/customPropTypes.js
@@ -1,43 +1,46 @@
 import {Children} from 'react';
 import _ from 'lodash';
 
-const customPropTypes = {
-  /*
-   * Ensures children are of a set of types
-   * Similar to `oneOfType` built-in validator
-   * @param {Array<string>} allowedTypes Collection of allowed Stardust component types
-   * @returns {Array} containing children of the specified type
-   */
-  ofComponentTypes: (allowedTypes) => {
-    return (props, propName, componentName) => {
-      const {children} = props;
-      const disallowed = Children.map(children, child => {
-        return _.includes(allowedTypes, _.get(child, 'type._meta.name')) ? null : child;
-      });
-      if (disallowed && disallowed.length !== 0) {
-        return new Error(
-          `\`${componentName}\` should only have children of type \`${allowedTypes}\`.`
-        );
-      }
-    };
-  },
-  /*
-   * Verifies exclusivity of a given prop
-   * @param {string} exclusives property used for exclusivity check
-   * @throws {Error} if named prop not mutually exclusive
-   * @returns no-op
-   */
-  mutuallyExclusive: exclusives => {
-    return (props, propName, componentName) => {
-      _.forEach(exclusives, exclusiveProp => {
-        if (props[propName] && props[exclusiveProp]) {
-          return new Error(
-            `\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\` not both.`
-          );
-        }
-      });
-    };
-  },
+/*
+ * Ensures children are of a set of types
+ * Similar to `oneOfType` built-in validator
+ * @param {Array<string>} allowedTypes Collection of allowed Stardust component types
+ * @returns {Array} containing children of the specified type
+ */
+export const ofComponentTypes = (allowedTypes) => {
+  return (props, propName, componentName) => {
+    const {children} = props;
+    const disallowed = Children.map(children, child => {
+      return _.includes(allowedTypes, _.get(child, 'type._meta.name')) ? null : child;
+    });
+    if (disallowed && disallowed.length !== 0) {
+      return new Error(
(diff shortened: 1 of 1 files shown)


eanplatter:
Does this change the way we use the customPropTypes util? :fallen_leaf:

dvdzkwsk:
Nope, the default export is still in there in case anything is using that (which I can't find any evidence of), and this fixes the utility for Babel 6 environments. My guess is that the `add-module-exports` plugin in stardust is preventing this issue from occurring during local development, but environments not using it break without this change.

We'll probably come across additional places where this is needed, but that can be saved for future PR's.

dvdzkwsk:
Also, why we don't want to be using `add-module-exports` currently: https://github.com/59naga/babel-plugin-add-module-exports/issues/20

src/utils/customPropTypes.js
```diff
@@ -1,43 +1,46 @@
...
+    });
+  };
 };
 
-export default customPropTypes;
+// Keeping this for backwards compatibility
+export default {
+  ofComponentTypes,
+  mutuallyExclusive,
+};
```
levithomason, line 46: There is no backward compatibility required.  All imports are destructure as it is, they should work with the new export syntax:

image

src/utils/customPropTypes.js
```diff
@@ -1,43 +1,46 @@
...
+ * @throws {Error} if named prop not mutually exclusive
+ * @returns no-op
+ */
+export const mutuallyExclusive = (exclusives) => {
+  return (props, propName, componentName) => {
+    _.forEach(exclusives, exclusiveProp => {
+      if (props[propName] && props[exclusiveProp]) {
         return new Error(
-          `\`${componentName}\` should only have children of type \`${allowedTypes}\`.`
+          `\`${componentName}\` should only have one of type \`${propName}\` or \`${exclusiveProp}\` not both.`
```
levithomason, line 35: Thanks for the fix here!

levithomason:
> Nope, the default export is still in there in case anything is using that (which I can't find any evidence of), and this fixes the utility for Babel 6 environments. My guess is that the add-module-exports plugin in stardust is preventing this issue from occurring during local development, but environments not using it break without this change.

Babel 6 (and future ES) supports destructured import of the default export.  The code as written before this PR will still work in Babel 6.  I'm using it in [other](https://github.com/dev-coop/anny/blob/master/src/Network.js#L4) projects that use Babel 6 without the add-module-exports plugin, too.  It's just regular ES module support.

`import {foo} from 'bar'` will import the default export and destructure its `foo` property, or it will import the named export `foo`.  Either case works.  The case that is no longer working in Babel 6 (because it isn't an ES module pattern) is an unnamed default export `export default {foo: 'bar'}`.  Also, the CommonJS interop `require('foo')` no longer requires the default export. You must `require('foo').default` to get it.

> Also, why we don't want to be using add-module-exports currently: 59naga/babel-plugin-add-module-exports#20

Double check [package.json](https://github.com/TechnologyAdvice/stardust/blob/master/package.json#L47) and [.babelrc](https://github.com/TechnologyAdvice/stardust/blob/master/.babelrc#L8), it's in there :)  But, there is no dist build so of course it only works in development (not when npm installed).

---

**Thinking out loud**
I'm indifferent to using multiple named exports over exporting a single object as the default and destructuring it.  They achieve the same thing.

I prefer we don't have to manage a default export object _in addition to_ the named exports.  Suggest we go with one or the other. A default default object or multiple named exports.  You can still import all named exports as an object without a default export:

``` jsx
// named exports
import * as customPropTypes from 'src/utils/customPropTypes'

// default export
import customPropTypes from 'src/utils/customPropTypes'
```

I also think we should be consistent with our export style throughout the codebase.

**Conclusion**
I don't think we should do this for the following:
- destructuring default exports on import works
- not wanting to maintain an extra default export object
- refactoring all default export objects into named exports outweighs any gain of doing so

dvdzkwsk:
Destructuring a default export object is not supported by Babel 6 without commonjs interop plugins: 
http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default/33506169#33506169

Unless I'm just not understanding correctly, it is midnight.

levithomason:
Nope, you are indeed correct.  The project I linked actually doesn't work if you actually try and use the destructured import...  In a double whammy, it was used in a `throw` case.  Tests were passing because it was throwing over the bad import, not cause it was throwing the correct error. -1 to weak tests :tired_face.

Yep, so with that cleared up this PR is obviously needed.  That said, it is midnight here and we should revisit on fresh brains.  Night :moon:

levithomason:
This util has since been merged with other prop utils.  The usages were also updated.  Closing.

PR #145, opened 2016-01-13
add plugin method to Dropdown component

athurman:
add plugin method to Dropdown component to be used in clients


ghost:
:shipit: Go ahead and merge as soon as tests pass. Then we can release the updated version

PR #147, opened 2016-01-20
chore(Dropdown): add props for semantic settings

dvdzkwsk:


src/modules/Dropdown/Dropdown.js
```diff
@@ -39,7 +124,7 @@ export default class Dropdown extends Component {
 
   render() {
     const options = _.map(this.props.options, (opt, i) => {
-      return <option key={i} value={opt.value}>{opt.text}</option>;
+      return <option key={opt.value} value={opt.value}>{opt.text}</option>;
```
dvdzkwsk, line 127: Minor fix while I'm in here, index as key is a risky anti-pattern, and since the `value` is a string value we can just use that as the key.

eanplatter, line 127: I swear React used to throw warnings in the console for this and then stopped. Not sure why they stopped.

eanplatter:
:fallen_leaf:

src/modules/Dropdown/Dropdown.js
levithomason, line 127: Only warns if key is missing, there is no way for React to know if it is an index or a real key.

levithomason:
:accept:

src/modules/Dropdown/Dropdown.js
```diff
@@ -5,18 +5,103 @@ import React, {Component, PropTypes} from 'react';
...
+      apiSettings: this.props.apiSettings,
+      fields: this.props.fields,
+      forceSelection: this.props.forceSelection,
+      glyphWidth: this.props.glyphWidth,
+      label: this.props.label,
+      match: this.props.match,
+      maxSelections: this.props.maxSelections,
+      on: this.props.on,
+      placeholder: this.props.placeholder,
+      saveRemoveData: this.props.saveRemoteData,
```
ghost, line 77: Looking at [Dropdown Settings - Remote Settings](http://semantic-ui.com/modules/dropdown.html#remote-settings) this defaults to true, meaning form values persist to session storage. Though I haven't actually seen it function yet, it'd be nice to have in the future.

ghost:
:dizzy:

PR #148, opened 2016-01-22
feat(Table): accept sort/onSortChange properties to display sorted state

dvdzkwsk:
Need this for MVP, figure we can have it implement more complex logic when we're not crunched for time.


ghost:
:ship:

athurman:
:octopus: :snowflake:

PR #151, opened 2016-01-24
Optimizes display performance of the Input render method

ghost:
https://github.com/erikras/redux-form/issues/123

While implementing a Redux form using Stardust I encountered a
performance bottleneck similar to the ones others are experiencing
in `redux-form` currently.

This pattern, pulled from the above issue, provides better performance
for forms with Inputs both in Redux and in general.

The most performant solution in the end is to use Immutables,
though that's very heavy-handed and not part of the general use case.

### DIFF
--- a/docs/app/Examples/collections/Form/Validation/FormSpecifyingValidationRulesExample.js
+++ b/docs/app/Examples/collections/Form/Validation/FormSpecifyingValidationRulesExample.js
@@ -11,7 +11,7 @@ export default class FormSpecifyingValidationRulesExample extends Component {
       skills: ['minCount[2]', 'empty'],
       terms: 'checked'
     }
-  }
+  };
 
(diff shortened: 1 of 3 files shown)


dvdzkwsk:
What about adding a `shouldComponentUpdate` with shallow equality checks on both the props and state to prevent unnecessary renders?

jhabdas:
@davezuko that makes way more sense. adjusting...

ghost:
I tried switching lifecycle methods, and pulled in and performed `shallowCompare` as suggested in the [React Advanced performance docs](https://facebook.github.io/react/docs/shallow-compare.html) without any luck.

levithomason:
This should be handled outside of Stardust.  This makes the component controlled by default.  Since props are spread on the input, you can make a controlled component from the outside.

PR #152, opened 2016-01-25
Revert "Optimizes display performance of the Input render method"

ghost:
This reverts commit 013ef7e7a9314404ce60dc7024577b8503b73c7b.


levithomason:
Assuming using a controlled Stardust Input worked for this use case, :accept:

PR #154, opened 2016-01-28
Automatic sortable table class

levithomason:


test/specs/collections/Table/Table-test.js
```diff
@@ -106,4 +106,22 @@ describe('Table', () => {
         });
     });
   });
+
+  describe('sortable', () => {
```
jhabdas: Haven't seen one of these in a while.

src/collections/Table/Table.js
```diff
@@ -82,6 +82,7 @@ export default class Table extends Component {
     const classes = classNames(
       'sd-table',
       'ui',
+      {sortable: !!this.props.onSortChange},
```
levithomason, line 85: If the sort handler is present, add the class so we get the proper formatting:

http://semantic-ui.com/collections/table.html#sortable

test/specs/collections/Table/Table-test.js
jhabdas: ;)

test/specs/collections/Table/Table-test.js
```diff
@@ -106,4 +106,22 @@ describe('Table', () => {
...
+
+    describe('without onSortChange prop', () => {
+      it('does not have class "sortable"', () => {
+        render(<Table data={tableData} />)
+          .findClass('sd-table')
+          .props.className
+          .should.not.include('sortable');
+      });
+    });
+  });
```
levithomason, line 126: Assert the class is added/removed along with the sort handler prop.

ghost:
:hamburger:

PR #155, opened 2016-01-28
Selectable table rows

levithomason:
Semantic UI supports a [`selectable` row table](http://semantic-ui.com/collections/table.html#selectable-row).  This just adds styling to the table.  There is also an [`active` row](http://semantic-ui.com/collections/table.html#active) style.

This PR combines the two, when you click a row (having added the `selectable` table class) the row will be set to `active`.  In addition, a callback prop `onSelectRow` will be called with the row data and the index so you can do what you may with it.  On table sort, the active row will be cleared.

---
- [x] enabled by `className='selectable'` and presence of `onSelectRow` prop
- [x] add onSelectRow callback, called with (rowData, index)
- [x] on select:
  - [x] clear active on all rows
  - [x] apply active to clicked row
- [x] toggle to deselect
- [x] accept a `selectedRows={[...index]}`
- [x] clear selectedRows on sort

### DIFF
--- a/karma.conf.babel.js
+++ b/karma.conf.babel.js
@@ -55,10 +55,6 @@ export default (config) => {
       },
       resolve: {
         root: paths.root,
-        modulesDirectories: [
-          'node_modules',
-          '.'
-        ],
         alias: {
           // When these key names are require()'d, the value will be supplied instead
           jquery: paths.testMocks + '/SemanticjQuery-mock.js',
--- a/karma.conf.babel.js
+++ b/karma.conf.babel.js
@@ -54,8 +54,10 @@ export default (config) => {
         ]
       },
       resolve: {
+        root: paths.root,
         modulesDirectories: [
           'node_modules',
+          '.'
         ],
         alias: {
           // When these key names are require()'d, the value will be supplied instead
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "babel-preset-stage-1": "^6.1.18",
     "chai": "^3.3.0",
     "del": "^2.0.2",
+    "dirty-chai": "^1.2.2",
     "doctrine": "^0.7.0",
     "eslint": "^1.5.1",
     "eslint-config-ta-webapp": "^2.1.0",
--- a/src/collections/Table/Table.js
+++ b/src/collections/Table/Table.js
@@ -40,8 +40,7 @@ export default class Table extends Component {
   }
 
   _isSelectable = () => {
-    const {className, onSelectRow} = this.props;
-    return _.includes(className, 'selectable') && !!onSelectRow;
+    return _.includes(this.props.className, 'selectable');
   };
 
   _deselectRow(index) {
--- a/src/collections/Table/Table.js
+++ b/src/collections/Table/Table.js
@@ -6,7 +6,6 @@ import META from '../../utils/Meta';
 
 export default class Table extends Component {
   static propTypes = {
-    basic: PropTypes.bool,
     children: ofComponentTypes(['TableColumn']),
     className: PropTypes.string,
     data: PropTypes.array,
@@ -20,7 +19,6 @@ export default class Table extends Component {
   };
 
   static defaultProps = {
-    selectedRows: [],
     sort: {
       key: null,
       direction: 'descending',
@@ -30,7 +28,7 @@ export default class Table extends Component {
   constructor(props, context) {
     super(props, context);
     this.state = {
-      selectedRows: this.props.selectedRows,
+      selectedRows: this.props.selectedRows || [],
     };
   }
 
--- a/src/collections/Table/Table.js
+++ b/src/collections/Table/Table.js
@@ -121,16 +121,9 @@ export default class Table extends Component {
       const classes = classNames('sd-table-row', {
         active: this._isRowSelected(rowIndex),
       });
+      const onClick = () => this._handleSelectRow(dataItem, rowIndex);
 
-      return (
-        <tr
-          className={classes}
-          key={rowIndex}
-          onClick={this._handleSelectRow.bind(this, dataItem, rowIndex)}
-        >
-          {cells}
-        </tr>
-      );
+      return <tr className={classes} key={rowIndex} onClick={onClick}>{cells}</tr>;
     });
   }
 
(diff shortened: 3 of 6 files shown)


karma.conf.babel.js
```diff
@@ -54,9 +54,7 @@ export default (config) => {
         ]
       },
       resolve: {
-        modulesDirectories: [
-          'node_modules',
-        ],
+        root: paths.root,
```
levithomason, line 57: Removed `modulesDirectories`, it was set to the default value.  Added `root` so we can require from root in tests.

package.json
```diff
@@ -53,6 +53,7 @@
     "babel-preset-stage-1": "^6.1.18",
     "chai": "^3.3.0",
     "del": "^2.0.2",
+    "dirty-chai": "^1.2.2",
```
levithomason, line 56: Function form for terminating assertion properties. https://github.com/prodatakey/dirty-chai

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
     children: ofComponentTypes(['TableColumn']),
     className: PropTypes.string,
     data: PropTypes.array,
+    onSelectRow: PropTypes.func,
     onSortChange: PropTypes.func,
+    selectedRows: PropTypes.arrayOf(PropTypes.number),
     sort: PropTypes.shape({
       key: PropTypes.string,
-      direction: PropTypes.string, // descending|ascending, defaults to descending
+      direction: PropTypes.oneOf(['descending', 'ascending']),
```
levithomason, line 17: Moved comments to formal enum `propTypes` and `defaultProps`.

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
...
     }),
   };
 
+  static defaultProps = {
+    selectedRows: [],
+    sort: {
+      key: null,
+      direction: 'descending',
+    },
+  }
```
levithomason, line 26: Also added sort props here, removes need for more complicated `_.get()` use later.

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
...
+
+  constructor(props, context) {
+    super(props, context);
+    this.state = {
+      selectedRows: this.props.selectedRows,
+    };
```
levithomason, line 32: Selected rows are tracked in `state` by their index.  The state is initialized from the prop of the same name.  This allows initializing the table with certain rows preselected.  The implementation of the `_selectRow` method only supports selecting a single row, though the component has been written to support selecting multiple rows for future extension.

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
...
-        sorted,
-        ascending: sorted && direction === 'ascending',
-        descending: sorted && direction === 'descending',
+        sorted: isSorted,
+        ascending: isSorted && sort.direction === 'ascending',
+        descending: isSorted && sort.direction === 'descending',
       });
 
-      return <th className={classes} key={key} onClick={onClick}>{content}</th>;
+      return <th className={classes} key={dataKey} onClick={onClick}>{content}</th>;
```
levithomason, line 102: No functional changes to this method.  Just destructuring, use of the defaultProp values, and variable name updates.

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
...
     return _.isObject(content) ? JSON.stringify(content) : content;
   }
 
+  _isRowSelected(index) {
+    return _.includes(this.state.selectedRows, index);
+  }
+
+  _isSelectable = () => {
+    return _.includes(this.props.className, 'selectable');
+  };
```
levithomason, line 46: Since classes are (for now) the single source of truth for component options, this is used here.  The class is auto applied if any of the "_select_" props are used.  See render method.

> Note, in the future we plan to get away from use of Semantic UI classes in our components.  They will be replaced by props.

src/collections/Table/Table.js
```diff
@@ -10,43 +10,97 @@ export default class Table extends Component {
...
+
+  _handleSelectRow(rowItem, rowIndex) {
+    this._toggleSelectRow(rowIndex);
+    if (this.props.onSelectRow) this.props.onSelectRow(rowItem, rowIndex);
+  }
+
   _handleSortHeaderChange(key, direction) {
     const {onSortChange} = this.props;
     if (onSortChange) {
+      this._unselectAllRows();
```
levithomason, line 81: Since sorting happens outside the component we can't track which items were selected reliably.  Instead, we simply clear selection on sort.

webpack.tests.babel.js
```diff
@@ -50,13 +50,7 @@ module.exports = {
...
   },
   resolve: {
-    root: [
-      paths.docsRoot
-    ],
-    modulesDirectories: [
-      'node_modules',
-      '.'
-    ],
+    root: paths.root,
```
levithomason, line 53: Also allow root require in webpack dev server tests, though, this config will likely go away in favor of phantomjs only tests soon.

karma.conf.babel.js
dvdzkwsk, line 57: :+1:

src/collections/Table/Table.js
dvdzkwsk, line 17: Much nicer, forgot about this.

src/collections/Table/Table.js
```diff
@@ -6,47 +6,100 @@ import META from '../../utils/Meta';
...
-      const content = column.props.headerRenderer
-        ? column.props.headerRenderer(this.props.data[0])
-        : _.startCase(column.props.dataKey);
-      const sorted = _.get(this, 'props.sort.key') === key;
-      const direction = _.get(this, 'props.sort.direction', 'descending');
+    const {children, data, sort} = this.props;
+
+    return Children.map(children, (column) => {
+      const {dataKey, headerRenderer} = column.props;
+      const content = headerRenderer ? headerRenderer(data[0]) : _.startCase(dataKey);
```
dvdzkwsk, line 91: You snuck this one in here.

dvdzkwsk:
image

PR #156, opened 2016-01-29
Remove unused s3 sync in build

levithomason:


scripts/circle_deploy.sh
```diff
@@ -16,13 +16,6 @@ echo "...generating changelog"
 ta-script circle_ci/create_changelog
 
 #
-# s3 sync
-#
-echo "...syncing with s3"
-ta-script aws/s3_sync -d ./dist -b "ta-stardust-assets/$NPM_PACKAGE_VERSION"
```
levithomason, line 22: We don't yet have s3 assets (maybe never will).  Removing as it is failing on CI.

PR #159, opened 2016-01-30
Change Dropdown to allow non-unique values.

eanplatter:
This is because it uses the `value` of the `option` as it's key.


levithomason:
Please update the title to something that will make sense in the [CHANGELOG](https://github.com/TechnologyAdvice/stardust/blob/master/CHANGELOG.md).  Also, tag the PR as `bug`.  This will make it show up in the right place with a good name that others will be able to understand.

Then, :accept:!

levithomason:
@eanplatter delete branches after merging, thanks!

PR #161, opened 2016-02-03
Support progress module

eanplatter:


levithomason:
:accept: Made a few updates, you can take a look through the commits.

PR #165, opened 2016-02-09
Support Form settings and callback props

### LINKED ISSUE #160: Form props are not passed to settings (prevent submit)
Preventing submission in Semantic UI forms requires returning false from the `onSuccess` callback.  Adding this callback to a Form in Stardust does not work.  This is because the Form is not passing valid settings to the plugin.  It should pass all valid settings props to the plugin.  See Checkbox for example.

Reference: https://github.com/Semantic-Org/Semantic-UI/issues/546#issuecomment-50775838

levithomason:
1. Fixes #160, handling form plugin props.
2. Adds deprecate utility to warn users of upcoming deprecations:
   
   ``` js
   constructor(props, context) {
     super(props, context)
     deprecate.props(this, {
       settings: 'Use a separate prop for each setting.',
     })
   }
   ```
   
   image


docs/app/Examples/collections/Form/Validation/FormSpecifyingValidationRulesExample.js
```diff
@@ -1,65 +1,57 @@
...
-          <Checkbox name='terms' className='hidden' label='I agree to the terms and conditions' />
-        </Field>
-        <Button className='blue submit'>Submit</Button>
-        <Message className='error' />
-      </Form>
-    )
-  }
-}
+const FormSpecifyingValidationRulesExample = (props) => (
+  <Form className='segment' fields={fields} onSuccess={() => false}>
```
levithomason: Use function form example component.  Pulls deprecated form `settings` prop into individual props (`fields`, `onSuccess`).

docs/app/Examples/collections/Form/Validation/FormValidatingOnBlurAndOtherEventsExample.js
```diff
@@ -1,73 +1,40 @@
...
+const fields = {
+  firstName: 'empty',
+  lastName: 'empty',
+  username: 'empty',
+  password: 'minLength[6]',
+  terms: 'checked',
 }
+
+const FormValidatingOnBlurAndOtherEventsExample = (props) => (
+  <Form className='segment' fields={fields} on='blur' inline onSuccess={() => false}>
```
levithomason: Note the form props used for form plugin settings.

src/collections/Form/Form.js
```diff
@@ -3,17 +3,54 @@ import $ from 'jquery'
...
+      fields: PropTypes.object,
+    }),
   };
 
+  constructor(props, context) {
+    super(props, context)
+    deprecate.props(this, {
+      settings: 'Use a separate prop for each setting.',
+    })
+  }
```
levithomason, line 53: Deprecated settings prop.

src/collections/Form/Form.js
```diff
@@ -3,17 +3,54 @@ import $ from 'jquery'
...
+  inline: PropTypes.bool,
+  transition: PropTypes.string,
+  duration: PropTypes.number,
+  // callbacks
+  onValid: PropTypes.func,
+  onInvalid: PropTypes.func,
+  onSuccess: PropTypes.func,
+  onFailure: PropTypes.func,
+  fields: PropTypes.object,
+}
```
levithomason, line 28: This is the pattern we are using for separating plugin props from component props.  There is a utility to pick `this.props` for only plugin props, useful when calling refresh on the plugin. There's also a util for omitting plugin props, useful when spreading user props or doing other component prop checks.

docs/app/Examples/collections/Form/Validation/FormValidatingOnBlurAndOtherEventsExample.js
ghost: Thoughts on setting a default prop on the `Form` component so this common use case of passing `onSuccess` does not require an understanding of the underlying implementation details in order to help avoid form implementations which cause a page refresh on Enter keypress within a form control by default when `onSuccess` is omitted?

TL;DR: Can we default onSuccess so it can be omitted in the simple use case?

src/collections/Form/Form.js
ghost, line 53: Good stuff

ghost:
See comment on practicalities and :guitar:

docs/app/Examples/collections/Form/Validation/FormValidatingOnBlurAndOtherEventsExample.js
levithomason: This is a great question.  I went back and forth with this actually.  I actually committed a `defaultProps` function that returned false and then reverted it.  I considered our current approach:

Stardust...
1. is a React port of Semantic UI
2. includes some addons and extra features made possible by JS components
3. is as close to vanilla React and Semantic UI as possible
   1. never sets default props
   2. never overrides styles
   3. spreads your props on the element of concern (`<input />` for `<Input />`)

I opted for consistency here in not defaulting.  I figured this is the "Semantic UI way" of doing things.  Also, I'm not sure that everyone's use case is to prevent submit.  For instance, Semantic UI actually disables submit automatically for you if you use their API setting.  I'm still on the fence with our default though.

What are your thoughts on keeping with some of this ^ vs defaulting in this one instance for our use cases?

**EDIT**
I am willing to gamble that it is true that most React users are submitting their forms via an API and therefore _probably_ want this as well.  Also, we could only include the prevent submit default prop if the action prop is not present.

ghost: Thanks for the thoughtful detail. I could go either way. Thinking 80/20 I'd swing towards the broader, but I'm comfortable with convention.

levithomason: 80/20 is enough to push me over the fence :P  default coming up.

PR #167, opened 2016-02-12
Add subcomponents

### LINKED ISSUE #99: Restructure sub components

levithomason:
Fixes #99. Refactors to subcomponents, i.e. `TableColumn` => `Table.Column`.

Also adds deprecation warnings when using the old components:

image

Grid components were also made into sub components:

image

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentExample.js
+++ b/docs/app/Components/ComponentDoc/ComponentExample.js
@@ -1,5 +1,5 @@
 import React, { Component, createElement, PropTypes } from 'react'
-import { Grid, Column, Button } from 'stardust'
+import { Grid, Button } from 'stardust'
 import Highlight from 'react-highlight'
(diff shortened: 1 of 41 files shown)


dvdzkwsk:
:ghost:

PR #168, opened 2016-02-13
Support full dropdown functionality

### LINKED ISSUE #4: Support all dropdown types
http://semantic-ui.com/modules/dropdown.html

### LINKED ISSUE #149: Dropdown fails to open on initial render with dynamic className change
I was attempting to add a loading indicator to the `Dropdown` component by dynamically adding a `loading` class using [`classNames`](https://www.npmjs.com/package/classnames) like so:

``` jsx
<Dropdown
  className={cx({loading: isDropdownLoading}, 'fluid search option')}
  name='clientId'
  options={dropdownOptions}
  defaultValue={client.id}
/>
```

Which resulted in the dropdown failing to open once the page rendered (only the first time the component rendered, not the second).

Reproduce with Stardust `v0.9.6` using semantic.js `v2.1.6`. Note the scope of this may apply more broadly to changes in props per @levithomason

levithomason:
### [Try It](http://deweybot.taplatform.net/qa/src/stardust/feature/dynamic-dropdown-options/#Dropdown-Types-Dropdown)

There are **3 examples** written.  More will be added in other PRs later.
### Not as large as it looks

This PR includes 3 primary things:
1. Doc Site Examples
2. Test updates
3. Dropdown rewrite
### Done

Fixes #4 supporting all dropdown permutations.  Also fixes #149 supporting dynamic options.
- [x] add dropdown menu, item, and divider sub components
- [x] support icons
- [x] support item text/value/descriptions
- [x] add extended dropdown example using dropdown subcomponents
- [x] handle regular dropdowns (menu only style)
- [x] handle selection dropdowns, with hidden input for values
- [x] separate component / plugin prop types
- [x] componentDidUpdate
- [x] add async options example

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -1,0 +1,48 @@
+import _ from 'lodash'
+import cx from 'classnames'
+import faker from 'faker'
+import React, { Component } from 'react'
+import { Button, Dropdown } from 'stardust'
+
+const getOptions = () => _.times(3, () => {
+  const name = faker.name.findName()
+  return { text: name, value: _.snakeCase(name) }
+})
+
+export default class DropdownAsyncOptions extends Component {
+  state = { isFetching: false, value: [], options: [] }
+
+  handleChange = value => this.setState({ value })
+
+  // fake api call
+  fetchOptions = () => {
+    this.setState({ isFetching: true })
+
+    setTimeout(() => {
+      this.setState({ isFetching: false, options: getOptions() })
+      this.selectRandom()
+    }, 500)
+  }
+
+  selectRandom = () => this.setState({ value: _.sample(this.state.options).value })
+
+  render() {
+    const { options, isFetching, value } = this.state
+    const classes = cx('search selection multiple', { 'disabled loading': isFetching })
+
+    return (
+      <div>
+        <Button onClick={this.fetchOptions}>Fetch</Button>
+        <Button onClick={this.selectRandom} disabled={_.isEmpty(options)}>Random</Button>
+        <Dropdown
+          defaultText='Add Users'
+          className={classes}
+          options={options}
+          value={value}
+          onChange={this.handleChange}
+        />
+        <pre>{JSON.stringify(this.state, null, 2)}</pre>
+      </div>
+    )
+  }
+}
--- a/docs/app/Examples/modules/Dropdown/States/Disabled.js
+++ b/docs/app/Examples/modules/Dropdown/States/Disabled.js
@@ -1,0 +1,16 @@
+import React, { Component } from 'react'
+import { Dropdown } from 'stardust'
+
+export default class DropdownDisabledExample extends Component {
+  render() {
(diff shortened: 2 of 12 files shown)


docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
```diff
@@ -0,0 +1,39 @@
...
+
+  render() {
+    const { options, isFetching } = this.state
+    return (
+      <div>
+        <Button onClick={this.updateItems}>
+          Fetch Items
+        </Button>
+        <Dropdown
+          className={`search selection multiple ${isFetching && 'disabled loading'}`}
```
ghost: I had opened an issue related to programmatic use of the loading indicator with https://github.com/TechnologyAdvice/stardust/issues/149. Can that be closed with this PR?

src/modules/Dropdown/Dropdown.js
```diff
@@ -1,111 +1,97 @@
...
-      onRemove: this.props.onRemove,
-    })
+    this.refresh()
   }
 
   componentDidUpdate(prevProps, prevState) {
-    this.element.dropdown('refresh')
+    if (prevProps === this.props) {
+      this.refresh()
+    }
```
ghost: When the component updates, wouldn't we want to refresh only if `props` did change? And, if so, wouldn't we do that with a shallow or deep compare given the `options` and `children` props are objects themselves?

ghost:
This is great! I left a couple of comments. Other than that the code looks great. Thanks for updating Examples given we have a new way to structure our menus now (e.g. `<Dropdown.Menu />`).

Ship when ready: :star:

docs/app/Examples/modules/Dropdown/Types/Dropdown.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react'
+import { Dropdown } from 'stardust'
+
+export default class DropdownDropdownExample extends Component {
+  render() {
+    return (
+      <Dropdown text='File' action='hide'>
```
levithomason, line 7: If you only want to show a menu, and not have the text update, set the correct [Semantic UI action](http://semantic-ui.com/modules/dropdown.html#specifying-select-action).  In this case, we just hide the dropdown on select.  This leaves the `File` text in the dropdown space so it looks and works like a normal dropdown menu (not a selection dropdown).

docs/app/Examples/modules/Dropdown/States/Disabled.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react'
+import { Dropdown } from 'stardust'
+
+export default class DropdownDisabledExample extends Component {
+  render() {
+    return (
+      <Dropdown className='disabled' text='Dropdown'>
+        <Dropdown.Menu>
+          <Dropdown.Item text='Choice 1' />
```
tslater, line 9: I can't get this example to work. the Dropdown.Items don't show up.

levithomason, line 9: Linking #346

PR #171, opened 2016-02-14
Fix conformance tests

levithomason:
Fixes a number of oversights and issues with the conformance tests.  Also fixes code to conform to now working conformance tests.


docs/app/ComponentGuidelines.md
```diff
@@ -190,28 +190,28 @@ All magic is noted in the documentation examples.
 
 ```jsx
 <Form.Field className='inherit-this' />
-// => <div className='sd-field inherit-this field>...
+// => <div className='sd-form-field inherit-this field>...
```
levithomason, line 193: Components were reorganized to sub components in #99:
`Field` => `FormField`.

Generated `className`s should also follow suit:
`className='sd-field'` => `className='sd-form-field'`

Fixing conformance tests pointed this out.

src/collections/Form/FormField.js
```diff
@@ -3,7 +3,7 @@ import classNames from 'classnames'
 import numberToWord from '../../utils/numberToWord'
 import META from '../../utils/Meta.js'
 
-export default class Field extends Component {
+export default class FormField extends Component {
```
levithomason, line 6: Class names were not matching filenames.

src/collections/Form/FormFields.js
```diff
@@ -17,22 +17,25 @@ export default class Fields extends Component {
...
-      if (_.get(child, 'type._meta.name') === 'Field') fieldCount += 1
-    })
-    fieldCount = numberToWord(fieldCount)
+    if (this.props.evenlyDivided) {
+      Children.forEach(this.props.children, child => {
+        if (_.get(child, 'type._meta.name') === 'FormField') {
+          fieldCount += 1
+        }
+      })
+    }
```
levithomason, line 33: Only count form field children if we are `evenlyDivided` so we are wasting cycles every render for no reason.

src/elements/Header/_Header.js
```diff
@@ -1,17 +1,22 @@
...
 export default class _Header extends Component {
   static propTypes = {
     _headerElement: PropTypes.string,
+    _sdClass: PropTypes.string,
     children: PropTypes.node,
     className: PropTypes.string,
     icon: PropTypes.node,
     image: PropTypes.node,
-  };
+  }
```
ghost, line 14: I love not having the `;` on static props. What makes this possible?

test/specs/Conformance-test.js
```diff
@@ -14,37 +12,20 @@ describe('Conformance', () => {
...
 
-    describe(name, () => {
+    describe(constructorName, () => {
       it('extends Component', () => {
-        expect(SDComponent.prototype).to.eql(Component.prototype)
-      })
-
-      it(`constructor name is "${name}" (matches component name)`, () => {
-        SDComponent.prototype.constructor.name.should.equal(name)
+        expect(Component.prototype).to.eql(React.Component.prototype)
```
ghost: I appreciate the clarity of `equal` over `eql`. Looks like we may be able to avoid API look-ups if we use the `deep` in the chain and stick with `equal`.

test/specs/Conformance-test.js
```diff
@@ -102,44 +83,9 @@ describe('Conformance', () => {
...
-              .findClass(sdClass)
-            if (isDOMComponent(rendered)) {
-              renderedClasses = rendered.getAttribute('class')
-            } else {
-              renderedClasses = rendered.props.className
-            }
-            const sdClassIndex = renderedClasses.indexOf(sdClass)
-            const classesIndex = renderedClasses.indexOf(classes)
-            expect(sdClassIndex).to.be.below(classesIndex)
-          })
```
ghost, line 118: Do we still check the order of `className`s somewhere? Do we need to?

test/specs/index-test.js
```diff
@@ -1,41 +1,43 @@
...
+describe('stardust (index.js)', () => {
+  _.each(componentInfo, (info) => {
+    const { _meta, constructorName, subComponentName } = info
+    const isPrivate = META.isPrivate(constructorName)
 
-const componentNames = _.map(componentCtx.keys(), key => {
-  return path.basename(key).replace('.js', '')
-})
+    // stardust.H1
+    const isStardustProp = _.has(stardust, constructorName)
```
ghost, line 12: Is this specific to `Header.H1`?

ghost:
Nothing glaring but the :sunny: in my face. :boat: with comments considered.

src/elements/Header/_Header.js
levithomason, line 14: Babel reverting their ridiculous breaking change where they threw errors if you didn't have them ;)

I actually meant to leave all them in and i'll pull them on another PR.  Will leave this here for now...

test/specs/Conformance-test.js
levithomason: Agreed on the clarity, done.

levithomason: Agreed on the clarity, done.

test/specs/Conformance-test.js
levithomason, line 118: No longer check order, with nested private components it is not reasonably feasible.  We still test that the classes are there, just not that the first child has them.

The `_Header` case was the deciding factor.  The first child is the private header.  The actual generated `sd-*` class is not the first child.  We could _assume_ that the second child will have the class, but that only works right now in this one case.  If we use 2 private wrappers for something else, it will break (`sd-*` would be child 3).

We dropped the ordering requirement for now after considering the other options.

karma.conf.babel.js
```diff
@@ -65,7 +65,7 @@ export default (config) => {
       progress: false,
       stats: statConfig,
       debug: true,
-      noInfo: true,
+      noInfo: false,
```
dvdzkwsk, line 68: por que?

src/collections/Grid/GridRow.js
```diff
@@ -10,14 +10,14 @@ export default class GridRow extends Component {
...
-    name: 'Row',
+    name: 'GridRow',
     type: META.type.collection,
     parent: 'Grid',
   };
 
   render() {
     const classes = classNames(
-      'sd-row',
+      'sd-grid-row',
```
dvdzkwsk, line 20: Just looking at these classes, it would be cool to have these be automated based off of `_meta.name`.

src/elements/Header/Header.js
```diff
@@ -22,9 +23,10 @@ export default class Header extends Component {
   static H5 = HeaderH5;
   static H6 = HeaderH6;
   static Subheader = HeaderSubheader;
+
   render() {
     return (
-      <_Header {...this.props} _headerElement='div'/>
+      <_Header {...this.props} />
```
dvdzkwsk, line 29: Seeing as this is done internally and you know all of the props being passed to it, I wonder if it would be (marginally) faster (meaning performance) to just apply them yourself rather than using the spread operator.

src/elements/Header/_Header.js
```diff
@@ -1,17 +1,22 @@
-import _ from 'lodash'
-import React, { Component, PropTypes } from 'react'
-import classNames from 'classnames'
+import cx from 'classnames'
```
dvdzkwsk, line 1: Not relevant to this line... I wonder if it would make sense to make this a pure component and not re-render unless props change. Since it's such a simple (and common) element it might be a nice future feature (or option) to do prop checks to prevent unneeded re-renders.

test/utils/componentInfo.js
```diff
@@ -0,0 +1,37 @@
...
+  const Component = componentCtx(key)
+  const { _meta, prototype } = Component
+
+  const constructorName = prototype.constructor.name
+  const filePath = key
+  const filename = path.basename(key)
+  const filenameWithoutExt = path.basename(key, '.js')
+  const subComponentName = _meta.parent ? _meta.name.replace(_meta.parent, '') : null
+  // HeaderH1 => sd-header-h1
+  const sdClass = 'sd-' + constructorName
```
dvdzkwsk, line 20: See my earlier comment, would be cool to generate these automatically (not during renders, of course) in the actual components.

webpack.tests.babel.js
```diff
@@ -58,7 +58,7 @@ module.exports = {
     progress: false,
     stats: statConfig,
     debug: true,
-    noInfo: true,
+    noInfo: false,
```
dvdzkwsk, line 61: por que?

dvdzkwsk:
:ghost: w/ comments

src/collections/Grid/GridRow.js
levithomason, line 20: This is something to think about for sure.  Since we are able to test for them, I believe should be able to generate them.  Will give it more thought.

src/elements/Header/Header.js
levithomason, line 29: This `Header` is use facing, so they may put any prop they like.

src/elements/Header/_Header.js
levithomason, line 1: Definitely, I've been skipping all optimizations while we crank toward 1.0.  This is a great point.

test/utils/componentInfo.js
levithomason, line 20: Again, totally agree.  Once we hit 1.0, let's profile and double back to make the biggest improvements.

PR #201, opened 2016-03-17
fix(FormFields): only apply "fields" class if "field" is not in class…

dvdzkwsk:
…Name


ghost:
:boat:

eanplatter:
:fallen_leaf:

dvdzkwsk:
Closed as unneeded since we can just not wrap in `Fields`.

PR #204, opened 2016-03-19
Stage branches and improve workflow

levithomason:
This PR improves our CI/CD and development processes:
1. All branches except master are now staged for testing
2. `lint` is run on pretest instead of every commit (slow)
3. `pre-commit` has been removed entirely (slow and error prone)
4. `gh-pages` is now deployed using the [`gh-pages`](https://github.com/tschaub/gh-pages) package


.npmignore
```diff
@@ -1,4 +0,0 @@
-docs/
-gulp/
-scripts/
-test/
```
levithomason, line 4: This is cruft since we are using the `files` key in `package.json`.  No need to ignore when we explicitly include files.

circle.yml
```diff
@@ -17,7 +17,13 @@ test:
     - nvm install 5 && npm test
 
 deployment:
+  development:
+    branch: /^(?!master).*$/
+    commands:
+      - gulp docs
+      - $(npm bin)/ta-script aws/stage -d docs/build -r $CIRCLE_PROJECT_REPONAME -b $CIRCLE_BRANCH
```
levithomason, line 24: Stage docs for all branches except master on AWS.

circle.yml
```diff
@@ -17,7 +17,13 @@ test:
...
   production:
     branch: master
     commands:
-      - bash scripts/circle_deploy.sh
+      - $(npm bin)/ta-script circle_ci/create_changelog
+      - npm run deploy:docs
```
levithomason, line 29: On `master` builds, update change log and deploy docs to `gh-pages`.

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "babel src -d dist",
     "clean": "while read line; do rm -rf $line; done < .gitignore",
+    "deploy:docs": "gulp docs && $(npm bin)/gh-pages -d docs/build -m 'deploy docs [ci skip]'",
```
levithomason, line 13: Build and deploy docs.  This makes a flat directory with just the doc site `build` dir.  Since there are no tests (or anything else) in there, we skip ci.

package.json
```diff
@@ -19,10 +20,9 @@
     "release:patch": "ta-script npm/release.sh patch",
     "start": "gulp",
     "karma": "babel-node $(npm bin)/karma start",
-    "test": "npm run karma  -- karma.conf.babel.js",
+    "test": "npm run lint && npm run karma -- karma.conf.babel.js",
```
levithomason, line 23: `lint` before `test`

package.json
```diff
@@ -94,7 +94,7 @@
     "run-sequence": "^1.1.4",
     "sinon": "git://github.com/cjohansen/Sinon.JS#b672042043517b9f84e14ed0fb8265126168778a",
     "sinon-chai": "^2.8.0",
-    "ta-scripts": "^2.3.0",
+    "ta-scripts": "^2.3.1",
```
levithomason, line 97: This patch version includes a fix to monkey patch relative html file attribute paths.  Previously, it only patched absolute paths.  Stardust uses relative paths for it's assets since it needs to work on `technologyadvice.github.io/stardust`.

ghost, line 97: Per SemVer this should happen automatically, no?

package.json
ghost, line 23: Nice to see this build target doing more for us

circle.yml
```diff
@@ -17,7 +17,13 @@ test:
     - nvm install 5 && npm test
 
 deployment:
+  development:
+    branch: /^(?!master).*$/
```
ghost, line 21: What does the period following the negative lookahead assertion add?

circle.yml
ghost, line 29: Slick

ghost:
:boat:

package.json
levithomason, line 97: Yes sir, did this strictly for CircleCI which _I think_ would've ran with old cache.  Not 100%.

circle.yml
levithomason, line 21: `*` - 0 or more of the preceding character
`.` - any character except line breaks

So without the `.` it will try to match 0 or more of _nothing_, which matches nothing.
You can play with it here and some example branch names: http://regexr.com/3d1vg

PR #205, opened 2016-03-23
Icon component

### LINKED ISSUE #178: <Icon /> Component
http://semantic-ui.com/elements/icon.html

levithomason:
Closes #178 includes docs.  Replaces all `<i />` with `<Icon />` and updates classes.

---

image

### DIFF
--- a/src/elements/Icon/Icon.js
+++ b/src/elements/Icon/Icon.js
@@ -1,0 +1,27 @@
+import React, { Component, PropTypes } from 'react'
+import cx from 'classnames'
+import META from '../../utils/Meta'
+
+export default class Icon extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+  }
+
+  static _meta = {
+    library: META.library.semanticUI,
+    name: 'Icon',
+    type: META.type.element,
+  }
+
+  render() {
+    const { className } = this.props
+    const classes = cx(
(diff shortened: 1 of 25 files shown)


src/elements/Icon/Icon.js
```diff
@@ -0,0 +1,27 @@
...
+
+  render() {
+    const { className } = this.props
+    const classes = cx(
+      'sd-icon',
+      className,
+      'icon',
+    )
+    return (
+      <i {...this.props} className={classes} />
```
dvdzkwsk, line 24: One thing I've seen with a lot of `<Icon />` components is the usage of props for icon type rather than classes, that way it's not just a no-value-added wrapper over an `<i>` element. It also makes it quicker (read: less code) to create them (compare `<Icon heart />` vs. `<Icon className='heart' />`)

Food for thought.

levithomason, line 24: Absolutely, we're going to move all Semantic UI class names into props once all components are done.  Like so:

``` js
<Icon user />
<Grid centerAligned padded />
```

levithomason:
Went to do this then pushed it back again.  Wanted to get some perspective though. How would you add a on click callback then? Do we parse props and only turn boolean props into classes? Spreading the rest?

PR #206, opened 2016-03-24
Remove jQuery dependency

levithomason:
# Migrated to #247

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -47,12 +47,11 @@ export default class DropdownAsyncOptions extends Component {
           disabled={isFetching}
           loading={isFetching}
         />
-        <p>
-          <label>
-            <input type='checkbox' value={search} onChange={this.toggleSearch} />
-            {' '}Search
-          </label>
-        </p>
+        {' '}
+        <label>
+          <input type='checkbox' checked={search} onChange={this.toggleSearch} />
+          {' '}Search
+        </label>
         <pre>{JSON.stringify(this.state, null, 2)}</pre>
       </div>
     )


package.json
```diff
@@ -43,6 +43,7 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "classnames": "^2.1.5",
+    "keyboardevent-key-polyfill": "github:technologyadvice/keyboardevent-key-polyfill#0eb142e8d6845a90f23d21b468a3c8c8162b5a25",
```
levithomason: The space key was missing from this package, until it is published, we'll be using our fork.  See here:

https://github.com/cvan/keyboardevent-key-polyfill/issues/10

docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
```diff
@@ -1,16 +1,19 @@
 import _ from 'lodash'
-import cx from 'classnames'
 import faker from 'faker'
 import React, { Component } from 'react'
 import { Button, Dropdown } from 'stardust'
 
-const getOptions = () => _.times(3, () => {
+const getOptions = () => _.times(5, () => {
```
dvdzkwsk, line 6: 5 > 3

docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
```diff
@@ -25,23 +28,34 @@ export default class DropdownAsyncOptions extends Component {
...
+          loading={isFetching}
         />
+        <p>
+          <label>
+            <input type='checkbox' value={search} onChange={this.toggleSearch} />
+            {' '}Search
+          </label>
+        </p>
         <pre>{JSON.stringify(this.state, null, 2)}</pre>
+        {/*
```
dvdzkwsk: Cruft? (the comments)

src/addons/Select/Select.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable valid-jsdoc */
```
dvdzkwsk, line 1: PRAISE JESUS

src/elements/Input/Input.js
```diff
@@ -59,6 +59,7 @@ export default class Input extends Component {
       'input'
     )
     const props = getUnhandledProps(this)
+    delete props.children
```
dvdzkwsk, line 61: This strikes me as strange/unintuitive. Not much else to add.

src/index.js
```diff
@@ -1,10 +1,15 @@
+import 'babel-polyfill'
```
dvdzkwsk: Not super cool with this if stardust relies on this polyfill being present. The furthest I'd go would be to tell people that certain features must be enabled (which can be done by them using the polyfill) but I'd really try to avoid including it yourself.

(Not only because of its weight, but because it pollutes the global scope which is fine for certain apps [case in point ours] but isn't something you want to apply to everyone using this library). Can it be avoided?

src/modules/Checkbox/Checkbox.js
```diff
@@ -45,7 +45,7 @@ export default class Checkbox extends Component {
   }
 
   componentWillUnmount() {
-    this.element.off()
+    _.invoke(this, 'element.off')
```
dvdzkwsk, line 48: Any specific reason for this change?

docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
levithomason, line 6: Needed the dropdown to scroll with options and have more items for search matches.  Again, this PR is not actually ready yet.

src/elements/Input/Input.js
levithomason, line 61: Yep, junk that is going away.

src/index.js
levithomason: Agreed, this has actually been in the lib but we should remove it.  We have lodash which covers most of what we're using polyfill for.

This will require a fair amount of changes, so we'll do this in another PR specifically for removing the polyfill.

PR #207, opened 2016-03-24
Update dependencies

levithomason:
This PR udpates bluebird and lodash.  There were a number of places where we didn't need lodash, they were updated.  Also, there was some ugly logic cleaned up along the way.

Glad to see this result in removing half the code touched:

image

### DIFF
--- a/src/elements/List/ListItem.js
+++ b/src/elements/List/ListItem.js
@@ -1,6 +1,5 @@
-import _ from 'lodash'
 import React, { Component, PropTypes } from 'react'
-import classNames from 'classnames'
+import cx from 'classnames'
 import META from '../../utils/Meta'
 
 export default class ListItem extends Component {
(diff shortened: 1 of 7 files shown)


test/specs/Conformance-test.js
```diff
@@ -50,7 +50,7 @@ describe('Conformance', () => {
         if (_.has(_meta, 'parent')) {
           describe('parent', () => {
             it('matches some component name', () => {
-              expect(_.pluck(componentInfo, 'constructorName')).to.contain(_meta.parent)
+              expect(_.map(componentInfo, 'constructorName')).to.contain(_meta.parent)
```
levithomason, line 53: note, `map` is the new `pluck`.  Uses property match shorthand.  `pluck` is gone.

src/elements/List/ListItem.js
```diff
@@ -21,31 +20,21 @@ export default class ListItem extends Component {
...
-    delete props.className
-    delete props.description
-    delete props.header
-    delete props.icon
-    delete props.image
+    const content = !header ? description : (
+      <div className='content'>
+        <div className='header'>{header}</div>
+        {description && <div className='description'>{description}</div>}
+      </div>
```
ghost, line 30: Looks like we only need the wrapping `div`s when the `description` has a `header`?

src/views/Items/Item.js
```diff
@@ -27,55 +27,27 @@ export default class Item extends Component {
...
-    delete props.children
-    delete props.className
-    delete props.description
-    delete props.extra
-    delete props.header
-    delete props.image
-    delete props.meta
+    const _description = children || description
+
+    const content = !header && !meta && !extra ? _description : [
```
ghost: It may be more clear to say `(header || meta || description)` , but I see why you did it in this order for easier code layout. It's just a little confusing.

src/elements/List/ListItem.js
levithomason, line 30: Yes sir, otherwise the content doesn't align in the middle with `middle aligned content`.  The description div has a top margin which pushes it down a bit.

test/specs/Conformance-test.js
ghost, line 53: Wow. We were only using it once! That's a simple refactor.

src/views/Items/Item.js
levithomason: Nice call.  I went back and forth here, I'll go back for the clarity win.

ghost:
:sake:

test/specs/Conformance-test.js
levithomason, line 53: Heh, there were lots of other [breaking lodash changes](https://github.com/lodash/lodash/wiki/Changelog#v400) but I mostly tried to remove them in favor of ES201\* features like spread and named destructuring.

PR #212, opened 2016-03-29
Default props

levithomason:
# Not ready

This PR will add default props so we don't have to repeat common prop types.


levithomason:
This has gotten stale and offers little gain over explicitly listing common props in each component.  Closing.

PR #215, opened 2016-03-29
Label component

levithomason:
Adds `<Label />` component and tests.  A separate PR will add docs.


README.md
```diff
@@ -37,7 +37,7 @@ export default class TryStardust extends Component {
 - [x] Icon
 - [x] Image
 - [x] Input
-- [ ] Label
+- [x] Label
```
levithomason, line 40: Mark it off in the support section of the README.

package.json
```diff
@@ -8,19 +8,21 @@
     "dist"
   ],
   "scripts": {
-    "build": "rimraf dist && babel src -d dist",
-    "docs": "gulp docs",
+    "docs": "gulp",
+    "build": "npm run build:dist && npm run build:docs",
+    "build:dist": "rimraf dist && babel src -d dist",
+    "build:docs": "gulp docs",
```
levithomason, line 14: `npm run docs` didn't run docs, not intuitive.  Renamed scripts to make more sense.

src/elements/Label/Label.js
```diff
@@ -0,0 +1,168 @@
...
+  type: META.type.element,
+  props: {
+    attached: ['top', 'bottom', 'top right', 'top left', 'bottom left', 'bottom right'],
+    size: sui.sizes,
+    color: sui.colors,
+    pointing: ['bottom', 'left', 'right'],
+    corner: ['left', 'right'],
+    right: ['left', 'right'],
+    ribbon: ['right'],
+  },
```
levithomason, line 88: Enum props are documented in our `_meta` tag.  This allows us to write shared tests against them.  Also will provide better doc site features, like generating forms that control and demonstrate component functionality.

src/utils/propUtils.js
```diff
@@ -60,3 +63,15 @@ export const customPropTypes = {
   },
 }
 
+// ------------------------------------
+// Prop Renderers
+// ------------------------------------
+export const iconPropRenderer = (val) => _.isString(val) ? <Icon className={val} /> : val
+export const imagePropRenderer = (val) => _.isString(val) ? <Image src={val} /> : val
```
levithomason, line 70: Props that should behave consistently across all components will have an associated render util.  This allows us to standardize prop functionality and features.

In this case, `icon` and `image` props can render either a string or a component.

src/utils/propUtils.js
```diff
@@ -60,3 +63,15 @@ export const customPropTypes = {
...
+
+// ------------------------------------
+// Prop To Class Name
+// ------------------------------------
+export const useKeyOnly = (val, key) => val && key
+export const useValueAndKey = (val, key) => val && val !== true && `${val} ${key}`
+export const useKeyOrValueAndKey = (val, key) => val && (val === true ? key : `${val} ${key}`)
```
levithomason, line 124: There are 4 prop patterns used to build up the `className` for a component.  The one not shown above is `valueOnly()` because it would simply return `val`, so the value is used inline instead.

**Key Only**
Props where the name only is used.

`<div class="ui tag label"></div>`
`<Label tag />`

**Value Only**
Props where only the value is used.

`<div class="ui big label"></div>`
`<Label size='big' />`

**Value And Key**
Props that require a value, where className requires value and key.

`<div class="ui left corner label"></div>`
`<Label corner='left' />`

**Key || Value And Key**
Boolean props with optional value, where className includes the key or value and key.

`<div class="ui pointing label"></div>`
`<Label pointing />`

`<div class="ui left pointing label"></div>`
`<Label pointing='left' />`

package.json
```diff
@@ -8,19 +8,21 @@
...
     "lint": "eslint .",
     "lint:fix": "npm run lint -- --fix",
     "lint:watch": "watch 'npm run lint --silent' docs gulp src test",
-    "prepublish": "npm run build",
+    "prepublish": "npm run build:dist",
     "release:major": "ta-script npm/release.sh major",
     "release:minor": "ta-script npm/release.sh minor",
     "release:patch": "ta-script npm/release.sh patch",
-    "start": "gulp",
+    "start": "npm run docs",
```
ghost, line 25: It seems it may be clearer to remove the `docs` script and call `npm run gulp` here. Is there a reason one may also want to run `npm run docs` separately?

levithomason, line 25: I think `npm start` should always start the project without a need to understand the underlying implementation.  Also, if we drop `gulp` (which we hope to do), `npm start` should still start the project.  If we tie the start script to gulp, then there is unnecessary coupling that would require updates to the README and dev workflow since they'll need to run a new command.

`npm start` might not always run the doc site, it will always "start the project".  There is an `npm run docs` that will always run the doc site.

This is my rationale at least.  Open to change if there is pretty good consensus that thinks otherwise.

src/utils/childrenUtils.js
```diff
@@ -0,0 +1,40 @@
+import _ from 'lodash'
+import { Children } from 'react'
```
athurman, line 2: Was curious as to the utilities associated with `React.Children` found [documentation here](https://facebook.github.io/react/docs/top-level-api.html#react.children) and posting here for other's convenience.

test/specs/elements/Label/Label-test.js
```diff
@@ -0,0 +1,272 @@
...
+        .find('Image')
+        .should.have.tagName('img')
+    })
+
+    it('adds the value to the Image className', () => {
+      const src = faker.internet.avatar()
+      shallow(<Label image={src} />)
+        .find('Image')
+        .should.have.prop('src', src)
+    })
```
ghost, line 151: `enzyme` looks delightful.

src/elements/Label/Label.js
```diff
@@ -0,0 +1,168 @@
...
+    size,
+    color,
+    useKeyOnly(floating, 'floating'),
+    useKeyOnly(horizontal, 'horizontal'),
+    useKeyOnly(pointing, 'pointing'),
+    useKeyOnly(tag, 'tag'),
+    useValueAndKey(attached, 'attached'),
+    useValueAndKey(corner, 'corner'),
+    useKeyOrValueAndKey(pointing, 'pointing'),
+    useKeyOrValueAndKey(ribbon, 'ribbon'),
```
eanplatter, line 45: I know we've bikeshedded our way around these names quite a bit, but this still seems a little weird, could just be the nature of the problem though. I guess I don't think about properties passed into markup in key/value sort of vocab, so I am probably the one off on this.

ghost:
:sake:

test/specs/elements/Label/Label-test.js
levithomason, line 151: Note, we're also using [`chai-enzyme`](https://github.com/producthunt/chai-enzyme) to give us these end of chain assertions like `should.have.prop()`.

If we had only `enzyme` we'd have awkward chai assertions made on enzyme values.

athurman:
LGTM 🐙

src/elements/Label/Label.js
levithomason, line 45: I really struggled with these, really.  Would love any input, I want it to be intuitive and clear.  The backstory is spelled out in [this comment](https://github.com/TechnologyAdvice/stardust/pull/215#discussion_r58302514) above.

The alternative is to write these out with no function helpers.  Slightly more performant, but more error prone and duplicates a lot of logic.  Every component will have a stack of these "prop to className" transformations.

Here is what it looks like without the helper functions:

``` js
const classes = cx('sd-label ui',
  size,
  color,
  // key only
  floating && 'floating',
  horizontal && 'horizontal',
  pointing && 'pointing',
  tag && 'tag',
  // value and key
  attached && attached !== true && `${attached} ${'attached'}`,
  corner && corner !== true && `${corner} ${'corner'}`,
  // key only || value and key
  pointing && (pointing === true ? 'pointing' : `${pointing} ${'pointing'}`),
  ribbon && (ribbon === true ? 'ribbon' : `${ribbon} ${'ribbon'}`),
  // ...others
  circular && (children && 'circular' || 'empty circular'),
  // TODO how to handle image child with no image class? there are two image style labels.
  (image || someChildType(children, Image) || someChildType(children, 'img')) && 'image',
  'label',
  className
)
```

Thoughts or ideas welcome.

eanplatter, line 45: I think the helper methods are valuable. I guess just adding some documentation might help de-obfuscate the syntax. I can just see myself coming back to this code in 6 month and having to re-learn what these helpers are for.

src/elements/Label/Label.js
```diff
@@ -0,0 +1,168 @@
...
+
+  /** Size of the label. */
+  size: PropTypes.oneOf(Label._meta.props.sizes),
+
+  /** Format the label like a product tag. */
+  tag: PropTypes.bool,
+
+  /** Primary text of the label, same as children. */
+  text: PropTypes.node,
+}
```
eanplatter, line 166: Have you found a lot of benefit in making the label a pure function?

src/utils/propUtils.js
eanplatter, line 124: This information is prime. Perhaps it might be a good idea to document it somewhere in the lib like I mentioned above.

test/specs/elements/Label/Label-test.js
```diff
@@ -0,0 +1,272 @@
...
+  common.propKeyOnlyToClassName(Label, 'tag')
+
+  common.propValueOnlyToClassName(Label, 'color')
+  common.propValueOnlyToClassName(Label, 'size')
+
+  common.propKeyAndValueToClassName(Label, 'attached')
+  common.propKeyAndValueToClassName(Label, 'corner')
+
+  common.propKeyOrValueToClassName(Label, 'pointing')
+  common.propKeyOrValueToClassName(Label, 'ribbon')
```
eanplatter, line 25: :+1: I like being able to see what common tests this component needs.

eanplatter:
Looks good: :fallen_leaf:

src/elements/Label/Label.js
levithomason, line 45: Added some docs and comments.

src/elements/Label/Label.js
levithomason, line 166: Nope, it is actually more LOC and limits us from using life cycle methods.  However, they will take advantage of extra optimization in future versions of React.  So, we're getting on the bandwagon now.

PR #217, opened 2016-04-01
Create common tests

levithomason:
This PR removes the conformance test and index.js test.  These tests previously looped over the Stardust keys and did magic tests.

They have replaced and extended with a `commonTests` module that includes:
- `isConformant`
- `hasUIClassName`
- `hasSubComponents`
- `rendersChildren`
- `propNameOnlyToClassName`
- `propValueOnlyToClassName`
- `propNameAndValueToClassName`
- `propNameOrValueToClassName`


test/specs/Conformance-test.js
```diff
@@ -1,93 +0,0 @@
-import _ from 'lodash'
```
levithomason, line 1: Conformance test is now an export of `commonTests`.

test/specs/commonTests.js
```diff
@@ -0,0 +1,361 @@
...
+    filenameWithoutExt,
+  }
+})
+
+/**
+ * Assert Component conforms to guidelines that are applicable to all components.
+ * @param {React.Component} Component A component that should conform.
+ * @param {Object} requiredProps Props required to render Component without errors or warnings.
+ */
+export const isConformant = (Component, requiredProps = {}) => {
```
levithomason, line 48: All exported common tests are well documented.

test/specs/commonTests.js
```diff
@@ -0,0 +1,361 @@
...
+      shallow(createElement(Component, props, child))
+        .should.include.text(child)
+    })
+  })
+}
+
+// ----------------------------------------
+// className from prop
+// ----------------------------------------
+const _definesPropOptions = (Component, propName) => {
```
levithomason: There are a four private tests here that are re-used in the exported tests.   This allows us to mix and match these specific private tests to create our common className to prop tests below.

test/specs/commonTests.js
```diff
@@ -0,0 +1,361 @@
...
+
+    it('adds prop value to className', () => {
+      const sample = _.sample(Component._meta.props[propName])
+      shallow(createElement(Component, { ...requiredProps, [propName]: sample }))
+        .should.have.className(sample)
+    })
+
+    it('does not add prop name to className', () => {
+      // silence propType warnings
+      consoleUtil.disableOnce()
```
levithomason, line 319: Note this util.  The test below will render this component with an intentionally incorrect prop value to assert that it does not have side effects.  This will throw big ugly propType warning logs in the middle of the tests.

This util let's us disable the console for one test.  It will automatically enable the console after this one test is complete.

PR #222, opened 2016-04-26
Component utils

levithomason:
This PR is a breakout of #206.  It adds utils that facilitate writing and testing components without Semantic UI jQuery plugins.

The source is well annotated so I'm skipping comment annotations on this one.  Here is the summary:

**childrenUtils**

With our transition, we also are implementing our [v.1.0 component API](https://github.com/TechnologyAdvice/stardust/issues/208).  This means no more `className` use for component behavior and styling.  We extend child utils to facilitate some of this.

**domEvent**

With jQuery out of the picture we now need to handle events.  Both creating and dispatching them in tests.  Simulant gives a cross browser way to create real browser events (we're in phantomjs).  `domEvent` is our wrapper implementation of `simulant`.

**syntheticEvent**

Real browser events alone won't do, our tests rely on React's synthetic events.  This util is just that, basically an enum of synthetic events and their shape.  These are used to test component events.

### DIFF
--- a/test/utils/domEvent.js
+++ b/test/utils/domEvent.js
@@ -1,0 +1,35 @@
+import simulant from 'simulant'
+
+// ----------------------------------------
+// Simulate DOM Events on real DOM nodes
+// ----------------------------------------
+
+/**
+ * Generic method for dispatching an event on a DOM node.
+ * @param {String|Object} node A querySelector string or DOM node.
+ * @param {String} eventType A DOMString
+ * @param {Object} [data] Additional event data.
+ * @returns {Object} The event
+ */
+export const fire = (node, eventType, data = {}) => {
+  const DOMNode = typeof node === 'string' ? document.querySelector(node) : node
+  const event = simulant(eventType, data)
+
+  return simulant.fire(DOMNode, event)
+}
+
+/**
+ * Dispatch a 'keydown' event on a DOM node.
+ * @param {String|Object} node A querySelector string or DOM node.
+ * @param {Object} [data] Additional event data.
+ * @returns {Object} The event
+ */
+export const keyDown = (node, data) => fire(node, 'keydown', data)
+
+/**
+ * Dispatch a 'click' event on a DOM node.
+ * @param {String|Object} node A querySelector string or DOM node.
+ * @param {Object} [data] Additional event data.
+ * @returns {Object} The event
+ */
+export const click = (node, data) => fire(node, 'click', data)
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "rimraf": "^2.5.2",
     "sass-loader": "^3.2.0",
     "semantic-ui-css": "^2.1.8",
+    "simulant": "^0.2.1",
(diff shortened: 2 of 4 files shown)


test/utils/domEvent.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Generic method for dispatching an event on a DOM node.
+ * @param {String|Object} node A querySelector string or DOM node.
+ * @param {String} eventType A DOMString
+ * @param {Object} [data] Additional event data.
+ * @returns {Object} The event
+ */
+export const fire = (node, eventType, data = {}) => {
+  const DOMNode = typeof node === 'string' ? document.querySelector(node) : node
+  // TODO remove hack once merged: https://github.com/Rich-Harris/simulant/pull/11
```
levithomason: Waiting, https://github.com/Rich-Harris/simulant/pull/11#issuecomment-214890331

kyleturco: Looks like he merged that in 👍

test/utils/domEvent.js
```diff
@@ -0,0 +1,46 @@
...
+  // simulant has issues
+  //   - throws if params is not defined with 'keydown'
+  //   - does not extend with 'code' param, which we need for 'keydown'
+  // to solve these we:
+  //   - use an empty params object
+  //   - assign our own params
+
+  const DONT_USE_THESE_PARAMS = {}
+  const event = simulant(eventType, DONT_USE_THESE_PARAMS)
```
kyleturco: Just curious - What is `DONT_USE_THESE_PARAMS` doing? Seems like an intense name

kyleturco:
🐓  Looks great!

test/utils/domEvent.js
```diff
@@ -0,0 +1,46 @@
...
+  return extendedEvent
+}
+
+/**
+ * Dispatch a 'keydown' event on a DOM node.
+ * @param {String|Object} node A querySelector string or DOM node.
+ * @param {Object} [data] Additional event data.
+ * @returns {Object} The event
+ */
+export const keyDown = (node, data) => fire(node, 'keydown', data)
```
dvdzkwsk, line 27: Just because I have to whenever I see repeated code like this:

``` js
const fire = curry((event, node, data) => { ... })

export const keyDown = fire('keydown')
export const click = fire('click')

click(node, data)
```

test/utils/domEvent.js
levithomason: Woop, updating.

test/utils/domEvent.js
levithomason: Aye, intense for exactly that reason.  The bug I fixed in `simulant` was centered around passing in event data in that arg.  Wanted to ensure no one ever put data in that object, it had to remain empty.

With the bug fixed merged and shipped, I can remove this now.

test/utils/domEvent.js
levithomason, line 27: May implement this on the cleanup PR once the breakout is done being merged.

PR #224, opened 2016-04-27
AutoControlledComponent

levithomason:
# AutoControlledComponent

This PR is a breakout of #206.  This PR is actually tiny. It is mostly runtime warnings, tests, and comments.  I've committed a "slim" version of the component with no runtime warnings nor comments so it can be reviewed easier.  **It is only 34 SLOC**.  I will remove it before merging.
## Why this thing?

Many of our components have props that are possibly [controlled](https://facebook.github.io/react/docs/forms.html#controlled-components) by the consumer.  Many of those props should also be self managed inside the component if there is no prop given.

Example, `<Dropdown />` has a `value` prop.  This is the `value` of the active item.  When the user clicks on an item, we set the value on state internally.  However, the consumer may have added a value prop to the `<Dropdown />` as well.  Now, when we want to access the value, we need to conditionally choose between which value we retrieve.  The prop or the state?

Also, it would be responsible to never set state for the value if we were passed a value prop.  This saves a render.  Yet, another conditional, if there is no value prop then set state.  This quickly becomes ugly and hard to maintain.
## Community Options

We reviewed and tested other libraries solving this same problem:
- [react-controllables](https://github.com/matthewwithanm/react-controllables)
- [uncontrollable](https://github.com/jquense/uncontrollable) (react-bootstrap)

These didn't work for us for the following:
- dictates a single handler for each prop
- dictates that each handler call back with the new prop value as the first argument to update props
- HOC/decorator pattern causes troubles for our static props/subcomponents
- HOC/decorator pattern cannot set computed props without a double initial render
- Overall too restrictive in the implementation
## Solution

<p align=center>
<br />
What we really want is a smart setState() that submits to controlled props.
<br />
<br />
</p>

This solves all our issues.  AutoControlledComponent is extended by your component, giving it a `trySetState` method and static `autoControlledProps` array.
#### `static autoControlledProps = []`

This array defines which props are auto controlled.

**`props` to `state`**
Auto controlled props are made available on `state`.  The value will automatically defer to prop values if present (controlled) or the internal state value if there is no prop (uncontrolled).
#### `trySetState(maybeState, [state])`

**`maybeState`**
When trying to set maybe state, controlled prop values always win.  Unnecessary renders are also prevented.

**`state`**
Allows setting regular state while simultaneously trying to set state for controlled values.  You often need to set both, this allows a single call to setState and a single render.
## Niceties

**Auto default props `defaultFoo`**

Each auto controlled prop also accepts a `default*` prop.  It works exactly like vanilla React, only applied on first render.

**Runtime Warnings**

The majority of the code is actually run time warnings to ensure proper use of the abstraction.  Warnings are removed in production by dead code elimination.
- prevent defaultFoo={} along side foo={} _(follows vanilla React)_
- require static autoControlledProps
- require propTypes for autoControlledProps
- require propTypes for corresponding default autoControlledProps
- prevent listing defaultProps in autoControlledProps _(automatically added)_
- prevent autoControlledProps in defaultProps _(trySetState on construction instead)_
- prevent trySetState for props that are not auto controlled

### DIFF
--- a/build/karma.conf.babel.js
+++ b/build/karma.conf.babel.js
@@ -1,6 +1,9 @@
 import { argv } from 'yargs'
+import webpack from 'webpack'
+
 import config from '../config'
 import webpackConfig from './webpack.config'
+
 const { paths } = config
 
 module.exports = (karmaConfig) => {
@@ -35,6 +38,9 @@ module.exports = (karmaConfig) => {
           ...webpackConfig.module.loaders,
         ],
       },
+      plugins: [
+        new webpack.DefinePlugin(config.compiler_globals),
+      ],
       resolve: {
         ...webpackConfig.resolve,
         alias: {
--- a/src/utils/AutoControlledComponent.js
+++ b/src/utils/AutoControlledComponent.js
@@ -1,0 +1,145 @@
+/* eslint-disable no-console */
+/**
+ * Why choose inheritance over a HOC?  Multiple advantages for this particular use case.
+ * In short, we need identical functionality to setState(), unless there is a prop defined
+ * for the state key.  Also:
+ *
+ * 1. Single Renders
+ *    Calling trySetState() in constructor(), componentWillMount(), or componentWillReceiveProps()
+ *    does not cause two renders. Consumers and tests do not have to wait two renders to get state.
+ *
+ * 2. Simple Testing
+ *    Using a HOC means you must either test the undecorated component or test through the decorator.
+ *    Testing the undecorated component means you must mock the decorator functionality.
+ *    Testing through the HOC means you can not simply shallow render your component.
+ *
+ * 3. Statics
+ *    HOC wrap instances, so statics are no longer accessible.  They can be hoisted, but this is more
+ *    looping over properties and storing references.  We rely heavily on statics for testing and sub
+ *    components.
+ *
+ * 4. Instance Methods
+ *    Some instances methods may be exposed to users via refs.  Again, these are lost with HOC unless
+ *    hoisted and exposed by the HOC.
+ */
+import _ from 'lodash'
+import { Component } from 'react'
+
+const getDefaultPropName = (prop) => `default${prop[0].toUpperCase() + prop.slice(1)}`
+
+export default class AutoControlledComponent extends Component {
+  componentWillMount() {
+    if (super.componentWillMount) super.componentWillMount()
+    const { autoControlledProps } = this.constructor
+
+    // Auto controlled props are copied to state.
+    // Set initial state by copying auto controlled props to state.
+    // Also look for the default prop for any auto controlled props (foo => defaultFoo)
+    // so we can set initial values from defaults.
+    this.state = _.transform(autoControlledProps, (res, prop) => {
+      const defaultPropName = getDefaultPropName(prop)
+
+      // apply default props if they exist
+      res[prop] = this.props[defaultPropName in this.props ? defaultPropName : prop]
+
+      if (!__PROD__) {
+        const { name } = this.constructor
+        // prevent defaultFoo={} along side foo={}
+        if (defaultPropName in this.props && prop in this.props) {
+          console.error(
+            `${name} prop "${prop}" is auto controlled. Specify either ${defaultPropName} or ${prop}, but not both.`
+          )
+        }
+      }
+    }, {})
+
+    if (!__PROD__) {
+      const { defaultProps, name, propTypes } = this.constructor
+      // require static autoControlledProps
+      if (!autoControlledProps) {
+        console.error(`Auto controlled ${name} must specify a static autoControlledProps array.`)
+      }
+
+      // require propTypes
+      _.each(autoControlledProps, (prop) => {
+        const defaultProp = getDefaultPropName(prop)
+        // regular prop
+        if (!_.has(propTypes, defaultProp)) {
+          console.error(`${name} is missing "${defaultProp}" propTypes validation for auto controlled prop "${prop}".`)
+        }
+        // its default prop
+        if (!_.has(propTypes, prop)) {
+          console.error(`${name} is missing propTypes validation for auto controlled prop "${prop}".`)
+        }
+      })
+
+      // prevent autoControlledProps in defaultProps
+      //
+      // When setting state, auto controlled props values always win (so the parent can manage them).
+      // It is not reasonable to decipher the difference between props from the parent and defaultProps.
+      // Allowing defaultProps results in trySetState always deferring to the defaultProp value.
+      // Auto controlled props also listed in defaultProps can never be updated.
+      const illegalDefaults = _.intersection(autoControlledProps, _.keys(defaultProps))
+      if (!_.isEmpty(illegalDefaults)) {
+        console.error([
+          'Do not set defaultProps for autoControlledProps,',
+          'use trySetState() in constructor() or componentWillMount() instead.',
+          `See ${name} props: "${illegalDefaults}".`,
+        ].join(' '))
+      }
+
+      // prevent listing defaultProps in autoControlledProps
+      //
+      // Default props are automatically handled.
+      // Listing defaults in autoControlledProps would result in allowing defaultDefaultValue props.
+      const illegalAutoControlled = _.filter(autoControlledProps, prop => prop.startsWith('default'))
+      if (!_.isEmpty(illegalAutoControlled)) {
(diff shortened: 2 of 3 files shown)


build/karma.conf.babel.js
```diff
@@ -35,6 +38,9 @@ module.exports = (karmaConfig) => {
           ...webpackConfig.module.loaders,
         ],
       },
+      plugins: [
+        new webpack.DefinePlugin(config.compiler_globals),
+      ],
```
levithomason, line 43: Needed compiler globals for `__PROD__` use.

src/utils/AutoControlledComponent-slim.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable no-console */
```
levithomason: Review this file, this is identical to actual component, but with no comments nor run time warnings.  It is only the production code.  This makes it about 1/3 the size and much easier to grok.

build/karma.conf.babel.js
dvdzkwsk, line 43: Recommend using `process.env.NODE_ENV`. It's more verbose, but is expected in webpack builds nowadays (esp. for React), whereas the former is just an internal helper. This way users won't have to have `__PROD__` as a global.

src/utils/AutoControlledComponent-slim.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable */
+import _ from 'lodash'
+import { Component } from 'react'
+
+const toDefault = (prop) => `default${prop.slice(0, 1).toUpperCase() + prop.slice(1)}`
```
dvdzkwsk: `.slice(0, 1)` === `prop[0]`, right?

src/utils/AutoControlledComponent.js
```diff
@@ -0,0 +1,140 @@
...
+    const { autoControlledProps } = this.constructor
+
+    this.state = _.transform(autoControlledProps, (res, prop) => {
+      const defaultProp = toDefault(prop)
+
+      // apply default props if they exist
+      res[prop] = this.props[defaultProp in this.props ? defaultProp : prop]
+
+      if (!__PROD__) {
+        const { name } = this.constructor
```
dvdzkwsk, line 46: grabbing `this.constructor.name` can be done outside of the loop.

dvdzkwsk, line 46: ignore me, it's prod only. I'm dumb.

src/utils/AutoControlledComponent.js
```diff
@@ -0,0 +1,140 @@
...
+        // prevent defaultFoo={} along side foo={}
+        if (defaultProp in this.props && prop in this.props) {
+          console.error(
+            `${name} prop "${prop}" is auto controlled. Specify either ${defaultProp} or ${prop}, but not both.`
+          )
+        }
+      }
+    }, {})
+
+    if (!__PROD__) {
```
dvdzkwsk, line 56: 👍  awesome developer utilities.

src/utils/AutoControlledComponent-slim.js
levithomason: don't school me publicly.... 😝

levithomason: updated!

src/utils/AutoControlledComponent-slim.js
```diff
@@ -0,0 +1,44 @@
...
+    this.state = _.transform(autoControlledProps, (res, prop) => {
+      const defaultProp = toDefault(prop)
+
+      // apply default props if they exist
+      res[prop] = this.props[defaultProp in this.props ? defaultProp : prop]
+    }, {})
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (super.componentWillReceiveProps) super.componentWillReceiveProps(nextProps)
```
ghost: Is there ever a case where `super.componentWillReceiveProps` would ever return falsy given `super` represents `React.Component` here?

levithomason: Hm, this seems intuitive.  However, when removing the checks we get this in tests:

``` shell
undefined is not an object (evaluating '_get(Object.getPrototypeOf(AutoControlledComponent.prototype), 'componentWillReceiveProps', this).call')
```

I'm not sure how React is constructing these classes, but apparently they do not have all their methods at least in tests.

build/karma.conf.babel.js
levithomason, line 43: Added a task to the main PR.  Will handle in the final cleanup/refactor.

src/utils/AutoControlledComponent-slim.js
levithomason: Note, I added the missing check/call to the `componentWillMount()` as well.

ghost:
⛵

src/utils/AutoControlledComponent-slim.js
```diff
@@ -0,0 +1,45 @@
...
+import { Component } from 'react'
+
+const toDefault = (prop) => `default${prop[0].toUpperCase() + prop.slice(1)}`
+
+export default class AutoControlledComponent extends Component {
+  componentWillMount() {
+    if (super.componentWillMount) super.componentWillMount()
+    const { autoControlledProps } = this.constructor
+
+    this.state = _.transform(autoControlledProps, (res, prop) => {
```
dvdzkwsk: Wouldw be cool to get some comments around this. It took me a couple minutes to fully grok it especially due to the generically named to `toDefault`. Not at all a blocker, just a note.

levithomason: react-controllables went with `fromDefaultName`, I'm indifferent but want it to make the most sense to the most people.

src/utils/AutoControlledComponent-slim.js
```diff
@@ -0,0 +1,45 @@
...
+    // props always win, update state with all auto controlled prop
+    const newState = _.pick(nextProps, this.constructor.autoControlledProps)
+    if (!_.isEmpty(newState)) this.setState(newState)
+  }
+
+  /**
+   * Safely attempt to set state for props that might be controlled by the user.
+   * Second argument is a state object that is always passed to setState.
+   * @param {object} maybeState State that corresponds to controlled props.
+   * @param {object} [state] Actual state, useful when you also need to setState.
```
dvdzkwsk: What's the rationale for also providing the old state here (especially if it will just override some of the newState)? Couldn't the caller just call `this.setState()` after `this.trySetState()` since they will be batched anyways?

levithomason: From the PR description:

> #### `trySetState(maybeState, [state])`
> 
> **`maybeState`**
> When trying to set maybe state, controlled prop values always win.  Unnecessary renders are also prevented.
> 
> **`state`**
> Allows setting regular state while simultaneously trying to set state for controlled values.  You often need to set both, this allows a single call to setState and a single render.

dvdzkwsk: I understand the description, but I don't understand _why_ it's necessary since the objective of the method is to only potentially apply some state.

``` js
foo() {
  this.trySetState(...)
  this.setState()

  // only 1 render is produced
}
```

It's just confusing to me to see the ability to _override_ `maybeState` with actual state in a method designed to simply apply `maybeState`.

levithomason: TL;DR - allows single render for setting state.

You can see it in action all over in the new dropdown, like [here](https://github.com/TechnologyAdvice/stardust/pull/226/files#diff-609214949026e594dbfd98481664c928R443).

Also, try set state is really just smart set state so it makes sense to set actual state along with maybe state in a single method.  I contemplated parsing a single state object into maybe state and normal state but this was cleaner, allowing for better warnings in dev as well.

dvdzkwsk:
👻  - still a bit weary of the state overrides in `trySetState`, but you're the designer here and probably have a good reason. Merge at will, but would be happy to continue the discussion.

src/utils/AutoControlledComponent-slim.js
levithomason: We solved this offline already, for the record, this http://www.react.run/4kJFdKoxb/27 shows how the single vs double render happens when using vanilla DOM events vs React events.

Since we have to have DOM events for the "click outside" and other features, we're in the pickle.

levithomason:
Addressed all comments in slim file and removed it, merging after tests pass.

PR #226, opened 2016-04-27
Dropdown without jQuery

levithomason:
# Hey...

This is an interim PR.  There are todos, comments, missing features, and ugly things.  There shouldn't be any bugs or cruft, though.

The rewrite thus far is >600 lines in the Dropdown and >700 lines in Dropdown tests.  I want to get some eyes and feedback on the functionality and approach thus far before cleaning up and abstracting things.

The `multiple` feature will be added in the next Dropdown PR.
## Areas of concern

These are the areas to watch and comment on.  These are some known ugly parts.  They've been added as TODOs to the main PR #206.
### Events

We have two event systems at play, React synthetic events and real DOM events.  You'll see lots of `document.add/removeEventListener` calls.  This is necessary for keyboard and mouse events that are dispatched on the document.  Example, clicking outside the dropdown or pressing escape to close an open dropdown.  Ideally, we'll write a clean abstraction for handling these.  Considering that we need to:
- prevent duplicate handlers
- cleanup all events on unmount
- provide once() functionality
- add/remove events at any time (not just lifecycle methods)
### Debug / Logging

Currently, nearly every method has a `console.debug()` when called and a `console.log()` for any relevant debugging data.  This would be pulled into an actual debug util.  Consider:
- [debug](https://www.npmjs.com/package/debug) module
- removal in production code
- runtime flag to enable/disable by module
### Spreading Consumer Props

Stardust components should spread the consumer's props.  We've experimented with a couple patterns for doing this effectively, though they each beak down.  Consider:
- every prop not defined in propTypes (public interface) is a consumer prop
- propTypes may be stripped from production code in the future
- using `...props` in `render()` isn't sufficient when breaking down many render methods

If it weren't for stripping propTypes in production we may be able to naively:

``` js
const getConsumerProps = (self) => _.omit(self.props, _.keys(self.constructor.propTypes))
```


docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
```diff
@@ -25,22 +28,31 @@ export default class DropdownAsyncOptions extends Component {
...
           onChange={this.handleChange}
+          disabled={isFetching}
+          loading={isFetching}
         />
+        <p>
+          <label>
+            <input type='checkbox' value={search} onChange={this.toggleSearch} />
+            {' '}Search
+          </label>
+        </p>
```
levithomason, line 55: The docs are getting an overhaul to better demonstrate the v1.0 component APIs.  Take this example with a grain of salt.  It is here only to show that the Dropdown works as expected.

docs/app/Examples/modules/Dropdown/Types/Dropdown.js
```diff
@@ -1,31 +0,0 @@
...
-            <Dropdown.Item text='Google Drive' />
-            <Dropdown.Item text='Dropbox' />
-            <Dropdown.Item text='Adobe Creative Cloud' />
-            <Dropdown.Item text='Private FTP' />
-            <Dropdown.Item text='Another Service...' />
-          </Dropdown.Menu>
-        </Dropdown.Item>
-        {/* item text can also be defined as children */}
-        <Dropdown.Item>E-mail Collaborators</Dropdown.Item>
-      </Dropdown>
```
levithomason, line 28: We are not going to ship v1.0 with support for building custom Dropdown components in this way.  There is work yet to be done in the way of handling the `options` prop in conjunction with `Dropdown.Item` children.  Especially when it comes to nested menus and searching features.

Currently, we only support the `options` prop which gives a SSOT and makes searching a breeze.

docs/app/Examples/modules/Dropdown/Types/Types.js
```diff
@@ -1,23 +0,0 @@
...
-          title='Dropdown'
-          description='A dropdown'
-          examplePath='modules/Dropdown/Types/Dropdown'
-        >
-          <Message>
-            Dropdown <code>children</code> are automatically wrapped in a
-            {' '}<code>{`<Dropdown.Menu />`}</code> for convenience.
-          </Message>
-        </ComponentExample>
-      </ExampleSection>
```
levithomason, line 20: Related custom dropdown doc component.

package.json
```diff
@@ -43,6 +43,7 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "classnames": "^2.1.5",
+    "keyboardevent-key-polyfill": "github:technologyadvice/keyboardevent-key-polyfill#0eb142e8d6845a90f23d21b468a3c8c8162b5a25",
```
levithomason, line 46: With every other event property you're used to using having been deprecated:

image

...and their successor property [`key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) being only in working draft status with limited support:

image

We are using a polyfill to ensure the `key` property is available on all keyboard events.  That said, this polyfill was missing the space key so we forked and updated.  Just [waiting on a release](https://github.com/cvan/keyboardevent-key-polyfill/issues/10) so we can go back to the official module.

src/index.js
```diff
@@ -1,5 +1,8 @@
+import keyboardeventKeyPolyfill from 'keyboardevent-key-polyfill'
 import { deprecateComponent } from './utils/deprecate'
 
+keyboardeventKeyPolyfill.polyfill()
```
levithomason, line 4: Considering the other comment surrounding babel-polyfill, I could see pulling the functionality of this polyfill into a util instead so that we're not touching the global namespace.  This polyfill of course [tampers with the KeyboardEvent.prototype](https://github.com/cvan/keyboardevent-key-polyfill/blob/master/index.js#L109).

docs/app/Examples/modules/Dropdown/States/Disabled.js
```diff
@@ -4,7 +4,7 @@ import { Dropdown } from 'stardust'
 export default class DropdownDisabledExample extends Component {
   render() {
     return (
-      <Dropdown className='disabled' text='Dropdown'>
+      <Dropdown className='disabled' text='Dropdown' options={[]}>
```
levithomason: Options are required.

src/modules/Dropdown/Dropdown.js
```diff
@@ -82,121 +66,611 @@ export default class Dropdown extends Component {
...
+    this.scrollSelectedItemIntoView()
+  }
+
+  // ----------------------------------------
+  // Behavior
+  // ----------------------------------------
+
+  scrollSelectedItemIntoView = () => {
+    const menu = document.querySelector('.ui.dropdown .menu.visible')
+    const item = document.querySelector('.ui.dropdown .menu.visible .item.selected')
```
levithomason: This won't work with multiple dropdowns on the page, consider using refs instead.

docs/app/Examples/modules/Dropdown/States/Disabled.js
ghost: I'm sure there's a good reason for this, but in a few words what is it?

levithomason: options are now required.  we dropped support for "dropdown file menu" type dropdowns for now. so every dropdown is a select style dropdown.

ghost: Rewording. It seems unexpected would we _require_ options. Shouldn't options be optional? If this is temporary no biggie. But it caught me off-guard.

src/index.js
ghost, line 4: Tamping is how Prototype got the job done! :trollface:

src/modules/Dropdown/Dropdown.js
```diff
@@ -82,121 +66,591 @@ export default class Dropdown extends Component {
...
+
+    /** Make the dropdown options searchable. */
+    search: PropTypes.bool,
+
+    /** The text displayed in the dropdown, usually for the active item. */
+    text: PropTypes.string,
 
-  static defaultProps = {
-    icon: 'dropdown',
+    // TODO 'searchInMenu' or 'search='in menu' or ???  How to handle this markup and functionality?
```
ghost, line 140: Just confirming we want this TODO in here.

src/modules/Dropdown/Dropdown.js
```diff
@@ -82,121 +66,591 @@ export default class Dropdown extends Component {
...
+    if (nextProps.value !== this.state.value) {
+      this.setValue(nextProps.value)
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) { // eslint-disable-line complexity
+    // console.debug('Dropdown.componentDidUpdate()')
+    // console.log('changed state:', _.transform(this.state, (res, val, key) => {
+    //   if (val !== prevState[key]) res[key] = val
+    // }, {}))
```
ghost, line 200: Also, should we pull out this commented out code, or is this addressed in one of the related PRs?

src/modules/Dropdown/Dropdown.js
```diff
@@ -82,121 +66,591 @@ export default class Dropdown extends Component {
...
+        e.preventDefault()
+        this.moveSelectionBy(1)
+        break
+      case 'ArrowUp':
+        e.preventDefault()
+        this.moveSelectionBy(-1)
+        break
+      default:
+        break
     }
```
ghost, line 277: Keyboard accessible drop-downs! Huzzah

ghost:
Lots of code in here, but I passed over and nothing other than the required `options` attribute stood out at me. Sweet looking tests! 🍶

src/index.js
levithomason, line 4: I'm not sure what my resolve on this is yet.  The polyfill does a good job of checking to see if it needs to be applied before running.  At the same time, any consumer who uses stardust is going to have their KeyboardEvent.prototype polyfilled underneath them.  That feels like a bad move.

The util would essentially be exactly what the polyfill is, a match from `keyCode` / `which` values to `key` values.  Just in a function instead of on the proto.

After talking through this, I'm leaning heavily toward util actually.  Will add a TODO to the main PR #206.

docs/app/Examples/modules/Dropdown/States/Disabled.js
levithomason: These are equivalent to `<select />` options.  I can see the argument in either case though.  I'll make them optional and provide a default empty array instead.

src/modules/Dropdown/Dropdown.js
levithomason, line 140: aye, thanks. this is a reminder to handle the [search in menu](http://semantic-ui.com/modules/dropdown.html#search-in-menu) dropdown.

src/modules/Dropdown/Dropdown.js
levithomason, line 200: Yep, from the description:

> ### Areas of concern
> 
> These are the areas to watch and comment on. These are some known ugly parts. They've been added as TODOs to the main PR #206.
> 
> ### Debug / Logging
> 
> Currently, nearly every method has a console.debug() when called and a console.log() for any relevant debugging data. This would be pulled into an actual debug util. Consider:
> - debug module
> - removal in production code
> - runtime flag to enable/disable by module
>   ...

PR #227, opened 2016-04-29
Remove eslint-loader, for performance

levithomason:
Build times are insane right now.  _Mostly_ due to `eslint-loader`:

With: `webpack built in 38492ms`
Without: `webpack built in 13282ms`

:scream: 65% reduction in build times without it!


dvdzkwsk:
Let it be so.

👻

ghost:
🐎

PR #228, opened 2016-04-29
Add babel-plugin-lodash for smaller bundles

levithomason:
Instead of importing the entire lodash lib (>300 funcs) [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) transforms our imports to only include the functions we actually use.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,5 +3,8 @@
     "es2015",
     "react",
     "stage-1"
+  ],
+  "plugins": [
+    "lodash"
   ]
 }
--- a/docs/app/Components/Sidebar/Sidebar.js
+++ b/docs/app/Components/Sidebar/Sidebar.js
@@ -1,4 +1,4 @@
-import _ from 'lodash'
+import _ from 'lodash/fp'
 import React, { Component } from 'react'
 import * as stardust from 'stardust'
 import META from 'src/utils/Meta'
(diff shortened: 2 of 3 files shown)


.babelrc
```diff
@@ -3,5 +3,8 @@
     "es2015",
     "react",
     "stage-1"
+  ],
+  "plugins": [
+    "lodash"
```
levithomason, line 8: Add the babel plugin.

docs/app/Components/Sidebar/Sidebar.js
```diff
@@ -11,23 +11,23 @@ export default class Sidebar extends Component {
...
 
   getComponentsByType = type => {
-    const items = _(this.getComponentsByQuery())
-      .filter(component => META.isType(component, type))
-      .sortBy((component, name) => name)
-      .map(component => {
+    const items = _.flow(
+      _.filter(component => META.isType(component, type)),
+      _.sortBy((component, name) => name),
+      _.map(component => {
```
levithomason, line 26: Chaining is not supported, this was the one place we used it.  Refactored to use [`lodash/fp`](https://github.com/lodash/lodash/wiki/FP-Guide) now with [`_.flow`](https://lodash.com/docs#flow).

This is simply composing functions.

PR #229, opened 2016-04-30
Remove babel-polyfill

levithomason:
As noted in #206, this PR removes the `babel-polyfill` for it's size and global namespace pollution.  We don't want to drop this on our consumers.

### DIFF
--- a/build/karma.conf.babel.js
+++ b/build/karma.conf.babel.js
@@ -13,10 +13,16 @@ module.exports = (karmaConfig) => {
     singleRun: !argv.watch,
     reporters: ['mocha'],
     files: [
-      require.resolve('babel-polyfill/browser'),
       './test/tests.bundle.js',
     ],
-    frameworks: ['mocha'],
+    frameworks: [
+      'phantomjs-shim',
+      'mocha',
+    ],
+    phantomjsLauncher: {
+      // exit on ResourceError, useful if karma exits without killing phantom
+      exitOnResourceError: true,
+    },
     preprocessors: {
       '**/*.bundle.js': ['webpack'],
     },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     "karma-mocha": "^0.2.2",
     "karma-mocha-reporter": "^2.0.0",
     "karma-phantomjs-launcher": "^1.0.0",
+    "karma-phantomjs-shim": "github:technologyadvice/karma-phantomjs-shim",
```
levithomason, line 91: Can go back to main branch once ~~String.repeat~~ and String.includes are merged:

https://github.com/tschaub/karma-phantomjs-shim/pull/6
~~https://github.com/tschaub/karma-phantomjs-shim/pull/7~~

levithomason:
We could also use lodash repeat/includes, however, chai-enzyme also uses `includes` for `.should.contain.text()` assertions.  So there is no way around that one.

build/karma.conf.babel.js
```diff
@@ -13,10 +13,16 @@ module.exports = (karmaConfig) => {
...
       './test/tests.bundle.js',
     ],
-    frameworks: ['mocha'],
+    frameworks: [
+      'phantomjs-shim',
+      'mocha',
+    ],
+    phantomjsLauncher: {
+      // exit on ResourceError, useful if karma exits without killing phantom
+      exitOnResourceError: true,
```
dvdzkwsk, line 24: Interesting, didn't know about this.

levithomason, line 24: Me either, just saw it in the docs while debugging something else.

dvdzkwsk:
👻

dvdzkwsk:
(w/ tests passing)

package.json
tschaub, line 91: `karma-phantomjs-shim@1.4.0` published with both `includes` and `repeat`.  Thanks for the contribution.

PR #231, opened 2016-05-07
Remove keyboard event polyfill

levithomason:
Addresses todo in #206.  Removes the keyboard-event-key-polyfill in favor of a ponyfill util.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "classnames": "^2.1.5",
-    "keyboardevent-key-polyfill": "github:technologyadvice/keyboardevent-key-polyfill#0eb142e8d6845a90f23d21b468a3c8c8162b5a25",
     "lodash": "^4.6.1"
   },
   "devDependencies": {
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,4 @@
-import keyboardeventKeyPolyfill from 'keyboardevent-key-polyfill'
 import { deprecateComponent } from './utils/deprecate'
-
-keyboardeventKeyPolyfill.polyfill()
 
 // ----------------------------------------
 // Addons
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -7,6 +7,7 @@ import React, { PropTypes } from 'react'
 import META from '../../utils/Meta'
 import { useKeyOnly, useKeyOrValueAndKey } from '../../utils/propUtils'
 import { iconPropRenderer } from '../../utils/propUtils'
+import keyboardKey from '../../utils/keyboardKey'
 
 // import as Component so eslint react/prop-types recognizes Dropdown
 import { default as Component } from '../../utils/AutoControlledComponent'
@@ -255,20 +256,20 @@ export default class Dropdown extends Component {
   }
 
   closeOnEscape = (e) => {
-    if (e.key !== 'Escape') return
+    if (keyboardKey.getCode(e) !== keyboardKey.Escape) return
     e.preventDefault()
     this.close()
   }
 
   moveSelectionOnKeyDown = (e) => {
     // console.debug('Dropdown.moveSelectionOnKeyDown()')
-    // console.log(`key: ${e.key}`)
-    switch (e.key) {
-      case 'ArrowDown':
+    // console.log(`key: ${keyboardKey.getCode(e)}`)
+    switch (keyboardKey.getCode(e)) {
+      case keyboardKey.ArrowDown:
         e.preventDefault()
         this.moveSelectionBy(1)
         break
-      case 'ArrowUp':
+      case keyboardKey.ArrowUp:
         e.preventDefault()
         this.moveSelectionBy(-1)
         break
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -44,7 +44,6 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "classnames": "^2.1.5",
-    "keyboardevent-key-polyfill": "github:technologyadvice/keyboardevent-key-polyfill#0eb142e8d6845a90f23d21b468a3c8c8162b5a25",
```
levithomason, line 47: Bye bye dep.

src/index.js
```diff
@@ -1,8 +1,5 @@
-import keyboardeventKeyPolyfill from 'keyboardevent-key-polyfill'
 import { deprecateComponent } from './utils/deprecate'
 
-keyboardeventKeyPolyfill.polyfill()
```
levithomason, line 4: No more tampering with the prototype.

src/modules/Dropdown/Dropdown.js
```diff
@@ -255,20 +256,20 @@ export default class Dropdown extends Component {
   }
 
   closeOnEscape = (e) => {
-    if (e.key !== 'Escape') return
+    if (keyboardKey.getCode(e) !== keyboardKey.Escape) return
```
levithomason, line 259: Util includes `getCode` and `getName` methods, as well as constants for all keyCode/which codes by key name.

src/index.js
dvdzkwsk, line 4: 👍

src/modules/Dropdown/Dropdown.js
```diff
@@ -354,9 +356,9 @@ export default class Dropdown extends Component {
 
   handleSearchKeyDown = (e) => {
     // console.debug('Dropdown.handleSearchKeyDown()')
-    // console.log(e.key)
-    switch (e.key) {
-      case 'ArrowDown':
+    // console.log(keyboardKey.getCode(e))
+    switch (keyboardKey.getCode(e)) {
+      case keyboardKey.Spacebar:
```
dvdzkwsk: ArrowDown -> Spacebar, was this intentional?

src/modules/Dropdown/Dropdown.js
```diff
@@ -278,20 +279,21 @@ export default class Dropdown extends Component {
   }
 
   openOnSpace = (e) => {
-    if (e.key !== ' ') return
+    if (keyboardKey.getCode(e) !== keyboardKey.Spacebar) return
```
dvdzkwsk, line 282: 👍 for Spacebar over `' '`

src/utils/keyboardKey.js
```diff
@@ -0,0 +1,295 @@
...
+  /**
+   * Get the `keyCode` or `which` value from a keyboard event or `key` name.
+   * @param {string|object} name A keyboard event like object or `key` name.
+   * @param {string} [name.key] If object, it must have one of these keys.
+   * @param {string} [name.keyCode] If object, it must have one of these keys.
+   * @param {string} [name.which] If object, it must have one of these keys.
+   * @returns {*}
+   */
+  getCode(name) {
+    return _.isObject(name) ? name.keyCode || name.which || this[name.key] : this[name]
```
dvdzkwsk: I get the intention, but could be more readable. Not a blocker.

src/utils/keyboardKey.js
```diff
@@ -0,0 +1,295 @@
...
+   * @param {number} [code.shiftKey] If object, it must have one of these keys.
+   * @returns {*}
+   */
+  getName(code) {
+    const isEvent = _.isObject(code)
+    let name = codes[isEvent ? code.keyCode || code.which : code]
+
+    if (Array.isArray(name)) {
+      if (isEvent) {
+        name = name[+code.shiftKey]
```
dvdzkwsk: is `shiftKey` a boolean or a number? I see you're either accessing `0` or `1`, but my morning brain is having a bit of trouble understanding how you're getting the correct one.

src/utils/keyboardKey.js
```diff
@@ -0,0 +1,295 @@
...
+  Control: 17,
+  Alt: 18,
+  Pause: 19,
+  CapsLock: 20,
+  Escape: 27,
+  Convert: 28,
+  NonConvert: 29,
+  Accept: 30,
+  ModeChange: 31,
+  ' ': 32,
```
dvdzkwsk, line 149: Spacebar?

dvdzkwsk, line 149: Oh, looks like that's below...

src/modules/Dropdown/Dropdown.js
levithomason: Hm, that seems wrong.  Checking, thanks.

src/utils/keyboardKey.js
levithomason: shiftKey is a boolean, this is just converting it to an integer (the index):

image

dvdzkwsk: Oh boy, never seen that done before, if that says anything about readability. Your call.

src/utils/keyboardKey.js
levithomason, line 149: Yep, this is the official `KeyboardKey.key` value.  All `key` values are the result of the typed key.  Only if the key does not produce a character is there a name, like `PageUp`.

The `Spacebar` and others down below are just aliases to `KeyboardKey.key` values that are not otherwise dot-notation accessible.

src/modules/Dropdown/Dropdown.js
levithomason: This method was actually cruft, removed.  The dropdown is opened anytime there is a query.  The document listeners handle the arrow down key as well.

src/modules/Dropdown/Dropdown.js
levithomason, line 282: Note, `' '` is the literal `KeyboardKey.key` value.

src/utils/keyboardKey.js
levithomason: Converted to if/else.

src/utils/keyboardKey.js
levithomason: I actually had to look this up myself, borrowed from the polyfill we were using.  Updated to a ternary.

PR #232, opened 2016-05-07
Use process.env instead of globals

levithomason:
Addresses the todo in #206.  Removes compiler global from source code.  Allows consumers to compile our source and still take advantage of the NODE_ENV.


dvdzkwsk:
👻

PR #233, opened 2016-05-09
Debug utility

levithomason:
This PR addresses the TODO in #206 to add a logging utility.  It is only used in the Dropdown for now as the Dropdown is our proving grounds for removing jQuery and setting our v1.0.0 precedence.

Two debug logs were generating object differences, see `componentWillReceiveProps` and `componentDidUpdate`.  This logic was pulled into an `objectDiff` util and tested.


src/modules/Dropdown/Dropdown.js
```diff
@@ -16,6 +18,8 @@ import DropdownItem from './DropdownItem'
 import DropdownMenu from './DropdownMenu'
 import Icon from '../../elements/Icon/Icon'
 
+const debug = makeDebugger('dropdown')
```
levithomason, line 21: Logs are auto namespaced to `stardust:<your-name>`

src/modules/Dropdown/Dropdown.js
```diff
@@ -184,21 +188,19 @@ export default class Dropdown extends Component {
 
   componentWillReceiveProps(nextProps) {
     super.componentWillReceiveProps(nextProps)
-    // console.debug('Dropdown.componentWillReceiveProps()')
-    // console.log('changed props:', _.transform(nextProps, (res, val, key) => {
-    //   if (val !== this.props[key]) res[key] = val
-    // }, {}))
+    debug.groupCollapsed('componentDidUpdate()')
+    debug('changed props:', objectDiff(nextProps, this.props))
+    debug.groupEnd()
```
levithomason, line 194: console methods `group`, `groupCollapsed`, and `groupEnd` are added for cleaner console activity.

src/modules/Dropdown/Dropdown.js
```diff
@@ -8,6 +8,8 @@ import META from '../../utils/Meta'
 import { useKeyOnly, useKeyOrValueAndKey } from '../../utils/propUtils'
 import { iconPropRenderer } from '../../utils/propUtils'
 import keyboardKey from '../../utils/keyboardKey'
+import { makeDebugger } from '../../utils/debug'
+import { objectDiff } from '../../utils/utils'
```
levithomason, line 12: New tested util for generating naive object diffs, for development / debugging only.

src/modules/Dropdown/Dropdown.js
```diff
@@ -184,21 +188,19 @@ export default class Dropdown extends Component {
...
       this.setValue(nextProps.value)
     }
   }
 
   componentDidUpdate(prevProps, prevState) { // eslint-disable-line complexity
-    // console.debug('Dropdown.componentDidUpdate()')
-    // console.log('changed state:', _.transform(this.state, (res, val, key) => {
-    //   if (val !== prevState[key]) res[key] = val
-    // }, {}))
+    debug.groupCollapsed('componentDidUpdate()')
```
dvdzkwsk, line 202: Seeing all of these methods that debug their own name, it might be nice to incorporate it as a decorator or something on the method or the class so they all do this automagically. Not a blocker, future idea.

src/modules/Dropdown/Dropdown.js
```diff
@@ -184,21 +188,19 @@ export default class Dropdown extends Component {
 
   componentWillReceiveProps(nextProps) {
     super.componentWillReceiveProps(nextProps)
-    // console.debug('Dropdown.componentWillReceiveProps()')
-    // console.log('changed props:', _.transform(nextProps, (res, val, key) => {
-    //   if (val !== this.props[key]) res[key] = val
-    // }, {}))
+    debug.groupCollapsed('componentDidUpdate()')
+    debug('changed props:', objectDiff(nextProps, this.props))
```
dvdzkwsk, line 193: This objectDiff will still happen unless the debug statements are actually stripped from the source code. `makeDebugger` simply won't log them unless told, but the arguments are still evaluated. You could consider passing it a callback function so that the inner content is lazily evaluated (i.e. then only runs in !prod).

src/modules/Dropdown/Dropdown.js
```diff
@@ -302,8 +308,9 @@ export default class Dropdown extends Component {
   }
 
   closeOnDocumentClick = (e) => {
-    // console.debug('Dropdown.handleDocumentClick()')
-    // console.log(e)
+    debug.groupCollapsed('handleDocumentClick()')
```
dvdzkwsk: text doesn't match the method name.

dvdzkwsk:
Couple of thoughts:

1) Even though the debugger will noop during production, the arguments are still always evaluated. This is needlessly expensive in production or even outside of debug mode, especially for more complex operations. Consider lazily running debug statements, i.e.:

``` js
debug(() => object.diff(...))
```

**The above is a solution if the following is not an option**

2) As a corollary, all of these debug statements are still _invoked_ in production, their argument evaluation notwithstanding, even if they are just noop. I know it's a lot of work to wrap everything in environment checks, but for a public library I think it could be optimized a bit further. The pattern used here is similar to what we do in our business apps, but that's a bit of a different realm than a public library. Not sure if there's an easy way to abstract this logic so it's removed at compile time, but it should be considered.

src/modules/Dropdown/Dropdown.js
levithomason, line 202: Yep, we're on the same page. In either case, I don't think env flags are enough to enable / disable logging. I'd like to have a transform the strips the debugs. Since, we could / should also strip propTypes and the static_meta property.

src/modules/Dropdown/Dropdown.js
levithomason, line 193: The goal is to strip them in production. However, I'll add a todo.

levithomason:
Agreed on all counts and considered these things prior to the PR. In the future we'll have an optimization phase which will include a transform to remove debugs, static propTypes, and static_meta from production code. 

I want to get more components moved to the 1.0 pattern before optimizing further. This PR moves us from console statements to a debug pattern we can iterate on.

PR #244, opened 2016-05-18
Use enzyme in all tests

levithomason:
This test removes our old `RenderTree` testing util in favor of `enzyme`.  The remaining tests using it were updated.  Some missing tests were added.  Very minimal component refactoring was made as well.

`shallow` rendering was used everywhere possible.  No test refactoring was done except that necessary to convert the old RenderTree syntax to the new enzyme.


src/addons/Confirm/Confirm.js
```diff
@@ -14,46 +14,29 @@ export default class Confirm extends Component {
     className: PropTypes.string,
     confirmLabel: PropTypes.string,
     header: PropTypes.string,
-    ref: PropTypes.string,
   }
 
   static defaultProps = {
     abortLabel: 'Cancel',
     confirmLabel: 'Yes',
-    ref: 'modal',
```
levithomason, line 23: There is no reason to set an internal access only `ref` name via props.  Anyway, it was not used.

src/addons/Confirm/Confirm.js
```diff
@@ -14,46 +14,29 @@ export default class Confirm extends Component {
...
   }
 
   state = {
     message: 'Are you sure?',
   }
 
-  componentDidMount() {
-    this.deferred = Promise.defer()
```
levithomason, line 31: Fixes promise.defer deprecation warning by using `new Promise()` instead.

src/elements/Button/Button.js
```diff
@@ -9,7 +9,6 @@ export default class Button extends Component {
   }
 
   static defaultProps = {
-    children: 'Click Here',
```
levithomason, line 12: Stardust should not set default content (nor styles).

src/modules/Modal/Modal.js
```diff
@@ -9,14 +9,9 @@ export default class Modal extends Component {
   static propTypes = {
     children: PropTypes.any,
     className: PropTypes.string,
-    ref: PropTypes.string,
     settings: PropTypes.object,
   }
 
-  static defaultProps = {
-    ref: 'modal',
-  }
```
levithomason, line 18: Again, ref props don't make sense.  Also unused.

dvdzkwsk:
👻 , nothing really jumping out to me.

PR #249, opened 2016-05-28
Setup doc site router

levithomason:
This is the doc site router in preparation for mirroring routes on our doc site with SUI docs.


docs/app/Components/Root.js
```diff
@@ -0,0 +1,75 @@
...
+
+    return (
+      <div style={style.container}>
+        <div style={style.menu}>
+          <DocsMenu />
+        </div>
+        <div style={style.main}>
+          {location.pathname !== '' ? (
+            <div>
+              {/* TODO will cleanup, debug to prove router works */}
```
levithomason, line 52: When visiting `elements/button` you can see the debug below, along with the button examples.  This is just to show the router is working.  Separate PRs will setup actual doc site routes and pages.

Issue #247, opened 2016-05-24
Remove jQuery dependency


levithomason:
# Hooray!

This issue is complete, it will be closed once the final PR, #400 is merged.

---

As we remove jQuery and SUI JS from these components, we're also updating their legacy component APIs to our v1 API.  You can see more of that work in #269.

The [CONTRIBUTING.md](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md) guide details how to go about designing a component API.  The [Spec out the API](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#spec-out-the-api) section is of primary importance.

You can reference the Dropdown for a complex jQuery free stateful component.  You can reference the Progress (#324) or Accordion for more simple jQuery free stateful components.
## PRs
- [x] ~~#220 doc site and _meta updates~~
- [x] ~~#221 List item prop renderers~~
- [x] ~~#222 component utils~~
- [x] ~~#223 Common tests~~
- [x] ~~#224 AutoControlledComponent~~
- [x] ~~#226 Dropdown without jQuery~~
- [x] ~~#225 Select component~~
- [x] ~~#260 Checkbox without jQuery~~
- [x] ~~#400 Form without jQuery~~
- [x] ~~#344 Message without jQuery~~
- [x] ~~#324 Progress without jQuery~~
- [x] ~~#400 Remove traces of jQuery~~
  - [x] ~~webpack config~~
  - [x] ~~karma config~~
  - [x] ~~_default config vendors~~
  - [x] ~~package dependencies~~
  - [x] l~~ib/jquery~~
## ~~TODOs~~

Follow up issues that sprung from the above PRs:
- [x] ~~#229 [remove babel-polyfil](https://github.com/TechnologyAdvice/stardust/pull/206/files#r61193441)~~
- [x] ~~#231 [remove keyboard event polyfill in favor of util](https://github.com/TechnologyAdvice/stardust/pull/226#discussion_r61350959)~~
- [x] ~~#??? [functional domEvent methods](https://github.com/TechnologyAdvice/stardust/pull/222/files#r61273736)~~ deeming this not necessary until we get other major priorities done
- [x] ~~#232 [use `process.env.NODE_ENV` instead of `__PROD__`](https://github.com/TechnologyAdvice/stardust/pull/224/files#r61321467)~~
[...]

PR #250, opened 2016-05-28
User prop spreading util

levithomason:
This PR addresses a todo in #247.  The `getUnhandledProps` util provides a way to capture all additional props passed to a component so they can be spread.  All props defined in the component's `propTypes`, `defaultProps`, and `autoControlledProps` are removed.  Leaving only extra props passed in by the user.


.eslintrc
```diff
@@ -4,10 +4,11 @@
     "ta/frontend"
   ],
   "rules": {
+    "complexity": [1, 10],
```
levithomason, line 7: Warn above 10.

.eslintrc
```diff
@@ -4,10 +4,11 @@
...
     "consistent-return": 0,
     "no-confusing-arrow": 0,
     "react/sort-comp": 1,
-    "valid-jsdoc": 1
+    "valid-jsdoc": 0
```
levithomason, line 11: JSDoc comments on classes do not need return statements, there is no way to tell this rule to ignore those.  Turned it off, IDEs mostly validate these anyway.

Issue #218, opened 2016-04-20
All event handlers should return the event data first


levithomason:
As we rewrite away from jQuery, #206, we are no longer bound to Semantic UI event callbacks.  In keeping things React friendly, all components will be required to return the event data as the first argument.

Any other helpful data can be returned after this.  This will be a breaking change to the current MenuItem component as it currently returns the menu item name.

``` js
this.props.onClick(name)
```

will become

``` js
this.props.onClick(e, name)
```

This is the only breaking change, there are several other non-breaking changes that need to be made however.  Note, the common test that has been written for this has currently commented out the actual assertion.  See, `commonTests.js` `it('handles events transparently'...` once merged.

This issue is a reminder to address the issues above and enable the assertion ^ for event data.

PR #254, opened 2016-06-03
Support multiple dropdown type

levithomason:
This PR adds support for the `multiple` dropdown type.  A multiple dropdown:
- accepts an array for the `value` (opposed to string/number)
- logs errors when !production if value is wrong type for the given dropdown type
- does not close when selecting a value
- allows removing values by clicking the &times; icon on the label for a value

Other updates and fixes:
- calls onChange with (e, value)
- only show "no results" message if `search`

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -12,7 +12,7 @@ export default class DropdownAsyncOptions extends Component {
   componentWillMount() {
     const options = getOptions()
     const value = _.sample(options).value
-    this.setState({ isFetching: false, search: true, value, options })
+    this.setState({ isFetching: false, search: true, multiple: true, value, options })
   }
 
   handleChange = value => this.setState({ value })
@@ -29,29 +29,36 @@ export default class DropdownAsyncOptions extends Component {
 
   selectRandom = () => this.setState({ value: _.sample(this.state.options).value })
   toggleSearch = (e) => this.setState({ search: e.target.checked })
+  toggleMultiple = (e) => this.setState({ multiple: e.target.checked })
 
   render() {
-    const { options, isFetching, search, value } = this.state
+    const { multiple, options, isFetching, search, value } = this.state
 
     return (
       <div>
-        <Button onClick={this.fetchOptions}>Fetch</Button>
-        <Button onClick={this.selectRandom} disabled={_.isEmpty(options)}>Random</Button>
+        <p>
+          <Button onClick={this.fetchOptions}>Fetch</Button>
+          <Button onClick={this.selectRandom} disabled={_.isEmpty(options)}>Random</Button>
+          <label>
+            <input type='checkbox' checked={search} onChange={this.toggleSearch} /> Search
+          </label>
+          {' '}
+          <label>
+            <input type='checkbox' checked={multiple} onChange={this.toggleMultiple} /> Multiple
+          </label>
+        </p>
         <Dropdown
           options={options}
           value={value}
           placeholder='Add Users'
+          multiple={multiple}
           search={search}
           selection
+          fluid
           onChange={this.handleChange}
           disabled={isFetching}
           loading={isFetching}
         />
-        {' '}
-        <label>
-          <input type='checkbox' checked={search} onChange={this.toggleSearch} />
-          {' '}Search
-        </label>
         <pre>{JSON.stringify(this.state, null, 2)}</pre>
       </div>
     )
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -11,7 +11,7 @@ const getOptions = () => _.times(5, () => {
 export default class DropdownAsyncOptions extends Component {
   componentWillMount() {
     const options = getOptions()
-    const value = [_.sample(options).value]
+    const value = []
     this.setState({ isFetching: false, search: true, multiple: true, value, options })
   }
 
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -11,11 +11,11 @@ const getOptions = () => _.times(5, () => {
 export default class DropdownAsyncOptions extends Component {
   componentWillMount() {
     const options = getOptions()
-    const value = _.sample(options).value
+    const value = [_.sample(options).value]
     this.setState({ isFetching: false, search: true, multiple: true, value, options })
   }
 
-  handleChange = value => this.setState({ value })
+  handleChange = (e, value) => this.setState({ value })
 
   // fake api call
   fetchOptions = () => {
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -26,7 +26,11 @@ export default class DropdownAsyncOptions extends Component {
     }, 500)
   }
 
-  selectRandom = () => this.setState({ value: _.sample(this.state.options).value })
+  selectRandom = () => {
+    const { multiple, options } = this.state
+    const value = _.sample(options).value
+    this.setState({ value: multiple ? [value] : value })
+  }
   toggleSearch = (e) => this.setState({ search: e.target.checked })
   toggleMultiple = (e) => {
     const { value } = this.state
--- a/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
+++ b/docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
@@ -11,8 +11,7 @@ const getOptions = () => _.times(5, () => {
 export default class DropdownAsyncOptions extends Component {
   componentWillMount() {
     const options = getOptions()
-    const value = []
-    this.setState({ isFetching: false, search: true, multiple: true, value, options })
+    this.setState({ isFetching: false, search: true, multiple: true, value: [], options })
   }
 
   handleChange = (e, value) => this.setState({ value })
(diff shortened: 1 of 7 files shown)


docs/app/Examples/modules/Dropdown/Content/AsyncOptions.js
```diff
@@ -29,29 +28,41 @@ export default class DropdownAsyncOptions extends Component {
 
   selectRandom = () => this.setState({ value: _.sample(this.state.options).value })
   toggleSearch = (e) => this.setState({ search: e.target.checked })
+  toggleMultiple = (e) => {
+    const { value } = this.state
+    // convert the value to/from an array
+    const newValue = e.target.checked ? _.compact([value]) : value[0]
+    this.setState({ multiple: e.target.checked, value: newValue })
+  }
```
levithomason, line 40: Since this example allows toggling the `multiple` prop, we need to convert the value to and from an array corresponding to the dropdown type.

src/elements/Label/Label.js
```diff
@@ -106,7 +105,7 @@ Label.propTypes = {
   /** Additional text with less emphasis. */
   detail: PropTypes.string,
 
-  /** Same as detail but formatted as a link. */
+  /** Format the detail as a link. */
   detailLink: PropTypes.string,
```
levithomason, line 109: Updated this behavior to match the `link` behavior.  Both just format the component as a link.

src/elements/Label/Label.js
```diff
@@ -136,13 +135,13 @@ Label.propTypes = {
...
   onClick: PropTypes.func,
 
-  /** Click callback for detail, called with (event, props). */
-  onClickDetail: PropTypes.func,
+  /** Click callback for detail, called with (event, props). Formats the detail as a link. */
+  onDetailClick: PropTypes.func,
 
   /** Adds an "x" icon, called with (event, props) when "x" is clicked. */
-  onClickRemove: PropTypes.func,
+  onRemove: PropTypes.func,
```
levithomason, line 142: Updated callbacks and click handlers to make more sense and be consistent with our others (`on*Click`)

src/modules/Dropdown/Dropdown.js
```diff
@@ -62,24 +62,24 @@ export default class Dropdown extends Component {
     /** Initial value of open. */
     defaultOpen: PropTypes.bool,
 
-    /** Current value, creates a controlled component. */
+    /** Current value or value array if multiple. Creates a controlled component. */
     value: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number,
-      PropTypes.arrayOf([
+      PropTypes.arrayOf(PropTypes.oneOfType([
```
levithomason, line 69: Fixed prop types for dropdown value / defaultValue.

src/modules/Dropdown/Dropdown.js
```diff
@@ -95,7 +95,7 @@ export default class Dropdown extends Component {
     /** Called with the React Synthetic Event on Dropdown blur. */
     onBlur: PropTypes.func,
 
-    /** Called with the current value on change. */
+    /** Called with the React Synthetic Event and current value on change. */
```
levithomason, line 98: **NOTE** this PR moves the dropdown to our "event first" callback argument pattern.  
See #218.

src/modules/Dropdown/Dropdown.js
```diff
@@ -222,13 +242,15 @@ export default class Dropdown extends Component {
 
     // opened / closed
     if (!prevState.open && this.state.open) {
+      this.open()
```
levithomason, line 238: Fixed bug where toggling the `open` prop did not open/close the menu.

src/modules/Dropdown/Dropdown.js
```diff
@@ -506,45 +600,44 @@ export default class Dropdown extends Component {
...
-  renderPlaceholder = () => {
-    const { placeholder, search } = this.props
-    const { searchQuery, value } = this.state
-
-    if (value || !placeholder) return null
-    const classes = cx('default text', search && searchQuery && 'filtered')
-
-    return <div className={classes}>{placeholder}</div>
-  }
-
```
levithomason, line 533: Combined `text` and `placeholder`.  They are the same, the only difference is whether or not there is a `default` class applied.  This text is gray when there is a default class.  Otherwise, it is the display text of the selected value (for !multiple dropdowns).

eanplatter:
🍂 sqursh enrd rbs tur maestr

src/elements/Label/Label.js
```diff
@@ -49,27 +46,29 @@ function Label(props) {
...
 
-  const _props = {
-    className: classes,
-    onClick: handleClick,
-    ...rest,
-  }
-
-  const _detail = detail || detailLink
-  const detailComponent = (detailLink || onClickDetail) && 'a' || detail && 'div'
+  const DetailComponent = (detailLink || onDetailClick) && 'a' || 'div'
```
dvdzkwsk, line 49: Would be nice to wrap the latter check in `()` for similar clarity to the initial expression.

src/elements/Label/Label.js
```diff
@@ -49,27 +46,29 @@ function Label(props) {
...
+    ),
+    remove: (removable || onRemove) && (
+      <Icon className='delete' onClick={handleRemove} />
+    ),
   })
 
-  // Do not destructure createElement import
-  // react-docgen only recognizes a stateless component when React.createElement is returned
-  return React.createElement(_component, _props, _children)
+  const LabelComponent = image || link || onClick ? 'a' : 'div'
```
dvdzkwsk, line 64: Just stating that this is not immediately clear to me, or at least leaves me questioning if it works exactly as I expect. Take that as you will.

`(image || link || onClick) ? 'a' : 'div'`

src/modules/Dropdown/Dropdown.js
```diff
@@ -188,14 +180,35 @@ export default class Dropdown extends Component {
...
+
+      /* eslint-disable no-console */
+      if (nextProps.multiple && !isNextValueArray) {
+        console.error(
+          'Dropdown `value` must be an array when `multiple` is set.' +
+          ` Received type: \`${Object.prototype.toString.call(nextProps.value)}\`.`,
+        )
+      } else if (!nextProps.multiple && isNextValueArray) {
+        console.error(
+          'Dropdown `value` must not be an array when `multiple` is not set.' +
```
dvdzkwsk, line 199: Double negation takes a re-read to understand.

`Dropdown 'value' can only been an array when multiple is set`

src/modules/Dropdown/Dropdown.js
```diff
@@ -373,63 +419,109 @@ export default class Dropdown extends Component {
...
+  getItemByValue = (value) => {
+    const { options } = this.props
+    return _.find(options, { value })
+  }
+
+  getMenuItemIndexByValue = (value) => {
+    const options = this.getMenuOptions()
 
-    return _.findIndex(options, ['value', value])
+    return _.findIndex(options, opt => opt.value === value)
```
dvdzkwsk: Previous places you are using shorthand (with `find`):

`return _.findIndex(options, { value })`

src/modules/Dropdown/Dropdown.js
```diff
@@ -373,63 +419,109 @@ export default class Dropdown extends Component {
...
-      selectedIndex: this.getItemIndexByValue(value),
-    })
+    }
+
+    // update the selected index
+    if (!selectedIndex) {
+      // Select the currently active item, if none, use the first item.
+      // Multiple selects remove active items from the list,
+      // their initial selected index should be 0.
+      newState.selectedIndex = multiple ? 0 : this.getMenuItemIndexByValue(value || _.get(options, '[0].value'))
```
dvdzkwsk, line 495: Why do you have to do this `_.get`, would this not just always equal 0 in this case?

src/modules/Dropdown/Dropdown.js
```diff
@@ -506,45 +598,44 @@ export default class Dropdown extends Component {
...
+    if (multiple) return
 
-    let _text
-    if (text) {
+    const classes = cx(
+      placeholder && !hasValue && 'default',
+      'text',
+      search && searchQuery && 'filtered'
+    )
+    let _text = placeholder
```
dvdzkwsk, line 610: 👎 for `_` prefix, since this is easy to mess up and is often used to indicate variable scoping. Don't think there'd be anything wrong doing `let { text } = this.props`. Not a blocker, minor comment.

src/modules/Dropdown/Dropdown.js
```diff
@@ -565,16 +656,50 @@ export default class Dropdown extends Component {
...
+    const { value } = this.state
+    if (!multiple || _.isEmpty(value)) {
+      debug.groupEnd()
+      return
+    }
+    const selectedItems = _.map(value, this.getItemByValue)
+    debug('selectedItems', selectedItems)
+    debug.groupEnd()
+
+    return _.map(selectedItems, (item) => {
```
dvdzkwsk: Seems like a filter -> map would be more logical.

dvdzkwsk:
👻 LGTM, will pull and test locally if you want before merging.

src/elements/Label/Label.js
levithomason, line 49: Not seeing where the extra parens can be added here.  Here is another way of writing this:

``` js
let DetailComponent = 'div'
if (detailLink || onDetailClick) DetailComponent = 'a'
```

src/modules/Dropdown/Dropdown.js
levithomason, line 495: The method used gets the index of a menu item by it's value.  The options is an array of objects with `text` and `value` props (or an empty array).  Since it may be empty, `options[0]` may be undefined, if not we need `options[0].value`.  Using _.get to avoid "cannot read prop value of undefined" errors.

src/modules/Dropdown/Dropdown.js
levithomason, line 610: `text` is already destructured in this context.  The final rendered value can be `this.props.text`, `null`, or the currently selected item's `text` prop (see logic below).

Suppose it could be something like `renderedText`?

src/modules/Dropdown/Dropdown.js
levithomason: Compacted the array first.

PR #255, opened 2016-06-09
Improve webpack performance

levithomason:
This PR cuts the build time by over 50% and the rebundle times by 80-90%.  This was done by using CDNs for larger libraries and adding the webpack DLL plugin for rarely changing libraries.

The CDN libraries are still listed as deps and installed for development. When offline, you can use them by running `npm start:local-modules`.

### DIFF
--- a/build/webpack.dll.js
+++ b/build/webpack.dll.js
@@ -1,0 +1,63 @@
+import webpack from 'webpack'
+
+import config from '../config'
+const webpackDllConfig = { module: {} }
+
+const { paths } = config
+
+// ------------------------------------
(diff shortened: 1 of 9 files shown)


build/webpack.dll.js
```diff
@@ -0,0 +1,68 @@
+import path from 'path'
+import webpack from 'webpack'
+
+import config from '../config'
+// import webpackConfig, { webpackHotMiddlewareEntry } from './webpack.config'
```
dvdzkwsk: crufteroni

PR #263, opened 2016-06-21
Upgrade React peer dependency versions

jaredhughes:
- Updates peer dependencies to React 15 (currently causing build failures in CircleCI)

**Please be sure to publish to NPM when/if merged.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
   },
   "peerDependencies": {
     "jquery": "^2.1.4",
-    "react": "^0.14.0",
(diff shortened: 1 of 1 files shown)


levithomason:
Since React `15.2.0-rc.1` is out, I think we should go more permissive on the support.  I'd be comfortable with `>=0.14.0 <= 15`.

levithomason:
Thanks for the note, fixed this along with the HMR issue and published it in `v0.14.1`.

PR #267, opened 2016-06-22
Update circle.yml

levithomason:


circle.yml
```diff
@@ -5,12 +5,7 @@ general:
 
 machine:
   node:
-    version: 5
-
-dependencies:
-  pre:
-    - echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
-    - npm install -g npm@3
```
levithomason, line 13: We no longer have private modules.

circle.yml
```diff
@@ -5,12 +5,7 @@ general:
...
+    version: 6
```
levithomason, line 8: Use node 6, comes with npm 3 also.

PR #268, opened 2016-06-22
Divider Component updated to v1 API

jhchill666:
- Divider Component updated to v1 API
- Documents updated to reflect
- Divider-test added to suite

### DIFF
--- a/src/elements/Divider/Divider.js
+++ b/src/elements/Divider/Divider.js
@@ -14,7 +14,7 @@ import {
 function Divider(props) {
   const {
     horizontal, vertical, inverted, fitted, hidden, section, clearing,
-    children
+    children, className
   } = props;
 
   const classes = cx('sd-divider ui',
@@ -25,7 +25,8 @@ function Divider(props) {
     useKeyOnly(hidden, 'hidden'),
     useKeyOnly(section, 'section'),
     useKeyOnly(clearing, 'clearing'),
-    'divider'
+    'divider',
+    className
   )
 
   const _children = createFragment({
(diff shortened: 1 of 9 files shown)


src/elements/Divider/Divider.js
```diff
@@ -1,30 +1,81 @@
...
+  const _children = createFragment({
+    children,
+  })
+
+  const DividerComponent = 'div'
+  const rest = getUnhandledProps(Divider, props)
+
+  return (
+    <DividerComponent className={classes} {...rest}>
+      {_children}
```
levithomason: Can we use the children directly here? Not sure why the fragment is required.

levithomason:
Wow, nice work! Thanks. Just the one question about the fragment then we can merge this guy.

test/specs/elements/Divider/Divider-test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+import Divider from 'src/elements/Divider/Divider'
+import * as common from 'test/specs/commonTests'
+
+describe('Divider', () => {
+  common.isConformant(Divider)
```
levithomason, line 7: Let's add common.rendersChildren and common.hasUIClassName.

levithomason, line 7: There are also common tests for the classNames, reference https://github.com/TechnologyAdvice/stardust/blob/master/test/specs/elements/Label/Label-test.js

src/elements/Divider/Divider.js
jhchill666: Of course.  I'd just whittled down the Label version, and not considered `createFragment` was now redundant

jhchill666:
Ok there you have it @levithomason

levithomason:
👍

levithomason:
I've added #269 to track the remaining components needing updated.  I've also added a section to the README outlining ["How can I help?"](https://github.com/TechnologyAdvice/stardust#how-can-i-help).

Your PR was much appreciated, would love to see more!

PR #273, opened 2016-06-23
Remove console.group* usage

### LINKED ISSUE #272: Remove console.group
in their words:

```
This feature is non-standard and is not on a standards track. Do not use it on production.
```

https://developer.mozilla.org/en-US/docs/Web/API/Console/group

our other option here is literally to fork stardust and start ripping things out (which we have already started)

levithomason:
Fixes #272 where console.group is not supported.  Instead of adding more fat to shim or ponyfill this, we're just removing it entirely.

The debug module itself is something that is still subject to lots of churn while we find the best solution.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
-coverage/
+coverage/*
 dist/*
 docs/build/*
+dll/*
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -180,7 +180,7 @@ export default class Dropdown extends Component {
 
   componentWillReceiveProps(nextProps) {
     super.componentWillReceiveProps(nextProps)
-    debug.groupCollapsed('componentWillReceiveProps()')
+    debug('componentWillReceiveProps()')
     // TODO objectDiff still runs in prod, stop it
     debug('changed props:', objectDiff(nextProps, this.props))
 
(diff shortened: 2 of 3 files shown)


.eslintignore
```diff
@@ -1,3 +1,4 @@
-coverage/
+coverage/*
 dist/*
 docs/build/*
+dll/*
```
levithomason, line 4: No need to lint the dll bundle.

src/modules/Dropdown/Dropdown.js
```diff
@@ -180,7 +180,7 @@ export default class Dropdown extends Component {
 
   componentWillReceiveProps(nextProps) {
     super.componentWillReceiveProps(nextProps)
-    debug.groupCollapsed('componentWillReceiveProps()')
+    debug('componentWillReceiveProps()')
```
levithomason, line 183: Group calls were converted to regular debug calls, so we still get the method logs.

src/modules/Dropdown/Dropdown.js
```diff
@@ -207,15 +207,12 @@ export default class Dropdown extends Component {
       debug('value changed, setting', nextProps.value)
       this.setValue(nextProps.value)
     }
-
-    debug.groupEnd()
```
levithomason, line 211: All groupEnd calls were removed.

src/utils/debug.js
```diff
@@ -19,23 +19,7 @@ if (process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test') {
...
-    inst.groupEnd = noop
-  } else {
-    /* eslint-disable no-console */
-    inst.group = console.group.bind(console, msg)
-    inst.groupCollapsed = console.groupCollapsed.bind(console, msg)
-    inst.groupEnd = console.groupEnd.bind(console)
-    /* eslint-enable no-console */
-  }
-
-  return inst
```
levithomason, line 38: Removed all console group functionality from the debug module.

athurman:
looks good to me 🐙

kyleturco:
🐧

PR #275, opened 2016-06-23
Remove create fragment dependency

### LINKED ISSUE #274: Unlisted peer dependency on react-addons-create-fragment
Create fragment is used in the Label.  When updating our internal apps, we're getting this error:

``` sh
ERROR in ./~/stardust/dist/elements/Label/Label.js
Module not found: Error: Cannot resolve module 'react-addons-create-fragment' in /Users/levithomason/src/unity/node_modules/stardust/dist/elements/Label
 @ ./~/stardust/dist/elements/Label/Label.js 31:33-72
```

This is the only place create fragment is used.  Instead of adding another dependency, we should refactor this bit of code to not rely on the addon and remove the dependency.

levithomason:
Fixes #274.  Removing `react-addons-create-fragment` dependency.

### DIFF
--- a/src/elements/Label/Label.js
+++ b/src/elements/Label/Label.js
@@ -1,6 +1,5 @@
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
-import createFragment from 'react-addons-create-fragment'
 
 import META from '../../utils/Meta'
 import * as sui from '../../utils/semanticUtils'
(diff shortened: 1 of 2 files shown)


src/elements/Label/Label.js
```diff
@@ -1,6 +1,5 @@
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
-import createFragment from 'react-addons-create-fragment'
```
dvdzkwsk, line 3: had never even known of this o.O

dvdzkwsk:
👻 looks identical

athurman:
🐙

PR #277, opened 2016-06-24
Container Component updated to v1 API

jhchill666:
- Container Component updated to v1 API
- Examples updated to match Semantic Docs
- Container-test added to suite


docs/app/Examples/elements/Container/Types/ContainerTextExample.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable max-len */
+
+import React, { Component } from 'react'
+import { Container, Message } from 'stardust'
+
+export default class ContainerContainerExample extends Component {
```
levithomason:
``` diff
-class ContainerContainerExample
+class ContainertextExample
```

docs/app/Examples/elements/Container/Types/ContainerTextExample.js
```diff
@@ -0,0 +1,21 @@
...
+  render() {
+    return (
+      <div>
+        <Message className='info'>
+          <p>A text container is a simpler markup alternative to using a grid with a single column, and is designed to have a reasonable maximum width for displaying flowing text</p>
+        </Message>
```
levithomason: Since example file text appears on the site below the example, we want to keep it as minimal as possible.  This `<Message />` should be moved up to the ContainerTypesExamples.js by adding a closing tag to the `< ComponentExample />` for this example file:

``` jsx
// ContainerTypesExamples.js ~ ln 14

<ComponentExample
   title='Text Container'
   description='A container can reduce its maximum width to more naturally accomodate a single column of text'
   examplePath='elements/Container/Types/ContainerTextExample'
 >
  <Message>
    // put the message here
    // it will be displayed above this example, but not in the example code
  </Message>
</ComponentExample>
```

docs/app/Examples/elements/Container/Types/ContainerTextExample.js
```diff
@@ -0,0 +1,21 @@
...
+        <Container text>
+          <h2>Header</h2>
```
levithomason: Prefer all stardust components in examples `<Header.H2 />`, for now.

docs/app/Examples/elements/Container/Variations/ContainerFluidExample.js
```diff
@@ -0,0 +1,22 @@
...
+import React, { Component } from 'react'
+import { Container, Message } from 'stardust'
+
+export default class ContainerFluidExample extends Component {
+  render() {
+    return (
+      <div>
+        <Message className='info'>
+          Fluid containers are useful for setting text alignment, or other variations on unstyled content
+        </Message>
```
levithomason: Same comment with the `<Message />` here as above.  Let's hoist it up out of the example code.

jhchill666: Ok, have actioned all of those.  Sorry still getting your code under my 
belt!

> Levi Thomason mailto:notifications@github.com
> 24 June 2016 at 17:09
> 
> In 
> docs/app/Examples/elements/Container/Variations/ContainerFluidExample.js 
> https://github.com/TechnologyAdvice/stardust/pull/277#discussion_r68421369:
> 
> > @@ -0,0 +1,22 @@
> > +/\* eslint-disable max-len */
> > +
> > +import React, { Component } from 'react'
> > +import { Container, Message } from 'stardust'
> > +
> > +export default class ContainerFluidExample extends Component {
> > - render() {
> > - return (
> > - <div>
> > - <Message className='info'>
> > - Fluid containers are useful for setting text alignment, or other 
> >   variations on unstyled content
> > - </Message>
> 
> Same comment with the |<Message />| here as above. Let's hoist it up 
> out of the example code.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> https://github.com/TechnologyAdvice/stardust/pull/277/files/bd4b8de6b915721a488337a4e1ce7a9194ce895a#r68421369, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AAWHs2B-2HIurEHZ2jwPvh7bkDBeQXRjks5qPAEfgaJpZM4I9hiP.

## 

Sent from Postbox 
https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach

src/elements/Container/Container.js
```diff
@@ -1,33 +1,68 @@
...
+    text, left, center, right, justified, fluid,
+    children, className,
+  } = props
+
+  const classes = cx('sd-container ui',
+    useKeyOnly(text, 'text'),
+    useKeyOnly(left, 'left aligned'),
+    useKeyOnly(center, 'center aligned'),
+    useKeyOnly(right, 'right aligned'),
+    useKeyOnly(justified, 'justified'),
```
levithomason: There are 4 types of possible class to props mappings for SUI classes.  The alignment class here follows this API pattern:

``` js
aligned="left"    // useValueAndKey(aligned, 'aligned')
aligned="right"   // useValueAndKey(aligned, 'aligned')
aligned="center"  // useValueAndKey(aligned, 'aligned')
```

One reason we use this key/value relationship is because two simultaneous alignment props are not valid:

``` js
<Container right left />
```

Where as making those values of the `aligned` prop means you can only have one at a time:

``` js
<Container aligned="left" />
```

However, the `justified` option breaks down however since the resulting class is not `justified aligned`.  We've not yet run into this matchup yet so I'm open to suggestions.  The best immediate thought I have is to still allow `aligned="justified"` for the API consistency, then in cx buildup:

``` js
cx(
  aligned === 'justified' ? 'justified' : useValueAndKey(aligned, 'aligned'),
)
```

Thoughts?

**EDIT**

Fixed incorrect className suggestion.

src/elements/Container/Container.js
```diff
@@ -1,33 +1,68 @@
...
+    useKeyOnly(fluid, 'fluid'),
+    'container',
+    className
+  )
+
+  const ContainerComponent = 'div'
+  const rest = getUnhandledProps(Container, props)
+
+  return (
+    <ContainerComponent className={classes} {...rest}>
```
levithomason: Since the `ContainerComponent` never changes, let's just use the `<div ...` inline here.

test/specs/elements/Container/Container-test.js
```diff
@@ -0,0 +1,27 @@
...
+
+  it('renders a <div /> element', () => {
+    shallow(<Container />)
+      .should.have.tagName('div')
+  })
+
+  it('adds the Container class', () => {
+    shallow(<Container />)
+      .should.have.className('container')
+  })
```
levithomason: This test is automated as part of the `common.isConformant()` test.  All components are required to include their Semantic UI component className.

src/elements/Container/Container.js
jhchill666: Absolutely.  Wasn't sure how stringently to follow your initial efforts 
on Label.  Will optimize as the default cause of action in future.

> Levi Thomason mailto:notifications@github.com
> 24 June 2016 at 17:24
> 
> In src/elements/Container/Container.js 
> https://github.com/TechnologyAdvice/stardust/pull/277#discussion_r68423440:
> 
> > - const classes = cx('sd-container ui',
> > - useKeyOnly(text, 'text'),
> > - useKeyOnly(left, 'left aligned'),
> > - useKeyOnly(center, 'center aligned'),
> > - useKeyOnly(right, 'right aligned'),
> > - useKeyOnly(justified, 'justified'),
> > - useKeyOnly(fluid, 'fluid'),
> > - 'container',
> > - className
> > - )
> >   +
> > - const ContainerComponent = 'div'
> > - const rest = getUnhandledProps(Container, props)
> >   +
> > - return (
> > - <ContainerComponent className={classes} {...rest}>
> 
> Since the |ContainerComponent| never changes, let's just use the |<div 
> ...| inline here.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> https://github.com/TechnologyAdvice/stardust/pull/277/files/f420da9c4eb10853132c436ba6a06faa82ffeba2#r68423440, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AAWHs1bthM95zj3FbrWdWrv0AFmfQBUbks5qPASrgaJpZM4I9hiP.

## 

Sent from Postbox 
https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach

test/specs/elements/Container/Container-test.js
jhchill666: So the common.propKeyOnlyToClassName are still required, but not the 2 
'it' blocks?

> Levi Thomason mailto:notifications@github.com
> 24 June 2016 at 17:26
> 
> In test/specs/elements/Container/Container-test.js 
> https://github.com/TechnologyAdvice/stardust/pull/277#discussion_r68423748:
> 
> > - common.propKeyOnlyToClassName(Container, 'text')
> > - common.propKeyOnlyToClassName(Container, 'left', { className: 
> >   'aligned' })
> > - common.propKeyOnlyToClassName(Container, 'center', { className: 
> >   'aligned' })
> > - common.propKeyOnlyToClassName(Container, 'right', { className: 
> >   'aligned' })
> > - common.propKeyOnlyToClassName(Container, 'justified')
> > - common.propKeyOnlyToClassName(Container, 'fluid')
> >   +
> > - it('renders a <div /> element', () => {
> > - shallow(<Container />)
> > - .should.have.tagName('div')
> > - })
> >   +
> > - it('adds the Container class', () => {
> > - shallow(<Container />)
> > - .should.have.className('container')
> > - })
> 
> This test is automated as part of the |common.isConformant()| test. 
> All components are required to include their Semantic UI component 
> className.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> https://github.com/TechnologyAdvice/stardust/pull/277/files/a083fd536fd2f416c6214e2af929e6c51b4e8d58#r68423748, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AAWHs8T9VMTH_JMod5KlJmmO1HaBUkeJks5qPAU9gaJpZM4I9hiP.

## 

Sent from Postbox 
https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach

src/elements/Container/Container.js
jamiehill: Understood.  But with only one possible instance of the 'aligned' prop 
being possible, be it 'left', 'center' or 'right', we'd only require one 
useValueAndKey mapping rather than 3?

I keep finding the api in terms of what should be Common props as enums, 
and bool for others, a little erratic.  Some props feel as though they 
should be one way, but only work the other.

For example, Button actions.  It feels right that you'd declare a 
<Button action/> but this breaks down when stipulating that it should be 
left or right <Button left action/>, but would seem even worse when 
combining it with an Icon:

<Button left action right icon/>

It seems like some commonality is required across all component props, 
as the same would apply for <Container className="left align"/>

Might defined props make things clearly for the end user?

<Container leftAlign />
<Container justify />
<Button leftAction />
<Button rightIcon/>

Or indeed, imposing defined props types with oneOf enums:

<Button align="left|center|right|justify" />
> Levi Thomason <mailto:notifications@github.com>
> 24 June 2016 at 17:23
>
> In src/elements/Container/Container.js 
> <https://github.com/TechnologyAdvice/stardust/pull/277#discussion_r68423298>:
>
> > - {this.props.children}
> > - </div>
> > - )
> > - }
> > +function Container(props) {
> > + const {
> > + text, left, center, right, justified, fluid,
> > + children, className,
> > + } = props
> > +
> > + const classes = cx('sd-container ui',
> > + useKeyOnly(text, 'text'),
> > + useKeyOnly(left, 'left aligned'),
> > + useKeyOnly(center, 'center aligned'),
> > + useKeyOnly(right, 'right aligned'),
> > + useKeyOnly(justified, 'justified'),
>
> There are 4 types of possible class to props mappings for SUI classes. 
> The alignment class here follows this API pattern:
>
> aligned="left" // useValueAndKey(aligned, 'aligned')
> aligned="right" // useValueAndKey(aligned, 'aligned')
> aligned="center" // useValueAndKey(aligned, 'aligned')
>
> One reason we use this key/value relationship is because two 
> simultaneous alignment props are not valid:
>
> |<Container right left />
> |
>
> Where as making those values of the |aligned| prop means you can only 
> have one at a time:
>
> |<Container aligned="left" />
> |
>
> However, the |justified| option breaks down however since the 
> resulting class is not |justified aligned|. We've not yet run into 
> this matchup yet so I'm open to suggestions. The best immediate 
> thought I have is to still allow |aligned="justified"| for the API 
> consistency, then in cx buildup:
>
> cx(
> useValueAndKey(aligned, 'aligned'),
> useValueAndKey(aligned, 'aligned'),
> useValueAndKey(aligned, 'aligned'),
> aligned === 'justified' && 'justified'
> )
>
> Thoughts?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> <https://github.com/TechnologyAdvice/stardust/pull/277/files/f420da9c4eb10853132c436ba6a06faa82ffeba2#r68423298>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/AAWHs4uiOF121pijncyPFGMdsccwEtpJks5qPARsgaJpZM4I9hiP>.
>

-- 
Sent from Postbox 
<https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach>

docs/app/Examples/elements/Container/Types/ContainerTextExample.js
jhchill666: Done

> Levi Thomason mailto:notifications@github.com
> 24 June 2016 at 16:58
> 
> In docs/app/Examples/elements/Container/Types/ContainerTextExample.js 
> https://github.com/TechnologyAdvice/stardust/pull/277#discussion_r68419885:
> 
> > @@ -0,0 +1,21 @@
> > +/\* eslint-disable max-len */
> > +
> > +import React, { Component } from 'react'
> > +import { Container, Message } from 'stardust'
> > +
> > +export default class ContainerContainerExample extends Component {
> > -class ContainerContainerExample
> > +class ContainertextExample
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> https://github.com/TechnologyAdvice/stardust/pull/277/files/06fde8b6601dcbdb89260c966a8eba1a11f55df0#r68419885, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AAWHs-APioSkl5c_w7RXiAMgUdImf-P9ks5qO_6mgaJpZM4I9hiP.

## 

Sent from Postbox 
https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach

test/specs/elements/Container/Container-test.js
levithomason: Sorry, should've been more explicit.  You can keep all tests in this file, just remove this one:

``` js
it('adds the Container class', ...
```

All components must include a class that matches the component name.  A `<Container />` must at least include `<div class="container" />`, for example.  Because this is true of all components, it is part of the base conformance test.

The conformance tests use the component's `_meta` object to ensure the className includes the component name.

src/elements/Container/Container.js
jhchill666: I think your suggestion of just using a ternary operator is probably the best option.  You could go down the route of adding unnecessary additions to propUtils but seems it's succinct enough to be used in this one isolated case.

levithomason: > we'd only require one useValueAndKey mapping rather than 3?

You are correct, I edited my typo shortly after submitting.  See above.

> For example, Button actions.  It feels right that you'd declare a <Button action/> but this breaks down when stipulating that it should be left or right <Button left action/>, but would seem even worse when combining it with an Icon:
> 
> `<Button left action right icon/>`

For a prop that may be boolean only _or_ enum, there is the `keyOrValueAndKey()` util which allows us to do this (I assume you meant Input instead of Button):

``` js
<Input action />                       // <input class="action" />
<Input action='left' />                // <input class="left action" />
<Input action='right' />               // <input class="right action" />
```

> Or indeed, imposing defined props types with oneOf enums:
> 
> `<Button align="left|center|right|justify" />`

I think this is the way to go.  You can see this in use in the [Label for the pointing](https://github.com/TechnologyAdvice/stardust/blob/master/src/elements/Label/Label.js#L142) prop.  A Label can be pointing alone or pointing with a direction, but never more than one.

I think with this one special handling for the justified case we have a pretty good solution:

``` js
cx(
  aligned === 'justified' ? 'justified' : useValueAndKey(aligned, 'aligned'),
)
```

If the `aligned` prop is `justified`, then the class will be `justified` only. Otherwise, the class will be the value of the prop and it's key:

``` js
<Container aligned='justified' />  // <div class="ui container justified"></div>
<Container aligned='left' />       // <div class="ui container left aligned"></div>
<Container aligned='right' />      // <div class="ui container right aligned"></div>
```

levithomason: Note, we won't be able to use the common test for `propValueAndKeyToClassName` since it has special handling for one of the values.

Since the alignment prop is pretty common, it may warrant it's own `common.implementsAlignmentProp()` test to handle the special `justified` case.

jhchill666: Yeah I'd digested more of your code after making that statement so 
please disregard.

Have amended the Container alignment as per your suggestion and updated test

Levi Thomason wrote:

test/specs/elements/Container/Container-test.js
```diff
@@ -0,0 +1,24 @@
...
+import Container from 'src/elements/Container/Container'
+import * as common from 'test/specs/commonTests'
+
+describe('Container', () => {
+  common.isConformant(Container)
+  common.rendersChildren(Container)
+  common.hasUIClassName(Container)
+
+  common.propKeyOnlyToClassName(Container, 'text')
+  common.propKeyAndValueToClassName(Container, 'aligned')
```
levithomason: This is the test that is failing.  This test will ensure that the component has the classNames:

`left aligned`
`right aligned`
`center aligned`
`justified aligned` <-- it is asserting this but it is wrong because we are overriding this value

I'm merging a new propUtil to handle the className logic.  I'm also merging a new common test to ensure the `aligned` prop is correctly implemented.  I think this makes sense since it has it's own className logic and it is used in many components.

Once merged, we'll be able to do this in `cx`:

``` js
cx(
  useAlignedProp(aligned),
)
```

And this in tests:

``` js
common.implementsAlignedProp(Container)
```

levithomason: I have now merged #285 which adds `useAlignedProp` propUtil and `implementsAlignedProp` common test.

The util and test account for the special handling of the `justified` value.  You should be able to update this PR now.

src/elements/Container/Container.js
```diff
@@ -1,33 +1,62 @@
...
-  }
+function Container(props) {
+  const {
+    text, aligned, fluid,
+    children, className,
+  } = props
+
+  const classes = cx('sd-container ui',
+    useKeyOnly(text, 'text'),
+    aligned === 'justified' ? 'justified' : useValueAndKey(aligned, 'aligned'),
```
levithomason: After updating this PR and getting the new tools merged in #285, replace line 21 here with:

``` js
useAlignedProp(aligned)
```

test/specs/elements/Container/Container-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+  it('renders a <div /> element', () => {
+    shallow(<Container />)
+      .should.have.tagName('div')
+  })
+
+  it('does not have aligned class when justified', () => {
+    shallow(<Container aligned='justified' />)
+      .should.not.have.className('aligned')
+  })
```
levithomason: After updating this PR and getting the new tools merged in #285, remove this test:

``` diff
-  it('does not have aligned class when justified', () => {
-    shallow(<Container aligned='justified' />)
-      .should.not.have.className('aligned')
-  })
```

Then, replace line 12 above:

``` diff
-common.propKeyAndValueToClassName(Container, 'aligned')
+common.implementsAlignedProp(Container)
```

docs/app/Examples/elements/Container/Types/ContainerTextExample.js
jhchill666: done

docs/app/Examples/elements/Container/Types/ContainerTextExample.js
jhchill666: done

test/specs/elements/Container/Container-test.js
jhchill666: done

test/specs/elements/Container/Container-test.js
jhchill666: implemented

src/elements/Container/Container.js
jhchill666: done

test/specs/elements/Container/Container-test.js
jhchill666: done

Issue #269, opened 2016-06-22
Update legacy component APIs to the v1 API


levithomason:
Right now, many components still use className for all styling and behavior. Our v1 API, decided in #208, is more declarative. You can see it at play in the Divider, Label, and Dropdown components.  As well as any completed task below.

The [CONTRIBUTING.md](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md) guide details how to go about designing a component API.  The [Spec out the API](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#spec-out-the-api) section is of primary importance. 

You can reference PR #262 as an example of how to upgrade a legacy component API to the v1 API.  Below are the remaining legacy components APIs.  Any PR to upgrade these components is much welcome.

**Collections**
- [x] ~~#351 Grid~~
- [x] ~~#382 Menu~~
- [x] ~~#451 Table~~

**Elements**
- [x] ~~#295 Button~~
- [x] ~~#277 Container~~
- [x] ~~#343 Icon~~
- [x] ~~#280 Image~~
- [x] ~~#281 Input~~
- [x] ~~#551 List~~
- [x] ~~#278 Header~~
- [x] ~~#287 Segment~~

~~**Modules**~~
- [x] ~~#379 Modal~~

~~**Views**~~
- [x] ~~#431 Item~~
- [x] ~~#334 Statistic~~

---

In addition to these component APIs, we're also removing jQuery and SUI JS from all components.  You can see that work in #247.

layershifter:
I'll make PR for `<Menu>` soon, it's mine :gem:

layershifter:
`Table` component hasn't own PR or issue, so I ask there.

---

It looks that `Table` can be stateless component, except one case, [sortable](http://semantic-ui.com/collections/table.html#sortable).

I think about making `Table.Sortable` that has state and allow it pass shorthands or only components `Table.Row`, `Table.Header`, etc. But, it is ugly. Any better ideas there?

levithomason:
Moved here https://github.com/TechnologyAdvice/stardust/pull/451

layershifter:
Unbelievable, it's almost done ✌️

levithomason:
I know it!  Very crazy amount of work.  So proud to be a part of this community effort.  🎉

jeffcarbs:
The closing of #551 checks the last box. Can we mark this issue as closed?

[...]

PR #281, opened 2016-06-24
feat(Input): update to v1 API

### LINKED ISSUE #291: feat(Dropdown): support label prop
### UPDATE

This will likely be solved by augmentation, see #414.

---

A Label in a labeled Input can actually be a dropdown, see [here](http://semantic-ui.com/elements/input.html#labeled).  This is not the same thing as a `labeled` dropdown, which separates a dropdown button icon from the button text.

**Dropdown Label**
image

It appears the only necessary extra dropdown class required is `label`.  I'm not entirely sure how we want to handle this yet.  We could simply add a `label` prop that adds this class.

I do not think we want to automatically add this when used inside an Input though. A dropdown can also appear as an Input action:

**Dropdown Button**
image

**Dropdown**
image

jhchill666:
Fixes #291


levithomason:
# Input API Proposal

Here is how I see the v1 API applying to the Input.  Would love feedback.  I'll update this comment as we narrow things down in the comments.

## Input

``` jsx
<Input />
```

``` html
<div class="ui input">
  <input type="text" >
</div>
```

## Focus

``` jsx
<Input focus />
```

``` html
<div class="ui input focus">
  <input type="text">
</div>
```

## Loading

In SUI, the loading style automatically adds a spinner icon, or overrides the provided icon.  It then requires either an `icon` or `left icon` class to correctly place the icon on the left or right.

**Loading only**

I propose we add the `icon` class by default when the `loading` prop is given.  This would place the spinner on the right inside of the input.

``` jsx
<Input loading />
```

``` html
<div class="ui icon input loading">
  <input type="text">
</div>
```

## Disabled

``` jsx
<Input disabled />
```

``` html
<div class="ui disabled input">
  <input type="text" disabled="disabled">
</div>
```

## Error

``` jsx
<Input error />
```

``` html
<div class="ui input error">
  <input type="text">
</div>
```

## Icon

I see two possible APIs for the input icons:
1. Makes use of the `iconPropRenderer`
2. Follows the `labeled` and `action` pattern (ie `icon` or `icon='left'`) and allows an Icon as an input child.

The two suggestions here follow the first option above.  After writing the rest of the examples, I am leaning toward API number 2 (see alternative below).

**Icon**

``` jsx
<Input icon='search' />
```

``` html
<div class="ui icon input">
  <input type="text">
  <i class="search icon"></i>
</div>
```

**Left Icon**

We might expect this to be `icon='left'` however, there is then no way to pass the icon string.  Open to more ideas on how to achieve left/right icon positions.

``` jsx
<Input leftIcon='search' />
```

``` html
<div class="ui left icon input">
  <input type="text">
  <i class="users icon"></i>
</div>
```

**Alternative**

The alternative is to allow Icons as children (as we do with actions and labels). The Input `icon` class would be inferred if there is an Icon child.

It would exclude the use of the `iconPropRenderer`.  Therefore excluding the simple `icon='search'` use as well.

Though it _would_ be more consistent with the `labeled` an `action` props.  Lastly, it follows the `cx` utilities as well `icon='left' => class="left icon"`.  Whereas, `iconLeft='search'` seems to break away from the pattern.

I think I prefer this alternative syntax to the `icon` prop but would love any feedback.

``` jsx
// same as <Input icon>
<Input>
  <Icon name='user' />
</Input>

<Input icon='left'>
  <Icon name='settings' />
</Input>

<Input action='right' icon='left'>
  <Icon name='search' />
  <Button>Search</Button>
</Input>
```

``` html
<div class="ui icon input">
  <input type="text">
  <i class="user icon"></i>
</div>

<div class="ui left icon input">
  <input type="text">
  <i class="settings icon"></i>
</div>

<div class="ui left icon right action input">
  <i class="search icon"></i>
  <input type="text">
  <button>Search</button>
</div>
```

## Labeled

Labeled Inputs can be pretty complex.  I've include all the SUI examples here to be sure we cover them all.

Our current Input supports most all of these already.  The approach was to use `children` for any buttons, dropdowns, or labels.

Again, any input or feedback is much welcomed.

**Label**

Just as I am proposing the `loading` input automatically adding the `icon` class, I propose a Label child automatically adds the `labeled` style.

``` jsx
// same as `<Input labeled>`, inferred by child
<Input>
  <Label>http://<Label>
</Input>
```

``` html
<div class="ui labeled input">
  <div class="ui label">http://</div>
  <input type="text">
</div>
```

**Label Dropdown**

The docs for [this input configuration](http://semantic-ui.com/elements/input.html#labeled) show use of a `label` class on the dropdown.  This is not documented in the dropdown itself.  I've added #291 to consider how we should support it.

``` jsx
<Input labeled='right'>
  <Dropdown label options={[...]} />
</Input>
```

``` html
<div class="ui right labeled input">
  <input type="text">
  <div class="ui dropdown label">
    <div class="text">.com</div>
    <i class="dropdown icon"></i>
    <div class="menu">
      <div class="item">.com</div>
      <div class="item">.net</div>
      <div class="item">.org</div>
    </div>
  </div>
</div>
```

**Right Label**

``` jsx
<Input labeled='right'>
  <Label basic>kg<Label>
</Input>
```

``` html
<div class="ui right labeled input">
  <input type="text">
  <div class="ui basic label">kg</div>
</div>
```

**Left and Right Label**

When two labels are given, the input must be in between them in the markup.  Notice also the HTML requires the `right labeled` class even though there are labels on both sides.

I propose the order of the label children should determine which is left/right.  Also, that the style `labeled='right'` should be inferred by the presence of two labels.

``` jsx
// same as <Input labeled='right' />
<Input>
  <Label basic>$<Label>
  <Label basic>kg<Label>
</Input>
```

``` html
<div class="ui right labeled input">
  <div class="ui label">$</div>
  <input type="text">
  <div class="ui basic label">.00</div>
</div>
```

**Left Corner Labeled**

Corner labeled inputs require HTML classes that match the corner position of the Label itself.  I propose we infer the Input classes from the Label props.

``` jsx
// same as <Input labeled='left corner' />
<Input>
  <Label corner='left'>
    <Icon name='asterisk' />
  <Label>
</Input>
```

``` html
<div class="ui left corner labeled input">
  <input type="text">
  <div class="ui left corner label">
    <i class="asterisk icon"></i>
  </div>
</div>
```

**Corner Labeled**

As with the left corner labeled Input above, I propose we infer the Input labeled classes to match the label props.

> See #292 for Label boolean corner prop support

``` jsx
// same as <Input labeled='corner' />
<Input>
  <Label corner>
    <Icon name='asterisk' />
  <Label>
</Input>
```

``` html
<div class="ui corner labeled input">
  <input type="text">
  <div class="ui corner label">
    <i class="asterisk icon"></i>
  </div>
</div>
```

## Action

I see the `action` prop behaving much like the `labeled` prop.  Since there are only a limited number of supported Input actions, I think we should infer the `action` prop settings for the Input.

I also notice `action` class is synonymous to `right action`.

**Action**

``` jsx
// same as <Input action />
<Input>
  <Button>Search<Button>
</Input>
```

``` html
<div class="ui action input">
  <input type="text">
  <button class="ui button">Search</button>
</div>
```

**Left Action**

``` jsx
<Input action='left'>
  <Button color='teal' labeled icon='cart'>
    Search
  <Button>
</Input>
```

``` html
<div class="ui left action input">
  <button class="ui teal labeled icon button">
    <i class="cart icon"></i>
    Checkout
  </button>
  <input type="text" value="$52.03">
</div>
```

**Right Action with Left Icon**

> This shows the _alternative_ form of the proposed icon API

``` jsx
<Input action='right' icon='left'>
  <Icon name='search' />
  <Dropdown basic floating button options={[...]} />
</Input>
```

``` html
<div class="ui right action left icon input">
  <i class="search icon"></i>
  <input type="text">
  <div class="ui basic floating dropdown button">
    <div class="text">This Page</div>
    <i class="dropdown icon"></i>
    <div class="menu">
      <div class="item">This Organization</div>
      <div class="item">Entire Site</div>
    </div>
  </div>
</div>
```

**Dropdown Action**

``` jsx
// same as <Input action>, inferred by children
<Input>
  <Dropdown compact selection options={[...]} />
  <Button type='submit'>Search</Button>
</Input>
```

``` html
<div class="ui action input">
  <input type="text">
  <select class="ui compact selection dropdown">
    <option value="all">All</option>
    <option selected="" value="articles">Articles</option>
    <option value="products">Products</option>
  </select>
  <div type="submit" class="ui button">Search</div>
</div>
```

## Transparent

``` jsx
<Input transparent />
```

``` html
<div class="ui transparent input">
  <input type="text">
</div>
```

## Inverted

``` jsx
<Input inverted />
```

``` html
<div class="ui inverted input">
  <input type="text">
</div>
```

## Fluid

``` jsx
<Input fluid />
```

``` html
<div class="ui fluid icon input">
  <input type="text">
</div>
```

## Size

``` jsx
<Input size='mini' />
<Input size='small' />
<Input size='large' />
<Input size='big' />
<Input size='huge' />
<Input size='massive' />
```

``` html
<div class="ui mini icon input">
  <input type="text">
</div>

<div class="ui small icon input">
  <input type="text">
</div>

<div class="ui large icon input">
  <input type="text">
</div>

<div class="ui big icon input">
  <input type="text">
</div>

<div class="ui huge icon input">
  <input type="text">
</div>

<div class="ui massive icon input">
  <input type="text">
</div>
```

jhchill666:
I've been having a look at labeled/action syntax and seems like a bit of a stumbling point, not only from an implementation standpoint, but also one of user clarity.

Leaving Icon out, how many labels/actions can an Input feasibly have?  Is it correct to assume one of either, one left and one right?  Assuming there is some limit, how do you handle other added components that don't fit the template?

I'm not a fan of making user assumptions, as more often than not they turn out to be incorrect. Is there no way of ring fencing left/right label/action's and to follow the emerging syntax a little more closely?  (just thinking out loud here):

1 Delegate responsibility for positioning to sub-components

```
<Input>
  <Input.Left icon="search" />
  <Input.Right action>
    <Dropdown compact selection options={[...]} />
  </Input.Right>
</Input>
```

2 Delegate responsibility for behavior to sub-components

```
<Input>
  <Input.Labeled left icon="search" />
  <Input.Action right>
    <Button type='submit'>Search</Button>
  </Input.Action>
</Input>
```

3 Delegate auxillary components to sub-components

```
<Input>
  <Input.Label left icon="search" />
  <Input.Button right>Search</Input.Button>
</Input>
```

`?`

jamiehill:
This syntax could easily be rolled out across similar components.  Was scratching my head looking at Button labeled/icons earlier, and this could fit the bill there quite well also:

```
<Button>
  <Button.Left labeled icon="pause"/>
</Button>
<Button>
  <Button.Right labeled icon="right arrow" />
</Button>
```

src/elements/Input/Input.js
```diff
@@ -1,74 +1,100 @@
...
+    const childIsAction = !isLabel && isButton || isDropdown
+
+    if (childIsAction) {
+      actionChildren.push(child)
+    } else if (isLabel) {
+      labelChildren.push(child)
+    }
+  })
+
+  const classes = cx('sd-input ui',
```
levithomason: Heads up, I've finished removing all `sd-*` classes, see #301.  Rebase to master to get the update.  The test for the `sd-*` class has been inverted to fail if any `sd-` class exists.

levithomason:
I've got an updated proposal that I intend to post here once I finish up the Form in #400.  My focus will then shift to Input and Button since they are very related.

levithomason:
Heads up!  I'm resuming work on this PR now 🍻

jeffcarbs:
@levithomason - I have a request/idea for this component: It's currently not possible to add props to the underlying input that are also props of the `Input` component (e.g. there's no way to add a class to the `input` since `className` is a handled prop of `Input`. This is preventing usage of the `Input` component within the `Search` component since the `input` tag needs the `prompt` class.

I'd propose an `input` shorthand prop that is passed to a factory to create the `<input>` tag. A string could be treated at the `type`, options would be passed as props. The defaultProp would be `text` so you'd get the current default (`<input type="text">`).

The weird thing here would be that you could technically pass `null` to not render an `<input>` tag at all, although I can't see any scenario where anyone would do that unintentionally.

levithomason:
Cool, I'll incorporate an `input` shorthand here.

src/elements/Input/Input.js
```diff
@@ -43,73 +49,113 @@ const inputPropNames = [
...
+  const unhandledProps = getUnhandledProps(Input, props)
+  const inputProps = _.pick(unhandledProps, inputPropNames)
+  const rest = _.omit(unhandledProps, inputPropNames)
+  const ElementType = getElementType(Input, props)
+  return (
+    <ElementType {...rest} className={classes}>
+      {isLeftLabeled && labelChildren}
+      {isLeftAction && actionChildren}
+      <input {...inputProps} type={type} />
+      {icon && <Icon name={icon} />}
```
jeffcarbs: This should use the `createIcon` factory.

src/elements/Input/Input.js
```diff
@@ -43,73 +49,113 @@ const inputPropNames = [
...
+  const classes = cx(
+    'ui',
+    size,
+    useKeyOnly(disabled, 'disabled'),
+    useKeyOnly(error, 'error'),
+    useKeyOnly(fluid, 'fluid'),
+    useKeyOnly(inverted, 'inverted'),
+    useKeyOnly(loading, 'loading'),
+    useKeyOnly(transparent, 'transparent'),
+    icon && 'icon',
```
jeffcarbs: Should this be `useKeyOnly`? Or is that reserved for boolean props?

levithomason: Should be, thanks.

src/elements/Input/Input.js
levithomason: Indeed, will update.

levithomason:
## Update

The newly updated API spec for customizing the HTML `<input />` and dealing with the labeled and action variations this looks like so:

``` jsx
<Input
  label='http://'
  labelPosition='left'
  input={<input />}
  action={<Button icon='heart' content='Like' />}
  actionPosition='right'
/>
```

Where `label`, `input`, and `action` are all shorthand props.  They will accept a literal, props object, or element. Also note, any input props passed are automatically spread on the `<input />` itself.  So the `input` shorthand prop here is mainly needed for adding classes / style props to the input.

This will be my main focus tomorrow.  The Button API is scheduled to be updated to keep consistent.  See #510.

levithomason:
The Input API spec of `<Input label='.com' labelPosition='right' />` will not work for inputs with a label on the left and on the right:

image

The same actions API `<Input action='Checkout' actionPosition='left' />` will not work for inputs with multiple actions:

image

The first thought was to go with `leftLabel` `rightLabel`, however, this does not extend to the actions very well.  There can be two actions on one side.  Ideally, the solution would be consistent for icons, labels, and actions and for all components.  Here is a start at the available patterns for each prop that will work with all permutations:

#### `icon` `iconPosition`

Works for all components, only ever one icon

#### `leftLabel` `rightLabel`

Works for all components, only ever one label per side

#### `actions` `actionsPosition`

If actions took an array of components, this would work.  can have multiple actions, but only ever on one side at a time

What I don't like is the inconsistency here.  I'd almost rather support a single icon, label, or action and go with `foo` `fooPosition` across the board.

@jcarbo @layershifter Help? :confused:

layershifter:
What about this?

``` jsx
const elements = [
  <Label>Left</Label>,
  {position: 'left', element: <Label>Left</Label>},
  {position: 'right', element: <Label>Rigth</Label>},
];

function Input () {
  const leftElements = _map(elements, element => {
    if(_.isObject(element) && element.position === 'right') return
    return element
  })

  const rightElements = _map(elements, element => {
    if(_.isObject(element) && element.position === 'right') {
      return element
    }
  })

  return (
    <div>
      {leftElements}
      <input>
      {rightElements}
   </div>
  )
}

const myInput = <Input elements={elements} />
```

It will be simply validatable with props, but there will be `key` problem.

levithomason:
It is an option, I actually suggested this same thing to our team offline :)

``` jsx
<Input labels={[
  { position: 'left', content:'$' },
  { position: 'right', content: '.00', basic: true },
]} />
```

It just seems too verbose / clunky IMO.  That said, it is better than having a different API for every prop.  I'm not sure if it is better than just supporting a single icon/label/action with a single `foo``fooPosition` API though.

layershifter:
Okay, what about full sub component markup? It will make everything simplier there.

``` jsx
<Input>
  <Label>1</Label>
  <Label>2</Label>
  <input type="text" />
  <Label>3</Label>
  <Label>4</Label>
</Input>
```

Or even this?

``` jsx
<Input>
  <Input.Left>
    <Label>1</Label>
    <Label>2</Label>
  </Input.Left>
  <Input.Right>
    <Label>3</Label>
    <Label>4</Label>
  </Input.Right>
</Input>
```

levithomason:
Thanks, I'll let this incubate a bit while I work on the prop standardization, #452.

owenoak:
I love the simplicity of `<Input label='foo'/>`.  I think 90%+ of the time you just need a single label, icon, etc next to an input.  Supporting more than one label elegantly is great, but I'd hate to lose declarative simplicity to support what seems like an edge case.

My preference would be to have 1 label be done with a `label` prop, with `labelPosition` to dictate the side.

(FWIW, I’d prefer `labelOn` over `labelPosition`, less to type, reads a bit cleaner IMO and fits with `on` prop proposal below) 

```
<Input label=“Foo” labelOn=“right”/>
```

yields

```
<div class=“ui right labeled input”>
  <input type=“text”/>
  <div class=“ui label”>Foo</div>
</div>
```

For simplicity, I’d make inline labels exclusive with the `label...` props, but supporting both would probably work.

Then if the developer wants more than one label or to do something fancy, they can just create inline elements and use an `on` prop in the child to set the order.  The parent element would “consume” the `on` prop of the nested elements.  I’d probably infer the order based on child order if `on` wasn’t specified.

**Fancy label**

```
<Input>
  <Label on=“right” pointing="left">That name is taken!</Label>
</Input>
```

yields

```
<div class=“ui right labeled input”>
  <input type=“text”/>
  <div class=“ui left pointing label”>That name is taken!</div>
</div>
```

**Multiple Labels**

```
<Input>
  <Label on=“left” basic>$</Label>
  <Label on=“right” basic>kg</Label>
</Input>
```

yields

```
<div class=“ui right labeled input”>
  <div class=“ui basic label”>$</div>
  <input type=“text”/>
  <div class=“ui basic label”>kg</div>
</div>
```

I’d do the same with `icon` and `iconOn` props:  they can easily get one icon with `props`, if they want more than one, use nested elements.

Same for actions: `action` and `actionOn` props for one auto-created action, then use inline elements with an “on” prop to specify more than one.  This extends nicely to the case where you have multiple actions on a side.

```
<Input>
  <Button on=“left” content=“Left"/>
  <Button on=“right” primary content=“Right1"/>
  <Button on=“right” content=“Right2"/>
</Input>
```

yields

```
</div class=“ui action input”>
  <button class=“ui button”>Left</button>
  <input type=“text”/>
  <button class=“ui primary button”>Right1</button>
  <button class=“ui button”>Right2</button>
</div>
```

The child `on` paradigm should extend well to other elements (Button, etc) which auto-create sub-elements given certain props.

I'd be happy to work on supporting this if there's interest.

levithomason:
I really like `labelOn` for its brevity.  I'll run this by our team for some more input.  I'd like to use this for the props API.

I also really love the idea of not having an `<input />` child by consuming the `on` prop for children.  However, the sub component API is the escape hatch for when the props API boxes you in.  We must provide full access to the markup in at least one of the APIs, otherwise, we create a trap that we're sure to fall into.

There are two (albeit unlikely) limitations we'd add with the `on` child prop API:
1. Using a child component that had an actual `on` prop.  If we consume it, then the component will never receive it.  We'd have to assume the user will never use a component child that requires an `on` prop.
2. Inability to access the `<input />` directly.  If we abstract away this element, then the user will never be able to access it (to wrap it or what have you).  Since shorthand props are not allowed with children, the user also would not be able to pass a shorthand `input` and the child components.  We could create a trap here.

Since the difference in the amount of code for adding `on` markup to children vs adding an actual `input` is negligible, but requires more logic and brings more limitations, I think we should avoid `on` for children.  Thanks for the detailed write up and engineering contribution, much appreciated.

levithomason:
Folks I respect have out voted me on `labelOn`, looks like we'll go with `labelPosition`!

owenoak:
I take your point about the `on` prop being consumed when sub-elements might want to use it.  It's non-optimal but it sure does feel natural to me when I read the JSX.

What's your current thinking for the ability to put a label on either side?  I'm really not wild about the "labels" option above -- if I'm specifying components in JSX, I want to specify sub-components in JSX as well, not JSON.  It also doesn't extend well to `action`s where you might want more than one on a side (unless you add a "postion" or "on" prop to your JSON spec, or use a map of arrays, etc).  Also, for the 90% case where I only want one `label`, having to use `labels` (and presumably pass in an array) is awkward.

What about the strategy of using an "empty" `<input>` as a placeholder value inside the `<Input>`, eg:

```
<Input>
    <Label>Foo</Label>
    <input>
    <Label>Bar</Label>
    <Dropdown>...</Dropdown>
    <Button>...</Button>
<Input>
```

The description of how to use this would be:

"If you want more than one label, icon, action, etc, place an `<input/>` sub-element **without any props** as a marker in the place where you want the input control to go, relative to the other subcomponents."

Eg:
- If they specify an `<input>` sub-element **without props,** we'll replace it, either with the automatically generated `<input>` element, or with the value of the `input` prop if they provide one.
- If they specify an `<input>` sub-element **with any props**, we'll use that as the main input element.
- If they _don't_ specify an `<input>` sub-element, all of the labels/etc will appear **before** the auto-generated input element.  If this happens and they get confused, a quick look at the docs tells them an easy way to get exactly what they want.

Actually, if we had this, I'd be content to drop the "labelPosition" prop and just default to putting the auto-generated `<Label>` from the `label` prop on one side or the other.  Keep the set of props as simple as possible, and let them use nested elements to do anything remotely fancy.

owenoak:
Also, re: `labelOn` vs `labelPosition` -- `position` feels more "techy" and doesn't seem to fit with the "make things as english-like as possible" philosophy of Semantic UI.  

If I was describing the component to someone, I'd likely say "put the label on the right" -- which translates pretty exactly to `labelOn="right"`.  

I'd be very unlikely to use `position` when describing something this to someone, if so I'd probably say "position the label on the right".  Mapping that to `labelPosition="right"` feels awkward to translate mentally and doesn't feel like english.  And I note that I'd still say "label on...right".  :-)

levithomason:
Most all the above comments are outdated and inaccurate.

## Current API (single label/action)

The majority use case of a single label/action can be achieved without child components:

image

This also translates well for complicated labels, like in SUI's ability to use a Dropdown as a Label:

image

## Current API (multiple labels/actions)

The current state of this PR (and vote) is to use the below API for multiple labels.  It will also be used for actions.

image

## `input` child

I want to make things as simple as possible.  Given the SUI constraints here, I think that means render children based on props:

``` jsx
<Input />
```

``` html
<div class="ui input">
  <input type="text" />
</div>
```

Otherwise render the children as is:

``` jsx
<Input>
  ...
</Input>
```

``` html
<div class="ui input">
  ...
</div>
```

This is very clear and consistent with our other components.  Any additional branch logic I fear will further complicate things.

## `labelOn` naming

I agree with your comments on SUI's philosophy here.  There will be a final sweep through all components and props once we finish the legacy component migration, #269.  I'll reconsider this at that time.  Appreciate your zeal regarding it.

levithomason:
Ah!  Long awaited, but I think it is finally done.

docs/app/Components/Sidebar/Sidebar.js
```diff
@@ -51,12 +51,21 @@ export default class Sidebar extends Component {
...
+  componentDidUpdate(prevProps, prevState) {
+    this.setSearchInput()
   }
 
   componentWillUnmount() {
     document.removeEventListener('keydown', this.handleDocumentKeyDown)
   }
 
+  setSearchInput() {
+    this._searchInput = findDOMNode(this).querySelector('.ui.input input')
```
levithomason, line 66: Input is now a stateless component, so no `refs` are available.  Updated the doc sidebar to use a regular query selector instead.

src/elements/Input/Input.js
```diff
@@ -1,118 +1,230 @@
...
-        {Icon.create(icon)}
-        {isRightLabeled && labelChildren}
-        {isRightAction && actionChildren}
-      </ElementType>
-    )
+  if (children) {
+    return <ElementType {...rest} className={classes}>{children}</ElementType>
   }
+
+  const actionElement = Button.create(action, elProps => ({
```
levithomason, line 96: Sometimes the defaultProps set by a shorthand factory are conditional based on the elements props.  Updated factory default props to take an object or a function that receives the props and returns the defaultProps.

test/specs/commonTests.js
```diff
@@ -571,7 +572,8 @@ export const implementsShorthandProp = (Component, options = {}) => {
     ShorthandComponent,
     mapValueToProps,
     requiredProps = {},
-    requiredShorthandProps = {},
+    shorthandDefaultProps = {},
+    alwaysPresent,
```
levithomason, line 576: Added an `alwaysPresent` flag to shorthand props.  This means it is not tested for "does not exist when null" and other cases.  The `Input` always has a shorthand `input` rendered.

levithomason (commented):
@jcarbo @layershifter this is ready for final.  If either of you could scan the code or pull this branch and try it out, I'd much appreciate it.  Thanks!

src/elements/Input/Input.js
```diff
@@ -1,118 +1,231 @@
...
+    className: cx(
+      // all action components should have the button className
+      !_.includes(elProps.className, 'button') && 'button',
+    ),
+  }))
+  const iconElement = Icon.create(icon)
+  const labelElement = Label.create(label, elProps => ({
+    className: cx(
```
dvdzkwsk, line 104: Any reason for not using what cx offers here?

``` js
className: cx({
  label: !_includes(...),
  [labelPosition]: _.includes(...),
})
```

src/lib/factories.js
```diff
@@ -30,23 +30,39 @@ const mergePropsAndClassName = (defaultProps, props) => {
  * @param {function|string} Component A ReactClass or string
  * @param {function} mapValueToProps A function that maps a primitive value to the Component props
  * @param {string|object|function} val The value to create a ReactElement from
- * @param {object} defaultProps Default props to add to the final ReactElement
+ * @param {object|function} [defaultProps] Default props object or function (called with regular props).
```
dvdzkwsk: Is there a clearer term that we could use rather than `regular` props?

src/modules/Dropdown/Dropdown.js
```diff
@@ -184,7 +187,7 @@ export default class Dropdown extends Component {
 
     inline: PropTypes.bool,
     labeled: PropTypes.bool,
-    linkItem: PropTypes.bool,
+    // linkItem: PropTypes.bool,
```
dvdzkwsk, line 190: cruft?

src/modules/Dropdown/Dropdown.js
```diff
@@ -903,7 +907,7 @@ export default class Dropdown extends Component {
       icon,
       inline,
       labeled,
-      linkItem,
+      // linkItem,
```
dvdzkwsk, line 910: ^^

src/modules/Dropdown/Dropdown.js
```diff
@@ -935,7 +940,7 @@ export default class Dropdown extends Component {
       // useKeyOnly(icon, 'icon'),
       useKeyOnly(labeled, 'labeled'),
       // TODO: linkItem is required only when Menu child, add dynamically
-      useKeyOnly(linkItem, 'link item'),
+      // useKeyOnly(linkItem, 'link item'),
```
dvdzkwsk, line 943: ^^

docs/app/Examples/elements/Input/States/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+import { Message } from 'src'
+
+const InputStates = () => (
```
layershifter, line 7: Var name might be `States`

docs/app/Examples/elements/Input/Types/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+const InputTypes = () => (
```
layershifter, line 5: ^^

docs/app/Examples/elements/Input/Variations/InputRightLeftLabeled.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Input, Label } from 'stardust'
+
+const InputRightLeftLabeled = () => (
+  <Input labelPosition='right' placeholder='mysite.com'>
```
layershifter, line 5: `labelPosition='right'` is cruft there?

docs/app/Examples/elements/Input/Variations/index.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+import { Message } from 'src'
+
+const InputVariations = () => (
```
layershifter, line 7: ^^

src/lib/factories.js
```diff
@@ -30,23 +30,39 @@ const mergePropsAndClassName = (defaultProps, props) => {
...
+  } else if (_.isString(val) || _.isNumber(val)) {
+    type = 'literal'
+    usersProps = mapValueToProps(val)
   }
 
-  // Create ReactElements from props objects
-  if (_.isPlainObject(val)) {
-    return <Component {...mergePropsAndClassName(defaultProps, val)} />
+  defaultProps = _.isFunction(defaultProps) ? defaultProps(usersProps) : defaultProps
+  const props = mergePropsAndClassName({ ...defaultProps }, usersProps)
```
jeffcarbs: I don't see any reason to spread `defaultProps` here. This could just be:

``` js
const props = mergePropsAndClassName(defaultProps, usersProps)
```

src/elements/Input/Input.js
levithomason: In early tests, the object form was 8-10x slower than the short circuited argument approach.  This is one reason we went with the lib utils we have for className buildup.  The timing difference will probably never be an issue.  I just continue it out of consistency and habit.

src/lib/factories.js
levithomason: I struggled with that, suggestions welcome.  It is one of 3 props.  The props from the element the user passed, a props object the user passed, or the props generated by the factory's `mapValueToProps` using string/number literal the user passed.

src/modules/Dropdown/Dropdown.js
levithomason: Thanks for the careful eye.  Intentionally disabled and will probably be removed entirely when the rest of the Dropdown examples are completed.

This PR sets a precedent for how we deal with components that require special classNames only when used in combinatory contexts (Input with `action` component).  This was done with the Input action and label.  Both of which should always have an `action` or `label` className respectively.

In the factory we can add classNames:

``` jsx
Button.create(val, { className: 'action' })
```

Or, we can use augmentation, which merges classNames:

``` jsx
<Dropdown as={Menu.Item} />
```

docs/app/Examples/elements/Input/States/index.js
levithomason: In order to aide in searching, let's always keep the component name in the default export name.  Let's also keep the filename the same as the default export (except for index.js files of course).

This way the default export matches the filename and we can easily search for either.  This makes automated refactors much easier to work with.  Once we're done with the legacy components I plan to update all doc files to be consistent.

docs/app/Examples/elements/Input/Variations/InputRightLeftLabeled.js
levithomason: I wish :(  The input has some very odd markup.  I think it needs some core SUI work to make more sense.  The `labeled` class alone formats the left label.  While adding `right labeled` also formats the right label, but the `labeled` portion of that class still formats the left label also.  Very confusing and unfortunate. 

http://semantic-ui.com/elements/input.html#labeled

image

src/lib/factories.js
levithomason: Ah, looks like `defaultProps` can be undefined, in which case that was creating an empty object so `mergePropsAndClassName` didn't throw.

I've instead just defaulted `defaultProps = {}` and removed the spread for clarity.

levithomason:
Whew, released in `stardust@0.48.0`!  Thanks to everyone for chipping in on this one in so many ways.  Much appreciated.

PR #283, opened 2016-06-24
Add ability to remove dropdown items on backspace

ianunruh:
Fairly simple change that only affects dropdowns with `search` and `multiple`. If the dropdown is focused, the search query is empty and backspace is pressed, the most recently selected item will be removed.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -222,12 +222,14 @@ export default class Dropdown extends Component {
       } else {
         document.addEventListener('keydown', this.moveSelectionOnKeyDown)
         document.addEventListener('keydown', this.selectItemOnEnter)
+        document.addEventListener('keydown', this.removeItemOnBackspace)
       }
     } else if (prevState.focus && !this.state.focus) {
       document.removeEventListener('keydown', this.openOnArrow)
       document.removeEventListener('keydown', this.openOnSpace)
       document.removeEventListener('keydown', this.moveSelectionOnKeyDown)
       document.removeEventListener('keydown', this.selectItemOnEnter)
+      document.removeEventListener('keydown', this.removeItemOnBackspace)
     }
 
     // opened / closed
@@ -236,6 +238,7 @@ export default class Dropdown extends Component {
       document.addEventListener('keydown', this.closeOnEscape)
       document.addEventListener('keydown', this.moveSelectionOnKeyDown)
       document.addEventListener('keydown', this.selectItemOnEnter)
+      document.addEventListener('keydown', this.removeItemOnBackspace)
       document.addEventListener('click', this.closeOnDocumentClick)
       document.removeEventListener('keydown', this.openOnArrow)
       document.removeEventListener('keydown', this.openOnSpace)
@@ -244,6 +247,7 @@ export default class Dropdown extends Component {
       document.removeEventListener('keydown', this.closeOnEscape)
       document.removeEventListener('keydown', this.moveSelectionOnKeyDown)
       document.removeEventListener('keydown', this.selectItemOnEnter)
+      document.removeEventListener('keydown', this.removeItemOnBackspace)
       document.removeEventListener('click', this.closeOnDocumentClick)
     }
   }
@@ -254,6 +258,7 @@ export default class Dropdown extends Component {
     document.removeEventListener('keydown', this.openOnSpace)
     document.removeEventListener('keydown', this.moveSelectionOnKeyDown)
     document.removeEventListener('keydown', this.selectItemOnEnter)
+    document.removeEventListener('keydown', this.removeItemOnBackspace)
     document.removeEventListener('keydown', this.closeOnEscape)
     document.removeEventListener('click', this.closeOnDocumentClick)
   }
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -335,6 +340,22 @@ export default class Dropdown extends Component {
     }
   }
 
+  removeItemOnBackspace = (e) => {
+    debug('remoteItemOnBackspace()')
```
levithomason: Typo `remote` here.

src/modules/Dropdown/Dropdown.js
```diff
@@ -335,6 +340,22 @@ export default class Dropdown extends Component {
...
+
+    const { multiple, search } = this.props
+    const { searchQuery, value } = this.state
+
+    if (searchQuery || !search || !multiple || !value) return
+
+    e.preventDefault()
+
+    // remove most recent value
+    this.onChange(e, _.dropRight(value))
```
levithomason: In case this component is not controlled (has no value / onChange props) then we need to handle updating the value ourselves.  We'll need a `this.setValue()` call just before this to support both controlled and uncontrolled dropdowns:

``` js
// remove most recent value
const newValue = _.dropRight(value)

this.setValue(newValue)
this.onChange(e, newValue)
```

If the component is not controlled, `setValue` will succeed in setting the new value to state.  If it is controlled, `setValue` attempt will be ignored.  In this case, broadcasting the change via onChange will change our props from the outside and we'll take that in as the new value.

ianunruh:
Thanks for the feedback so far. I'm pretty new to React and ES2015, so my code quality might not be the best :)

levithomason:
We'll need some coverage in order to merge this.  Could you write some tests?  We can add a new section to `Dropdown-test.js` for this:

``` js
describe('removing items on backspace', () => {
  it('removes the last item when there is no search query', () => {
    // ...
  })
  it('does not remove the last item when there is a search query', () => {
    // ...
  })
  it('does not remove items for multiple dropdowns without search', () => {
    //...
  })
})
```

Something like this to get started.  `npm test` to get running.  I'd suggest flagging your test with `.only()` so as to not be overwhelmed by the other tests:

``` js
describe.only('removing items on backspace', () => {
...
```

levithomason:
> Thanks for the feedback so far. I'm pretty new to React and ES2015, so my code quality might not be the best :)

No worries, thoroughly pleased to see this PR.  I knew this feature was missing but just didn't have the time to implement it.  Will gladly walk you through.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -620,6 +620,62 @@ describe('Dropdown Component', () => {
...
+      const expected = _.without(value, _.last(value))
+      wrapperMount(<Dropdown {...requiredProps} value={value} multiple search onChange={spy} />)
+
+      // open
+      wrapper.simulate('click')
+
+      domEvent.keyDown(document, { key: 'Backspace' })
+
+      spy.should.have.been.calledOnce()
+      spy.firstCall.args[1].should.deep.equal(expected)
```
levithomason, line 650: Looks good, can we add a state assertion to ensure the component is being self controlled:

``` js
wrapper.should.have.state('value', expected)
```

levithomason:
This is lookin superb, just the one request for a state assertion on the test here https://github.com/TechnologyAdvice/stardust/pull/283#discussion_r68460171 and I think we can merge this guy.  Thanks much!

src/modules/Dropdown/Dropdown.js
```diff
@@ -335,6 +340,25 @@ export default class Dropdown extends Component {
...
 
+  removeItemOnBackspace = (e) => {
+    debug('removeItemOnBackspace()')
+    debug(keyboardKey.getName(e))
+    if (keyboardKey.getCode(e) !== keyboardKey.Backspace) return
+
+    const { multiple, search } = this.props
+    const { searchQuery, value } = this.state
+
+    if (searchQuery || !search || !multiple || !value) return
```
levithomason: Found a little bug here.  When there are not items left and you hit backspace, this method still calls onChange and setValue.  This is because `!value` here is still true since it is an empty array.

Being that multiple dropdowns have arrays for value we should use `_.isEmpty(value)` here instead.  This will be true if empty, false if there are any items in the array.

test/specs/modules/Dropdown/Dropdown-test.js
ianunruh: I'll need to add a test with an uncontrolled component. However, I'm having an issue where even though `expected` and `state.value` have the same elements, the assertion fails with `expected <Dropdown /> to have a 'value' state with the value [ Array(4) ], but the value was [ Array(4) ]`. Is there something other than `should.have.state` that I should be using instead? Looking through other tests in Stardust, I don't see any usable of `should.have.state` with arrays.

levithomason, line 650: Hm, I wonder if it is a deep equality issue.  You could try using the getter form of `state` and then deep equal:

``` js
wrapper
  .state('value')
  .should.deep.equal(expected)
```

If this still gives you troubles, don't worry about adding it, we can iterate on this.

ianunruh, line 650: That did the trick.

dvdzkwsk:
👻 looks great

levithomason:
Thanks much for the PR @ianunruh.  Hope to see more of your great work here.

PR #287, opened 2016-06-26
Segment: update to v1 API

jhchill666:


docs/app/Examples/elements/Segment/Variations/SegmentAttachedComplexExample.js
```diff
@@ -5,25 +5,25 @@ export default class SegmentAttachedComplexExample extends Component {
...
-        <Header.H5 className='attached header'>
+        <Header.H5 attached>
           Lions
         </Header.H5>
-        <Segment className='attached segment'>
+        <Segment attached>
           Humans don't think of lions as being like cats, but they are.
         </Segment>
-        <Message className='bottom attached warning'>
+        <Message attached='bottom' className='warning'>
```
levithomason: `<Message />` has not yet been updated to v1 API, so the `attached` prop won't work here.

test/specs/elements/Segment/Segment-test.js
```diff
@@ -2,16 +2,15 @@ import faker from 'faker'
...
 
 describe('Segment', () => {
   common.isConformant(Segment)
-  common.hasUIClassName(Segment)
-  common.rendersChildren(Segment)
-  common.hasSubComponents(Segment, [Segments])
+  common.hasSubComponents(Segment, [SegmentGroup])
 
-  describe('heading', () => {
+  // TODO re-enable when Header v1 merged to master
```
levithomason: Done

src/elements/Segment/SegmentGroup.js
```diff
@@ -0,0 +1,42 @@
...
+  const classes = cx(
+    className
+  )
+
+  const rest = getUnhandledProps(SegmentGroup, props)
+
+  return (
+    <_Segment className={classes} {...rest} _sdClass='sd-segment-group' _segmentClass='segments'>
+      {children}
+    </_Segment>
```
levithomason: A SegmentGroup only supports `horizontal`, `raised`, `stacked` and `piled` classes.  Whereas the individual Segment will not support these.

Because of this, I think we should just use a `div` here in the SegmentGroup and move those props from the _Segment to the SegmentGroup.  Then, rename the _Segment to Segment.

---

As an aside it is worth noting how the docs work here.  Any non-private class has documentation generated for it.  The doc blocks on props generate the props table, the class/function doc block generates the primary component heading.

This means if a component is marked private by an underscore, it will not show up in the docs.  This also means, if another component imports a private a component and uses it (Segment => _Segment), the exposed component (Segment) will not appear to have props or description documentation because the docs are hidden in the private component.

src/elements/Segment/_Segment.js
```diff
@@ -0,0 +1,116 @@
...
+  children: PropTypes.any,
+
+  /** Class names for custom styling. */
+  className: PropTypes.string,
+
+  /** Optional segment heading. */
+  heading: PropTypes.any,
+
+  /** */
+  raised: PropTypes.bool,
```
levithomason: ~~This is a SegmentGroup prop only.~~ lies, this is OK

src/elements/Segment/_Segment.js
```diff
@@ -0,0 +1,116 @@
...
+
+  /** Formatted to show it contains multiple pages. */
+  stacked: PropTypes.bool,
```
levithomason: ~~This is a SegmentGroup prop only.~~ lies, this is OK

src/elements/Segment/_Segment.js
```diff
@@ -0,0 +1,116 @@
...
+
+  /** Formatted to look like a pile of pages. */
+  piled: PropTypes.bool,
```
levithomason: ~~This is a SegmentGroup prop only.~~ lies, this is OK

docs/app/Examples/elements/Segment/Groups/SegmentHorizontalSegmentsExample.js
```diff
@@ -4,11 +4,11 @@ import { Segment } from 'stardust'
 export default class SegmentHorizontalSegmentsExample extends Component {
   render() {
     return (
-      <Segment.Segments className='horizontal'>
+      <Segment.Group aligned='horizontal'>
```
levithomason: Since the [`horizontal`](http://semantic-ui.com/elements/segment.html#horizontal-segments) class here applies to the layout of the segments and there is only one option, I think we should use a boolean prop:

``` js
<Segment.Group horizontal />
```

image

Please let me know if I'm over looking anything.

docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
```diff
@@ -5,9 +5,9 @@ export default class SegmentVerticalSegmentExample extends Component {
   render() {
     return (
       <Segment>
-        <Segment className='vertical'>Te eum doming eirmod, nominati pertinacia argumentum ad his.</Segment>
-        <Segment className='vertical'>Pellentesque habitant morbi tristique senectus.</Segment>
-        <Segment className='vertical'>Eu quo homero blandit intellegebat. Incorrupte consequuntur mei id.</Segment>
+        <Segment aligned='vertical'>Te eum doming eirmod, nominati pertinacia argumentum ad his.</Segment>
+        <Segment aligned='vertical'>Pellentesque habitant morbi tristique senectus.</Segment>
+        <Segment aligned='vertical'>Eu quo homero blandit intellegebat. Incorrupte consequuntur mei id.</Segment>
```
levithomason: [Vertical](http://semantic-ui.com/elements/segment.html#vertical-segment) here applies to the vertical spacing and lack of vertical borders between consecutive segments.  I think this is another case of a boolean prop:

``` js
<Segment vertical />
```

image

This is what makes sense to me at least.  Let me know if you disagree and why.

src/elements/Segment/_Segment.js
```diff
@@ -0,0 +1,116 @@
...
+  )
+}
+
+_Segment._meta = {
+  library: META.library.semanticUI,
+  name: '_Segment',
+  type: META.type.element,
+  props: {
+    attached: ['top', 'bottom'],
+    floated: ['left', 'right'],
```
levithomason: I have merged #288 which includes two new sui utils: `sui.textAlignments` and `sui.floats`.

docs/app/Examples/elements/Segment/Variations/SegmentAttachedComplexExample.js
jhchill666: My mistake, have pushed a revert

test/specs/elements/Segment/Segment-test.js
jhchill666: Re-enabled the test and updated to match Segment header logic

src/elements/Segment/SegmentGroup.js
jhchill666: I had noticed how the docs are generated, but wasn't sure how private class documentation should be handled.  In the instance of Segment, which is basically a wrapper class for the logic laden _Segment, non of _Segments documentation has any method of being exposed for the root class Segment.

The re-structuring you suggested above, has solved this in the case, but doesn't for other similarly nested components.    What do you think the best way of remedying this is ?

jhchill666: On another note on your first point, I've pretty much been going on the Stardust docs for figuring out props for each component.  These docs say Segment can be raised, piled, stacked etc.  I guess these examples need relocated under Segments?

levithomason: > On another note on your first point, I've pretty much been going on the Stardust docs for figuring out props for each component. These docs say Segment can be raised, piled, stacked etc. I guess these examples need relocated under Segments?

I was almost completely wrong here.  I doubled checked the SUI docs and the individual Segment does support raised, stacked and piled.  Just not horizontal, that appears to only be for groups of segments.

I would make a habit of referencing the official SUI docs (maybe even over the Stardust docs) when determining component APIs.  This will help ensure we cover all the bases.

> The re-structuring you suggested above, has solved this in the case, but doesn't for other similarly nested components. What do you think the best way of remedying this is ?

We're able to teach react-docgen how to resolve doc blocks by reference.  This would allow us to infer docs by component reference.  If possible, I'd like to keep it as straight forward as possible and put the documentation on the component implementing the private component API.

docs/app/Examples/elements/Segment/Groups/SegmentHorizontalSegmentsExample.js
jhchill666: Should 'vertical' be supported here as well, or inferred as the default, 
by the omission of 'horizontal' ?

> Levi Thomason mailto:notifications@github.com
> 26 June 2016 at 22:39
> 
> In 
> docs/app/Examples/elements/Segment/Groups/SegmentHorizontalSegmentsExample.js 
> https://github.com/TechnologyAdvice/stardust/pull/287#discussion_r68510275:
> 
> > @@ -4,11 +4,11 @@ import { Segment } from 'stardust'
> > export default class SegmentHorizontalSegmentsExample extends 
> > Component {
> > render() {
> > return (
> > - <Segment.Segments className='horizontal'>
> > - <Segment.Group aligned='horizontal'>
> 
> Since the |horizontal| 
> http://semantic-ui.com/elements/segment.html#horizontal-segments 
> class here applies to the layout of the segments and there is only one 
> option, I think we should use a boolean prop:
> 
> <Segment.Group horizontal />
> 
> image 
> https://cloud.githubusercontent.com/assets/5067638/16365125/b53c0eae-3bab-11e6-8a78-00810e7c74bb.png
> 
> Please let me know if I'm over looking anything.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub 
> https://github.com/TechnologyAdvice/stardust/pull/287/files/a628afbe7bc8b86e4cc139252db0d32941a9e279#r68510275, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AAWHsy5NzDjxjcY32cSo1-PmIc6bqqo6ks5qPvGogaJpZM4I-lPR.

## 

Sent from Postbox 
https://www.postbox-inc.com/?utm_source=email&utm_medium=siglink&utm_campaign=reach

jhchill666: Amended to horizontal:bool, disregard comment asking whether `vertical` would be supported.

docs/app/Examples/elements/Segment/Types/SegmentVerticalSegmentExample.js
jhchill666: Done

src/elements/Segment/_Segment.js
jhchill666: Re-implemented after initially removing, as per

> I was almost completely wrong here. I doubled checked the SUI docs and the individual Segment does support raised, stacked and piled. Just not horizontal, that appears to only be for groups of segments.

src/elements/Segment/_Segment.js
jhchill666: Implemented

src/elements/Header/HeaderH1.js
```diff
@@ -15,4 +15,12 @@ HeaderH1._meta = {
   type: META.type.element,
 }
 
+HeaderH1.propTypes = {
+  _sdClass: PropTypes.string,
+}
+
+HeaderH1.defaultProps = {
+  _sdClass: 'sd-header-h1',
```
levithomason: Heads up, I've finished removing all `sd-*` classes, see #301.  Rebase to master to get the update.  The test for the `sd-*` class has been inverted to fail if any `sd-` class exists.

levithomason:
Heads up, I'm resuming work on this PR along side the Button PR.

levithomason:
I've rebased this to master, squashed to 2 commits, finished features, and added/fixed tests.

The last thing to do is round out the examples.  There are a few missing.  Also, something is odd with the circular example.

levithomason:
Released in `stardust@0.41.0`, thanks for all the help!

PR #295, opened 2016-06-27
feat(Button): support all features, update to v1 API

### LINKED ISSUE #43: Add toggle button functionality
Let's add the functionality for the toggle button. Here's the toggle button on the semantic docs - http://semantic-ui.com/elements/button.html#toggle

### LINKED ISSUE #46: Create attached button element
We want to be able to use a Button element attached a div. Currently, when set as a Button it doesn't adopt the styling of the attached div. See PR #45

jhchill666:
Fixes #43 
Fixes #46


src/elements/Button/ButtonGroup.js
```diff
@@ -0,0 +1,79 @@
...
+  icon: PropTypes.bool,
+
+  /** A button can have different sizes */
+  size: PropTypes.oneOf(ButtonGroup._meta.props.size),
+
+  /** Groups can be formatted to appear vertically */
+  vertical: PropTypes.bool,
+
+  /** Groups can have their widths divided evenly */
+  widths: PropTypes.oneOf(ButtonGroup._meta.props.widths),
```
jhchill666, line 92: Wasn't sure what this prop should be named???

levithomason, line 92: I think the real intention here is to have buttons be equal width.  In the FormFields we have `evenlyDivided` which adds the number-word class to the group (ie `five fields`) by counting the field children.  [Evenly divided](http://semantic-ui.com/collections/form.html#evenly-divided) is the name of that section in the SUI docs.

I could see going with `equalWidth` here and counting the button children.  It is also consistent with the grid class `equal width` which scales column widths.

In order to not tie the user's hands should they want odd sized button groups, we could offer both `equalWidth` and `widths`.  Or, just use something like `widths='equal'`.

levithomason, line 92: It is worth mentioning here that we also want to do away with the word number classes in this library.  Or at least offer a side by side alternative.  You can see this in the Grid Column where we use `width={3}` instead of `width='three'`.

We could do this easily by extending the `sui.widths` to include the numbers.  Then, extend the `numberToWord` util to include a map of words to words.  Finally, always use `utils/numberToWord.js` on the `width` prop values.

This way, users can use either numbers or words to define widths wherever they appear.

docs/app/Examples/elements/Button/Groups/ButtonColoredButtonsExample.js
```diff
@@ -0,0 +1,20 @@
...
+      <Button.Group color='blue'>
+        <Button>
+          One
+        </Button>
+        <Button>
+          Two
+        </Button>
+        <Button>
+          Three
+        </Button>
```
levithomason: Can we collapse these solely to save doc site example space?

``` js
<Button>One</Button>
<Button>Two</Button>
<Button>Three</Button>
```

jhchill666: Sure, we're already expanded I previous examples and conscious about changing you guys stuff!

docs/app/Examples/elements/Button/Types/ButtonLabeledIconExample.js
```diff
@@ -1,16 +1,14 @@
...
     return (
       <div>
-        <Button className='labeled icon'>
-          <Icon className='pause' />
+        <Button labeled='icon' icon='pause'>
           Pause
         </Button>
-        <Button className='right labeled icon'>
-          <Icon className='right arrow' />
+        <Button labeled='right icon' icon='right arrow'>
```
levithomason: Hm, I recall some left/right classes in SUI are order dependent.  I decided to test the classes generated by `labeled='right icon'` and indeed, this will not work :(  The classes generated will be `right icon labeled`.  The combination `right labeled` no longer appears.  Here's what happens:

---



---

Because of this, I propose we only allow `labeled` and `labeled='right'` use of the labeled prop.  This describes the label itself.  The icon then is a separate matter, I'll comment on that elsewhere so we have two threads to discuss.

docs/app/Examples/elements/Button/Groups/ButtonColoredButtonsExample.js
levithomason: At this point, I'd like to see more of your ideas as they have been great.  😉

docs/app/Examples/elements/Button/Variations/ButtonSocialExample.js
```diff
@@ -1,36 +1,29 @@
...
-import { Button, Icon } from 'stardust'
+import { Button } from 'stardust'
 
 export default class ButtonSocialExample extends Component {
   render() {
     return (
       <div>
-        <Button className='facebook'>
-          <Icon className='facebook' />
+        <Button className='facebook' icon='facebook'>
```
levithomason: I almost want to put these button social color classes into the color options:

``` js
<Button color='facebook' />
```

Thoughts?

levithomason:
As part of this PR let's take on:
- [x] #43 Toggle functionality
- [x] #46 Nested attached button issue

src/elements/Button/Button.js
```diff
@@ -1,34 +1,136 @@
...
+ * Button
+ */
+function Button(props) {
+  const {
+    children, color, size, className, animated, floated, labeled, icon, inverted, fluid, basic, active, disabled,
+    loading, circular, compact, attached, toggle, type,
+  } = props
+
+  const classes = cx('sd-button', 'ui',
+    icon && 'icon',
```
levithomason: We have a tricky bit here to deal with.  The presence of an icon will not always necessitate the icon class.  The icon class collapses padding to work with individual icon buttons:



I'm not sure of a good solution here yet.  I'll go through this the examples and your suggestion about `Button.Left`, etc in more detail later.

src/elements/Button/ButtonGroup.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Button.Group
+ */
+function ButtonGroup(props) {
+  const {
+    basic, className, children, color, icon, size, vertical, widths,
+  } = props
+
+  const classes = cx('sd-button-group', 'ui',
```
levithomason: Heads up, I've finished removing all `sd-*` classes, see #301.  Rebase to master to get the update.  The test for the `sd-*` class has been inverted to fail if any `sd-` class exists.

docs/app/Examples/collections/Form/Validation/FormValidationExamples.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react/jsx-curly-spacing */
```
levithomason: This rule has been turned off entirely for the project.  I do not think it works correctly with the spread operators.  You should be able to remove this after rebasing to master.

levithomason:
It's pretty late, I hope this isn't full of errors!

---

# Button API Proposal

I was taking a look at [Material-UI Buttons](http://www.material-ui.com/#/components/flat-button) and I noticed they separated `FlatButton`, `IconButton`, `RaisedButton`, and `FloatingActionButton`.

This got me thinking if we could benefit from splitting up `AnimatedButton`, `LabeledButton`, `IconButton`, and `LabeledIconButton` components.  Each of these varies pretty wildly in markup.

It is worth looking at. After writing these examples out and grokking SUI "labeled" as it related to Labels and Icons, I still come back to the approach of a `labeled='left'` prop.  I think this makes the most sense now.  Let me know what you think.

## Animated

The `animated` prop feels OK to me here.  The `.Content` sub component follows the pattern of other components.  They just feel weird together here.  Not so sure on this one.

image

``` jsx
<Button animated>
  <Button.Content visible>Next</Button.Content>
  <Button.Content hidden>
    <Icon name='right arrow' />
  </Button.Content>
</Button>
<Button animated='vertical'>
  <Button.Content hidden>Shop</Button.Content>
  <Button.Content visible>
    <Icon name='shop' />
  </Button.Content>
</Button>
<Button animated='fade'>
  <Button.Content visible>Sign-up for a Pro account</Button.Content>
  <Button.Content hidden>
    $12.99 a month
  </Button.Content>
</Button>
```

``` html
<div class="ui animated button" tabindex="0">
  <div class="visible content">Next</div>
  <div class="hidden content">
    <i class="right arrow icon"></i>
  </div>
</div>
<div class="ui vertical animated button" tabindex="0">
  <div class="hidden content">Shop</div>
  <div class="visible content">
    <i class="shop icon"></i>
  </div>
</div>
<div class="ui animated fade button" tabindex="0">
  <div class="visible content">Sign-up for a Pro account</div>
  <div class="hidden content">
    $12.99 a month
  </div>
</div>
```

## Labeled

> In SUI, the "labeled" part of the button is either a Label or an Icon, never both.  If both are present, the Label is put in the labeled position and the Icon is left inside the button.

Since the labels inside a button can easily be complex, I think it makes sense to allow the Label as a child.  We can also take in a `label` prop for `basic` label text like we do with `icon`.

image

``` jsx
<Button labeled icon='heart'>
  <Label basic>2,048</Label>
  Like
</Button>
<Button labeled='left' icon='heart'>
  <Label basic pointing='right'>2,048</Label>
  Like
</Button>
// basic labeled buttons could be achieved with a `label` prop
<Button labeled='left' icon='fork' label='1048 />
```

``` html
<div class="ui labeled button" tabindex="0">
  <div class="ui button">
    <i class="heart icon"></i> Like
  </div>
  <a class="ui basic label">
    2,048
  </a>
</div>
<div class="ui left labeled button" tabindex="0">
  <a class="ui basic right pointing label">
    2,048
  </a>
  <div class="ui button">
    <i class="heart icon"></i> Like
  </div>
</div>
<div class="ui left labeled button" tabindex="0">
  <a class="ui basic label">
    1,048
  </a>
  <div class="ui icon button">
    <i class="fork icon"></i>
  </div>
</div>
```

image

``` jsx
<Button labeled color='red' icon='heart'>
  <Label basic color='red' pointing='left'>1,048</Label>
  Like
</Button>
<Button labeled basic color='blue' icon='fork'>
  <Label basic color='blue' pointing='left'>1,048</Label>
  Forks
</Button>
```

``` html
<div class="ui labeled button" tabindex="0">
  <div class="ui red button">
    <i class="heart icon"></i> Like
  </div>
  <a class="ui basic red left pointing label">
    1,048
  </a>
</div>
<div class="ui labeled button" tabindex="0">
  <div class="ui basic blue button">
    <i class="fork icon"></i> Forks
  </div>
  <a class="ui basic left pointing blue label">
    1,048
  </a>
</div>
```

## Icon

An Icon button is a button that _only_ has an icon.  This is what adds the `icon` class.  Otherwise, it is just a button with an Icon inside of it, which may also have text beside it.

image

``` jsx
<Button icon='cloud' />
// or
<Button>
  <Icon name='cloud' />
</Button>
```

``` html
<button class="ui icon button">
  <i class="cloud icon"></i>
</button>
```

## Labeled Icon

Per SUI, the `labeled` prop applies to the Icon if there is no Label also.  Buttons with `labeled` and `icon` props receive an `icon` class as well.

image

``` jsx
<Button labeled='left' icon='pause'>
  Pause
</Button>
<Button labeled='right' icon='pause'>
  Pause
</Button>
```

``` html
<button class="ui labeled icon button">
  <i class="pause icon"></i>
  Pause
</button>
<button class="ui right labeled icon button">
  <i class="right arrow icon"></i>
  Next
</button>
```

## Basic

I throw this one in just to show that a Button with text and an Icon does not get an `icon` class.  The `icon` class is assigned if it has _only_ an Icon, or has an Icon _and_ is `labeled`.

image

``` jsx
<Button basic icon='user'>
  Add Friend
</Button>
// or
<Button basic>
  <Icon name='user' />
  Add Friend
</Button>
```

``` html
<button class="ui basic button">
  <i class="icon user"></i>
  Add Friend
</button>
```

docs/app/Examples/elements/Button/Content/ButtonConditionalsExample.js
```diff
@@ -1,14 +1,14 @@
...
+import { Button } from 'stardust'
 
 export default class ButtonConditionalsExample extends Component {
   render() {
     return (
-      <Buttons>
+      <Button.Group>
         <Button>Cancel</Button>
         <div className='or' />
         <Button className='positive'>Save</Button>
```
levithomason: We should pull `positive` into a prop as well.

docs/app/Examples/elements/Button/States/ButtonLoadingExample.js
```diff
@@ -5,10 +5,10 @@ export default class ButtonLoadingExample extends Component {
...
     return (
       <div>
-        <Button className='loading' />
-        <Button className='basic loading' />
-        <Button className='primary loading' />
-        <Button className='secondary loading' />
+        <Button loading />
+        <Button basic loading />
+        <Button loading className='primary' />
+        <Button loading className='secondary' />
```
levithomason: Regarding `primary` and `secondary`.  My first thought was to have these as boolean props, however, we could also do a value only prop:

``` js
<Button primary />
<Button emphasis='primary' />
```

I like the first for brevity, however, there is also `positive` and `negative` which we could put under emphasis.  Thoughts?

jhchill666:
I'm going to have to go for the first option, just feels cleaner and more succinct.  I'm sure I recall seeing other primary/secondary props on another component(?) is this correct? If so how would the same apply to it?

levithomason:
> I'm sure I recall seeing other primary/secondary props on another component(?) is this correct? If so how would the same apply to it?

Yep, the Segment for example also has [`emphasis`](http://semantic-ui.com/elements/segment.html#emphasis).  Though, in that case the options are `secondary` and `tertiary`.

Even though these are mutually exclusive attributes, you cannot have both `secondary` and `tertiary` nor `primary` and `secondary`, I think I'm with you in choosing the first option.  It just doesn't seem like `emphasis` makes intuitive sense as an enum prop.

src/elements/Button/Button.js
```diff
@@ -1,33 +1,136 @@
...
+    useKeyOnly(loading, 'loading'),
+    useKeyOnly(toggle, 'toggle'),
+    'button',
+    className
+  )
+
+  const rest = getUnhandledProps(Button, props)
+
+  return (
+    <button type={type} className={classes} {...rest}>
```
levithomason: When the `attached` prop is used, the button component needs to be a `div`.  This fixes #46.  The CSS does not work on a `button` element for attached buttons.

``` html
<div class="ui top attached button" tabindex="0">Top</div>
<div class="ui attached segment">
  <p></p>
</div>
<div class="ui bottom attached button" tabindex="0">Bottom</div>
```

Also, the `tabIndex={0}` prop will need to be added so users can still tab to the button `div`.

src/elements/Button/Button.js
```diff
@@ -1,33 +1,136 @@
...
+  /**  */
+  type: PropTypes.string,
+
+  /** A button can have different sizes */
+  size: PropTypes.oneOf(Button._meta.props.size),
 }
+
+Button.defaultProps = {
+  type: 'button',
+}
```
levithomason, line 254: Let's pull the type prop and it's default value.  The user can set this if they wish.

src/elements/Button/Button.js
jhchill666: Now uses `div` tag when `!!attached && type != 'submit'` as per [#46](https://github.com/TechnologyAdvice/stardust/issues/46)

jhchill666: `tabIndex=0` added when tag type is `div`

src/elements/Button/ButtonGroup.js
levithomason, line 92: See here for the width API decision https://github.com/TechnologyAdvice/stardust/issues/183#issuecomment-232728080

levithomason:
@jamiehill hope all is well.  Did you plan on picking this PRs back up or is it OK if I resume work on them?

jhchill666:
Hey Levi. Absolutely, been on hols such not much productivity of late. If you're keen to get going straight away, you take it but I'll be back on board in next few days

levithomason:
Thanks for the quick reply.  I'll work on other things in the interim, if I run out I'll let you know before stepping on your work.

levithomason:
Per our emails, I've rebased and updated this PR to the latest master.  There are some failing tests to fix with the Button prop to className.

Note all component utils are now imported from one place, `../../lib`.  One of those utils is `getElementType`.  It is similar to `getUnhandledProps`, except instead of returning the rest of the props, it returns the Element type the component should render as.

There is one feature needed for this new util.  Currently, it accepts a hash map of props to element types.  Example, `{ attached: 'div' }` will render as a div when the attached prop is present.  Though, the button requires the ability to render a particular element type based on the the _value_ of the `type` prop, not simply the presence of it.

I've added a todo in Button.js noting this.  This PR should extend the `getElementType` lib function to be able to select an element type based on the value of a prop, not just the presence of the prop only.

LMK if you have any issues/questions otherwise.  Cheers!

levithomason:
@jamiehill ^

levithomason:
Heads up, I'm resuming work on this PR.

levithomason:
I've finished the labeled button API. Just need to add tests now.

levithomason:
Woop!  I think she's ready.  Just got done hashing the remaining APIs with our team and making updates.  Hope to merge when tests pass.

PR #301, opened 2016-06-28
Remove sd-* classNames

### LINKED ISSUE #293: Remove sd-* classes
The `sd-*` classes are for `stardust-*`.  These will no longer be needed since the library is now the official React SUI library.  We should remove these classes and tests.  There are common tests for these as well.

levithomason:
Fixes #293.  This PR removes all `sd-*` classNames.  This was done as we are now officially a SUI project.  There will be no more "Stardust" once migrated.

The test ensuring the `sd-*` class exists has been inverted to ensure it does not exist.  This is only until we are done merging other PRs which are still using the `sd-*` class.

### DIFF
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -25,11 +25,6 @@ const componentInfo = componentCtx.keys().map(key => {
   const subComponentName = _.has(_meta, 'parent') && _.has(_meta, 'name')
     ? _meta.name.replace(_meta.parent, '')
     : null
-  // HeaderH1 => sd-header-h1
-  const sdClass = `sd-${constructorName
-    .replace('_', '')                 // remove underscore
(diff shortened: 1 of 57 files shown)


test/specs/commonTests.js
```diff
@@ -239,9 +232,19 @@ export const isConformant = (Component, requiredProps = {}) => {
...
+      const children = wrapper.children()
+
+      // component itself
+      if (className) className.should.not.contain('sd-')
+
+      // children
+      children.forEach(c => {
+        const childClassName = c.prop('className')
+        if (childClassName) childClassName.should.not.contain('sd-')
+      })
```
levithomason, line 247: I've inverted this test for now to help catch any further sd-\* classes from other PRs that have not yet been scrubbed.  After a while, we can remove this test.

PR #310, opened 2016-06-28
Fix dropdown set value and error

levithomason:


src/modules/Dropdown/Dropdown.js
```diff
@@ -187,14 +187,15 @@ export default class Dropdown extends Component {
     if (process.env.NODE_ENV !== 'production') {
       // in development, validate value type matches dropdown type
       const isNextValueArray = Array.isArray(nextProps.value)
+      const hasValue = _.has(nextProps, 'value')
 
       /* eslint-disable no-console */
-      if (nextProps.multiple && !isNextValueArray) {
+      if (hasValue && nextProps.multiple && !isNextValueArray) {
```
levithomason, line 193: We should not throw these dev errors regarding value unless there is actually a value.

dvdzkwsk:
👻

src/modules/Dropdown/Dropdown.js
```diff
@@ -328,13 +329,15 @@ export default class Dropdown extends Component {
...
 
-    this.setValue(value)
 
     // notify the onChange prop that the user is trying to change value
     if (multiple) {
       // state value may be undefined
-      this.onChange(e, _.union(this.state.value, [value]))
+      const newValue = _.union(this.state.value, [value])
+      this.setValue(newValue)
+      this.onChange(e, newValue)
```
levithomason, line 338: This should arguably be handled by trySetState, but we're handling in the component for now.

PR #312, opened 2016-06-28
Fix getComponentDocInfo docPath on windows

### LINKED ISSUE #304: File docgenInfo.json missing from 'docs/app'
I assume this file is generated; however, when running the initial 'npm start' a breaking error occurs where webpack notes that docgenInfo.json is missing. This file is referenced by [doc/app/utils/getComponentDocInfo.js](https://github.com/TechnologyAdvice/stardust/blob/master/docs/app/utils/getComponentDocInfo.js) `import docgenInfo from '../docgenInfo.json'`. Perhaps this file aught to be included in the repo perhaps with some default content?

levithomason:
Fixes #304.  This PR removes the `/` in the path check logic.  On Windows this is a `\`.

### DIFF
--- a/docs/app/utils/getComponentDocInfo.js
+++ b/docs/app/utils/getComponentDocInfo.js
@@ -2,13 +2,15 @@ import _ from 'lodash'
 import docgenInfo from '../docgenInfo.json'
 
 /**
- * This util extracts and formats a doc object from docgenInfo.json for a single `component` and merges it with
- * the component's meta information.
+ * This util extracts and formats a doc object from docgenInfo.json for a single `component`.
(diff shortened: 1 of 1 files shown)


dvdzkwsk:
👻  -- will be curious to see what this resolves in regard to the original issue.

levithomason:
If not, this would have also been an issue.  Then this is at least one of the problems.  His docgenInfo.json keys look like this:

``` json
"C:\\Users\\Joshua\\Dropbox\\GitHub\\stardust\\src\\addons\\Confirm\\Confirm.js": {}
```

There would never be a match on that ^ using the `/` as you noted.

PR #316, opened 2016-06-29
Supporting env var assignments on Windows

joshuazmiller:
Trying to resolve #313

PR #320, opened 2016-07-01
Drop babel-node from build process

levithomason:
This PR dials back the build and config syntax to work with vanilla node 6.  This means no more `babel-node` in order to run Karma.

This is good in general because it reduces complexity and deps.  The main reason it was done though was in attempt for Windows compatibility.  See #316, where we were having issues running karma by referencing the npm bin path.  On windows, this needs to be the Windows version which I believe is `karma.cmd`.

Having removed babel-node means we also remove the need to reference the path directly.  Windows _should_ be able to execute the test script now and NPM _should_ do its cross platform magic and run the correct version.

We shall see.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,14 @@ CONTRIBUTING
 
 ## Getting Started
 
+Make sure you have at least [Node.js v6][11]:
+
+```sh
+node -v
+
+v6.2.1
+```
+
 ### Clone & Install
 
 Start by cloning this repo and installing dependencies:
@@ -510,3 +518,4 @@ Adding documentation for new components is a bit tedious.  The best way to do th
 [8]: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit
 [9]: http://semantic-ui.com/introduction/glossary.html
 [10]: http://semantic-ui.com/elements/label.html
+[11]: https://nodejs.org/
--- a/build/karma.conf.babel.js
+++ b/build/karma.conf.babel.js
@@ -1,7 +1,6 @@
-import { argv } from 'yargs'
-
-import config from '../config'
-import webpackConfig from './webpack.config'
+const { argv } = require('yargs')
+const config = require('../config')
+const webpackConfig = require('./webpack.config')
 
 const { paths } = config
 
@@ -42,8 +41,7 @@ module.exports = (karmaConfig) => {
     },
     webpack: {
       devtool: 'cheap-module-source-map',
-      module: {
-        ...webpackConfig.module,
+      module: Object.assign({}, webpackConfig.module, {
         loaders: [
           {
             test: /sinon\.js$/,
@@ -51,14 +49,10 @@ module.exports = (karmaConfig) => {
           },
           ...webpackConfig.module.loaders,
         ],
-      },
-      plugins: [
-        ...webpackConfig.plugins,
-      ],
-      resolve: {
-        ...webpackConfig.resolve,
-        alias: {
-          ...webpackConfig.resolve.alias,
+      }),
+      plugins: webpackConfig.plugins,
+      resolve: Object.assign({}, webpackConfig.resolve, {
+        alias: Object.assign({}, webpackConfig.resolve.alias, {
           jquery: `${paths.test('mocks')}/SemanticjQuery-mock.js`,
           sinon: 'sinon/pkg/sinon',
           // These are internal deps specific to React 0.13 required() by enzyme
@@ -69,8 +63,8 @@ module.exports = (karmaConfig) => {
           // this is a React 0.13 dep required by enzyme
           // ignore it since we don't have it
           'react/addons': 'empty/object',
-        },
-      },
+        }),
+      }),
     },
     webpackServer: {
       progress: false,
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -1,9 +1,8 @@
-import config from '../config'
-import webpack from 'webpack'
-import HtmlWebpackPlugin from 'html-webpack-plugin'
-import _ from 'lodash'
-import yargs from 'yargs'
-const { argv } = yargs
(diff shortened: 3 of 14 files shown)


CONTRIBUTING.md
```diff
@@ -40,6 +40,14 @@ CONTRIBUTING
 
 ## Getting Started
 
+Make sure you have at least [Node.js v6][11]:
+
+```sh
+node -v
+
+v6.2.1
```
levithomason, line 48: Added a note to contributing to use Node 6 since babel-node is no longer transpiling build and config code.

build/webpack.config.js
```diff
@@ -27,7 +27,7 @@ const webpackConfig = {
 
 const webpackHotPath = `${config.compiler_public_path}__webpack_hmr`
 
-export const webpackHotMiddlewareEntry = `webpack-hot-middleware/client?${_.map({
+const webpackHotMiddlewareEntry = `webpack-hot-middleware/client?${_.map({
```
levithomason, line 29: This export was never used.  I searched the code base as well as tested all the build scripts.  All is well.

build/karma.conf.babel.js
```diff
@@ -1,9 +1,9 @@
-import { argv } from 'yargs'
-
-import config from '../config'
-import webpackConfig from './webpack.config'
+const yargs = require('yargs')
+const config = require('../config')
+const webpackConfig = require('./webpack.config')
 
 const { paths } = config
+const { argv } = yargs
```
eanplatter: Does the style guide say anything about just doing `const argv = require('yargs').argv`, actually that's silly, I am pretty sure you can do this:

``` js
const { argv } = require('yargs')
```

Yeah I am pretty sure that's doable.

dvdzkwsk: Get out of here Platter.

<3

levithomason: Since I'm never using `yargs` in the scope, this makes sense.  Update coming.

build/karma.conf.babel.js
```diff
@@ -42,23 +42,19 @@ module.exports = (karmaConfig) => {
     },
     webpack: {
       devtool: 'cheap-module-source-map',
-      module: {
-        ...webpackConfig.module,
+      module: Object.assign({}, webpackConfig.module, {
```
eanplatter, line 44: Would it make sense to use a spread operator here? I _believe_ they're in node6.

build/webpack.config.js
```diff
@@ -1,8 +1,8 @@
...
-import webpack from 'webpack'
-import HtmlWebpackPlugin from 'html-webpack-plugin'
-import _ from 'lodash'
-import yargs from 'yargs'
+const config = require('../config')
+const webpack = require('webpack')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const _ = require('lodash')
+const yargs = require('yargs')
 const { argv } = yargs
```
eanplatter: same paths and yargs comment as before

build/karma.conf.babel.js
levithomason, line 44: Man, I could have sworn it was though I'm on `6.2.1` and Node swears otherwise.  Could be I need a flag or `use strict` or some other something enabled.

build/webpack.config.js
```diff
@@ -36,19 +36,18 @@ export const webpackHotMiddlewareEntry = `webpack-hot-middleware/client?${_.map(
...
-    ...APP_ENTRY,
+    APP_ENTRY,
   ],
   vendor: [
     webpackHotMiddlewareEntry,
-    ...config.compiler_vendor,
-  ],
+  ].concat(
+    config.compiler_vendor
+  ),
```
eanplatter: I'm confused, why isn't this just:

``` js
vendor: [
  webpackHotMiddlewareEntry,
  config.compiler_vendor,
]
```

build/karma.conf.babel.js
eanplatter, line 44: https://kangax.github.io/compat-table/es6/

build/webpack.dll.js
```diff
@@ -1,6 +1,6 @@
-import webpack from 'webpack'
+const webpack = require('webpack')
 
-import config from '../config'
+const config = require('../config')
 const webpackDllConfig = { module: {} }
 
 const { paths } = config
```
eanplatter, line 6: this too can be done in the import

eanplatter:
Just some syntax comments, other than that LGTM! 🍂

build/webpack.config.js
levithomason: `webpackHotMiddlewareEntry` is a string and `config.compiler_vendor` is an array.

build/karma.conf.babel.js
levithomason, line 44: I know, but riddle me this:

**spread.js**

``` js
const objA = {
  foo: 'bar'
}

const objB = { ...objA }
```

``` sh
$ node -v
v6.2.1

$ node spread.js
/Users/levithomason/src/stardust/spread.js:5
const objB = { ...objA }
               ^^^
SyntaxError: Unexpected token ...
    at Object.exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:513:28)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Function.Module.runMain (module.js:575:10)
    at startup (node.js:160:18)
    at node.js:456:3
```

There must be some flag or some other config required.  I just don't want to risk it not running for anyone when `Object.assign()` is available.

levithomason, line 44: OK, it works with arrays.  Also seems to be with rest params, but **not** with objects....

image

source: http://node.green/

levithomason, line 44: Updated all arrays to use it and it's good.  Thanks for the push on it!

PR #279, opened 2016-06-24
Icon Component updated to v1 API (bad merge)

jhchill666:

PR #321, opened 2016-07-02
Add Breadcrumb component

### LINKED ISSUE #184: <Breadcrumb /> Component
http://semantic-ui.com/collections/breadcrumb.html

layershifter:
# Under Construction

Fixes #184


levithomason:
Thanks for the PR, it's looking great so far! It looks like you've probably already read the contributing guidelines. Looking forward to the API spec for this component. 

The [CONTRIBUTING.md](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md) guide details how to go about designing a component API.  The [Spec out the API](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#spec-out-the-api) section is of primary importance. You can also reference other PRs for example specs, like #281. 

Cheers!

layershifter:
# Breadcrumb API Proposal

## Breadcrumb

Regarding Semantic UI docs, Breadcrumb can have only variation, so I think there will no be  
exceptions about following syntax.

``` js
<Breadcrumb>...</Breadcrumb>
<Breadcrumb size='massive'>...</Breadcrumb>
```

---

Each breadcrumb contains sections and dividers.

## Section

There is a trouble with section, it can be text or link, so I propose to add `isLink` property (I don't found anything like this in other components).

``` js
<Breadcrumb.Section>Home</Breadcrumb.Section>
<Breadcrumb.Section active>Home</Breadcrumb.Section>
<Breadcrumb.Section isLink>Home</Breadcrumb.Section>
<Breadcrumb.Section active isLink>Home</Breadcrumb.Section>
```

## Divider

Breadcrumb component also has divider, but it's not `ui divider`, so it might be subcomponent.

``` html
<span class="divider">/</span>
<i class="right chevron icon divider"></i>
```

I think of such an implementation:

``` js
<Breadcrumb.Divider>/</Breadcrumb.Divider>
<Breadcrumb.Divider icon='right chevron'>/</Breadcrumb.Divider>
```

# All together

``` js
<Breadcrumb>
  <Breadcrumb.Section isLink>Home</Breadcrumb.Section>
  <Breadcrumb.Divider icon='right chevron'>/</Breadcrumb.Divider>
  <Breadcrumb.Section active>Catalog</Breadcrumb.Section>
</Breadcrumb>
```

Feedback welcome :smile:

levithomason:
I think you're right about the sub components and props here.  Here are my thoughts.

## Section

Since props without a value default to `true`, let's not prefix `isLink` with `is*`.  We'll use `link` instead.  The [Label](https://github.com/TechnologyAdvice/stardust/blob/master/src/elements/Label/Label.js) component has a `link` prop which changes the tag to an `a` tag.  It also changes to an `a` tag if you provide an `onClick` prop, this way you get the pointer cursor if you have a click handler.

I think we can do a similar thing here with the Breadcrumb.  I see 3 ways of formatting the Section as a link:
1. add a `link` prop
2. add an `onClick` prop
3. add an `href` prop

If any of these three are present, I think we should use an `a` tag:

``` jsx
<Breadcrumb.Section link>Home</Breadcrumb.Section>
<Breadcrumb.Section onClick={this.handleClick}>Home</Breadcrumb.Section>
<Breadcrumb.Section href='google.com'>Home</Breadcrumb.Section>
```

``` html
<a class="section">Home</a>
<a class="section">Home</a>
<a class="section" href="google.com">Home</a>
```

## Divider

The default divider is `/`.  When an icon is used, there is no `/` child.  I believe we should default the child to `/` when there is no `icon` prop given.

``` jsx
<Breadcrumb.Divider />
// and
<Breadcrumb.Divider icon='right chevron' />
```

``` html
<div class="divider"> / </div>
<i class="right chevron icon divider"></i>
```

> To ensure the user doesn't put both an `icon` prop _and_ children, we have a `customPropTypes` in `propUtils` called `mutuallyExclusive` to prevent the user from defining two conflicting props.

### Icon

~~We may also extend Icon to allow a `divider` prop:~~  The more I thought about this the less it made sense. This prop wouldn't be used anywhere else that I can see. It isn't an attribute of all icons so I no longer think we should add this. 

``` jsx
<Icon name='right chevron' divider />
```

``` html
<i class="right chevron icon divider"></i>
```

See #279 for that component update. I'll /cc @jamiehill for this addition.

## Using Props

We've also been allowing components to define basic markup through props.  We would make `children` `mutuallyExclusive` with each of these props since these props will generate the `children` for you.

I can see allowing something like this:

``` jsx
const sections = []

<Breadcrumb sections={sections} />
```

### Regular

``` jsx
const sections = [
  { text: 'Home' },
  { text: 'Store' },
  { text: 'T-Shirt', active: true }
]

<Breadcrumb sections={sections} />
```

``` html
 <div class="ui breadcrumb">
   <a class="section">Home</a>
   <div class="divider"> / </div>
   <a class="section">Store</a>
   <div class="divider"> / </div>
   <div class="active section">T-Shirt</div>
   </div>
```

### Icons

We could use a `divider` prop to define the divider type.  This doesn't allow mixing divider types.  Though, props are meant to be simple.  The sub components are available if the user needs more control.

``` jsx
const sections = [
  { text: 'Home' },
  { text: 'Store' },
  { text: 'T-Shirt', active: true }
]

<Breadcrumb sections={sections} divider='right angle' />
```

``` html
 <div class="ui breadcrumb">
  <a class="section">Home</a>
  <i class="right angle icon divider"></i>
  <a class="section">Store</a>
  <i class="right angle icon divider"></i>
  <div class="active section">T-Shirt</div>
</div>
```

### Links

Just as with the Breadcrumb.Section, you could provide a `link`, `onClick`, or `href` key to the section object to create a link style:

``` jsx
const sections = [
  { text: 'Home', link: true },
  { text: 'Store', onClick: handleClick },
  { text: 'T-Shirt', href: 'google.com' }
]

<Breadcrumb sections={sections} />
```

``` html
 <div class="ui breadcrumb">
  <a class="section">Home</a>
  <i class="divider"></i>
  <a class="section">Store</a>
  <i class="divider"></i>
  <div class="active" href='google.com'>T-Shirt</div>
</div>
```

---

Let me know what you think!

layershifter:
## Section

I fully agree with your thoughts :+1: 

## Divider

I agree with default value, but it may to be set custom:

``` html
<div class="divider">/</div>
<div class="divider">></div>
```

``` js
<Breadcrumb.Divider />
<Breadcrumb.Divider delimiter='>' />
<Breadcrumb.Divider symbol='>' />
<Breadcrumb.Divider value='>' />
```

May be one of variants?

### Icon

I think better will be extend `Icon` component. What about this?

``` js
<Breadcrumb.Icon name='right chevron' />
// or
<Breadcrumb.IconDivider name='right chevron' />
```

## Using props

I think this will be more prefered:

``` js
<Breadcrumb sections={sections} divider='>' />
<Breadcrumb sections={sections} icon-divider='right angle' />
```

levithomason:
> I agree with default value, but it may to be set custom

I like this but think we should use children instead of a delimiter, symbol, or value prop. It will also be more flexible incase someone would like to pass another component as the divider child. 

For the Icon, I like 

``` jsx
<Breadcrumbs icon name='right angle' />
```

This can just be an Icon where we add the "divider" className. 

I'm on mobile right now and it's a holiday here. If you'd like to start with this much, we can iterate on the rest. I'll be offline for the rest of the day. 

Thanks!

levithomason:
Just noticed I copied the wrong JSX in my last comment. I meant to say for the Breadcrumb.Icon I like the idea of re-using the Icon component with this syntax:

``` jsx
<Breadcrumb.Icon name='right angle' />
```

Where this would be exposed like so:

``` js
// BreadcrumbIcon.js

function BreadcrumbIcon(props) { 
  // add the 'divider' class to the Icon
}
```

``` js
// Breadcrumb.js
import BreadcrumbIcon from './BreadcrumbIcon'

function Breadcrumb(props) { /* ... */ }

Breadcrumb.Icon = BreadcrumbIcon
```

jhchill666:
Seems to me inconsistent with other Apis, Button for example, eg <Button icon={icon} /> where icon is an Icon instance or icon class name, no?

levithomason:
This was my original intuition as well. I've not given this one a fair amount of time or thought today. Starting tomorrow, I'll be able to look more at this. Thanks for the input and concern for the APIs.

layershifter:
## Breadcrumb.Section

I added component that implements described functionality. But here is problem with props.

I think that only one prop (`link`, `href`, `onClick`) can be passed to component. I need write custom validator for this? I seen `mutuallyExclusive`, but it doesn't performs type check of prop.

## Breadcrumb.Divider

I think about this, what about this?

``` js
<Breadcrumb.Divider />
<Breadcrumb.Divider>/</Breadcrumb.Divider>
<Breadcrumb.Divider icon='right chevron' />

<Breadcrumb sections={sections} />
<Breadcrumb sections={sections} divider='/' />
<Breadcrumb sections={sections} icon-divider='right angle' />
<Breadcrumb sections={sections} icon divider='right angle' />
```

levithomason:
> I think that only one prop (link, href, onClick) can be passed to component.

I can see link and href being mutually exclusive, but I don't think the onClick should be.  A user may want a click handler along with link/href.

> I seen mutuallyExclusive, but it doesn't performs type check of prop.

This is a short coming of `mutuallyExclusive`, I'm extending it now to take a prop types function as its second parameter so we can use it like so:

``` js
static propTypes = {
  link: customPropTypes.mutuallyExclusive(['href'], PropTypes.bool),
  href: customPropTypes.mutuallyExclusive(['link'], PropTypes.string),
}
```

levithomason:
For the divider, how about this?:

``` jsx
<Breadcrumb.Divider />
<Breadcrumb.Divider>/</Breadcrumb.Divider>
<Breadcrumb.Divider icon='right chevron' />

<Breadcrumb sections={sections} />
<Breadcrumb sections={sections} divider='/' />
<Breadcrumb sections={sections} icon='right angle' />
```

levithomason:
I just merged #323 which adds the `all()` custom prop type.  You can use it do define mutually exclusive and other types:

``` js
panels: customPropTypes.all([
  customPropTypes.mutuallyExclusive(['children']),
    active: PropTypes.bool,
    title: PropTypes.string,
    content: PropTypes.string,
    onClick: PropTypes.func,
  })),
]),
```

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Handles onClick event. If onClick is not defined, does nothing.
+   * */
+  const handleClick = (e) => {
+    if (!_.isUndefined(e)) e.preventDefault()
+    if (!_.isFunction(onClick)) return
+
+    onClick()
+  }
```
levithomason: Few notes on event handlers.
1. Never preventDefault (the user can prevent default if they want)
2. Always pass back the event
3. propTypes validates the onClick as being a function for us, so we don't have to include validation

The refactored `handleClick` would look like this:

``` js
const handleClick = (e) => {
  if (onClick) onClick(e)
}
```

Since now all this function does is pass the event from one function to another, it can be removed  entirely.  We can just spread the `onClick` prop on the returned component.  When called, it will also be called with the event.

There are times that we want to first capture a click handler.  This would be when we want to pass some additional arguments after the event argument.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const classes = cx(
+    active && 'active',
```
levithomason: This is arguably cleaner than the using the util we have for this, and we may do away with this util in the future.  However, for consistency let's use the util here.  We can refactor later if we do decide to remove this simple util:

``` js
import { useKeyOnly } from '../../utils/propUtils'

// ...

const classes = cx(
  useKeyOnly(active, 'active')
)
```

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import Breadcrumb from 'src/collections/Breadcrumb/Breadcrumb'
+import * as common from 'test/specs/commonTests'
+import sandbox from 'test/utils/Sandbox-util'
+
+describe('BreadcrumbSection', () => {
+  common.isConformant(Breadcrumb.Section)
+  common.rendersChildren(Breadcrumb.Section)
+
+  describe('active', () => {
```
levithomason: For props that convert to classNames, we have `propUtils` for extracting the classes and `commonTests` for testing them.  Have a look at CONTRIBUTING.md under the [Prop Utils](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#prop-utils) and the following section [Testing className](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#testing-classname) for how to use them.  Let me know if you have any questions after that.

levithomason:
Before I forget, go ahead and check off Breadcrumb in the README.md under support.

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import Breadcrumb from 'src/collections/Breadcrumb/Breadcrumb'
```
levithomason: Let's import the actual component here so we don't have a dependency on how it is exposed to the user.

``` js
import BreadcrumbSection from 'src/collections/Breadcrumb/BreadcrumbSection'
```

layershifter:
It's okay that following JSX will be generate this HTML?

``` js
<Breadcrumb.Section link>Home</Breadcrumb.Section>
```

``` html
 <a class="section" href="javascript:void(0)">Home</a>
```

May be better solution?

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -6,46 +6,26 @@ import sandbox from 'test/utils/Sandbox-util'
 describe('BreadcrumbSection', () => {
   common.isConformant(Breadcrumb.Section)
   common.rendersChildren(Breadcrumb.Section)
+  common.propValueOnlyToClassName(Breadcrumb.Section, 'active')
```
layershifter: Can you point me? Why test is failed?

layershifter:
> For the divider, how about this?:

I agree, will implemented soon.

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
levithomason: I believe you meant to use key only instead of value only:

``` diff
-common.propValueOnlyToClassName(Breadcrumb.Section, 'active')
+common.propKeyOnlyToClassName(Breadcrumb.Section, 'active')
```

Key only will use the key name (`active`) as the className whereas value only will use the value.  Example, `color='red'` uses value only, since we only the value `red` as the className.

Also, it is helpful to run `npm run test:watch` and use [exclusive tests](https://mochajs.org/#exclusive-tests) to reduce the clutter and identify issues.

Lastly, you can always click `details` on the ci/circle ci check on the PR to see the issue: https://circleci.com/gh/TechnologyAdvice/stardust/1315

levithomason:
> It's okay that following JSX will be generate this HTML?

Let's just leave the href attribute out.  It is [valid HTML 5](https://www.w3.org/TR/html-markup/a.html#placeholder-hyperlink) and considered a placeholder hyperlink.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+ * @see Breadcrumb
+ */
+function BreadcrumbSection(props) {
+  const {
+    active, children, className, link, href, onClick, ...rest,
+  } = props
+
+  /**
+   * Handles onClick event. If onClick is not defined, does nothing.
+   * */
```
levithomason: Reserve doc block comments (`/**`) for docs and public methods/members.  Since this function cannot be used outside of this class, the doc block can be removed.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+    if (onClick) onClick(e)
+  }
+
+  const classes = cx(
+    useKeyOnly(active, 'active'),
+    className,
+    'section',
+  )
+
+  // Handle link and onClick props.
```
levithomason: I think the code is clear enough here, let's remove the comment and examples for link/onClick, href, and the default variant.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
+import META from '../../utils/Meta'
+import { customPropTypes, useKeyOnly } from '../../utils/propUtils'
+
+/**
+ * A section sub-component for Breadcrumb component.
+ *
+ * @see Breadcrumb
```
levithomason: The `@see` tag will show a link to a related component on the doc site:

image

Subcomponents however do not yet show up on the doc site.  Let's remove the `@see` tag here since it will not do anything.

---

Side note, we will likely expose subcomponents in the future, but it will be automated.  We can do this since the _meta property already defines the parent component.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+BreadcrumbSection._meta = {
+  library: META.library.semanticUI,
+  name: 'BreadcrumbSection',
+  type: META.type.collection,
+  parent: 'Breadcrumb',
+}
+
+BreadcrumbSection.propTypes = {
+  /** Optional props that adds class active */
+  active: PropTypes.bool,
```
levithomason, line 56: PropTypes are self documented as optional/required via the `.isRequired` property so the "optional" word is not needed.

We tend to try to use the SUI doc descriptions for features as well.  In this case, the SUI doc description is not too helpful:

image

Borrowing from DropdownItem, I suggest updating this description to:

``` js
/** Style as the currently active section. */
```

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Primary content of the Breadcrumb.Section */
+  children: PropTypes.node,
+
+  /** Additional classes added to the element. */
+  className: PropTypes.string,
+
+  /** Makes element link (<a>) instead of text (<div>) */
```
levithomason: Suggested edit:

``` diff
-Makes element link (<a>) instead of text (<div>)
+Render as an `a` tag instead of a `div`.
```

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+  link: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['href']),
+    PropTypes.bool,
+  ]),
+
+  /** Makes element link (<a>) instead of text (<div>) and adds href attribute. */
```
levithomason: Suggested edit:

``` diff
-Makes element link (<a>) instead of text (<div>) and adds href attribute.
+Render as an `a` tag instead of a `div` and adds the href attribute.
```

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,98 @@
...
+  href: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['link']),
+    PropTypes.string,
+  ]),
+
+  /** Makes element link (<a>) instead of text (<div>) and adds onClick event. */
+  onClick: PropTypes.func,
```
levithomason, line 77: I believe we can do away with this prop entirely.  Since props are spread, it will be spread on the component and still work.

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -0,0 +1,45 @@
...
+import BreadcrumbSection from 'src/collections/Breadcrumb/BreadcrumbSection'
+import * as common from 'test/specs/commonTests'
+import sandbox from 'test/utils/Sandbox-util'
+
+describe('BreadcrumbSection', () => {
+  common.isConformant(BreadcrumbSection)
+  common.rendersChildren(BreadcrumbSection)
+  common.propKeyOnlyToClassName(BreadcrumbSection, 'active')
+
+  it('is div by default and does not have `href` attr', () => {
```
levithomason: Let's split this into two tests:

``` js
it('renders as a div by default', ...)
```

And move the `href` assertion to the `describe('href')` section:

``` js
describe('href prop', () => {
  it('is not present by default', () => {
    // ...
  })
})
```

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -0,0 +1,45 @@
...
+    const section = shallow(<BreadcrumbSection>Home</BreadcrumbSection>)
+
+    section.should.have.tagName('div')
+    section.should.not.have.attr('href')
+  })
+
+  describe('link prop', () => {
+    it('is should be `a` when has prop link', () => {
```
levithomason, line 17: For link prop, href prop, and onClick prop tests `is should` => `should`.

src/collections/Breadcrumb/BreadcrumbSection.js
layershifter, line 77: Can you tell us more? If I remove `onClick` from propTypes, eslint fails with: `error  'onClick' is missing in props validation  react/prop-types`

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * A section sub-component for Breadcrumb component.
+ */
+function BreadcrumbSection(props) {
+  const {
+    active, children, className, link, href, onClick, ...rest,
+  } = props
+
+  const handleClick = (e) => {
+    if (onClick) onClick(e)
```
layershifter: Code coverage reports that this branch not tested. How I make it tested properly?

levithomason: You need to click on the component while it does not have have an `onClick` prop.  You can just assert that it does not throw any errors.

Reference the [Accordion test for onClick](https://github.com/TechnologyAdvice/stardust/blob/master/test/specs/modules/Accordion/Accordion-test.js#L162).

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+    </div>
+  )
+}
+
+Breadcrumb._meta = {
+  library: META.library.semanticUI,
+  name: 'Breadcrumb',
+  type: META.type.collection,
+  props: {
+    size: _.without(sui.sizes, 'medium'),
```
levithomason, line 53: Nice catch with the missing 'medium' class.

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+
+  let dividerJSX = <Breadcrumb.Divider />
+
+  if (divider) {
+    dividerJSX = <Breadcrumb.Divider>{divider}</Breadcrumb.Divider>
+  }
+
+  if (icon) {
+    dividerJSX = <Breadcrumb.Divider icon={icon} />
+  }
```
levithomason: **EDIT** _See next comment, this one no longer relevant._

Can we use `else` chaining here to prevent extra function calls and conditional checks?:

``` jsx
let dividerJSX

if (divider) {
  dividerJSX = <Breadcrumb.Divider>{divider}</Breadcrumb.Divider>
} else if (icon) {
  dividerJSX = <Breadcrumb.Divider icon={icon} />
} else {
  dividerJSX = <Breadcrumb.Divider />
}
```

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+      {sections.map((section, index) => {
+        const isLast = index === (sections.length - 1)
+
+        return [
+          <Breadcrumb.Section {...section}>{section.text}</Breadcrumb.Section>,
+          isLast ? null : dividerJSX,
+        ]
+      })}
+    </div>
+  )
```
levithomason: Regarding the `render()` function:
- prefer to pull js expressions into variables or smaller render methods, i.e. `renderSections()`.
- children in an array need a unique `key`, see [dynamic children](https://facebook.github.io/react/docs/multiple-components.html#dynamic-children)
- `children` here is a 2 dimensional array with a the last nested child being `null`, this could be a flat array with no null values.
- spreading section `{ ...section }` is also going to spread props we don't need, like `text`.  suggest destructuring `text` and spreading the `...rest`.

Some refactoring to address the above may look something like this:

``` jsx
const sectionsJSX = []

sections.forEach(({ text, ...restSection }, index) => {
  const key = `${text}-${index}`

  sectionsJSX.push(
    <Breadcrumb.Section {...restSection} key={key}>{text}</Breadcrumb.Section>
  )

  if (index !== sections.length - 1) {
    sectionsJSX.push(React.cloneElement(dividerJSX, { key }))
  }
})

return <div className={classes} {...rest}>{sectionsJSX}</div>
```

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+    PropTypes.string,
+  ]),
+
+  /** Array of props for Breadcrumb.Section. */
+  sections: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['children']),
+    customPropTypes.all([
+      React.PropTypes.arrayOf(React.PropTypes.shape({
+        text: React.PropTypes.string.isRequired,
+        link: customPropTypes.all([
```
levithomason: I hadn't considered nested use of the `all()` method like this.  If this works, that is awesome.  Were you able to try this out and confirm it works?

levithomason: Actually, since the section object is spread on the BreadcrumbSection, the BreadcrumbSection prop types can take care of the validation.  Just in case the mutually exclusive doesn't work with nested shape keys like this.

src/collections/Breadcrumb/Breadcrumb.js
levithomason: Looking at the `Breadcrumb.Divider` render method, it has conditional logic to return the proper component if an icon or children are passed.  It also has mutuallyExclusive prop types for these.

The `Breadcrumb` prop types mutually exclude `icon` and `divider`.  This means the only valid props are `icon` only, `divider` only or neither.  So we can simply do this:

``` jsx
const dividerJSX = <Breadcrumb.Divider icon={icon}>{divider}</Breadcrumb.Divider>
```

Because, if nothing is passed, these will both be `undefined` and the default divider `/` will render.  Otherwise, only `icon` or `divider` can be passed in which case the other will be `undefined` and the component will still be valid.  There is no need for the if/else logic in this area.

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+Breadcrumb.propTypes = {
+  /** Primary content of the Breadcrumb */
+  children: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['sections', 'icon', 'divider']),
+    PropTypes.node,
+  ]),
+
+  /** Primary content of the Breadcrumb.Divider. */
```
levithomason: Suggest adding, `For use with the sections prop.`

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,106 @@
...
+  divider: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['icon']),
+    PropTypes.string,
+  ]),
+
+  /** Render as an `Icon` component with `divider` class instead of a `div` in Breadcrumb.Divider. */
```
levithomason: Suggest adding, `For use with the sections prop.`

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -0,0 +1,55 @@
...
+    children, icon, className, ...rest,
+  } = props
+
+  const classes = cx(
+    className,
+    'divider',
+  )
+
+  if (icon) {
+    return <Icon className='divider' name={icon} />
```
levithomason: Let's spread `...rest` on the Icon as well.

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if (children) {
+    return <div className={classes} {...rest}>{children}</div>
+  }
+
+  return <div className={classes} {...rest}>/</div>
```
levithomason: If we use `defaultProps` for `children`, the `/` will show up in documentation.

``` js
  // ...

  if (icon) {
    return <Icon className={classes} name={icon} {...rest} />
  }

  return <div className={classes} {...rest}>{children}</div>
}

BreadcrumbDivider.defaultProps = {
  children: '/'
}
```

I've also added `classes` to the icon here so the users classes are still applies in the case of an icon divider.

test/specs/collections/Breadcrumb/BreadcrumbSection-test.js
```diff
@@ -0,0 +1,53 @@
...
+    })
+
+    it('is called with (event) on click', () => {
+      const handleClick = sandbox.spy()
+      const wrapper = shallow(<BreadcrumbSection onClick={handleClick}>Home</BreadcrumbSection>)
+
+      wrapper.should.have.tagName('a')
+
+      wrapper.simulate('click')
+      handleClick.should.have.been.calledOnce()
```
levithomason, line 50: Since the test states that the spy is called with an (event), we should also assert the first argument matches an object:

``` js
handleClick.should.have.been.calledOnce()
handleClick.should.have.been.calledWithMatch({})
```

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -25,28 +25,23 @@ function Breadcrumb(props) {
...
-
-        return [
-          <Breadcrumb.Section {...section}>{section.text}</Breadcrumb.Section>,
-          isLast ? null : dividerJSX,
-        ]
-      })}
-    </div>
-  )
+    if (index !== sections.length - 1) {
+      sectionsJSX.push(React.cloneElement(dividerJSX, { key: dividerKey }))
```
layershifter, line 41: You proposed

``` js
sectionsJSX.push(React.cloneElement(dividerJSX, { key }))
```

But it seems incorrect (dividers are not rendered), I added another key for divider. It's correct?

layershifter:
Code coverage shows 100% on `Breadcrumb`. Do I need more tests for `icon` and `divider` props on `Breadcrumb`?

---

Do I need examples to docs?

src/collections/Breadcrumb/Breadcrumb.js
levithomason, line 41: Whoops!  You're right, I reused the same `key` on the divider as I did on the section.  React uses the `key` to identify components, so it thought this was one component.

There may be a typo though:

``` js
const dividerKey = `$divider-${index}`
//=> `$divider-0`, `$divider-1`, etc.
```

You may have intended to do this?

``` js
const dividerKey = `${divider}-${index}`
//=> `/-0`, `/-1`, etc.
```

I would suggest we just add `divider` to the `key` for the `dividerKey`. Given a breadcrumb like so: `Home / Music / Playlist`, we'd have:

``` js
const key = `${text}-${index}`
const dividerKey = `${key}-divider`
//=> key: `Home-0`, `Music-1`, `Playlist-1`
//=> dividerKey: `Home-0-divider`, `Music-1-divider`, `Playlist-1-divider`
```

This should be reasonably unique for now.  Each element just needs to have a unique `key`.  Open to better ideas.

levithomason:
> Do I need more tests for icon and divider props on Breadcrumb?

It would be fantastic to have tests in `Breadcrumb-test.js` that assert that the icon/divider props generate the markup they are intended to generate.

> Do I need examples to docs?

Yes please, ❤️.  We don't have much in the way of [docs for adding examples](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#examples).  If you get stuck please LMK.

test/specs/collections/Breadcrumb/Breadcrumb-test.js
```diff
@@ -12,14 +12,29 @@ describe('Breadcrumb', () => {
...
   })
+
+  it('renders defined divider with `divider` prop', () => {
+    const wrapper = mount(<Breadcrumb sections={sections} divider='>' />)
+    const divider = wrapper.find('BreadcrumbDivider').first()
+
+    divider.text().should.to.equal('>')
+  })
+
+  it('renders divider as `Icon` with `icon` prop', () => {
```
levithomason: Since `icon` and `image` props should behave consistently across the entire project, there is a common tests for them.  You can remove this test and replace it with:

``` js
common.implementsIconProp(Breadcrumb)
```

test/specs/collections/Breadcrumb/Breadcrumb-test.js
```diff
@@ -12,14 +12,29 @@ describe('Breadcrumb', () => {
...
 
     wrapper.find('BreadcrumbDivider').should.have.length(1)
     wrapper.find('BreadcrumbSection').should.have.length(2)
   })
+
+  it('renders defined divider with `divider` prop', () => {
+    const wrapper = mount(<Breadcrumb sections={sections} divider='>' />)
+    const divider = wrapper.find('BreadcrumbDivider').first()
+
+    divider.text().should.to.equal('>')
```
levithomason: We use the [chai-enzyme](https://github.com/producthunt/chai-enzyme) plugin.  This allows us to move the [enzyme](https://github.com/airbnb/enzyme/) method to the end of the assertion chain:

``` diff
-divider.text().should.to.equal('>')
+divider.should.contain.text('>')
```

This makes for much more readable errors as well.  Compare the assertion syntax here with the error messages:

``` jsx
shallow(<Button>foo</Button>)
  .text().should.equal('bar')

//=> expected 'foo' to equal 'bar'
```

``` jsx
shallow(<Button>foo</Button>)
  .should.contain.text('bar')

//=>  expected <Button /> to contain text 'bar', but it has 'foo'      

         HTML:

         <button type="button" class="ui button">foo</button>
```

test/specs/collections/Breadcrumb/Breadcrumb-test.js
```diff
@@ -0,0 +1,40 @@
...
+  const sections = [
+    { text: 'Home', link: true },
+    { text: 'T-Shirt', href: 'google.com' },
+  ]
+
+  it('renders children with `sections` prop', () => {
+    const wrapper = shallow(<Breadcrumb sections={sections} />)
+
+    wrapper.find('BreadcrumbDivider').should.have.length(1)
+    wrapper.find('BreadcrumbSection').should.have.length(2)
```
levithomason: As noted in [#321 (comment)](https://github.com/TechnologyAdvice/stardust/pull/321/files#r70023249) let's use `chai-enzyme` assertions for much better assertions and error messages:

``` diff
-wrapper.find('BreadcrumbDivider').should.have.length(1)
+wrapper.should.have.exactly(1).descendants('BreadcrumbDivider')
```

``` diff
-wrapper.find('BreadcrumbSection').should.have.length(2)
+wrapper.should.have.exactly(2).descendants('BreadcrumbSection')
```

Compare the difference in output between `should.have.length(x)` and `should.have.exactly(x).descendants(...)`:

``` jsx
shallow(<Textarea />)
  .find('textarea').should.have.length(2)

//=> expected { Object (component, root, ...) } to have a length of 2 but got 1
```

``` jsx
shallow(<Textarea />)
  .should.have.have.exactly(2).descendants('textarea')

//=> expected <Textarea /> to have 2 descendants 'textarea' but actually found 1

         HTML:

         <textarea data-reactroot=""></textarea>
```

test/specs/collections/Breadcrumb/BreadcrumbDivider-test.js
```diff
@@ -0,0 +1,30 @@
...
+describe('BreadcrumbDivider', () => {
+  common.isConformant(BreadcrumbDivider)
+  common.rendersChildren(BreadcrumbDivider)
+
+  it('renders as a div by default', () => {
+    shallow(<BreadcrumbDivider />)
+      .should.have.tagName('div')
+  })
+
+  describe('icon prop', () => {
```
levithomason: See [#321 (comment)](https://github.com/TechnologyAdvice/stardust/pull/321/files#r70022927) for `common.implementsIconProp()`.

layershifter: Yes, it's good solution, but seems it will not work without #279.

image

test/specs/collections/Breadcrumb/Breadcrumb-test.js
layershifter: I tried, but seems it's not solution. Breadcrumb will not become `<i>` with prop.

layershifter:
I made copy for `Breadcrumb` from SUI docs

docs/app/Examples/collections/Breadcrumb/BreadcrumbExamples.js
```diff
@@ -0,0 +1,60 @@
...
+            examplePath='collections/Breadcrumb/Types/BreadcrumbExample'
+          />
+          <ComponentExample examplePath='collections/Breadcrumb/Types/BreadcrumbDividerExample' />
+          <ComponentExample examplePath='collections/Breadcrumb/Types/BreadcrumbPropExample' />
+        </ExampleSection>
+
+        <ExampleSection title='Content'>
+          <ComponentExample
+            title='Divider'
+            description='A breadcrumb can contain a divider to show the relationship between sections, this can be
```
levithomason: Strings in JS cannot span more than one line unless you use backticks, but this preserves whitespace including line breaks.  Suggest using an array and joining the lines:

``` js
description={[
  'A breadcrumb can contain a divider to show the relationship between sections,',
  'this can be formatted as an icon or text.',
].join(' ')}
```

docs/app/Examples/collections/Breadcrumb/Content/BreadcrumbIconDividerPropExample.js
```diff
@@ -0,0 +1,17 @@
...
+
+  sections = [
+    { text: 'Home', link: true },
+    { text: 'Registration', link: true },
+    { text: 'Personal Information', active: true },
+  ];
+
+  render() {
+    return (
+      <Breadcrumb icon='right-angle' sections={this.sections} />
```
levithomason:
``` diff
-icon='right-angle'
+icon='right angle'
```

levithomason: Could we refactor the sections from the component instance to the global scope?  I think it helps illustrate that prop configured components are well suited to receive their values from outside the component:

``` js
import { Breadcrumb } from 'stardust'

const sections = [
  { text: 'Home', link: true },
  { text: 'Registration', link: true },
  { text: 'Personal Information', active: true },
]

export default class BreadcrumbDividerPropExample extends Component {
  render() {
    return (
      <Breadcrumb divider='/' sections={sections} />
    )
  }
}
```

docs/app/Examples/collections/Breadcrumb/Content/BreadcrumbSectionPropExample.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react'
+import { Breadcrumb } from 'stardust'
+
+export default class BreadcrumbSectionPropExample extends Component {
+
+  sections = [
+    { text: 'Home', link: true },
+    { text: 'Search', active: true },
+  ];
```
levithomason: Same comment here regarding moving sections to the global scope.  Gasp!  Is that a semicolon 😝

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -0,0 +1,55 @@
...
+    children, icon, className, ...rest,
+  } = props
+
+  const classes = cx(
+    className,
+    'divider',
+  )
+
+  if (icon) {
+    return <Icon className={classes} name={icon} {...rest} />
```
levithomason: Oh right, until the `<Icon />` v1 API is merged, there is no `name` prop yet.  I know it is really close, LMK check on that.

FYI, this is one reason why Breadcrumb tests are failing on CI.

levithomason:
Wow, nice work on the docs.  Very thankful for all your PRs!  Left a couple comments on the docs.

I'll see what we can do to get the Icon merged quickly so we can merge this one too 👍

docs/app/Examples/collections/Breadcrumb/Content/BreadcrumbIconDividerPropExample.js
levithomason: While we're at the refactors, we might as well drop React.Component and use stateless functional examples as well:

``` js
import React from 'react'
import { Breadcrumb } from 'stardust'

const sections = [...]

const BreadcrumbDividerPropExample = () => (
  <Breadcrumb divider='/' sections={sections} />
)

export default BreadcrumbDividerPropExample
```

src/collections/Breadcrumb/BreadcrumbDivider.js
levithomason: If you'd like to wait on this PR, let's leave the Icon with the name prop.  If you'd like to get this one merged before continuing on with the others, then just use `className` here and the Icon PR can update it.

I'll leave the allocation up to you.  I'd expect the Icon PR to be merged very soon, shouldn't take the whole weekend.

layershifter:
I think we can wait until Icon's PR will be merged :smile:

levithomason:
Icon may have some more doc refactors coming.  How about we update this PR and merge it to keep pace?

**EDIT** I'd hate to hold up this great work for a single use of the Icon :)

layershifter:
@levithomason I've that `Icon` is ready to merge, it seems that it can be today. So I think that we can wait with merge of Breadcrumb :)

levithomason:
Hoping to get the conflict resolved! https://github.com/TechnologyAdvice/stardust/pull/279#issuecomment-232087055

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -18,7 +18,10 @@ function BreadcrumbDivider(props) {
   const rest = getUnhandledProps(BreadcrumbDivider, props)
 
   if (icon) {
-    return <Icon className={classes} name={icon} {...rest} />
+    // TODO: After update <Icon> to API replace with this code:
+    // return <Icon className={classes} name={icon} {...rest} />
+
+    return <Icon className={[icon, classes].join(' ')} {...rest} />
```
layershifter: I've updated usage of `Icon`, but it's still failing `common.implementsIconProp`. It seems that this test it's not correct for such situation.

layershifter:
ping @levithomason

levithomason:
Hmm, I'll pull this and see what's going on.

src/collections/Breadcrumb/BreadcrumbDivider.js
levithomason: Icon/Image props need to support both strings and components being passed.  We use the `iconPropRenderer` util for this.  This case is a little more tricky because we also need to add the users classes as well as the `divider` class to the icon after it is rendered.

The common test was over asserting things, I've updated it.  I've also pushed an update the prop renderers to support passing props.  I've tested this with the new renderers and it works:

``` diff
-import { customPropTypes, getUnhandledProps } from '../../utils/propUtils'
-import Icon from 'src/elements/Icon/Icon'
+import { customPropTypes, getUnhandledProps, iconPropRenderer } from '../../utils/propUtils'

/**
 * A divider sub-component for Breadcrumb component.
 */
function BreadcrumbDivider(props) {
  const {
    children, icon, className,
  } = props
  const classes = cx(
    className,
    'divider',
  )

  if (icon) {
    // TODO: After update <Icon> to API replace with this code:
    // return <Icon className={classes} name={icon} {...rest} />

-  return <Icon className={[icon, classes].join(' ')} {...rest} />
+    const iconClasses = _.isString(icon)
+      ? cx(icon, classes)
+      : cx(icon.props.className, classes)
+
+    return iconPropRenderer(icon, { ...rest, className: iconClasses })
}
```

Note I also used `cx` to build up the icon classes.

layershifter:
:wolf:

levithomason:
Fantastic, merging!

Side note, I'm going to also extend the prop renderers to merge the className when a component is passed.  This way, we don't have to check if `icon`/`image` is a string and merge classNames every time ourselves.

PR #322, opened 2016-07-05
Add Flag component 

### LINKED ISSUE #177: <Flag /> Component
http://semantic-ui.com/elements/flag.html

layershifter:
## API discussion

This component is pretty simple. There is only one changable property. I propose use `name` property as in `Icon` #279.

``` html
<i class="ae flag"></i>
<i class="france flag"></i>
<i class="myanmar flag"></i>
```

``` js
<Flag name='ae' />
<Flag name='france' />
<Flag name='myanmar' />
```

Feedback welcome :smile:


layershifter:
Fixes #177

src/elements/Flag/Flag.js
```diff
@@ -0,0 +1,30 @@
...
+
+function Flag(props) {
+  const { className, name, ...rest } = props
+  const classes = cx(
+    name,
+    className,
+    'flag',
+  )
+
+  return (<i className={classes} {...rest} />)
```
levithomason: Parens are only required for multi line JSX.  They can be removed here.

src/elements/Flag/Flag.js
```diff
@@ -0,0 +1,30 @@
...
+  library: META.library.semanticUI,
+  name: 'Flag',
+  type: META.type.element,
+}
+
+Flag.propTypes = {
+  /** Classes to add to the Flag className. */
+  className: PropTypes.string,
+
+  /** Name for element, can use the two digit country code, the full name, or a common alias */
```
levithomason: `Name of element` seems like it should be `Flag name`.

test/specs/elements/Flag/Flag-test.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('Flag', () => {
+  common.isConformant(Flag)
+
+  it('renders an <i /> element', () => {
+    shallow(<Flag />)
+      .should.have.tagName('i')
+  })
+
+  it('adds the "flag" class', () => {
```
levithomason: This test is covered by the `isConformant()` test.  The `isConformant()` test ensures the className includes has the component name.

levithomason:
Looks good to me 👍 Let's add the flag names, ISO codes, and aliases in the prop validation for `name` to finish it off.

layershifter:
>  Let's add the flag names, ISO codes, and aliases in the prop validation for name to finish it off.

Do you mean something like this?

``` js
const names = [
  'andorra',
  'ad',
  ...,
  'burma',
  'mm',
  'myanmar'
];

Flag._meta = {
  props: {
    name: names
  }
}

Flag.propTypes = {
 size: PropTypes.oneOf(Flag._meta.props.name)
}
```

levithomason:
Yep, that will do it.  As far as syntax, you can list them out horizontally since there will be many.  We allow 120 columns per line.

levithomason:
Great, lmk if there is anything else you'd like to do here.  Otherwise, I'll merge this once tests pass.

levithomason:
Oh, almost forgot.  Go ahead and check it off in the README.md under support :)

layershifter:
Updated branch and README :smile:

test/specs/elements/Flag/Flag-test.js
```diff
@@ -0,0 +1,18 @@
...
+
+describe('Flag', () => {
+  common.isConformant(Flag)
+
+  it('renders an <i /> element', () => {
+    shallow(<Flag />)
+      .should.have.tagName('i')
+  })
+
+  it('passes property "name" to class', () => {
```
levithomason: Sorry I missed this earlier, this test should be replaced with the common tests for "value only" props:

``` js
common.propValueOnlyToClassName(Flag, 'name')
```

This test ensures the prop is defined in _meta, is added to className when it should be, and is not added when it shouldn't be.

levithomason:
Thanks for the component!

PR #323, opened 2016-07-05
Add "all" custom prop type and fix issues

levithomason:
This PR adds the `all()` custom prop type.  This allows passing multiple prop type checkers.  It is needed in #321 and elsewhere to validate mutually exclusive props.

I also fixed a number of prop type issues and updated components using mutuallyExclusive.

### DIFF
--- a/docs/app/Examples/elements/Header/Variations/HeaderColoredExample.js
+++ b/docs/app/Examples/elements/Header/Variations/HeaderColoredExample.js
@@ -13,7 +13,7 @@ export default class HeaderColoredExample extends Component {
         <Header.H4 color='teal'>Teal</Header.H4>
         <Header.H4 color='blue'>Blue</Header.H4>
         <Header.H4 color='purple'>Purple</Header.H4>
-        <Header.H4 color='violent'>Violent</Header.H4>
+        <Header.H4 color='violet'>Violet</Header.H4>
         <Header.H4 color='pink'>Pink</Header.H4>
         <Header.H4 color='brown'>Brown</Header.H4>
         <Header.H4 color='grey'>Grey</Header.H4>
--- a/src/elements/Header/_Header.js
+++ b/src/elements/Header/_Header.js
@@ -77,7 +77,7 @@ _Header.propTypes = {
   ]),
 
   /** Color of the header. */
-  color: PropTypes.oneOf(_Header._meta.props.colors),
+  color: PropTypes.oneOf(_Header._meta.props.color),
 
   /** Align header content */
   aligned: PropTypes.oneOf(_Header._meta.props.aligned),
--- a/src/modules/Accordion/Accordion.js
+++ b/src/modules/Accordion/Accordion.js
@@ -22,7 +22,10 @@ export default class Accordion extends AutoControlledComponent {
     activeIndex: PropTypes.number,
 
     /** Accordion.Title and Accordion.Content components.  Mutually exclusive with the panels prop. */
-    children: customPropTypes.mutuallyExclusive('panels'),
+    children: customPropTypes.all([
+      customPropTypes.mutuallyExclusive(['panels']),
+      PropTypes.node,
+    ]),
 
     /** Classes to add to the Accordion className. */
     className: PropTypes.string,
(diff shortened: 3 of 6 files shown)


docs/app/Examples/elements/Header/Variations/HeaderColoredExample.js
```diff
@@ -13,7 +13,7 @@ export default class HeaderColoredExample extends Component {
         <Header.H4 color='teal'>Teal</Header.H4>
         <Header.H4 color='blue'>Blue</Header.H4>
         <Header.H4 color='purple'>Purple</Header.H4>
-        <Header.H4 color='violent'>Violent</Header.H4>
+        <Header.H4 color='violet'>Violet</Header.H4>
```
levithomason, line 16: Fixed typo.

src/elements/Header/_Header.js
```diff
@@ -77,7 +77,7 @@ _Header.propTypes = {
   ]),
 
   /** Color of the header. */
-  color: PropTypes.oneOf(_Header._meta.props.colors),
+  color: PropTypes.oneOf(_Header._meta.props.color),
```
levithomason, line 80: The `violent` color typo made it through because this prop type validator was pointed at the wrong prop name as well.

src/modules/Accordion/Accordion.js
```diff
@@ -22,7 +22,10 @@ export default class Accordion extends AutoControlledComponent {
     activeIndex: PropTypes.number,
 
     /** Accordion.Title and Accordion.Content components.  Mutually exclusive with the panels prop. */
-    children: customPropTypes.mutuallyExclusive('panels'),
+    children: customPropTypes.all([
+      customPropTypes.mutuallyExclusive(['panels']),
+      PropTypes.node,
+    ]),
```
levithomason, line 28: Here is how we use the `all()` validator.  It takes an array of prop type functions.

src/utils/propUtils.js
```diff
@@ -10,36 +10,83 @@ import Image from '../elements/Image/Image'
...
-      const disallowed = Children.map(children, child => {
+      if (propName !== 'children') {
+        throw new Error(`ofComponentTypes can only be used on the \`children\` prop, not ${propName}`)
+      }
+      if (!_.isArray(allowedTypes)) {
+        throw new Error([
+          'Invalid argument supplied to ofComponentTypes, expected an instance of array.'
+            ` See ${componentName} prop \`${propName}\`.`,
+        ].join(''))
+      }
```
levithomason, line 26: Added validation to all custom prop types arguments.  Some of these were being improperly used, so there was no validation happening at all.

src/utils/propUtils.js
```diff
@@ -10,36 +10,83 @@ import Image from '../elements/Image/Image'
...
-   * Verifies exclusivity of a given prop
-   * @param {string} exclusives property used for exclusivity check
+   * Verifies exclusivity of a given prop.
+   * @param {string[]} exclusives An array of props that cannot be used with this prop.
    */
   mutuallyExclusive: exclusives => {
     return (props, propName, componentName) => {
-      _.forEach(exclusives, exclusiveProp => {
-        if (props[propName] && props[exclusiveProp]) {
-          return new Error(
```
levithomason, line 38: Prop type functions must return an error.  This function however was returning an error inside of a forEach.  The validation error was never returned to the prop type function.  It now properly returns the errors.

src/utils/propUtils.js
```diff
@@ -10,36 +10,83 @@ import Image from '../elements/Image/Image'
...
+        throw new Error([
+          'Invalid argument supplied to mutuallyExclusive, expected an instance of array.'
+            ` See ${componentName} prop \`${propName}\`.`,
+        ].join(''))
+      }
+
+      // mutually exclusive
+      const disallowed = _.transform(exclusives, (acc, exclusive) => {
+        if (_.has(props, propName) && _.has(props, exclusive)) {
+          return acc.push(exclusive)
```
dvdzkwsk: `.push` returns what you pushed. Should it be `return acc.concat(exclusive)`?

Sidenote: feel like reduce or filter is more declarative, but is tangential to the above comment.

levithomason: ack! thanks, had a spread operator here earlier.

dvdzkwsk:
👻

kyleturco:
🐧 looks good

PR #327, opened 2016-07-07
Temporarily bundle jQuery and SUI JS

### LINKED ISSUE #326: Form: Uncaught TypeError: this.element.form is not a function
_Uncaught TypeError: this.element.form is not a function_ at https://github.com/TechnologyAdvice/stardust/blob/master/src/collections/Form/Form.js#L81.

Variable values at the moment of error:

``` js
this.element == $(this.refs.element)
this.element.form === undefined
```

I'm using stardust with webpack-dev-server and this issue occurs every time I use Form, including

``` js
<Form />
```

levithomason:
Fixes #326.  This PR address the issue of two copies of jQuery being loaded.  One on the the window via script tag, and another via the webpack bundle.

To simplify things for users for now, we simply bundle jQuery and SUI JS with Stardust.  This is only temporary as we are moving quickly to remove jQuery in #247.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,6 @@ export default class TryStardust extends Component {
 
   <!-- SUI CSS -->
   <link href="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.8/semantic.css" rel="stylesheet">
-
-  <!-- Temporary script dependencies until we're done removing jQuery -->
-  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.2.4/jquery.js"></script>
-  <script src="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.8/semantic.js"></script>
 </head>
 <body>
   <script src="bundle.js"></script>
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -66,11 +66,6 @@ webpackConfig.output = {
 // ------------------------------------
 webpackConfig.plugins = [
   new webpack.DefinePlugin(config.compiler_globals),
-  new webpack.ProvidePlugin({
-    $: 'jquery',
-    jQuery: 'jquery',
-    'window.jQuery': 'jquery',
-  }),
   new webpack.DllReferencePlugin({
     context: paths.base('node_modules'),
     manifest: require(paths.base('dll/vendor-manifest.json')),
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -47,7 +47,9 @@
     "bluebird": "^3.3.4",
     "classnames": "^2.1.5",
     "debug": "^2.2.0",
-    "lodash": "^4.6.1"
+    "jquery": "^3.1.0",
+    "lodash": "^4.6.1",
+    "semantic-ui-css": "^2.2.1"
```
levithomason, line 52: I'll be so glad the day I delete these two lines.

src/index.js
```diff
@@ -1,3 +1,5 @@
+import 'jquery'
+import 'semantic-ui-css/semantic.js'
```
levithomason: We load jQuery then SUI JS now.

README.md
```diff
@@ -58,10 +58,6 @@ export default class TryStardust extends Component {
 
   <!-- SUI CSS -->
   <link href="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.8/semantic.css" rel="stylesheet">
-
-  <!-- Temporary script dependencies until we're done removing jQuery -->
-  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.2.4/jquery.js"></script>
-  <script src="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.8/semantic.js"></script>
```
levithomason, line 64: No more need to include any deps, just `npm i stardust`.

docs/app/Components/Root.js
```diff
@@ -1,4 +1,3 @@
-import 'semantic-ui-css/semantic.js'
```
levithomason, line 1: Our doc site no longer needs to load `semantic.js`.

build/webpack.config.js
```diff
@@ -66,11 +66,6 @@ webpackConfig.output = {
 // ------------------------------------
 webpackConfig.plugins = [
   new webpack.DefinePlugin(config.compiler_globals),
-  new webpack.ProvidePlugin({
-    $: 'jquery',
-    jQuery: 'jquery',
-    'window.jQuery': 'jquery',
-  }),
```
levithomason, line 73: We don't expose jQuery on the window up front.  We attach it to the window ourselves after `require()`ing it.  We do this so we can throw an error if the user has already loaded it on the window before we load.

build/webpack.config.js
```diff
@@ -160,7 +155,8 @@ webpackConfig.module.loaders = [{
 // ----------------------------------------
 // For faster builds in dev, rely on prebuilt libraries
 // Local modules can still be enabled (ie for offline development)
-if (argv.localModules) {
+// in TEST we need local modules because karma uses a different index.html (no CDNs)
+if (__TEST__ || argv.localModules) {
```
levithomason, line 159: In tests we use local modules, there are no CDNs.

Issue #180, opened 2016-03-16
<Loader /> Component


levithomason:
http://semantic-ui.com/elements/loader.html

layershifter:
## API

Loader can be basic and with text.

``` html
<div class="ui loader"></div>
<div class="ui text loader">Loading</div>
```

``` js
<Loader />
<Loader text='Loading' />
```

Loader has following states: active, disabled and indeterminate.

``` html
<div class="ui active loader"></div>
<div class="ui disabled loader"></div>
<div class="ui indeterminate text loader">Preparing Files</div>
```

``` js
<Loader active />
<Loader disabled />
<Loader indeterminate text='Preparing Files' />
```

Loader has following variations: inline, centered inline, inverted and size.

``` html
<div class="ui active inline loader"></div>
<div class="ui active centered inline loader"></div>
<div class="ui active inverted loader"></div>
<div class="ui mini text loader">Loading</div>
```

``` js
<Loader active inline />
<Loader active centered inline />
<Loader active centered-inline />
<Loader size='mini' text='Loading' />
```

Did I miss any? :smile:

levithomason:
Woop!  Nice job here, we're ready to start this one considering these comments:

### Text

Let's also support `mutuallyExclusive` children.  This will allow users to put more complicated markup in the loader:

``` jsx
<Loader>
  <p>Taking a while?</p>
  <Button size='mini' icon='refresh'>Reload</Button>
</Loader>
```

``` html
<div class="ui mini text loader">
  <p>Taking a while?</p>
  <button class="ui mini button">
    <i class="refresh icon" />
    Reload
  </button>
</div>
```

> Note the presence of `children` should also add the `text` class.

### Inline

The API for this one looks like this. Use the `keyOrValueAndKey()` propUtil in the `cx` buildup to extract the classes.

``` jsx
<Loader inline />
<Loader inline='centered' />
```

### Size

Just clarifying it does not support `big` or `massive` sizes.

layershifter:
Thanks for feedback :smile: I'll make PR soon

layershifter:
It seems, issue may be closed :smile:

PR #328, opened 2016-07-08
Add Loader component

layershifter:
### Types

Loader can be basic and with text.

``` html
<div class="ui loader"></div>
<div class="ui text loader">Loading</div>
```

``` js
<Loader />
<Loader text='Loading' />
```

Also support `mutuallyExclusive` children. This will allow users to put more complicated markup in the loader:

``` html
<div class="ui mini text loader">
  <p>Taking a while?</p>
  <button class="ui mini button">
    <i class="refresh icon" />
    Reload
  </button>
</div>
```

``` js
<Loader>
  <p>Taking a while?</p>
  <Button size='mini' icon='refresh'>Reload</Button>
</Loader>
```
### States

Loader has following states: active, disabled and indeterminate.

``` html
<div class="ui active loader"></div>
<div class="ui disabled loader"></div>
<div class="ui indeterminate text loader">Preparing Files</div>
```

``` js
<Loader active />
<Loader disabled />
<Loader indeterminate text='Preparing Files' />
```
### Variations

Loader has following variations: inline, centered inline, inverted and size.

``` html
<div class="ui active inline loader"></div>
<div class="ui active centered inline loader"></div>
<div class="ui active inverted loader"></div>
<div class="ui mini text loader">Loading</div>
```

``` js
<Loader inline />
<Loader inline='centered' />
<Loader inverted />
<Loader size='mini' text='Loading' />
```

> Component supports all sizes.

### DIFF
--- a/docs/app/Examples/elements/Loader/Variations/InlineCenterLoaderExample.js
+++ b/docs/app/Examples/elements/Loader/Variations/InlineCenterLoaderExample.js
@@ -1,0 +1,10 @@
+import React, { Component } from 'react'
+import { Loader } from 'stardust'
+
+export default class InlineCenterLoaderExample extends Component {
+  render() {
+    return (
+      <Loader active inline='center' />
+    )
+  }
+}
--- a/docs/app/Examples/elements/Loader/Variations/InlineCenterLoaderExample.js
+++ b/docs/app/Examples/elements/Loader/Variations/InlineCenterLoaderExample.js
@@ -4,7 +4,7 @@ import { Loader } from 'stardust'
 export default class InlineCenterLoaderExample extends Component {
   render() {
     return (
-      <Loader active inline='center' />
+      <Loader active inline='centered' />
     )
   }
 }
--- a/docs/app/Examples/elements/Loader/Variations/InvertedSizeLoaderExamples.js
+++ b/docs/app/Examples/elements/Loader/Variations/InvertedSizeLoaderExamples.js
@@ -1,0 +1,58 @@
+import React, { Component } from 'react'
+import { Loader, Segment } from 'stardust'
+
+export default class SizeLoaderExamples extends Component {
+  render() {
+    return (
+      <div>
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='mini' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='tiny' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='small' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='medium' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='large' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='big' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='huge' text='Loading' />
+          </div>
+        </Segment>
+
+        <Segment className='very padded'>
+          <div className='ui active inverted dimmer'>
+            <Loader size='massive' text='Loading' />
+          </div>
+        </Segment>
+      </div>
+    )
+  }
+}
--- a/docs/app/Examples/elements/Loader/Variations/InvertedSizeLoaderExamples.js
+++ b/docs/app/Examples/elements/Loader/Variations/InvertedSizeLoaderExamples.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react'
 import { Loader, Segment } from 'stardust'
 
-export default class SizeLoaderExamples extends Component {
+export default class InvertedSizeLoaderExamples extends Component {
   render() {
     return (
       <div>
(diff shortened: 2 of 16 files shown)


layershifter:
Fixes #180

src/elements/Loader/Loader.js
```diff
@@ -0,0 +1,75 @@
...
+  /** Show that the Loader is indeterminate. */
+  indeterminate: PropTypes.bool,
+
+  /** Show that the Loader is inverted. */
+  inverted: PropTypes.bool,
+
+  /** Show that the Loader is inline. */
+  inline: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.string,
```
levithomason:
``` diff
-PropTypes.string,
+PropTypes.oneOf(Loader._meta.props.inline)
```

test/specs/elements/Loader/Loader-test.js
```diff
@@ -0,0 +1,35 @@
...
+describe('Loader', () => {
+  common.isConformant(Loader)
+  common.hasUIClassName(Loader)
+  common.propKeyOnlyToClassName(Loader, 'active')
+  common.propKeyOnlyToClassName(Loader, 'disabled')
+  common.propKeyOnlyToClassName(Loader, 'indeterminate')
+  common.propKeyOrValueToClassName(Loader, 'inline')
+  common.propKeyOnlyToClassName(Loader, 'inverted')
+  common.propKeyOnlyToClassName(Loader, 'text')
+  common.rendersChildren(Loader)
```
levithomason, line 16: Looks like we need the `size` test in here as well:

``` js
common.propValueOnlyToClassName(Loader, 'size')
```

src/elements/Loader/Loader.js
```diff
@@ -0,0 +1,75 @@
...
+  ]),
+
+  /** Size of the Loader. */
+  size: PropTypes.oneOf(Loader._meta.props.sizes),
+
+  /** Text content of the Loader. Mutually exclusive with the children prop. */
+  text: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['children']),
+    PropTypes.string,
```
levithomason: Since `text` is placed in the component as children, any valid child node is technically also valid as a `text` value.  Let' permit all nodes here:

``` diff
-PropTypes.string,
+PropTypes.node,
```

levithomason:
Added a couple inline comments, looking great!  Let's scratch it off in the README.md under Support.

The last bit is to add some examples.  LMK if you need any more information on this.

layershifter:
I'll add docs soon

layershifter:
I've ported docs from SUI and added entry to README.md.

---

But, here is note, docs for `<Loader>` require `<Message>`, `<Segment>` and `<Dimmer>` (I've added TODOs).

levithomason:
> docs for <Loader> require <Message>, <Segment> and <Dimmer> (I've added TODOs).

Let's use the existing APIs for Message and Segment. Then, just use html markup for the Dimmer. We can update these in the PRs to update and create those.

levithomason:
Whoops, ignore my last as it looks like this is already done :+1:

layershifter:
What about merge? :yum:

levithomason:
Woop!  Thanks for the note, taking a final look at this guy now.

docs/app/Examples/elements/Loader/Variations/InlineCenterLoaderExample.js
```diff
@@ -0,0 +1,10 @@
+import React, { Component } from 'react'
+import { Loader } from 'stardust'
+
+export default class InlineCenterLoaderExample extends Component {
```
levithomason:
``` diff
-InlineCenterLoaderExample
+InlineCenteredLoaderExample
```

Please update file name as well.

docs/app/Examples/elements/Loader/Variations/InvertedSizeLoaderExamples.js
```diff
@@ -0,0 +1,58 @@
+import React, { Component } from 'react'
+import { Loader, Segment } from 'stardust'
+
+export default class SizeLoaderExamples extends Component {
```
levithomason: Filename and class name are mismatched here, not sure which is correct.

src/elements/Loader/Loader.js
```diff
@@ -0,0 +1,75 @@
...
+  inverted: PropTypes.bool,
+
+  /** Show that the Loader is inline. */
+  inline: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.oneOf(Loader._meta.props.inline),
+  ]),
+
+  /** Size of the Loader. */
+  size: PropTypes.oneOf(Loader._meta.props.sizes),
```
levithomason: This is not validating sizes at the moment because the property is misspelled and therefore `undefined`:

``` diff
-Loader._meta.props.sizes
+Loader._meta.props.size
```

src/elements/Loader/Loader.js
```diff
@@ -0,0 +1,75 @@
...
+    useKeyOnly(indeterminate, 'indeterminate'),
+    useKeyOrValueAndKey(inline, 'inline'),
+    useKeyOnly(inverted, 'inverted'),
+    useKeyOnly(text, 'text') || useKeyOnly(children, 'text'),
+    size,
+    className,
+    'loader',
+  )
+
+  return <div className={classes} {...rest}>{ children || text }</div>
```
levithomason, line 24: For consistency, use `getUnhandledProps()` propUtil instead of `...rest`

``` js
const rest = getUnhandledProps(Loader, props)
```

levithomason:
Ready to merge following the clean up comments added.

levithomason:
Ignore that test fail, I've re-triggered your build without cache.  I think this should resolve it.

layershifter:
Wait with merge, I rename examples with pattern:

> ${componentName}${exampleName}Example.js

levithomason:
> Wait with merge, I rename examples with pattern:

Either way is OK, the docs are getting refactored soon.  The filenames will likely change.

layershifter:
Ok :+1: I will not touch names. So wait for CI finish

levithomason:
Another one down, thanks much!

PR #333, opened 2016-07-13
Fix Safari private mode localStorage

### LINKED ISSUE #332: Setting localStorage throws in Safari private mode
Per the gitter issue from @eliasliedholm:

> Hi, I’m running into a problem when using Stardust in safari private mode, because safari does not allow the debugger tool to write to the localStorage object, and will throw an exception. I found that in runtime, process.env.NODE_ENV is undefined, so the debugger will be added even in production, which means safari private mode will always crash.
> 
> Was wondering if you knew a way to get around it?

We use [debug](https://github.com/visionmedia/debug) for this and this is a known issue with the way it is configured in the browser: https://github.com/visionmedia/debug/issues/167.

I'm adding this as a bug.  We should wrap our localStorage use in a try/catch.

levithomason:
Fixes #332.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,3 @@
+coverage:
+  ignore:
+    - src/utils/*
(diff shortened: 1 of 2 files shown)


dvdzkwsk:
👻

eliasliedholm:
🙌 thank you!

Issue #183, opened 2016-03-16
<Step /> Component


levithomason:
http://semantic-ui.com/elements/step.html

layershifter:
I'll work about this issue after #184.

layershifter:
# API

## Types

### Step

A single step

``` html
<div class="ui steps">
  <div class="step">
    Shipping
  </div>
</div>
```

``` js
<Step.Group>
  <Step.Item>Shipping</Step.Item>
</Step.Group>
// or
const items = [
  {text: 'Shipping'}
];

<Step.Group steps={items}>
```

## Groups

### Steps

A set of steps

``` html
<div class="ui steps">
  <div class="step">
    <i class="truck icon"></i>
    <div class="content">
      <div class="title">Shipping</div>
      <div class="description">Choose your shipping options</div>
    </div>
  </div>
  <div class="active step">
    <i class="payment icon"></i>
    <div class="content">
      <div class="title">Billing</div>
      <div class="description">Enter billing information</div>
    </div>
  </div>
  <div class="disabled step">
    <i class="info icon"></i>
    <div class="content">
      <div class="title">Confirm Order</div>
    </div>
  </div>
</div>
```

``` js
<Step.Group>
  <Step.Item icon='truck'>
    <Step.Item.Title>Shipping</Step.Item.Title>
    <Step.Item.Description>Shipping</Step.Item.Description>
  </Step.Item>
  <Step.Item active icon='payment'>
    <Step.Item.Title>Billing</Step.Item.Title>
    <Step.Item.Description>Confirm Order</Step.Item.Description>
  </Step.Item>
  <Step.Item disabled icon='info' title='Confirm Order' />
</Step.Group>
// or
const items = [
  {icon: 'truck', title: 'Shipping', description: 'Shipping'},
  {active: true, icon: 'payment', title: 'Billing', description: 'Confirm Order'},
  {disabled: true, icon: 'info', title: 'Confirm Order'},
];

<Step.Group steps={items}>
```

### Ordered

A step can show a ordered sequence of steps

``` html
<div class="ui ordered steps">
  <div class="completed step">
    <div class="content">
      <div class="title">Shipping</div>
      <div class="description">Choose your shipping options</div>
    </div>
  </div>
...
</div>
```

``` js
<Step.Group ordered>
  <Step.Item completed>
    <Step.Item.Title>Shipping</Step.Item.Title>
    <Step.Item.Description>Shipping</Step.Item.Description>
[...]

Issue #203, opened 2016-03-18
Rename plural group components (Fields) -> Group


dvdzkwsk:
We would like to nest groupings under the component (as static properties), so that:

``` js
import { Form } from 'stardust'

// old
<Form.Field>
<Form.Fields>

// new
<Form.Field>
<Form.Field.Group>
```

This changes the naming of `Fields` -> `Group` since it is standardized across all components and is more intuitive overall.
### Todo

> @levithomason added by edit  on Jul 11, 2016
- [x] ~~#400 Fields~~
- [x] ~~#295 Buttons~~
- [x] ~~#279 Icons~~
- [x] ~~#287 Segments~~
- [x] ~~#431 Items~~
- [x] ~~#334 Statistics~~

jhchill666:
If standardising across groupable components, should Group be component agnostic?

levithomason:
Could you give an example @jamiehill?

jhchill666:
@davezuko suggested in the inital comment that "group is standardized across all components", and as such I wondered whether a generic Group Class was what he was thinking, or just a standardised api for such grouped components

levithomason:
Initially, this would just be a standardized API.  This issue was partly raised as we've decided to first build out our components with a sub-component approach.  Example, `<Dropdown />`, `<Dropdown.Item />`, etc.  This allows us to handle the special use case of an `Item` in a `Dropdown`, which is quite different than many other `Item`s.  This means the "group" components look like this: `<Button.Buttons />`.

Consider a group of [buttons](http://semantic-ui.com/elements/button.html#buttons).  In SUI we need a div with class `buttons` that wraps a set of `ui button` elements:

``` html
<div class="ui buttons">
  <button class="ui button">One</button>
  <button class="ui button">Two</button>
  <button class="ui button">Three</button>
</div>
```

We currently handle this pattern in Stardust like this:

``` js
import { Buttons, Button } from 'stardust'

const group = (
  <Buttons>
    <Button>One</Button>
    <Button>Two</Button>
    <Button>Three</Button>
  </Buttons>
)
```

Every component that can be grouped (which is most) has a component with the plural form of the name for grouping.  Button / Buttons, Statistic / Statistics, etc.  This issue is proposing that we do two things:
1. Move group components to sub components: `<Button.Buttons />`, `<Statistic.Statistics />`
2. Rename all group components to "Group": `<Button.Group />`, `<Statistic.Group />`

The above would then become:

``` js
[...]

PR #334, opened 2016-07-15
Update Statistic to v1 API

layershifter:
# API
### Statistic

A statistic can display a value with a label above or below it.

``` html
<div class="ui statistic">
  <div class="value">5,550</div>
  <div class="label">Downloads</div>
</div>
<div class="ui statistic">
  <div class="label">Views</div>
  <div class="value">40,509</div>
</div>
```

``` jsx
<Statistic>
  <Statistic.Value>5,550</Statistic.Value>
  <Statistic.Label>Downloads</Statistic.Label>
</Statistic>
<Statistic>
  <Statistic.Label>Views</Statistic.Label>
  <Statistic.Value>40,509</Statistic.Value>
</Statistic>
```
### Statistic Group

A group of statistics

``` html
<div class="ui statistics">
  <div class="statistic">
    <div class="value">22</div>
    <div class="label">Faves</div>
  </div>
  ...
</div>
```

``` jsx
<Statistic.Group>
  <Statistic>
    <Statistic.Value>22</Statistic.Value>
    <Statistic.Label>Faves</Statistic.Label>
  </Statistic>
  ...
</Statistic.Group>
```
### Value

A statistic can contain a numeric, icon, image, or text value

``` html
<div class="statistic">
    <div class="value">22</div>
   <div class="label">Saves</div>
</div>
<div class="statistic">
  <div class="text value">Three<br>Thousand</div>
  <div class="label">Signups</div>
</div>
<div class="statistic">
  <div class="value"><i class="plane icon"></i> 5</div>
  <div class="label">Flights</div>
</div>
<div class="statistic">
    <div class="value">
      <img src="/images/avatar/small/joe.jpg" class="ui circular inline image"> 42
    </div>
    <div class="label">Team Members</div>
</div>
```

``` jsx
<Statistic>
  <Statistic.Value>22</Statistic.Value>
  <Statistic.Label>Saves</Statistic.Label>
</Statistic>
<Statistic>
  <Statistic.Value text>Three<br>Thousand</Statistic.Value>
  <Statistic.Label>Signups</Statistic.Label>
</Statistic>
<Statistic>
  <Statistic.Value><Icon name='plane' /> 5</Statistic.Value>
  <Statistic.Label>Flights</Statistic.Label>
</Statistic>
<Statistic>
  <Statistic.Value><Image src='/images/avatar/small/joe.jpg' ... /></Statistic.Value>
  <Statistic.Label>Team Members</Statistic.Label>
</Statistic>
```
### Label

A statistic can contain a label to help provide context for the presented value

_Already showed_
### Horizontal Statistic

A statistic can present its measurement horizontally

``` html
<div class="ui horizontal statistic">...</div>
<div class="ui horizontal statistics">...</div>
```

``` jsx
<Statistic horizontal>...</Statistic>
<Statistic.Group horizontal>...</Statistic.Group>
```
### Colored

A statistic can be formatted to be different colors

``` html
<div class="red statistic"></div>
```

``` jsx
<Statistic color='red'>...</Statistic>
```

> Colors will be used from `semanticUtils`
### Inverted

A statistic can be formatted to fit on a dark background

``` html
<div class="ui inverted statistic">...</div>
```

``` jsx
<Statistic inverted>...</Statistic>
```
### Evenly Divided

A statistic group can have its items divided evenly

Will be taken from #183
### Floated

An statistic can sit to the left or right of other content

``` html
<div class="ui right floated statistic">...</div>
```

``` jsx
<Statistic floated='left'>...</Statistic>
<Statistic floated='right'>...</Statistic>
```
### Size

A statistic can vary in size

``` jsx
<Statistic size='mini'>...</Statistic>
```

> There is no `medium`, `big` and `massive` sizes.


layershifter:
Refers #269

layershifter:
What about this?

``` jsx
<Statistic label='Faves' value='22' />
```

levithomason:
API looks great.  The simplified props version look right as well:

``` js
<Statistic label='Faves' value='22' />
```

We probably just accept the limitation that using props only like this will generate markup with the value on top and the label on bottom.  I assume this is the most common configuration.

image

Any intuitive ideas on how to allow reordering label/value with the props API?

layershifter:
> Any intuitive ideas on how to allow reordering label/value with the props API?

What about boolean `reverse`/`labeled`/`valued` prop?

layershifter:
What about short syntax for label and value?

``` jsx
<Statistic.Label text='Faves' />
<Statistic.Value text='22' />
```

**UPD:**
Oops, `text` prop is already taken in `Statistic.Value`, what about `content`?

layershifter:
I've implemented all features, added tests and docs. @levithomason wait for feedback :smile:

levithomason:
> What about boolean reverse/labeled/valued prop?

After thinking it through, I think we should just provide a single layout, value on top, label on bottom.  If a user wants more control, they can use the sub components.  This seems to be the cleanest solution, for now.

> Oops, `text` prop is already taken in Statistic.Value, what about `content`?

Nice catch, I never saw the `text` option buried in the docs.  We've been using text in many places for shorthand configuration, which as you've noted will not work here.  I'd like for shorthand props to be as consistent as possible between all components.  My worry is that since `content` is also a common SUI term, this will have conflicts elsewhere.  This makes me think we should not provide shorthand props for label/value sub components here just yet.

Perhaps add a TODO in each component to find a shorthand prop for each that is not `text` but that is also consistent with shorthand props in other components.  Then, once we get through the v1 updates, we can scan through the components and find the most appropriate prop name.

levithomason:
Some more ideas on props for Statistic sub components.  The Statistic already accepts `label` and `value` shorthand props.  We could allow these props on the subcomponents:

``` jsx
<Statistic.Value value={99} />
<Statistic.Label label='Likes' />
```

Regarding label/value order for Statistic shorthand props, React props are order dependent.   As noted in the [React docs](https://facebook.github.io/react/docs/transferring-props.html#consuming-and-transferring-the-same-prop):

> **NOTE:**
> Order matters. By putting the `{...other}` before your JSX props you ensure that the consumer of your component can't override them. In the example above we have guaranteed that the input will be of type `"checkbox"`.

So we could simply do this:

``` jsx
<Statistic label='Label first' value={3} />
<Statistic value={99} label='Label last' />
```

See it in action here: http://www.react.run/HJxagQKP/2

src/views/Statistic/Statistic.js
```diff
@@ -1,42 +1,95 @@
...
+  /** Label content of the Statistic. Mutually exclusive with the children prop. */
+  label: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['children']),
+    PropTypes.node,
+  ]),
+
+  /** A statistic can vary in size. */
+  size: PropTypes.oneOf(Statistic._meta.props.size),
+
+  /** Shows that StatisticValue is text. */
```
levithomason: Suggest something along the lines of:

``` js
/** Format the value with smaller font size to fit nicely beside number values. */
```

src/views/Statistic/StatisticGroup.js
```diff
@@ -0,0 +1,68 @@
...
+import numberToWord from '../../utils/numberToWord'
+import META from '../../utils/Meta'
+import Statistic from './Statistic'
+
+function StatisticGroup(props) {
+  const { children, className, horizontal, items, widths } = props
+  const classes = cx(
+    'ui',
+    useKeyOnly(horizontal, 'horizontal'),
+    _.isNumber(widths) ? numberToWord(widths) : widths,
```
levithomason: Currently, if `widths='5'` is provided, the class `5` will be added because it isn't of type number.  I think we should support widths as strings as well.  This will be common when getting the number from an API or from the value of a DOM control.

One easy way to do this is to change the return statement in `numberToWord` to return the argument itself if it did not have a map for the number:

``` diff
-return map[number]
+return map[number] || number
```

This way, `number` can be a string or number and it will still find the word equivalent.  If there is no match found (i.e. in case the user passed `two`) then it will be returned.

src/views/Statistic/StatisticGroup.js
```diff
@@ -0,0 +1,68 @@
...
+  items: customPropTypes.all([
+    customPropTypes.mutuallyExclusive(['children']),
+    PropTypes.array,
+  ]),
+
+  /** A statistic group can have its items divided evenly. */
+  widths: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
```
levithomason: Let's extend `semanticUtils` to include `widths` so we can validate with `oneOf`.  We should include numbers in both number and string types, and also the word forms.  Including the previous suggested refactor to `numberToWord`, we can generate these like so:

``` js
// numberToWord.js
export const numberToWordMap = {
  1: 'one',
  2: 'two',
  3: 'three',
  4: 'four',
  5: 'five',
  6: 'six',
  7: 'seven',
  8: 'eight',
  9: 'nine',
  10: 'ten',
  11: 'eleven',
  12: 'twelve',
  13: 'thirteen',
  14: 'fourteen',
  15: 'fifteen',
  16: 'sixteen',
}

export default function numberToWord(number) {
  return numberToWordMap[number] || number
}
```

``` js
// semanticUtils.js
import { numberToWordMap } from './numberToWord'

export const widths = [
  ..._.keys(numberToWordMap),             // '1', '2' ...
  ..._.keys(numberToWordMap).map(Number), // 1, 2 ...
  ..._.values(numberToWordMap),           // 'one', 'two', ...
]
```

levithomason: Note, the reasoning behind these enums is that I'll be adding support for prop validation that provides suggested prop values based on similarity comparison.  So if you added `widths='twevle` we will say `Did you mean: "twelve"?`.

This is especially helpful for things like the icon names.

src/views/Statistic/StatisticLabel.js
```diff
@@ -4,12 +4,17 @@ import React, { PropTypes } from 'react'
 import { customPropTypes, getUnhandledProps } from '../../utils/propUtils'
 import META from '../../utils/Meta'
 
+// TODO: This file has disabled shorthand props
+// @see https://github.com/TechnologyAdvice/stardust/pull/334
```
levithomason, line 8: 👍

docs/app/Examples/views/Statistic/Variations/StatisticGroupHorizontalExample.js
```diff
@@ -0,0 +1,14 @@
...
+import { Statistic } from 'stardust'
+
+const items = [
+  { label: 'Views', value: '2,204' },
+  { label: 'Downloads', value: '3,322' },
+  { label: 'Tasks', value: 22 },
+]
+
+const StatisticGroupHorizontalExample = () => (
+  <Statistic.Group horizontal items={items} />
```
levithomason: Nice shorthand here.

src/views/Statistic/Statistic.js
```diff
@@ -1,42 +1,95 @@
...
+    return <div className={classes} {...rest}>{children}</div>
   }
 
-  static Statistics = StatisticStatistics
-  static Label = StatisticLabel
-  static Value = StatisticValue
+  return (<div className={classes} {...rest}>
+    <Statistic.Value text={text}>{value}</Statistic.Value>
+    <Statistic.Label>{label}</Statistic.Label>
+  </div>)
```
levithomason: Since props are applied in order, we need to spread the `rest` before applying the className.  The `classes` includes the user's className values.  If we apply the `rest` after the `classes`, then their classNames will override the `classes` we've built up.

Also, I think the linter should have caught this but let's return multiline JSX like this:

``` jsx
return (
  <div {...rest} className={classes}>
    <Statistic.Value text={text}>{value}</Statistic.Value>
    <Statistic.Label>{label}</Statistic.Label>
  </div>
)
```

levithomason: https://facebook.github.io/react/docs/transferring-props.html#consuming-and-transferring-the-same-prop

src/views/Statistic/StatisticGroup.js
levithomason: Now that the utils are updated:

``` diff
-    _.isNumber(widths) ? numberToWord(widths) : widths,
+    numberToWord(widths),
```

src/views/Statistic/StatisticGroup.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  let itemsJSX = []
+
+  items.forEach((item, index) => {
+    const key = `${item.label}-${item.value}-${index}`
+    itemsJSX.push(<Statistic key={key} {...item} />)
+  })
+
+  return <div className={classes} {...rest}>{itemsJSX}</div>
```
levithomason: As noted in Statistic, spread before applying className to prevent user's className from overriding the className build up.

Conformance tests ensure props are spread, I should add a test to ensure className is not override  during spread ;)

levithomason: Have to correct myself here, while this is true regarding prop order `rest` actually does not include className in this example.  That is because the className prop is being handled in the propTypes, so `getUnhandledProps` is returning an object of props that does not include `className`.

Let's still spread them first, and I'm adding the common test now. I just wanted to point out that this would actually work as is and my original comment was incorrect.

layershifter:
Okay, here is some problems:
- `common.isConformant(StatisticGroup)` fails, seems test might be updated:
  image
- `common.propValueOnlyToClassName(StatisticGroup, 'widths')` fails, seems there needed new test in commonTests:
  image

test/specs/views/Stastistic/StatisticGroup-test.js
```diff
@@ -1,15 +1,13 @@
...
-import numberToWord from 'src/utils/numberToWord'
-
 import StatisticGroup from 'src/views/Statistic/StatisticGroup'
 
 describe('StatisticGroup', () => {
   common.isConformant(StatisticGroup)
   common.hasUIClassName(StatisticGroup)
   common.rendersChildren(StatisticGroup)
   common.propKeyOnlyToClassName(StatisticGroup, 'horizontal')
+  common.propValueOnlyToClassName(StatisticGroup, 'widths')
```
levithomason: This test is _mostly_ accurate, but not entirely.  Consider when a user passes `widths='3'`, the value `'3'` is not actually added to the className, the word `three` is.

We should likely add a common test for `implementsNumberToWordProp(Statistic, 'widths')`.  This test would assert that the `numberToWord` value is added to the `className`.  It would also assert that it works with numbers, strings, and number words (i.e. 'two').

The reason I think we should allow the second param to this common tests is that on other components, like `FormFields`, the `numerToWord` prop is `width`.  This way, we can point this test to a specific prop name.

layershifter:
I've made sort, seems it make all ugly :crying_cat_face: 

image

layershifter:
I've created `implementsNumberToWordProp`, take a look

Need fix for `common.isConformant(StatisticGroup)` and decision for props order

levithomason:
> I've made sort, seems it make all ugly

Oh boy, yep. That isn't good nor intuitive. Let's not sort.

levithomason:
I'll be tied up the next couple hours. Will check on the rest then. Thanks much for all the refactors and rapid responses!

layershifter:
It seems that `componentClassName` definition in `componentInfo` needs to be modified for #203

levithomason:
Whoops, missed this one.  Checking...

levithomason:
Rebase and give this a shot.  https://github.com/TechnologyAdvice/stardust/commit/c5b0425234cbc7b0718a86560425f44ae46c6581

Tested on this branch and master.  It uses the plural name of the parent for componentClassName on sub component groups.  So, `statistics` for `StatisticGroup`.

layershifter:
Seems it works :+1: Waiting for review

levithomason:
On it

Issue #325, opened 2016-07-07
RFC: Standardize Component Parts


levithomason:
See #345 for experimental work against this issue. 

---

This is a continuation of https://github.com/TechnologyAdvice/stardust/pull/311#issuecomment-231006785 which discusses how to standardize and reuse components and component parts.
## Components vs Component Parts

The confusing thing for many people is that a `header` "component part" is not a `ui header` "component". It has the class `header` because, semantically, that is what it is in its parent component's context.  The glossary tries to explain this under [`ui`](http://semantic-ui.com/introduction/glossary.html#project-terminology).  Really, the only shared/reusable aspect is the class `header`.

> ...Would it not be convenient to be able to utilize the Header sub comps:
> 
> ``` jsx
> <Card>
>    <Card.Header.H1>Large Card Header</Card.Header.H1>
> </Card>
> ```

Intuitively, you'd think this should work.  But it doesn't, the Card `header` is a "component part".  Semantically, it is the header part of the Card component.  The lack of the `ui` class is supposed make this distinction.  Since it is not a `ui header` component, it does not support h\* tags (nor other `ui header` component features like color, size, etc):

![http://g.recordit.co/MnogaDygxZ.gif](http://g.recordit.co/MnogaDygxZ.gif)

> _Notice how different h\* tags only change the bottom margin, but not the font size._

If a user would like to use `ui header` "component" features (h\* tags, size, color, etc.), they would have to forgo the Card.Header "component part" and use a Header instead:

``` jsx
<Card>
   <Header.H1>Large Card Header</Header.H1>
</Card>
```

While a user can do this, it is not supported markup and does break certain layouts. Look what happens when a `ui header` component is used in a Card with an image:

![http://g.recordit.co/aAfLgdpGoI.gif](http://g.recordit.co/aAfLgdpGoI.gif)

> _H\* tag sizing works, but the top margin breaks.  The image has to be deleted in order for the Header to layout correctly._
## Updated Parts Proposal

My proposal is that we standardize all "component parts", but that we should _not reuse "components" with the same name_.  This is because they do not share propTypes/features/markup.  We would also have to strip the `ui` class.
### Analysis

I again reviewed markup for all SUI components, these "component parts" are shared by many SUI components:

`actions` `content` `description` `divider` `extra` `header` `item` `menu` `meta` `label` `value`

> _Though there are also Header, Item, Menu, and Label "components" these are not the same as the "component parts".  They merely share the same name._

[...]

PR #335, opened 2016-07-15
Add Step component

### LINKED ISSUE #183: <Step /> Component

layershifter:
# API
### Step

A single step

``` html
<div class="ui steps">
  <div class="step">
    Shipping
  </div>
</div>
```

``` js
<Step.Group>
  <Step>Shipping</Step>
</Step.Group>
// or
const items = [
  {text: 'Shipping'}
];

<Step.Group steps={items}>
```
### Steps

A set of steps

``` html
<div class="ui steps">
  <div class="step">
    <i class="truck icon"></i>
    <div class="content">
      <div class="title">Shipping</div>
      <div class="description">Choose your shipping options</div>
    </div>
  </div>
  <div class="active step">
    <i class="payment icon"></i>
    <div class="content">
      <div class="title">Billing</div>
      <div class="description">Enter billing information</div>
    </div>
  </div>
  <div class="disabled step">
    <i class="info icon"></i>
    <div class="content">
      <div class="title">Confirm Order</div>
    </div>
  </div>
</div>
```

``` js
<Step.Group>
  <Step icon='truck'>
    <Step.Title>Shipping</Step.Title>
    <Step.Description>Shipping</Step.Description>
  </Step>
  <Step active icon='payment'>
    <Step.Title>Billing</Step.Title>
    <Step.Description>Confirm Order</Step.Description>
  </Step>
  <Step disabled icon='info' title='Confirm Order' />
</Step.Group>
// or
const items = [
  {icon: 'truck', title: 'Shipping', description: 'Shipping'},
  {active: true, icon: 'payment', title: 'Billing', description: 'Confirm Order'},
  {disabled: true, icon: 'info', title: 'Confirm Order'},
];

<Step.Group steps={items}>
```
### Ordered

A step can show a ordered sequence of steps

``` html
<div class="ui ordered steps">
  <div class="completed step">
    <div class="content">
      <div class="title">Shipping</div>
      <div class="description">Choose your shipping options</div>
    </div>
  </div>
...
</div>
```

``` js
<Step.Group ordered>
  <Step completed>
    <Step.Title>Shipping</Step.Title>
    <Step.Description>Shipping</Step.Description>
  </Step>
</Step.Group>
// or
const items = [
  {completed: true, title: 'Billing', description: 'Confirm Order'},
];

<Step.Group ordered steps={items}>
```
### Vertical

A step can be displayed stacked vertically

``` html
<div class="ui vertical steps">...</div>
```

``` js
<Step.Group vertical>...</Step.Group>
<Step.Group vertical steps={items}>
```
### Description

A step can contain a description

_Already described_
### Icon

A step can contain an icon

_Already described_
### Link

A step can link

``` html
// Case 1
<div class="link step"></div>
// Case 2
<a class="active step">...</a>
```

``` js
// Case 1
<Step link>...</Step>

// Case 2
<Step active href='http://google.com'>...</Step>
<Step active onClick={handler}>...</Step>
```
### Active

A step can be highlighted as active

_Already described_
### Completed

A step can show that a user has completed it

_Already described_
### Disabled

A step can show that it cannot be selected

``` html
<div class="disabled step"></div>
```

``` js
<Step disabled>...</Step>
```
### Stackable

A step can stack vertically only on smaller screens
### Vertical

A step can be displayed stacked vertically

``` html
<div class="ui tablet stackable steps">...</div>
```

``` js
<Step.Group stackable='tablet'>...</Step.Group>
<Step.Group stackable='tablet' steps={items}>
```
### Fluid

A fluid step takes up the width of its container

``` html
<div class="ui fluid steps">...</div>
```

``` js
<Step.Group fluid>...</Step.Group>
```
### Attached

Steps can be attached to other elements

``` html
<div class="ui top attached steps">...</div>
```

``` js
<Step.Group attached='top'>...</Step.Group>
```
### Evenly Divided

Steps can be divided evenly inside their parent

``` html
<div class="ui two steps"></div>
```
#### Equal Width

``` jsx
<Step.Group widths='equal'>
  <Step />
  <Step />
</Step.Group>
<div class="ui two steps">
  <div class="step"></div>
  <div class="step"></div>
</div>
```
#### Explicit Widths

``` jsx
<Step.Group widths={2}> // or widths='two'
  <Step />
  <Step />
</Step.Group>
<div class="ui two steps">
  <div class="step"></div>
  <div class="step"></div>
</div>
```
### Size

Steps can have different sizes

``` html
<div class="ui mini steps">...</div>
```

``` js
<Step.Group size='mini'>...</Step.Group>
```

Fixes #183


src/elements/Step/Step.js
```diff
@@ -19,26 +19,31 @@ function Step(props) {
...
+  const getChildren = () => {
+    if (children) return children
+
+    return [<Step.Title>{title}</Step.Title>, <Step.Description>{description}</Step.Description>].map((item) => item)
+  }
+  const getContent = () => {
+    if (!icon) return <div className='content'>{getChildren()}</div>
+
+    return [iconPropRenderer(icon), <div className='content'>{getChildren()}</div>].map((item) => item)
+  }
```
layershifter: @levithomason may be cleaner solution?

levithomason: Hm, can you elaborate a bit on the goal here?  One thing I notice is that the `.map((item) => item)` looks to be redundant since it just returns every element in the array without modification:

image

layershifter:
Refactored this fragment, it might be more understantable now :+1:

levithomason:
I see, it looks like we're abstracting away the `Step.Content` by creating this automatically.  I'm on the fence about this one still.  I'll have to give this some more thought.  Here are my current thoughts.

There are many components that have `content`.  Usually, `content` also can contain `header`, `description`, `meta`, and `extra` component part children.  It can also sometimes have additional definition classes, like `extra content` (Card), `image content` (Modal), `bottom aligned content` (Item).

Because of the need to sometimes configure content, and given our sub component API pattern, we may likely end up supporting Content parts something like this:

``` jsx
<Card>
  <Card.Content>...</Card.Content>
  <Card.Content extra>...</Card.Content>
</Card>

<Modal>
  <Modal.Content>...</Modal.Content>
  <Modal.Content image>...</Modal.Content>
</Modal>

<Item>
  <Item.Content>...</Item.Content>
  <Item.Content aligned='bottom'>...</Item.Content>
</Item>
```

This would mean for the Step:

``` jsx
<Step>
  <Step.Content>...</Step.Content>
</Step>
```

Again, I'm open for ideas and input here but this is where I'm leaning so far.

**EDIT**

To be clear, `content` would be auto created for you when using the props API for creating your component.  However, if we go with the above sub component API for Content, you would have to manually wrap your children in Content when appropriate.  In most cases, anytime there is markup that is adjacent to the content (an icon, image, etc) then the Content wrapper is required around the header, description, etc.  But, if there is no icon/image/etc that is adjacent to the header/description/etc, then there is no content wrapper required.  See semantic-ui.com/views/item.html example markup for more.

> Linking #325

layershifter:
@levithomason Seems it make code cleaner, I'll refactor `<Content>` to separate component

levithomason:
Sounds good

layershifter:
**Updated**, I undertand my miss ^_^

layershifter:
See last commit, it's working solution, but it looks like :hankey:  Is there another solution for inheritance?

levithomason:
I'll pull this and take a look.

I've got a branch going to address #325 reusable component parts.  I'll open that PR so that we can further hash that out there.  This will require some updates to how the docs are created as well as some testing considerations.

My intuition says we don't hold up current PRs for reusability optimizations.  I think we should continue merging these with current pattern.  Then, we can let the component parts PR handle any refactoring.

All said, I'll look at this now and also open the component parts PR in its current state.

src/elements/Step/Step.js
```diff
@@ -0,0 +1,103 @@
...
+      className,
+      'step',
+    )
+    const handleClick = (e) => {
+      if (onClick) onClick(e)
+    }
+    const rest = getUnhandledProps(Step, this.props)
+
+    // if (href) return <a {...rest} className={classes} href={href} onClick={handleClick}>{getChildren(props)}</a>
+    // if (onClick) return <a {...rest} className={classes} onClick={handleClick}>{getChildren(props)}</a>
```
eanplatter: Are these intentional?

levithomason:
I've opened #345, all reusable component part conversation should be moved there.  Let's not implement reusable component parts until a verdict is reached in that PR.

Pulling and taking a look at the solution started here as well now.

layershifter:
> Let's not implement reusable component parts until a verdict is reached in that PR.

OK, wait for pull, I will update components

layershifter:
Made cleanup, take a look. I need to add tests and docs for `<Step.Content>`

docs/app/Examples/elements/Step/Content/StepLinkExamples.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Step } from 'stardust'
 
 // eslint-disable-next-line no-alert
-const handleClick = () => alert('Clicked!')
+const handleClick = (e) => {alert('Clicked!'), console.log(e.target)}
```
levithomason: Would prefer we avoid `alert()` for examples. We've been using stateful examples when handling interactivity like this.  We update the example with some kind of output to show what is going on.  I think it is a cleaner representation, but also more complete in that it shows users how to fully integrate the component with their own view.  Here are some examples:

### Radio Groups

http://technologyadvice.github.io/stardust/addons/radio#radio-group

### Table Selectable

http://technologyadvice.github.io/stardust/collections/table#selectable

### Accordion activeIndex

http://technologyadvice.github.io/stardust/modules/accordion#active-index

### Dropdown

http://technologyadvice.github.io/stardust/modules/dropdown#async-options

src/elements/Step/Step.js
```diff
@@ -9,43 +9,36 @@ import StepGroup from './StepGroup'
...
+    useKeyOnly(link, 'link'),
+    className,
+    'step',
+  )
+  const rest = getUnhandledProps(Step, props)
+
+  const content = [
+    icon && iconPropRenderer(icon),
+    <StepContent {...{ children, description, title, ...rest }} />,
+  ]
```
levithomason: Most `content` component parts are only required if there is an adjacent icon or image.  Checking the Step markup, it looks like content is not included when there is [no icon](http://semantic-ui.com/elements/step.html#step).  This includes when there is only a [title and description](http://semantic-ui.com/elements/step.html#description).  In some cases, adding the `content` without an adjacent icon/image will break layouts.

Our current API spec does not allow mixing children when using smart props for markup.  This is because smart props generate children, including conditional `content` and other hierarchy.  There is then no sure fire way for us to know where to insert the user's children to prevent issues. So, the API here would  support one of these two cases but not both:

``` jsx
<Step
  icon='truck'
  title='Shipping'
  description='Choose your shipping options'
/>

<Step>
  <Icon name='truck' />
  <Step.Content>
    <Step.Title>Shipping</Step.Title>
    <Step.Description>Choose your shipping options</Step.Description>
  </Step.Content>
</Step>
```

Lastly, the array children will need unique `key` props.  This might end up looking something more like this:

``` jsx
const content = icon ? [
  iconPropRenderer(icon),
  (title || description) && <StepContent key='content' {...{ description, title }} />,
] : [
  title && <StepTitle key={'title'} />,
  description && <StepDescription key={'description'} />,
]
```

---

As an aside, I'm on the fence with allowing children as the "primary content" of a component while also using smart props.  This has been considered (and used in the past, Item.js) but under our current spec is not supported.  I haven't decided yet if it is good to support this, or if it just adds confusion and complexity.

Here is what it could look like:

``` jsx
// title provided, use children as "primary content" (i.e. description)
<Step icon='truck' title='Shipping'>
  Choose your shipping options
</Step>
```

This looks nice and saves on markup, but would require a lot more conditional logic and propTypes to be done correctly.  For now, I think we should wait on it and just go with the very clear and explicit "use only smart props or use only children" approach.

layershifter:
@levithomason 
It seems, that it will not look so pretty :crying_cat_face: See, last commit.
Also, seems `ordered` prop might be passed from `<StepGroup>`. What about [Context](https://facebook.github.io/react/docs/context.html)? But, `<Step>` must be class for this.

image

levithomason:
Checking

levithomason:
Here's a solution without context.  The main change I made to get it working was to pass the `ordered` prop down from the StepGroup to each Step (for both props and children):

### StepGroup.js

``` jsx
import { mapChildType } from '../../utils/childrenUtils'

// render()

  if (items) {
    return (
      <div {...rest} className={classes}>
        {items.map((item, index) => <Step key={index} {...item} ordered={ordered} />)}
      </div>
    )
  }

  return (
    <div {...rest} className={classes}>
      {mapChildType(children, Step, (step, index) => (
        <Step {...step.props} key={index} ordered={ordered} />
      ))}
    </div>
  )

// propTypes

  children: customPropTypes.all([
    customPropTypes.mutuallyExclusive(['description', 'title']),
    customPropTypes.ofComponentTypes(['Step']),    // <-- only allow Step children
    PropTypes.node,
  ]),
```

### Step.js

Tha main change I made here was to use the ordered prop to wrap content.  I reworked this component a bit to clean it up as well:

``` jsx
function Step(props) {
  const {
    active, className, children, completed, description, disabled, icon, href, link, onClick, ordered, title,
  } = props

  // CHANGES: removed ui className
  const classes = cx(
    useKeyOnly(active, 'active'),
    useKeyOnly(completed, 'completed'),
    useKeyOnly(disabled, 'disabled'),
    useKeyOnly(link, 'link'),
    className,
    'step',
  )
  const rest = getUnhandledProps(Step, props)

  const handleClick = (e) => {
    if (onClick) onClick(e)
  }

  // CHANGES: remove function, use assignment, wrap if ordered
  const contentJSX = completed || icon || ordered ? [
    icon && iconPropRenderer(icon, { key: icon }),
    // CHANGES: remove children ? logic, StepContent handles this
    <StepContent key='content' title={title} description={description}>
      {children}
    </StepContent>,
  ] : [
    // CHANGES: include children + title/desc, they cannot be used togehter
    // so this will result in only one or the other
    children && children,
    title && <StepTitle key='title' title={title} />,
    description && <StepDescription key='description' description={description} />,
  ]

  // CHANGES: remove duplicate JSX return statements in favor
  // of a conditional component tagName
  const StepComponent = href || onClick ? 'a' : 'div'

  // CHANGES: use a single return
  // always apply href, if it is undefined, the prop will not exist
  return (
    <StepComponent
      {...rest}
      className={classes}
      href={href}
      onClick={handleClick}
    >
      {contentJSX}
    </StepComponent>
  )
}
```

image

levithomason:
As a heads up, I'm headed out for a vacation at the moment.  I'll be back next week.  See you then!

src/elements/Step/StepGroup.js
```diff
@@ -48,6 +57,7 @@ StepGroup.propTypes = {
   /** Primary content of the StepGroup. Mutually exclusive with items prop. */
   children: customPropTypes.all([
     customPropTypes.mutuallyExclusive(['description', 'title']),
+    customPropTypes.ofComponentTypes(['Step']),    // <-- only allow Step children
```
levithomason: The inline comment can be removed here, this was just to point it out in the PR comment.

layershifter:
Awesome work on Step :+1:

src/elements/Step/StepGroup.js
layershifter: Yep, copypaste is evil :smile:

layershifter:
I think it's ready for review :)

---

I've updated `StepLinkExamples.js`, as you can see example with onClick is non-working. Take a look, it seems usage of `mapChildType` in `<StepGroup/>` might be updated.

layershifter:
@levithomason seems there is only one problem to solve, waiting for your opinion

layershifter:
ping @levithomason

levithomason:
Taking a look!

levithomason:
Would you mind rebasing this PR to the latest master?  It would help with the review.

layershifter:
Yes, give me five minutes

layershifter:
Made :yum:

levithomason:
Thanks much, looking at Feed right now.  Will get to this next.

levithomason:
I think your fork needs to update it's master to the upstream master.  I'm still able to review now however.

levithomason:
## Fixing Links

There 2 are changes necessary to fix the links example:

**1. StepLinkExamples.js**

Spread props on the `ClickableStep` so the icon, title, and description are passed.

``` diff
-    return <Step active={this.state.active} onClick={this.handleClick} />
+    return <Step {...this.props} active={this.state.active} onClick={this.handleClick} />
```

**2. StepGroup.js**

Because `mapChildType` will only map `Step`, the `ClickableStep` components are filtered out.  I've made a few changes to the `StepGroup` to fix this and some other things:

``` diff
-if (items) {
-  return (
-    <div {...rest} className={classes}>
-      {items.map((item, index) => <Step key={index} {...item} ordered={ordered} />)}
-    </div>
-  )
-}
-
-return (
-  <div {...rest} className={classes}>
-    {mapChildType(children, Step, (step, index) => (
-      <Step {...step.props} key={index} ordered={ordered} />
-    ))}
-  </div>
-)
+const content = items
+  ? items.map((item, index) => <Step key={index} {...item} ordered={ordered} />)
+  : children
+
+return <div {...rest} className={classes}>{content}</div>
```

## Ordered Updates

Since we no longer map Steps in the Group and add the ordered prop, we need to handle this now.  The Step looks like it was only using `ordered` to wrap the Step title/description in `content`.  I think we just go with your earlier proposal and always wrap title/description _props_ in content.  If the user uses subcomponent children, they'll need to wrap it themselves:

**Step.js**

``` diff
const {
-  active, className, children, completed, description, disabled, icon, href, link, onClick, ordered, title,
+  active, className, children, completed, description, disabled, icon, href, link, onClick, title,
} = props

-const contentJSX = completed || icon || ordered ? [
-  icon && iconPropRenderer(icon, { key: icon }),
-  children && <StepContent key='content' children={children} />,
-  (title || description) && <StepContent key='content' description={description} title={title} />,
-] : [
-  children && children,
-  title && <StepTitle key='title' title={title} />,
-  description && <StepDescription key='description' description={description} />,
-]
 const StepComponent = href || onClick ? 'a' : 'div'

 return (
   <StepComponent
     {...rest}
     className={classes}
     href={href}
     onClick={handleClick}
   >
-     {contentJSX}
+     {!children && iconPropRenderer(icon)}
+     {children || <StepContent description={description} title={title} />}
   </StepComponent>
 )
```

For consistency, simplicity, and due to other component conflicts, the `children` will always be mutually exclusive with props shorthand for v1:

``` diff
  /** A step can contain an icon. */
-  icon: PropTypes.string,
+  icon: customPropTypes.all([
+    PropTypes.node,
+  ]),
```

We could technically allow `<Step icon='truck'><Title>Shipping</Title></Step>` by wrapping children in `content` when there is an `icon` prop. Though I want to ship with consistency for v1 and there are components that this is much more difficult or doesn't make sense for.  Lastly, after shipping a more straightforward v1 prop API, we can always release non-breaking updates to allow mixing prop markup APIs and subcomponent APIs.  Though, we cannot do the same in reverse.

## Update Examples

With the above changes, we'll need to use either the props API or subcomponent API, but not both simultaneously.  We also need to include the `content` component when there is an icon and using subcomponents.  Example:

**StepGroupExample.js**

``` diff
-import { Step } from 'stardust'
+ import { Icon, Step } from 'stardust'

-<Step icon='truck'>
+<Step>
+  <Icon name='truck' />
+  <Step.Content>
    <Step.Title>Shipping</Step.Title>
    <Step.Description>Choose your shipping options</Step.Description>
+  </Step.Content>
</Step>
```

---

With all these changes, everything seems to work in the docs and components.  It also seems more simple, which is good.  LMK how the updates go!

layershifter:
Thanks for review.

## Fixing Links

### 1. StepLinkExamples.js

Agree, it works.

### 2. StepGroup.js

``` jsx
items.map((item, index) => <Step key={index} {...item} ordered={ordered} />)
```

Antipattern with `key` is there :fire_engine: I think about extended validation of `items` prop with `PropTypes.shape` and form key from them:

``` js
const {description, title} = item
const key = `${title}-${description}`
```

## Ordered Updates

I don't like these updates because they will break this nice layout:

``` jsx
<Step icon='truck'>
  <Title>Shipping</Title>
  <Description>Choose your shipping options</Description>
</Step>
```

May be we can use [React.Content](https://facebook.github.io/react/docs/context.html) for passing `ordered` prop to children. Is any reason to reject this solution (except, `StepGroup` will cease to be stateless)?

levithomason:
> ...I think about extended validation of items prop with PropTypes.shape and form key from them

That is much better.  We can allow them to pass a key in the items array.  How about something like this (accounts for undefined title/description):

``` jsx
items.map((item) => {
  const key = item.key || [item.title, item.description].join('-')
  return <Step key={key} {...item} />
})
```

> I don't like these updates because they will break this nice layout...

**TL;DR at the bottom 😄**

The markup Step > Title, Description is definitely cleaner than Step > Content > Title, Description.  However, consider other components with `content`.  The `content` wrapper is usually required anytime there is an adjacent `icon` or `image`:

``` html
<h2 class="ui header">Account Settings</h2>

<h2 class="ui icon header">
  <i class="settings icon"></i>
  <div class="content">
    Account Settings
  </div>
</h2>
```

``` html
<div class="ui card">
  <div class="image">
    <img src="/images/avatar2/large/kristy.png">
  </div>
  <div class="content">...</div>
</div>
```

``` html
<div class="ui comments">
  <div class="comment">
    <a class="avatar">
      <img src="/images/avatar/small/joe.jpg">
    </a>
    <div class="content">... </div>
  </div>
</div>
```

At first, it seems safe to abstract away the `content` wrapper based on the presence of an `icon` or `image` prop.  Though, we'd have to accept some limitations.  The user will never be able access the `content` wrapper nor insert any markup between the root node and the `content` node.  We are assuming then that our API will cover all possible use cases of markup between the root node and `content`.  We are also assuming that our props API can handle all use cases for props on components between this space.  We are removing access to this part of the render tree as users are limited to passing nodes through the `icon`/`image` props.

I see three places this could break down and cause issues. First, if a user needed to pass something more complicated than a basic icon name or image src through the props then it could get ugly and painful.  Second, I think it would be an issue if only some of our components allowed use of `icon`/`image`/etc prop with `children` and _automatically_ inserted the `content` wrapper, while other components did not allow `icon`/`image`/etc with `children` and required the user to _manually_ insert the `content` wrapper. I think consistency in the library should take priority over convenience.  Last, if the user ever wanted/needed to add props to the `content`, then it could be an issue as we'd have to also accept content props through a component prop.

I think all of these issues are already present. At TA we had a fair number of use cases where we had to pass complicated components through shorthand props that fully abstracted away access to parts of the render tree.  It made for very ugly code and markup issues that would not be present if we had full access and control to the component markup.  

We also have components in the library with `content` parts that cannot be easily configured with only an `icon`/`image` prop value.  Modal `content` is adjacent to a `header`.  Imagine the user wants to add a header `close icon` with a click handler to close the modal.  (We had a similar issue at TA with passing markup through `<ListItem icon='...' />`)

``` jsx
const header = (
  // extra node to wrap text :( sometimes breaks styles
  <span>
    Here is my header
    <Icon style={closeIconStyle} name='close icon' onClick={this.hideModal} />
  </span>
)

<Modal header={header}>...</Modal>

<div class="ui modal">
  <div class="header">
    <span>
      Here is my header
      <Icon name='close icon' />
    </span>
  </div>
  <div class="content">...</div>
</div>
```

Whereas, if we did not abstract away access to the `content` node, then the user could have simply added the Icon.  Also, the generated markup does not require any extra `span` node.  (The same is true in the case of a `ListItem` with extra markup in the icon/image part of the tree):

``` jsx
<Modal>
  <Modal.Header>
    This is my header
    <Icon style={closeIconStyle} name='close icon' onClick={this.hideModal} />
  </Modal.Header>
  <Modal.Content>
    ...
  </Modal.Content>
</Modal>

<div class="ui modal">
  <div class="header">
      Here is my header
      <Icon name='close icon' />
  </div>
  <div class="content">...</div>
</div>
```

There are also components which take props on the content.  Like [vertically aligned lists](http://semantic-ui.com/elements/list.html#vertically-aligned), modals with [image content](http://semantic-ui.com/modules/modal.html#image-content), [animated buttons](http://semantic-ui.com/elements/button.html#animated), and probably others.  These could be hoisted up to the top level, but this again is making the assumption that there will never be conflicting props between the top component and the content.

# TL;DR

In conclusion, _completely_ abstracting away `Content` offers cleaner markup for simple use cases.  Anything beyond simple use causes problematic and extra complicated markup.  The `Content` subcomponent API is more verbose, but does not become problematic or complicated for use cases beyond the simple.  Mixing the props API with the subcomponent API removes the benefits of the subcomponent API while adding the issues of the props API.

Because of this, I think our first version should offer the props API for simple use cases, offer the subcomponent API for use cases beyond the smiple, and make these two mutually exclusive.

---

P.S.

Personally, I like mixing the props API and sub component APIs to abstract away `Content`.  I'd be very happy to see a solution that does this while also solving the above noted issues.  My colleagues and I have not yet figured this out and I haven't seen a proposal that does.  I would accept an API proposal that does this.  If not, I hope it is figured out for v2.

layershifter:
@levithomason thanks for a detailed point of view :+1:

levithomason:
No problem, explaining helps me to better understand it myself!

levithomason:
Once the step/examples are updated, I think we can merge this one

layershifter:
I've updated all, so let's go :car:

levithomason:
Woop!  I'll make some doc updates soon.  Need to remove `ordered` from `Step` propTypes.  I also started adding descriptions to the props API variant of the docs per [these comments](https://github.com/TechnologyAdvice/stardust/pull/344#discussion_r74088351).

PR #338, opened 2016-07-18
Cleanup docs for Loader

layershifter:
Clean up for `<Loader>` docs:
- examples are stateless;
- examples were renamed;
- examples looks better :+1:


levithomason:
Great, pulling and trying this out now.

docs/app/Examples/elements/Loader/Variations/LoaderInlineCenteredExample.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react'
+import { Loader } from 'stardust'
+
+export default () => <Loader active inline='centered' />
```
levithomason: Just for consistency, perhaps we should assign these to a const and export that.

docs/app/Examples/elements/Loader/Variations/LoaderInlineExample.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react'
+import { Loader } from 'stardust'
+
+export default () => <Loader active inline />
```
levithomason: Same here, regarding const assignment/export.

levithomason:
Two minor comments, looks great otherwise.

PR #342, opened 2016-07-19
Fixed broken reference to utils/Meta

tslater:
I ran the build and the tests. This should fix the problem I'm having.


levithomason:
👍

PR #344, opened 2016-07-21
Update Message to v1 API

levithomason:
This PR updates the Message component to our v1 API and removes jQuery.  Running the doc site and looking at the example usages is a big help to reviewing.


docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -52,10 +52,10 @@ export default class ComponentExample extends Component {
         <Grid.Column>
           <Grid>
             <Grid.Column width={12}>
-              <Header.H3 style={{ marginBottom: 0 }}>
+              {this.props.title && <Header.H3 style={{ marginBottom: 0 }}>
                 {this.props.title}
-              </Header.H3>
-              <p>{this.props.description}</p>
+              </Header.H3>}
+              {this.props.description && <p>{this.props.description}</p>}
```
levithomason, line 58: Some examples do not have a `title` or `description` as they are under the title/desc of the example immediately above it.  We needed to hide the title to prevent empty anchor links from showing up.  Also, the description created extra unnecessary whitespace.

docs/app/Examples/collections/Message/Types/MessageIconExample.js
```diff
@@ -1,18 +1,14 @@
...
-          Get the best news in your e-mail every day.
-        </Message>
+const MessageIconExample = () => (
+  <Message>
+    <Icon name='notched circle' loading />
+    <Message.Content>
+      <Message.Header>Just one second</Message.Header>
+      We are fetching that content for you.
+    </Message.Content>
+  </Message>
```
levithomason: Icon messages require an `icon` class.  We look for an `<Icon />` child and add this class.  I'm not 100% pleased with this as using another icon element or decorated `<Icon />` will not work.  The alternative is to require the icon prop:

```jsx
<Message icon>
  <Icon name='info' />
<Message>
```

This is also odd.  It is awkward in light of the intended usage of the icon prop:

```jsx
<Message icon='info' />
```

In this use case we add the icon className.  Since the prop API for markup is mutually exclusive children, allowing both a bool `icon` prop and children seems like something we shouldn't do.

### TL;DR

I'm not 100% happy with how we're adding the icon class, but I think it is the least of the evils.  Feedback welcome.

docs/app/Examples/collections/Message/Types/MessageListExample.js
```diff
@@ -0,0 +1,14 @@
...
+import { Message } from 'stardust'
+
+const MessageListExample = () => (
+  <Message>
+    <Message.Header>New Site Features</Message.Header>
+    <Message.List>
+      <Message.List.Item>You can now have cover images on blog pages</Message.List.Item>
+      <Message.List.Item>Drafts will now auto-save while writing</Message.List.Item>
+    </Message.List>
+  </Message>
```
levithomason, line 11: This is super ugly and painful.  Alternatives are being discussed.

src/collections/Message/Message.js
```diff
@@ -1,66 +1,198 @@
...
-        {this.props.dismissable && closeIcon}
-        {this.props.icon && icon}
-        {this.props.icon && content}
-        {!this.props.icon && this.props.header && header}
-        {!this.props.icon && this.props.children && children}
-      </div>
-    )
-  }
+  /** Primary content.  Mutually exclusive with children. */
+  content: PropTypes.node,
```
dvdzkwsk: Since this content is wrapped in a `p` tag, should it be restricted to `string`, and if they need to control the markup children should be used instead?

levithomason: Yes it should, forgot to change this back.  Fixing.

src/collections/Message/Message.js
```diff
@@ -1,66 +1,198 @@
...
+    useKeyOnly(error, 'error'),
+    'message',
+    className,
+  )
 
-  static _meta = {
-    name: 'Message',
-    type: META.type.collection,
-  }
+  const dismissIcon = (dismissable || onDismiss) && <Icon name='close' onClick={onDismiss || _.noop} />
```
dvdzkwsk: Should `dismissable` even be a thing if it's automatically dismissible when the handler is provided? It's noop with just `dismissible`, which doesn't seem to make much sense.

dvdzkwsk:
Looks good to me, left just a couple comments. My only other comment is that, given how almost every property is applied, it would make more sense to write `<Message type='warning' />` over `<Message warning />`, as well as conceivably improve prop validation, so take that for what it's worth.

docs/app/Examples/collections/Message/Variations/MessageInfoPropsExample.js
```diff
@@ -0,0 +1,12 @@
...
+
+const MessageInfoPropsExample = () => (
+  <Message
+    info
+    header='Was this what you wanted?'
+    content="Did you know it's been a while?"
+  />
+)
+
+export default MessageInfoPropsExample
```
kyleturco, line 12: Without inspecting the code snippet, this example (MessageInfoPropsExample) and the previous (MessageInfoExample) look identical. I think it would be good to add either a title to each example or have the text within the example acknowledge the difference. On first scroll, it just looks a bug in the docs.

docs/app/Examples/collections/Message/Variations/MessageWarningPropsExample.js
```diff
@@ -0,0 +1,12 @@
...
+
+const MessageWarningPropsExample = () => (
+  <Message
+    warning
+    header='You must register before you can do that!'
+    content='Visit our registration page, then try again.'
+  />
+)
+
+export default MessageWarningPropsExample
```
kyleturco, line 12: Same visual issue as the `MessageInfoExample` and `MessageInfoPropsExample`

src/collections/Message/Message.js
levithomason: I think that makes sense.  This pattern also exists in the Label (removeable onRemove), so we can update it there as well.

See #384

levithomason: FWIW I toyed with the idea of allowing auto controlled state here to toggle the `hide` prop.  In this case, it would make sense to just say "dismissable" and not handle clicks yourself. However, it seemed like a bad practice to let users hide Messages but leave them in memory.  I think it is better to manage a message stack yourself somewhere.

docs/app/Examples/collections/Message/Variations/MessageInfoPropsExample.js
levithomason, line 12: Valid point.  Without the code snippet showing, it is not clear what the difference is.  I'll add some description text.

levithomason:
> Looks good to me, left just a couple comments. My only other comment is that, given how almost every property is applied, it would make more sense to write <Message type='warning' /> over <Message warning />, as well as conceivably improve prop validation, so take that for what it's worth.

Agreed.  We originally did not add a `state` or `type` prop as the first few components we handled were capable of multiple SUI "types" like:

``` jsx
<Dropdown multiple search selection />
<Button inverted basic icon />
```

Also "states" 

``` jsx
<Dropdown disabled loading />
<Button disabled loading />
```

Looking back, there are some props that we can be reasonably sure that they will be forever mutually exclusive (i.e. warning + info).  These probably should be pulled into a single prop key.  I'm not exactly sure what a good key name would be however since `type`, `state`, and `variation` are defined in the SUI glossary to have a particular meaning in the framework.  Here's how the Message classes are broken down:

Types: list, icon, dismisable
States: hidden, visible
Variations: floating, compact, attached, warning, info, positive, success, negative, error, color, size

In order to better validate the props, we'd need to consider the color as well.  Since, you cannot have a `red positive` or `green error` message.  I think to really validate and restrict the component to proper use we'd have to put all these props values into the `color` prop:

``` jsx
<Message color='red' />
<Message color='error' />
<Message color='negative' />

<Message color='green' />
<Message color='success' />
<Message color='positive' />
```

Thoughts?

dvdzkwsk:
In adherence to Semantic, your `color` proposal makes sense. In reality, it would take some adjusting to treat `success` not as a variation or a type, but a color. The benefit of your idea, however, is that it ultimately becomes more intuitive over time; there'd be no need to remember if something is a variation, type, or color.

Still, I'm leaning more toward keeping type/variation separate from color, just because color is such a distinct and targeted property (after all, `warning` is not in fact a color). I honestly don't know where to draw the line, because you could apply that same argument to separating `type` and `variation`, but, personally, I find that distinction less meaningful for the end user than the API developer. Meaning, `color` makes a lot of sense to a user: "oh, ok, I put the actual name of a color here". On the other hand, the distinction, as a user, between `variation` and `type` doesn't carry as much weight. Kinda feels like a "why should I care?" type of thing.

As far as `state`, that still makes sense as standalone props, since it generally resembles common attributes on vanilla HTML (`<input type="text" disabled />`).

Just for kicks, here's how I'd approach writing a `<Message />` component if not given any direction besides what props to apply.

``` js
<Message type='success' floating hidden />
```

Yes, I could, in theory, make the mistake of adding both `type` and `color`, but I don't think it's necessarily valid to design an API around this scenario since `success`, `warning`, et al. have long been stand-ins for explicit color assignments.

levithomason:
Cool, I think we should implement something like this. Here are some more considerations before doing so.  What about positive/negative which are closely related to success/error?  Using `type`:

``` jsx
<Message type='positive' />
<Button type='negative' />
```

Semantic UI calls primary/secondary, success/positive, error/negative "emphasis" in the docs.  So it could look something like this:

``` jsx
<Message emphasis='positive' />
<Button emphasis='negative' />
```

The `type` prop would then break down with the [Segment](http://semantic-ui.com/elements/segment.html) which has SUI "raised", "stacked", "piled", and "vertical" types.  If the "type" prop is extended to "positive/negative/primary/secondary" then the other mutually exclusive types will need to be bool or have a different key.

``` jsx
<Segment stacked type='secondary' />
// vs
<Segment type='stacked' emphasis='tertiary' />
```

I think `type` also breaks down for the [Progress](http://semantic-ui.com/modules/progress.html#active) component, where active/success/warning/error/disabled are all states:

``` jsx
<Progress type='error' />
// vs
<Progress error />
```

All said, rather than trying to make all props mean the same thing for all components, there is also the possibility that it is more intuitive to have props mean different things on different components. Example:

``` jsx
<Message type='(success|info|warning|error)' />
<Progress state='(success|warning|error)' active disabled />
<Button emphasis='(primary|secondary|positive|negative)' />
<Segment emphasis='(secondary|tertiary)' />
```

Lastly, the other serious consideration is parity with the core framework.  I wonder if it will just add more confusion to have the SUI docs showing `types`, `states`, `variations` only to have SUI-React map these values to different names.  Especially given that the SUI docs are linked from our docs for every component.

The flip side here being, the SUI components only use className.  All other groupings are strictly for doc site headings.  Our use is quite different, we are defining components, not necessarily doc site headings.

docs/app/Examples/collections/Message/Variations/MessageWarningPropsExample.js
levithomason, line 12: Done, added descriptions above each.

image

kyleturco, line 12: Nice! Looks good to me

levithomason:
We had some further offline discussion on success/info/warning/error.  Gonna merge as bools for now.  We'll continue to think this through as we approach v1.

PR #345, opened 2016-07-22
perf(parts): add reusable component parts

### LINKED ISSUE #325: RFC: Standardize Component Parts

levithomason:
## Experiment - Fixes #325

This PR addresses the issue of reusable component parts.  These are parts of a component that do not have the `ui` className.  See #325 for more detail.

Please offer any comments, ideas, or flames as I think any input on this would be helpful.


src/ComponentParts.js
```diff
@@ -0,0 +1,72 @@
...
+ * They do not share capabilities. They only share the same name.
+ */
+import _ from 'lodash'
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import META from './utils/Meta'
+import { getUnhandledProps } from './utils/propUtils'
+
+function makePart(partName) {
```
levithomason: See PR description, this is the component part factory.

src/ComponentParts.js
```diff
@@ -0,0 +1,72 @@
...
+    children: PropTypes.node,
+
+    /** Classes to add to className. */
+    className: PropTypes.string,
+  }
+
+  return ComponentPart
+}
+
+export const Actions = makePart('Actions')
```
levithomason: This is how the factory exports reusable parts.

src/make_parts.zsh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env zsh
+
+rm parts/*
+files=(Actions Content Description Divider Extra Header Item Menu Meta Label Value Avatar Bar Date Event ImageContent Summary Title)
+
+for file in $files; do
```
levithomason: This was the original script that generated full component part files.  This is now done away with by the component part factory above

src/modules/Accordion/AccordionContent.js
```diff
@@ -1,38 +1,34 @@
 import React, { PropTypes } from 'react'
 import cx from 'classnames'
 
+import Content from '../../parts/Content'
```
levithomason: This is an example implementation of a reusable component part, created by the factory.

src/parts/Actions.js
```diff
@@ -0,0 +1,35 @@
+import cx from 'classnames'
```
levithomason: All files in `src/parts/*` are the output of the component part script.  You can ignore these since that was replaced by the component factory function.

src/make_parts.zsh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env zsh
```
dvdzkwsk: Might run into issues with Windows/non-zsh users trying to use this.

levithomason: Aye, this was just for my local use to generate the first draft files.  The script and files were outdated by the factory function.

levithomason:
Updated with more robust part factories and notes.  This is far from done, but will continue to churn on the ideas.  Feedback welcome.

levithomason:
This is getting stale, closing.

I think PR also suffered from scope creep.  I think we should only implement a reusable component for the basic "className only" use case.  This is something that will also likely shake out of solving #419, since the base component will handle the same functionality as a reusable component.

PR #351, opened 2016-07-27
Update Grid to v1 API

layershifter:
Update `<Grid>` to v1 API.
http://semantic-ui.com/collections/grid.html

---
# Types
## Grid

A basic grid

``` html
<div class="ui grid">
  <div class="column"></div>
</div>
```

``` jsx
<Grid>
  <Grid.Column></Grid.Column>
</Grid>
```
## Divided

A grid can have dividers between its columns

``` html
<div class="ui three column divided grid">
  <div class="column"></div>
</div>
```

``` jsx
<Grid column={3} divided>
  <Grid.Column></Grid.Column>
</Grid>
<Grid column='three' divided>
  <Grid.Column></Grid.Column>
</Grid>
```
## Vertically Divided

A grid can have dividers between rows

``` html
<div class="ui vertically divided grid"></div>
```

``` jsx
<Grid column={3} divided='vertically' />
```
## Celled

A grid can have rows divided into cells

``` html
<div class="ui celled grid"></div>
```

``` jsx
<Grid celled />
```
## Internally Celled

A grid can have rows divisions only between internal rows

``` html
<div class="ui internally celled grid"></div>
```

``` jsx
<Grid celled='internally' />
```
# Content

_Skip_
# Variations
## Floated

A column can sit flush against the left or right edge of a row

``` html
<div class="ui grid">
  <div class="left floated column"></div>
  <div class="right floated column"</div>
</div>
```

``` jsx
<Grid>
  <Grid.Column floated='left' />
  <Grid.Column floated='right' />
</Grid>
```
## Column Width

A column can vary in width taking up more than a single grid column

_Skip_
## Column Count

A grid can have a different number of columns per row

_Skip_
## Equal Width

A row can automatically resize all elements to split the available width evenly

``` html
<div class="ui equal width grid"></div>
```

``` jsx
<Grid width='equal' />
```
## Stretched

An can stretch its contents to take up the entire column height

``` html
<div class="stretched row"></div>
```

``` jsx
<Grid.Row stretched />
```
## Padded

A grid can preserve its vertical and horizontal gutters on first and last columns

``` html
<div class="ui padded grid"></div>
<div class="ui vertically padded grid"></div>
<div class="ui horizontally padded grid"></div>
```

``` jsx
<Grid padded />
<Grid padded='vertically' />
<Grid padded='horizontally' />
```
## Relaxed

A grid can increase its gutters to allow for more negative space

``` html
<div class="ui relaxed grid"></div>
<div class="ui very relaxed grid"></div>
```

``` jsx
<Grid relaxed />
<Grid relaxed='very' />
```
## Colored

A grid row or column can be colored

``` html
<div class="ui red column"></div>
<div class="ui red row"></div>
```

``` jsx
<Grid.Column color='red' />
<Grid.Row  color='red' />
```
## Centered

A grid can increase its gutters to allow for more negative space

``` html
<div class="ui centered grid"></div>
```

``` jsx
<Grid centered />
```
# Text Alignment

A grid, row, or column can specify its text alignment

``` html
<div class="ui left aligned grid"></div>
<div class="center aligned row"></div>
<div class="right aligned column"></div>
<div class="justified row"></div>
```

``` jsx
<Grid aligned='left' />
<Grid.Row aligned='center' />
<Grid.Column aligned='right' />
<Grid.Row aligned='justified' />
```
## Vertical Alignment

A grid, row, or column can specify its vertical alignment to have all its columns vertically centered

**(!)** Problem. Vertical aligment uses same class name as text alignment, but we can't pass two `aligned` props :smile:  So, I propose use `vertical` prop.

``` html
<div class="ui bottom aligned grid"></div>
<div class="middle aligned row"></div>
<div class="top aligned column"></div>
```

``` jsx
<Grid aligned='bottom' />
<Grid.Row aligned='middle' />
<Grid.Column aligned='top' />
```
# Responsive Variations
## Doubling

A grid can double its column width on tablet and mobile sizes

``` html
<div class="ui doubling grid"></div>
```

``` jsx
<Grid doubling />
```
## Stackable

A grid can have its columns stack on-top of each other after reaching mobile breakpoints

``` html
<div class="ui stackable grid"></div>
```

``` jsx
<Grid stackable />
```
# Reversed

A grid or row can specify that its columns should reverse order at different device sizes

_See docs_
# Device Visibility

A columns or row can appear only for a specific device, or screen sizes

_See docs_
# Responsive Width

A column can specify a width for a specific device

_See docs_


layershifter:
Reference #269

layershifter:
Docs finished :smile: Open for discussion.

Take a look, pay attention on `Responsive Variations`, there are mistakes in SUI docs.

levithomason:
Awesome! Will review tomorrow :D

layershifter:
I've added test, you can see problems with `prop` naming, so I'm open to ideas :smile:

docs/app/Examples/collections/Grid/Types/GridExample.js
```diff
@@ -0,0 +1,15 @@
+import _ from 'lodash'
+import React from 'react'
+import { Image, Grid } from 'stardust'
+
+const columns = _.range(0, 16).map((i) => (
```
levithomason: [`_.times`](https://lodash.com/docs#times) can be used here:

``` js
_.range(0, 16).map((i) => {})

// becomes

_.times(16, (i) => {})
```

docs/app/Examples/collections/Grid/Types/GridDividedExample.js
```diff
@@ -0,0 +1,40 @@
...
+      </Grid.Row>
+
+      <Grid.Row>
+        <Grid.Column>{image}</Grid.Column>
+        <Grid.Column>{image}</Grid.Column>
+        <Grid.Column>{image}</Grid.Column>
+      </Grid.Row>
+    </Grid>
+
+    <Divider />
```
levithomason: To keep examples clean and concise, let's use separate example files instead of a `<Divider />` between examples.  Then when adding the `<ComponentExample />` to the index.js for the related examples, just omit the `title` and `description`.  This way they'll appear all under one heading but have some whitespace between them and separate code toggle buttons.

levithomason:
Here's some initial feedback to keep things moving:

### Vertical Alignment

Your snippet here shows `aligned` instead of `vertical`, however, I see the issue you are referring to.  I've talked this over with @davezuko and we're leaning toward:

``` jsx
<Grid textAlign='left' verticalAlign='top' />
```

The confusing thing about `aligned` is that it is not clear which axis it is operating on.  Especially given a grid column where `aligned='left'` seems like it might shift the column itself to the left most position in the grid when it will actually only align the content of the column.

This API would require refactoring other components using the `aligned` prop.  We would also need to update/add common tests.  I want to sleep on this and see if anything better comes up.  More feedback also welcomed.

### Column Count & Equal Width

I think we should use the plural name for grid columns prop and support `equal` there as well. This will be consistent with what we're doing with `<Foo.Group widths={...} />` (i.e. Step, Statistics, etc), except the prop name will be `columns` instead of `widths`:

``` diff
-<Grid column={3} />
+<Grid columns={3} />

-<Grid width='equal' />
+<Grid columns='equal' />
```

### Reversed, Device Visibility & Responsive Width

These all look great.

> Take a look, pay attention on Responsive Variations, there are mistakes in SUI docs.

Can you elaborate on this one?  I think I'm missing the error.  I pulled and tested your branch and I believe it looks correct.

layershifter:
### Vertical Alignment

Pretty idea, I like it. Working on it.

layershifter:
### Vertical Alignment

I've updated usage of `aligned` prop, renamed `commonTests` and `propUtils`, added new functions.

layershifter:
### Column Count & Equal Width

Updated as you asked :smile:

layershifter:
### Reversed, Device Visibility & Responsive Width

> I think I'm missing the error

`Reverved` might be `Resersed` and some other errors in SUI docs.

---

Take a look on `Responsive Width` implementation, prop naming really correct (`mobile`, `largeScreen`, `widescreen`, etc)? If it's okay, I've update `implementsNumberToWordProp` test to support different class name (see commented test on `GridColumn-test.js`).

docs/app/Examples/collections/Grid/ResponsiveVariations/GridReversedComputerVerticallyExample.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Grid } from 'stardust'
+
+const { Column, Row } = Grid
+
+const GridReversedMobileVerticallyExamples = () => (
```
levithomason: Mobile should be computer here.

docs/app/Examples/collections/Grid/ResponsiveVariations/GridReversedMobileExample.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Grid } from 'stardust'
+
+const { Column } = Grid
+
+const GridReversedMobileVerticallyExamples = () => (
```
levithomason: No "Vertically" here.

docs/app/Examples/collections/Grid/Variations/GridTextAlignmentExample.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { Grid, Menu } from 'stardust'
+
+// TODO: Update <Menu> usage after v1 API
+
+const { Column, Row } = Grid
+
+const GridTextAlignmentExample = () => (
+  <Grid textAlign='center' columns={3}>
```
levithomason: I believe this file should be `GridTextCenterAlignmentExample` (added Center) to match the others, yea?

docs/app/Examples/collections/Grid/Variations/GridTextAlignmentJustifiedExample.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { Grid, Menu } from 'stardust'
+
+// TODO: Update <Menu> usage after v1 API
+
+const { Column, Row } = Grid
+
+const GridTextAlignmentJustifiedExample = () => (
+  <Grid textAlign='center'>
```
levithomason: Should this prop be `textAlign='justified'` to match the file and const name?

layershifter: No, example's name is about `<Row>`.

levithomason: What I mean is that the filename and the const name are called "Text Alignment Justified" but the prop is set to `textAlign='center'`.  These two are conflicting.  I would think the filename and const name should be "Text Align Center" or the prop should be `textAlign='justified'` so they are matching.

levithomason: Ah, I see line 29 has the justified prop.  The Grid prop here was confusing me.

layershifter: What about rename to `GridTextAlignmentMixedExample.js`?

src/utils/propUtils.js
```diff
@@ -248,6 +250,24 @@ export const useKeyOrValueAndKey = (val, key) => val && (val === true ? key : `$
...
+ * <Grid columns='equal' />
+ * <div class="ui equal width grid"></div>
+ *
+ * @example
+ * <Grid columns={4} />
+ * <div class="ui four column grid"></div>
+ */
+export const useColumnsProp = (val, canEqual = false) => val === 'equal' && canEqual
+  ? 'equal width'
+  : useValueAndKey(numberToWord(val), 'column')
```
levithomason, line 268: Thanks for adding these!

src/utils/propUtils.js
```diff
@@ -248,6 +250,24 @@ export const useKeyOrValueAndKey = (val, key) => val && (val === true ? key : `$
...
+
+/**
  * The "aligned" prop follows the useValueAndKey except when the value is "justified'.
```
levithomason:
``` diff
-  * The "aligned" prop
+  * The "textAlign" prop
```

test/specs/commonTests.js
```diff
@@ -415,32 +415,30 @@ const _classNamePropValueBeforePropName = (Component, propKey, requiredProps) =>
...
+    _definesPropOptions(Component, 'columns')
+    _noDefaultClassNameFromProp(Component, 'columns', requiredProps)
+    _noClassNameFromBoolProps(Component, 'columns', requiredProps)
 
-    _.each(Component._meta.props.aligned, (propVal) => {
-      if (propVal === 'justified') {
-        it('adds "justified" without "aligned" to className', () => {
-          shallow(<Component { ...requiredProps } aligned='justified' />)
-            .should.have.className('justified')
+    it('adds prop value to className', () => {
```
levithomason: It looks like this test now uses `numberToWord` to convert the value.  Let's update the assertion to

``` diff
-    it('adds prop value to className',
+    it('adds numberToWord value to className'
```

test/specs/commonTests.js
```diff
@@ -528,6 +526,56 @@ export const implementsNumberToWordProp = (Component, propKey, requiredProps = {
...
+ * @param {Object} [requiredProps={}] Props required to render the component.
+ */
+export const implementsVerticalAlignProp = (Component, requiredProps = {}) => {
+  describe('verticalAlign (common)', () => {
+    _definesPropOptions(Component, 'verticalAlign')
+    _noDefaultClassNameFromProp(Component, 'verticalAlign')
+    _noClassNameFromBoolProps(Component, 'verticalAlign', requiredProps)
+
+    _.each(Component._meta.props.verticalAlign, (propVal) => {
+      it(`adds "${propVal} verticalAlign" to className`, () => {
```
levithomason: Typo `adds "${propVal} verticalAlign"` should be `aligned`

docs/app/Examples/collections/Grid/Variations/GridTextAlignmentJustifiedExample.js
levithomason: It seems the left aligned column inside of the center aligned grid is also not working.  This might add to the confusion.

Since there is already an example showing alignment on the grid itself, what if we remove `textAlign='center'` from the grid here.  Then, change the column `textAlign='left'` below to `textAlign='center'` so it shows column override alignment.  Finally, leave the row `textAlign='justified'` as is.  Then this example will show left, center, and justified alignments.

I like `GridTextAlignmentMixedExample.js` as well.

test/specs/collections/Grid/GridColumn-test.js
```diff
@@ -4,4 +4,18 @@ import * as common from 'test/specs/commonTests'
...
+  common.propValueOnlyToClassName(GridColumn, 'color')
+  // common.implementsNumberToWordProp(GridColumn, 'computer')
+  common.propKeyAndValueToClassName(GridColumn, 'floated')
+  // common.implementsNumberToWordProp(GridColumn, 'largeScreen')
+  // common.implementsNumberToWordProp(GridColumn, 'mobile')
+  common.propKeyAndValueToClassName(GridColumn, 'only')
+  // common.implementsNumberToWordProp(GridColumn, 'tablet')
+  common.implementsTextAlignProp(GridColumn)
+  common.implementsVerticalAlignProp(GridColumn)
+  // common.implementsNumberToWordProp(GridColumn, 'widescreen')
```
levithomason: What do we need to do in order to get these tests working?

layershifter: Seems, it was my mistake, I'll make commit soon

levithomason:
I think we're good to go on this PR.  Is there anything else you wanted to hit before merging?  If not, merge master and I'll give it a final once over and merge.

layershifter:
Updated to origin, seems we're ready :smile_cat:

layershifter:
Updated again :v:

levithomason:
I believe this is good to go, I've looked at it too many times!  Merging, we can iterate on any oversights.  Thanks for the _massive_ amount of work here 🍻

levithomason:
Released in `v0.28.0`

PR #353, opened 2016-07-28
fix(jQuery): weaken window.jQuery error

### LINKED ISSUE #352: Don't throw error if jquery is detected
The error that is thrown here was preventing our app from loading:
https://github.com/TechnologyAdvice/stardust/blob/3d3dab56d1de1a87b946f77c7f04ffc7cfbb48fa/src/utils/jquery.js#L14

jquery might be loaded for any number of reasons. I don't think it's reasonable to throw an error if it exists. Just not using it within stardust should be enough. In our case, we're using stardust in a Meteor app.  Meteor makes it notoriously difficult to remove jquery completely, although we're trying.

dvdzkwsk:
Fixes #352

### DIFF
--- a/src/utils/jquery.js
+++ b/src/utils/jquery.js
@@ -11,10 +11,15 @@ if (window.jQuery) jQuery = window.jQuery
 else if (_.get(window, '$.prototype.jquery')) jQuery = window.$
 
 if (jQuery) {
-  throw new Error([
-    'Do not load jQuery, it is bundled with Stardust.',
+  /* eslint-disable no-console */
+  console.error([
+    'Do not load jQuery on the window, it is bundled with Stardust.',
     'Stardust will soon be jQuery free: https://github.com/TechnologyAdvice/stardust/issues/247.',
+    'Note that, until this is complete, including jQuery on the page will mean that jQuery ' +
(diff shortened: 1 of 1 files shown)


src/utils/jquery.js
```diff
@@ -11,15 +11,17 @@ if (window.jQuery) jQuery = window.jQuery
...
     'Do not load jQuery, it is bundled with Stardust.',
     'Stardust will soon be jQuery free: https://github.com/TechnologyAdvice/stardust/issues/247.',
   ].join(' '))
+  /* eslint-enable no-console */
 }
 
 debug('Loading jQuery')
 // load jQuery, then load SUI jQuery
-window.jQuery = window.$ = require('jquery')
+window.jQuery = window.$ = jQuery || require('jquery')
```
dvdzkwsk: Prevents overriding `window.jQuery` if it exists, which was the source of jQuery extension issues in the past.

levithomason:
Could you confirm the doc site form validation works as expected with this setup?

levithomason:
:accept:

dvdzkwsk:
Confirmed it works with and without jQuery on the window @levithomason .

PR #356, opened 2016-08-01
Dropdown: add support for allowAdditions

dylankiss:


levithomason:
Thanks, I will review this tomorrow. Could you please add an example to the doc site so we can try it out? You can reference the other Dropdown examples in `docs/app/Examples/modules/Dropdown`. Just add an example file (probably `types/AllowAdditions.js`), then update the index.js in that directory.

dylankiss:
@levithomason I added a simple example to the docs.

docs/app/Examples/modules/Dropdown/Types/AllowAdditions.js
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <div>
+        <Dropdown
+          options={[
+            { text: 'English', value: 'English' },
+            { text: 'French', value: 'French' },
+            { text: 'Spanish', value: 'Spanish' },
+            { text: 'German', value: 'German' },
+            { text: 'Chinese', value: 'Chinese' },
+          ]}
```
levithomason: Let's pull these options into a `const` above the example.  This way we can reuse them.  Also, they'll be created once instead of every render.

Then, let's add a single and multiple dropdown that use the same options and both allow additions.  Similar to http://semantic-ui.com/modules/dropdown.html#tagging-and-user-additions.

This raised the first bug, multiple dropdowns do not allow adding multiple additions.

src/modules/Dropdown/Dropdown.js
```diff
@@ -98,6 +98,9 @@ export default class Dropdown extends Component {
     /** Name of the hidden input which holds the value. */
     name: PropTypes.string,
 
+    /** Allow user additions to the list of options */
+    allowAdditions: PropTypes.bool,
```
levithomason: This prop requires props `options`, `selection`, and `search`.  See the `selection` propType definition for an example on how to enforce bool while also requiring those other props.

src/modules/Dropdown/Dropdown.js
```diff
@@ -458,6 +479,10 @@ export default class Dropdown extends Component {
     this.setState({
       selectedIndex: 0,
       searchQuery: newQuery,
+      addition: {
+        text: newQuery,
+        value: newQuery,
+      },
```
levithomason: When backspacing the search query, an empty option is left added to the dropdown menu:



I'm thinking this feature may be better suited for a callback than state.  So, when a user attempts to add an item, we callback `onAddition` with the item value being added.  Then, the user can merge this item into the list of options themselves.  We would still fire `onChange` with addition value and also `trySetState` to add the addition to the dropdown value.

If we did this, all the mechanisms in place for getting items and setting the active item would work as is since the addition would simply be added to the actual list of options.  It would also give the developer opportunity to perform other tasks before allowing the item to be added.  For instance, they might want to disable the dropdown, show the loading state, and save the item back to the server.  Or, perform some client side validation before allowing the item in the options list.

Here is an example of what I am explaining:

``` jsx
import { Dropdown } from 'stardust'

export default class AllowAdditionsExample extends Component {
  state = { options: [{ text: 'English', value: 'English' }] }

  handleAddition = (value) => {
    // validate, save to server, show a msg to the user, etc.
    this.setState({
      options: [{ text: value, value }, ...this.state.options]
    })
  }

  render() {
    return (
      <Dropdown
        options={this.state.options}
        placeholder='Choose Language'
        search
        selection
        allowAdditions
        onAddition={this.handleAddition}
      />
    )
  }
}
```

This is not a full working example, but it illustrates the point.  Let me know your thoughts.

dylankiss:
@levithomason Thanks for your very useful remarks! I will look into it as soon as possible.

dylankiss:
@levithomason I'm trying to add tests for my component, but when I try to run them using `npm test`, I get the following error:

```
05 08 2016 12:44:04.986:ERROR [config]: Invalid config file!
  SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.parseConfig (/Users/dkiss/projects/stardust/node_modules/karma/lib/config.js:284:22)
    at new Server (/Users/dkiss/projects/stardust/node_modules/karma/lib/server.js:57:20)
    at Object.exports.run (/Users/dkiss/projects/stardust/node_modules/karma/lib/cli.js:243:7)
    at requireCliAndRun (/Users/dkiss/projects/stardust/node_modules/karma-cli/bin/karma:44:16)
    at /Users/dkiss/projects/stardust/node_modules/karma-cli/bin/karma:54:12
    at /Users/dkiss/projects/stardust/node_modules/resolve/lib/async.js:44:21
    at ondir (/Users/dkiss/projects/stardust/node_modules/resolve/lib/async.js:187:31)
    at /Users/dkiss/projects/stardust/node_modules/resolve/lib/async.js:153:39
    at onex (/Users/dkiss/projects/stardust/node_modules/resolve/lib/async.js:93:22)
```

I haven't changed anything in the config file. Do you know immediately what this problem could be?

dylankiss:
@levithomason I have changed the behavior to your suggestions and added an example for single and multi-select. I also added a few test cases to cover the code changes.

Still unable to run the tests locally, so it was just: write a test -> push ... :)

levithomason:
Which version of Node do you have? This project requires Node 6.  See CONTRIBUTING.md [Getting Started](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#getting-started) for more.

dylankiss:
Well, that will be it. We're working with Node 4. Will try with Node 6 next time :)

dylankiss:
@levithomason While adding this feature I noticed that there is this behavior in the Dropdown that automatically selects the first suggestion in the list if you press enter. This makes that you can't add a substring of an existing option as a new option. I don't know what the best behavior would be to allow this, because you will still want to select your first option by pressing enter also when you are allowed to add a new one.

levithomason:
> This makes that you can't add a substring of an existing option as a new option.

Hm, some refactors are probably necessary. On mobile but I'll try to guide you. 

There is a single "selected" item in a selection Dropdown. This is the highlighted item in the menu. It is made "active" on pressing Enter. The active items make up the Dropdown value. 

There is a "selectedIndex" state key. The menu item at the index matching the selectedIndex is "selected", meaning it is highlighted. The up/down arrow keys decrement and increment the selectedIndex. When the search query changes, the selectedIndex is set to 0 (the first search result). When the search query doesn't match anything, the selectedIndex is set to -1 (iirc) so no item is selected. 

Here's where the refactor may be needed. We currently match on string "includes". Perhaps when there is not an exact match, we include an item at the bottom of the list that says "`Add: <search query >`". If the user chooses this item, we call the "onAddition" callback. 

When there are no matching results, this would replace the "No results" message that is currently shown. 

Hope this helps. I'll be out the rest of today but will  look more into this as soon as I can.

dylankiss:
I was also considering something like that. I'd rather put it at the top of the list instead of at the bottom though. Because you could possibly have a lot of options that start with the same letter(s) and that would make the `Add: ...` option only visible when you scroll to the bottom. When it's at the top it's immediately clear you can add that value. The first predefined option is just one "arrow down" away then :)

What's your opinion on this?

BTW: The vanilla Semantic UI dropdown has the same behavior for user additions (also unable to add a substring).

levithomason:
> What's your opinion on this?

My intuition is that the most common use case will be selecting existing items and the least common use case will be adding items.  Because of this, I think the more common use case should have fewer keystrokes and more visibility.  Meaning, the `Add: ...` would be at the bottom.

> Because you could possibly have a lot of options that start with the same letter(s) and that would make the Add: ... option only visible when you scroll to the bottom.

This is a really valid point.  I think when adding an item you'll have a search query that will match little to no existing items.  In this case you'll be able to see the `Add: ...` as either the first item, or just a minimal number of items down the list.  Being this should be the exception, not the norm, I think it makes sense to offer the existing items first to the user.  Then, in those less common edge cases that they are searching for a non-existent item with many substring matches, they may have to scroll to find the `Add: ..`.  I think that will be the intuitive behavior of a person looking for their match in the list when they can't see it.

> BTW: The vanilla Semantic UI dropdown has the same behavior for user additions (also unable to add a substring).

Had no idea 😄, though it seems like a common enough problem and great feature we should add still.  Thoughts?

dylankiss:
> My intuition is that the most common use case will be selecting existing items and the least common use case will be adding items.

I guess this depends on the specific use case for this control. For my use case it is the other way round. I basically want to use it as a text field with a history list of used values that the user could re-use, but primarily he should be able to simply add a new value here.

Ideally just tabbing from the control after having entered his custom value should call `onAddition` and `onChange`, as does the vanilla SUI Dropdown too. But that's another issue, I think.

What are your thoughts on adding another prop to select whether you want to go 'insert-first' or 'select-first' and depending on that, show the `Add: ...` option at the top or bottom of the list?

levithomason:
> What are your thoughts on adding another prop to select whether you want to go 'insert-first' or 'select-first' and depending on that, show the Add: ... option at the top or bottom of the list?

Sure, something like `additionPosition={'top'|'bottom'}` or similar would be great.

dylankiss:
Alright, I will look into it probably tonight or tomorrow.

dylankiss:
@levithomason Another discrepancy between Stardust and Semantic UI Dropdown is the way it handles blurs (and tab).

In Semantic UI, when you use the keyboard to browse through dropdown list, the placeholder changes to the entry you are currently highlighting. When losing focus after that (by clicking somewhere else or pressing tab) the highlighted value gets selected.

In Stardust
- it does not update the placeholder value when browsing trough the list with the up and down keys
- when clicking outside the component the value stays what it was before, or the search text stays in the search box when there was text in there
- when pressing tab, the dropdown menu does not collapse and for the rest it has the same behavior as clicking outside the component

Is there a reason for these discrepancies, or are they just things that were overseen?

levithomason:
There is an issue with closing on blur.  There are todos in the code and tests explaining this.

All the other features can be implemented afaik, PRs welcome!  It might make sense to open a separate issue/PR since the allowAdditions is entirely separate from these features.

dylankiss:
Okay, I might work on these issues then after this one is ready :)

dylankiss:
@levithomason Sorry for the delay. I managed to incorporate the `Add: ...` option (that now allows substrings) with an optional `additionPosition`, being `'top'` or `'bottom'` (default). The single select example uses no `additionPosition` (defaults to bottom) and the multi select uses `'top'`. I also added tests to cover all the changes.

Could you take a look to see whether this looks good to you?
Thanks :)

levithomason:
Awesome!  Pulling and reviewing now.

levithomason:
Looking much better, thanks for docs as well.  There is one issue remaining.  When typing a query that doesn't exists, then removing it, then "Add: " text is still shown:

image

This is also true for multiple dropdowns, but only when it already has a value:

image

src/modules/Dropdown/Dropdown.js
```diff
@@ -100,6 +100,21 @@ export default class Dropdown extends Component {
...
+    allowAdditions: customPropTypes.every([
+      customPropTypes.demand(['options', 'selection', 'search']),
+      PropTypes.bool,
+    ]),
+
+    /** Called with the new value added by the user. Use this to update the options list. */
+    onAddition: PropTypes.func,
+
+    /** Position of the `Add: ...` option in the dropdown list ('top' or 'bottom'). */
+    additionPosition: PropTypes.string,
```
levithomason:
``` diff
-PropTypes.string,
+PropTypes.oneOf(_meta.props.additionPosition),
```

At the top of the file:

``` diff
const _meta = {
  name: 'Dropdown',
  type: META.TYPES.MODULE,
  props: {
    pointing: ['bottom left', 'bottom right'],
+    additionPosition: ['top', 'bottom'],
  },
}
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -227,6 +246,25 @@ export default class Dropdown extends Component {
...
+      } else if (_.last(this.state.options) !== _.last(this.props.options)) {
+        // `Add: ...` option at the bottom of the list
+        this.setState({
+          options: [...nextProps.options, _.last(this.state.options)],
+        })
+      } else {
+        // No `Add: ...` option -> just copy the list
+        this.setState({
+          options: [...nextProps.options],
+        })
```
levithomason: Instead of putting `options` on state, I think we should insert the `Add:` item into the array of items returned by `getMenuOptions()` method.  This should be the only place it is required.  Anytime the menu renderers, it will have the `Add:` item in it if necessary.

This would allow us to remove the state and extra logic from many places. The current code for handling the `onAddition` callback will still work.

``` js
getMenuOptions = () => {
  // ... at the bottom of the function:

  // insert the "add" item
  if (allowAdditions && search && searchQuery && !_.some(filteredOptions, { text: searchQuery })) {
    const addItem = {
      text: `Add: ${searchQuery}`,
      value: searchQuery,
    }
    if (additionPosition === 'top') filteredOptions.unshift(addItem)
    else filteredOptions.push(addItem)
  }

  return filteredOptions
}
```

We can then save a render and we can remove all the code that is handling `options` on state.

src/modules/Dropdown/Dropdown.js
```diff
@@ -747,7 +802,13 @@ export default class Dropdown extends Component {
...
         onClick={this.handleItemClick}
         selected={selectedIndex === i}
-        text={opt.text}
+        text={
+          allowAdditions
+          && (additionPosition === 'top' ? i === 0 : i === options.length - 1)
+          && opt.text.startsWith('Add: ') ?
+          <span><strong><em>Add:</em></strong> {opt.text.substring(5)}</span> :
+          opt.text
+        }
```
levithomason: With the update to `getMenuOptions`, this prop would also be reverted back to `text={text}`.  That is because the `getMenuOptions` method will be inserting an item with the text already set.

`renderOptions` gets its list of items from the `getMenuOptions`.

src/modules/Dropdown/Dropdown.js
levithomason: Note, this also resolves the bug I noted about backspacing a query.  That is because the logic here includes `searchQuery` as a condition for injecting the `add` item.

src/modules/Dropdown/Dropdown.js
```diff
@@ -100,6 +100,21 @@ export default class Dropdown extends Component {
...
+     * Allow user additions to the list of options (boolean).
+     * Requires the use of `selection`, `options` and `search`.
+     */
+    allowAdditions: customPropTypes.every([
+      customPropTypes.demand(['options', 'selection', 'search']),
+      PropTypes.bool,
+    ]),
+
+    /** Called with the new value added by the user. Use this to update the options list. */
+    onAddition: PropTypes.func,
```
levithomason: Looking back on this now, this prop probably makes more sense as `onAddItem`.  Mind updating?

_Side note, I plan on updating the `options` prop to `items` so it follows SUI closer.  Then the, `Select` component will use `options`._

levithomason:
> Could you take a look to see whether this looks good to you?

All comments in!  Let me know if you'd like any help on this PR.  I don't want to over burden you with changes.  If you would like some help, just add me as a collaborator on your fork and I can make updates along with you.

Thanks much, cheers!

src/modules/Dropdown/Dropdown.js
dylankiss: This was also true before, but the extra thing I do here is add some styling to the `Add:` part, such that it stands out from the rest.

dylankiss:
@levithomason Valid point concerning the state ;) I changed it. Also renamed the onAddition prop as per your proposal. I agree it is better.

Could you have another look please?

levithomason:
Will do, thanks

src/modules/Dropdown/Dropdown.js
```diff
@@ -471,7 +493,18 @@ export default class Dropdown extends Component {
...
+    if (allowAdditions && !_.some(options, { value: newQuery })) {
+      if (additionPosition === 'top') {
+        newOptions = [{ text: `Add: ${newQuery}`, value: newQuery }, ...options]
+      } else {
+        newOptions = [...options, { text: `Add: ${newQuery}`, value: newQuery }]
+      }
+    }
+
     this.setState({
+      options: newOptions,
```
levithomason: This can be removed since there is now no use of state for options.

src/modules/Dropdown/Dropdown.js
```diff
@@ -471,7 +493,18 @@ export default class Dropdown extends Component {
...
+    let newOptions = options
+
+    if (allowAdditions && !_.some(options, { value: newQuery })) {
+      if (additionPosition === 'top') {
+        newOptions = [{ text: `Add: ${newQuery}`, value: newQuery }, ...options]
+      } else {
+        newOptions = [...options, { text: `Add: ${newQuery}`, value: newQuery }]
+      }
+    }
+
```
levithomason: This block of logic can also be removed.

src/modules/Dropdown/Dropdown.js
```diff
@@ -453,7 +475,7 @@ export default class Dropdown extends Component {
     debug(e.target.value)
     // prevent propagating to this.props.onChange()
     e.stopPropagation()
-    const { search, onSearchChange } = this.props
+    const { search, onSearchChange, options, allowAdditions, additionPosition } = this.props
```
levithomason: This can now be reverted.

src/modules/Dropdown/Dropdown.js
```diff
@@ -527,11 +570,11 @@ export default class Dropdown extends Component {
   setValue = (value) => {
     debug('setValue()')
     debug('value', value)
-    const { multiple } = this.props
+    const { multiple, options } = this.props
     const { selectedIndex } = this.state
-    const options = this.getMenuOptions()
     const newState = {
       searchQuery: '',
+      options,
```
levithomason: Can remove options from state here as well.

src/modules/Dropdown/Dropdown.js
```diff
@@ -527,11 +570,11 @@ export default class Dropdown extends Component {
   setValue = (value) => {
     debug('setValue()')
     debug('value', value)
-    const { multiple } = this.props
+    const { multiple, options } = this.props
     const { selectedIndex } = this.state
-    const options = this.getMenuOptions()
```
levithomason, line 537: Should be reverted so that the selectedIndex is based on the filtered list of options in the menu rather than the full list of options on the prop.

src/modules/Dropdown/Dropdown.js
levithomason: I like the look it provides, but here are my concerns:
1. It fails `DropdownItem` `text` prop validation, unless we allow nodes on `text`. However doing so would go against our current shorthand prop standards (string and numbers only) #391.
2. There is a lot of brittle logic here which I'm not sure justifies the gain (bold italic).  There are also a lot of tests required to test all the code branches.
3. I think this is better suited for core SUI CSS since it is purely stylistic.

So again, I'm not opposed to this feature and I agree the look it gives far better than without it.  But, I don't think the required tradeoffs are worth the bold/italic styling in this case.

I'm open to any other ideas or counter arguments before making a final call.  Let me know what you think.

dylankiss: Agreed that it adds a lot of bulk for a simple styling thing. Maybe instead of the `Add: newValue`, we just want to show `newValue` so that it looks like one of the other options. Or again make it optional (introduces another prop though).

I think without other styling it's a little confusing that your custom value is prepended by `Add:`. Might be just me though.

dylankiss:
Changed to all your comments. Stupid that I didn't see these :)

(Sorry for the delay in response btw, but there's a huge time difference ...)

dylankiss:
@levithomason What do you think of removing the `Add:` prefix for our custom option and show it as a regular option? Or make it configurable?

If I think of my use case I definitely don't need the `Add:` prefix, but maybe other use cases require it to be clear that you are adding something.

levithomason:
Don't fret over the changes and responsiveness, it's just how software goes.  I appreciate all your time and effort here 👍 

A configurable addition label would be superb.  Maybe `additionLabel` and default it to `Add:`?  What do you think?

dylankiss:
@levithomason Sounds good to me. I added the prop and tests for it. This feature is now fully usable for me. What do you think?

levithomason:
Awesome, let's get tests passing and merge it.

dylankiss:
I just rebased before pushing without re-running the tests :) Since you added defaultProps and I did too (at another position), mine overruled yours. Should be fixed now.

levithomason:
Woop! Thanks for the work here, merging.

levithomason:
Released in `v0.34.4`

PR #357, opened 2016-08-01
Add Flag examples

fodurrr:
Hi,

I am quite a newbie on Github and in general, but after 6 months research, reading hundreds of articles and watching hundreds of videos I settled down in General with React, Redux, etc. 

Then I started to look for my stack a good component UI library, I must say that was really a hard part with almost every library I found different issues which I did not like, except this one.

I love your work, love the idea that you eliminate JQuery, etc.

I am super convinced that this library will be the defacto standard for React very soon :))

I am really amazed, keep up the good work.

As I mentioned I am newbie but I would like to contribute and I thought a good place to start is the documentation since anyway I am learning.

I would like to contribute to the examples part in the documentation.

Please feedback, I hope I got my first small open source contribution right :)

Thanks,
Peter

PR #358, opened 2016-08-02
Flag: fix names of countries

layershifter:
Replaced dots with spaces as it might be :smile: 

Refs PR #357


levithomason:
Makes sense to me.  Merging!

PR #363, opened 2016-08-03
Feed: add component

### LINKED ISSUE #188: <Feed /> Component
http://semantic-ui.com/views/feed.html

layershifter:
# WIP

http://semantic-ui.com/views/feed.html
- [x] Components
- [x] Docs
- [x] Tests

---

Fixes #188

### DIFF
--- a/docs/app/Examples/views/Feed/Content/Date.js
+++ b/docs/app/Examples/views/Feed/Content/Date.js
@@ -39,7 +39,7 @@ const Date = () => {
         <Feed.Label image={imageSrc} />
         <Feed.Content
           date='3 days ago'
-          summary={<span>You added <a>Jenny Hess</a> to your <a>coworker</a> group.</span>}
+          summary='You added Jenny Hess to your coworker group.'
         />
       </Feed.Event>
     </Feed>
--- a/docs/app/Examples/views/Feed/Content/Date.js
+++ b/docs/app/Examples/views/Feed/Content/Date.js
@@ -1,13 +1,13 @@
 import React from 'react'
 import { Feed } from 'stardust'
 
-const image = 'http://semantic-ui.com/images/avatar/small/jenny.jpg'
+const imageSrc = 'http://semantic-ui.com/images/avatar/small/jenny.jpg'
 
 const Date = () => {
   return (
     <Feed>
       <Feed.Event>
-        <Feed.Label image={image} />
+        <Feed.Label image={imageSrc} />
         <Feed.Content>
           <Feed.Date>3 days ago</Feed.Date>
           <Feed.Summary>
@@ -17,7 +17,7 @@ const Date = () => {
       </Feed.Event>
 
       <Feed.Event>
-        <Feed.Label image={image} />
+        <Feed.Label image={imageSrc} />
         <Feed.Content>
           <Feed.Date date='3 days ago' />
           <Feed.Summary>
@@ -27,7 +27,7 @@ const Date = () => {
       </Feed.Event>
 
       <Feed.Event>
-        <Feed.Label image={image} />
+        <Feed.Label image={imageSrc} />
         <Feed.Content date='3 days ago'>
           <Feed.Summary>
             You added <a>Jenny Hess</a> to your <a>coworker</a> group.
--- a/docs/app/Examples/views/Feed/Content/Date.js
+++ b/docs/app/Examples/views/Feed/Content/Date.js
@@ -1,0 +1,41 @@
+import React from 'react'
+import { Feed } from 'stardust'
+
+const image = 'http://semantic-ui.com/images/avatar/small/jenny.jpg'
+
+const Date = () => {
+  return (
+    <Feed>
+      <Feed.Event>
+        <Feed.Label image={image} />
+        <Feed.Content>
+          <Feed.Date>3 days ago</Feed.Date>
+          <Feed.Summary>
+            You added <a>Jenny Hess</a> to your <a>coworker</a> group.
+          </Feed.Summary>
+        </Feed.Content>
+      </Feed.Event>
+
+      <Feed.Event>
+        <Feed.Label image={image} />
+        <Feed.Content>
+          <Feed.Date date='3 days ago' />
+          <Feed.Summary>
+            You added <a>Jenny Hess</a> to your <a>coworker</a> group.
+          </Feed.Summary>
+        </Feed.Content>
+      </Feed.Event>
+
+      <Feed.Event>
+        <Feed.Label image={image} />
+        <Feed.Content date='3 days ago'>
+          <Feed.Summary>
+            You added <a>Jenny Hess</a> to your <a>coworker</a> group.
+          </Feed.Summary>
+        </Feed.Content>
+      </Feed.Event>
+    </Feed>
+  )
+}
+
+export default Date
--- a/docs/app/Examples/views/Feed/Content/Date.js
+++ b/docs/app/Examples/views/Feed/Content/Date.js
@@ -34,6 +34,14 @@ const Date = () => {
           </Feed.Summary>
         </Feed.Content>
       </Feed.Event>
+
+      <Feed.Event>
+        <Feed.Label image={imageSrc} />
+        <Feed.Content
+          date='3 days ago'
+          summary={<span>You added <a>Jenny Hess</a> to your <a>coworker</a> group.</span>}
+        />
+      </Feed.Event>
     </Feed>
   )
 }
--- a/docs/app/Examples/views/Feed/Types/Props.js
+++ b/docs/app/Examples/views/Feed/Types/Props.js
@@ -5,30 +5,30 @@ const events = [
   {
     date: '1 Hour Ago',
     image: 'http://semantic-ui.com/images/avatar/small/elliot.jpg',
-    meta: <Feed.Like icon='like'>4 Likes</Feed.Like>,
-    summary: <div><Feed.User>Elliot Fu</Feed.User> added you as a friend</div>,
+    meta: '4 Likes',
+    summary: 'Elliot Fu added you as a friend',
   },
   {
     date: '4 days ago',
     image: 'http://semantic-ui.com/images/avatar/small/helen.jpg',
-    meta: <Feed.Like>1 Like</Feed.Like>,
-    summary: <div><a>Helen Troy</a> added <a>2 new illustrations</a></div>,
+    meta: '1 Like',
+    summary: 'Helen Troy added 2 new illustrations',
     extraImages: [
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
+      'http://semantic-ui.com/images/wireframe/image.png',
+      'http://semantic-ui.com/images/wireframe/image.png',
     ],
   },
   {
     date: '2 Days Ago',
     image: 'http://semantic-ui.com/images/avatar/small/jenny.jpg',
-    meta: <Feed.Like>8 Likes</Feed.Like>,
-    summary: <div><Feed.User>Jenny Hess</Feed.User> added you as a friend</div>,
+    meta: '8 Likes',
+    summary: 'Jenny Hess added you as a friend',
   },
   {
     date: '3 days ago',
     image: 'http://semantic-ui.com/images/avatar/small/joe.jpg',
-    meta: <Feed.Like>8 Likes</Feed.Like>,
-    summary: <div><a>Joe Henderson</a> posted on his page</div>,
+    meta: '8 Likes',
+    summary: 'Joe Henderson posted on his page',
     extraText: [
       "Ours is a life of constant reruns. We're always circling back to where we'd we started, then starting all",
       "over again. Even if we don't run extra laps that day, we surely will come back for more of the same another",
@@ -38,11 +38,11 @@ const events = [
   {
     date: '4 days ago',
     image: 'http://semantic-ui.com/images/avatar/small/justen.jpg',
-    meta: <Feed.Like>41 Likes</Feed.Like>,
-    summary: <div> <a>Justen Kitsune</a> added <a>2 new photos</a> of you</div>,
+    meta: '41 Likes',
+    summary: 'Justen Kitsune added 2 new photos of you',
     extraImages: [
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
+      'http://semantic-ui.com/images/wireframe/image.png',
+      'http://semantic-ui.com/images/wireframe/image.png',
     ],
   },
 ]
--- a/docs/app/Examples/views/Feed/Types/Props.js
+++ b/docs/app/Examples/views/Feed/Types/Props.js
@@ -1,0 +1,52 @@
+import React from 'react'
+import { Feed } from 'stardust'
+
+const events = [
+  {
+    date: '1 Hour Ago',
+    image: 'http://semantic-ui.com/images/avatar/small/elliot.jpg',
+    meta: <Feed.Like icon='like'>4 Likes</Feed.Like>,
+    summary: <div><Feed.User>Elliot Fu</Feed.User> added you as a friend</div>,
+  },
+  {
+    date: '4 days ago',
+    image: 'http://semantic-ui.com/images/avatar/small/helen.jpg',
+    meta: <Feed.Like>1 Like</Feed.Like>,
+    summary: <div><a>Helen Troy</a> added <a>2 new illustrations</a></div>,
+    extraImages: [
+      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
+      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
+    ],
+  },
+  {
+    date: '2 Days Ago',
+    image: 'http://semantic-ui.com/images/avatar/small/jenny.jpg',
+    meta: <Feed.Like>8 Likes</Feed.Like>,
+    summary: <div><Feed.User>Jenny Hess</Feed.User> added you as a friend</div>,
+  },
+  {
+    date: '3 days ago',
+    image: 'http://semantic-ui.com/images/avatar/small/joe.jpg',
+    meta: <Feed.Like>8 Likes</Feed.Like>,
+    summary: <div><a>Joe Henderson</a> posted on his page</div>,
+    extraText: [
+      "Ours is a life of constant reruns. We're always circling back to where we'd we started, then starting all",
+      "over again. Even if we don't run extra laps that day, we surely will come back for more of the same another",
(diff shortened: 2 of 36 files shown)


levithomason:
Note on docs, I've updated the doc example logic and file naming patterns.  Doc files are now organized like so:

```
/docs
  /app
    /Examples
      /views
        index.js  <--- previously "FeedExamples.js"
        /Types
          index.js <-- (optional, can use root index) previously "FeedTypesExamples.js"
```

You can still put all the examples in the root `index.js`, just the filename pattern has been updated.  I'll be updating the old doc files to use this new naming pattern.  All new components should follow it.

layershifter:
@levithomason Did I correctly named doc files?

levithomason:
The naming is only dictated for the `index.js` files, so you are good.  The other files just need to make sense.  I think what you have is OK 👍

layershifter:
It seems we're ready for initial discussion :smile: 
- there are huge amount of subcomponents, it's not looks good;
- `like` designed to be used only with `like` icon, but we use `stars` in production, so I've made small overhead in `<Like>`;

``` html
<a class="like">
  <i class="like icon"></i> 4 Likes
</a>
```
- I'm using `imagePropRender()` on Label, it produces html with `class="ui image"`, while there are no classes on `` in SUI docs;
- TAL on `<FeedExtra>`: seems prop validation for text is ugly, what about shorthand for images, ideas?
- seems `<Summary>` and `<Meta>` can be shorthands on `<Content>`, what you think?
- do we need render with passing object on `<Feed>`, like:

``` jsx
const events = [
  { content: 'Text' },
  ...
]

() => <Feed events={events}>
```

---

Something else? :smile:

levithomason:
@davezuko and I had an offline discussion regarding similar difficulties with sub components in the Message component PR. 

I'm out today but will check back tonight or tomorrow and give an update.

levithomason:
> there are huge amount of subcomponents, it's not looks good;

Indeed. I'll take any better ideas on how to improve this.  Either way, once we get all the components in, we'll revisit these and see if we can optimize them.  We may be able to reuse a single set of component parts.  I want to get all the components completed with explicit parts before we start to optimize though.  This way we can see all the requirements of all the components before making final decisions on how it should be optimized.  If you have any ideas or suggestions, please add them to #345 where this is being worked on.

> like designed to be used only with like icon, but we use stars in production, so I've made small overhead in `<Like>`;

Makes sense to me.  I think the user should be able to pick the icon as well.

> I'm using imagePropRender() on Label, it produces html with class="ui image", while there are no classes on `` in SUI docs;

It looks like this `img` does not support any of the `<Image />` props nor any classNames (i.e. `ui`, `image`, etc.).  It is simply an `img` tag.  Because of this, it seems to me we should handle rendering this prop internally in the component.  If the prop `_.isString()` use it as the `src` for a regular `img`.  If it `_.isObject()` (an element), use the element itself.

Side note, I plan on improving the icon/prop renderers a bit.  Right now they accept either an icon name or image src, or an Icon or Image element.  I think it would be good to allow a string (name/src), a plain props object (`_.isPlainObject() && !React.isValidElement`), or an element (`React.isValidElement`).  I'll probably double back and make these updates a little later.  We just need a util to distinguish

> TAL on `<FeedExtra>`: seems prop validation for text is ugly, what about shorthand for images, ideas?

The `text` prop validation looks correct to me.  Can you elaborate on the ugly part you are seeing?

Regarding shorthand for images, we could accept an array of image `src` strings or image elements.  This would make the validation `oneOfType([bool, arrayOf(string), arrayOf(node)])`.

> seems `<Summary>` and `<Meta>` can be shorthands on `<Content>`, what you think?

Makes sense to me.  Seems a component is component parts should allow a props to configure those parts.

> do we need render with passing object on `<Feed>`, like:

I also thinks this also makes great sense.  Just like `<Accordion panels={} />`, `<Breadcrumb sections={} />`, etc.

> Something else?

Probably! 😄 This is a really great start though, let's keep momentum and just tackle these items for now.

src/views/Feed/Feed.js
```diff
@@ -0,0 +1,56 @@
...
+function Feed(props) {
+  const { children, className, size } = props
+  const classes = cx('ui', className, size, 'feed')
+  const rest = getUnhandledProps(Feed, props)
+
+  return <div {...rest} className={classes}>{children}</div>
+}
+
+Feed._meta = {
+  library: META.library.semanticUI,
```
levithomason: Heads up, I've deprecated library as of today in #380.  See that PR for rationale.  There is also a failing test for including it now.

layershifter:
> It looks like this img does not support any of ...

Updated.

> Can you elaborate on the ugly part you are seeing?

Nevermind :smile: 

> Regarding shorthand for images, we could accept an array of image src strings or image elements.

There is problem with key handling. TAL on `FeedExtra.js`, any ideas?

levithomason:
Sorry, not sure what `TAL` is and I need a little more info to help on this one.  Can you describe the issue with key handling?

src/views/Feed/FeedExtra.js
```diff
@@ -14,7 +15,17 @@ function FeedExtra(props) {
   )
   const rest = getUnhandledProps(FeedExtra, props)
 
-  return <div {...rest} className={classes}>{children || text}</div>
+  if (!images || _.isBoolean(images)) {
+    return <div {...rest} className={classes}>{children || text}</div>
+  }
+
+  const imagesJSX = images.map((image, index) => {
+    const key = `${index}`
```
layershifter: What might be key for image, when image is string? And what when is object?

levithomason: How about the image `src` as it is required and will be unique?

layershifter: In docs src's are same (`/images/wireframe/image.png`), so I have:
image

---

How I can deal with this, when I have element at input?

``` jsx
const images = [<img src='http://semantic-ui.com/images/wireframe/image.png' />, ...];
```

``` jsx
const imagesJSX = images.map(image => {
  return _.isString(image) ? <img key={image} src={image} /> : image
})
```

I think that `React.cloneElement(image, {key: image.src})` is smelly solution.

levithomason: Hm, I see.

/cc @davezuko we we're just talking about `weak-key`, here we have an array of objects with possibly no unique key/value pairs.  Ideas for generating unique React keys?

**EDIT**

I prefer not to add the polyfill for WeakMaps...

src/views/Feed/Feed.js
```diff
@@ -16,18 +16,18 @@ import FeedSummary from './FeedSummary'
...
   const rest = getUnhandledProps(Feed, props)
 
-  if (!items) {
+  if (!events) {
     return <div {...rest} className={classes}>{children}</div>
   }
 
-  const eventsJSX = items.map(({ key, event }, index) => {
+  const eventsJSX = events.map(({ key, ...eventData }, index) => {
     const finalKey = key || index
```
layershifter: Same problem with key is here.

layershifter:
Seems, docs and components finished. I'll start writting test after review. So, feedback is welcome :smile:

src/views/Feed/Feed.js
levithomason: Let's use some property of the `eventData` instead of the `index`.  Using `index` at all in a `key` is an anti-pattern we're moving away from.

You can see more on this in #376

layershifter: I've made comment [there](https://github.com/TechnologyAdvice/stardust/pull/376#issuecomment-238654442), seems there is small problem with it.

levithomason: Thanks, I replied there.

layershifter: > Limit the props to strings only.

I don't understand how I can limit this. Proptypes validation?

---

Also, it will break doc example (see `Feed/Types/Props.js`). May be it might be simplified by more shorthand props.

---

In any case, it is better to make review before. I can't escape from feeling that I did something   nasty there.

levithomason: Ah, I see the example now.  I'm working on the Modal updates right now, almost done. Then, I'll pull the Feed down and come up with some suggestions.

docs/app/Examples/views/Feed/Content/Date.js
```diff
@@ -39,7 +39,7 @@ const Date = () => {
         <Feed.Label image={imageSrc} />
         <Feed.Content
           date='3 days ago'
-          summary={<span>You added <a>Jenny Hess</a> to your <a>coworker</a> group.</span>}
+          summary='You added Jenny Hess to your coworker group.'
```
levithomason, line 42: Reduced all shorthand props, except icon/image, to strings only.

docs/app/Examples/views/Feed/Types/Props.js
```diff
@@ -38,11 +38,11 @@ const events = [
...
     image: 'http://semantic-ui.com/images/avatar/small/justen.jpg',
-    meta: <Feed.Like>41 Likes</Feed.Like>,
-    summary: <div> <a>Justen Kitsune</a> added <a>2 new photos</a> of you</div>,
+    meta: '41 Likes',
+    summary: 'Justen Kitsune added 2 new photos of you',
     extraImages: [
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
-      <a><img src='http://semantic-ui.com/images/wireframe/image.png' /></a>,
+      'http://semantic-ui.com/images/wireframe/image.png',
+      'http://semantic-ui.com/images/wireframe/image.png',
```
levithomason, line 45: Even though we support Icon/Image components elsewhere, I'm considering only supporting strings for simplicity and clarity in v1.  Will leave these examples as strings even though our prop renderer supports both.

src/views/Feed/Feed.js
```diff
@@ -54,7 +54,16 @@ Feed.propTypes = {
...
+    PropTypes.arrayOf(PropTypes.shape({
+      key: PropTypes.string,
+      date: PropTypes.string,
+      image: PropTypes.node,
+      icon: PropTypes.node,
+      meta: PropTypes.string,
+      summary: PropTypes.string,
+      extraText: PropTypes.string,
+      extraImages: PropTypes.arrayOf(PropTypes.node),
+    })),
```
levithomason, line 78: Validated entire feed object, added `key` to validation.

src/views/Feed/FeedContent.js
```diff
@@ -52,16 +52,13 @@ FeedContent.propTypes = {
 
   /** Shorthand for FeedExtra with prop images. */
   extraImages: customPropTypes.all([
-    customPropTypes.mutuallyExclusive(['children', 'content', 'extraText']),
-    PropTypes.oneOf([
-      PropTypes.array,
-      PropTypes.node,
-    ]),
+    customPropTypes.mutuallyExclusive(['children', 'content']),
```
levithomason: Removed exclusive with extraText.  Extra images and Extra text can appear together:

image

src/views/Feed/FeedContent.js
```diff
@@ -52,16 +52,13 @@ FeedContent.propTypes = {
...
+    PropTypes.arrayOf(PropTypes.node),
```
levithomason: Changed to always require an array, rather than node/array.

src/views/Feed/FeedContent.js
```diff
@@ -52,16 +52,13 @@ FeedContent.propTypes = {
...
   ]),
 
   /** Shorthand for FeedExtra with prop text. */
   extraText: customPropTypes.all([
-    customPropTypes.mutuallyExclusive(['children', 'content', 'extraImages']),
+    customPropTypes.mutuallyExclusive(['children', 'content']),
```
levithomason: Removed exclusive with extraImages, see above.

src/views/Feed/FeedExtra.js
```diff
@@ -45,7 +45,6 @@ FeedExtra.propTypes = {
     PropTypes.oneOfType([
       PropTypes.bool,
       PropTypes.arrayOf(PropTypes.node),
-      PropTypes.arrayOf(PropTypes.string),
```
levithomason: Node type covers string type since strings are renderable.  Node is anything that can be rendered.

levithomason:
Updated, take a look at the last commit and comments.

docs/app/Examples/views/Feed/Content/Date.js
levithomason, line 42: Renders a simple read only feed like so:

image

layershifter:
Thanks for review, all is nice, I'll fix it soon and add tests.

layershifter:
Oh, I missed that you've already applied all :smile_cat: Working on tests, seems we can merge it today.

levithomason:
Awesome, I've got the Modal updates and docs done.  It was blocked by the Image, so finishing that up as well.

src/views/Feed/FeedExtra.js
```diff
@@ -0,0 +1,62 @@
...
+    className,
+    useKeyOnly(images, 'images'),
+    useKeyOnly(text, 'text'),
+    'extra'
+  )
+  const rest = getUnhandledProps(FeedExtra, props)
+
+  if (Array.isArray(images)) {
+    const imagesJSX = images.map((image, i) => {
+      return _.isString(image) ? <img key={i} src={image} /> : image
```
layershifter: @levithomason It seems that you like antipattern with `key` more than I :laughing: 

I think:
- allow pass only string, not nodes;
- use pair of `index` and `src`.

Better ideas?

levithomason: > @levithomason It seems that you like antipattern with key more than I 😆

I must admit, I once took it very lightly :S

> - allow pass only string, not nodes;
> - use pair of index and src.

There is a chance that the strings might be the same, for instance a placeholder image for multiple images.  Also, adding index into the string will affectively do the same thing as having an index only, since the key changes when the child order changes.  We would need to accept objects in order to accept a `key`: `images={[{key: 'foo', src: '...'}]}`.

I think for now, we allow a string only and fix the duplicate image src string edge case if it arises.

src/views/Feed/Feed.js
```diff
@@ -0,0 +1,83 @@
...
+  const { children, className, events, size } = props
+  const classes = cx('ui', className, size, 'feed')
+  const rest = getUnhandledProps(Feed, props)
+
+  if (!events) {
+    return <div {...rest} className={classes}>{children}</div>
+  }
+
+  const eventsJSX = events.map(({ key, ...eventData }, index) => {
+    const finalKey = key || index
```
layershifter: `Key` antipattern is here :oncoming_police_car: 

What about this?

``` js
const finalKey = key || `${date}-${meta}-${summary}`
```

layershifter:
I think I blurted out too much about **today** :cry: 
I've finish with tests tomorrow and will wait for final review.

src/views/Feed/Feed.js
levithomason: I think you mean `index` is an antipattern here.  In either case, 👍 to the proposed change.

levithomason:
No worries, I'll review once tests are in then.  Thanks.

levithomason:
Heads up, I've done some heavy utils refactoring in #388.  I left many comments there to help get up to speed.  That PR greatly improved things and unblocks the Image and Modal work.  Really glad with the results and I think you will be too after updating.

LMK if you have any questions.  I'll leave comments on this PR where it needs updated to meet the new refactors later tonight.

levithomason:
I'm updating this branch and will push in a few minutes.

levithomason:
I updated this PR to use the new utils.  It might be a good idea to familiarize yourself with the changes made in the last commit.  I also fixed tests and lint issues.

I'm not sure why PhantomJS is crashing in watch mode though.  The branch was doing that before the updates.  Master branch does not crash.  Looking into it.

levithomason:
Suddenly after the rebase test watching is working again.  LMK if you have any issues.

layershifter:
> Suddenly after the rebase test watching is working again.

_Offtop_: Did you have issues with PhantomJS's crash, too?

---

### Feed.js

I've updated `key` prop handling as we discussed.

``` jsx
const eventsJSX = events.map(eventProps => {
    const { childKey, date, meta, summary, ...eventData } = eventProps
    const finalKey = childKey || `${date}-${meta}-${summary}`

    return <FeedEvent
      date={date}
      key={finalKey}
      meta={meta}
      summary={summary}
      {...eventData}
    />
  })
```

### FeedExtra.js
1. Only array of string can be passed now.

``` diff
- PropTypes.arrayOf(PropTypes.node),
+ PropTypes.arrayOf(PropTypes.string),
```
1. Updated `key` prop (better ideas?) and used new factory.

``` jsx
const imagesJSX = images.map((image, index) => {
  const key = [index, image].join('-')

  return createImg(image, {key})
})
```

### FeedLabel-test.js

Are tests for `image prop` correct (`createImg` factory)?

---

Tests are ready. Show must go on :ghost:

levithomason:
> Did you have issues with PhantomJS's crash, too?

Indeed, not sure what fixed it in the rebase but it seems to be fine now?

> I've updated key prop handling as we discussed.

We'll have to pick a different prop name as `key` won't work, see here: https://github.com/TechnologyAdvice/stardust/pull/390#discussion_r74691048

src/views/Feed/Feed.js
```diff
@@ -31,13 +31,13 @@ function Feed(props) {
...
 
-    return <FeedEvent
+    return (<FeedEvent
       date={date}
       key={finalKey}
       meta={meta}
       summary={summary}
       {...eventData}
-    />
+            />)
```
levithomason: Indent issue here, needs some new lines inserted.

levithomason:
FeedExtra 👍 

FeedLabel-test 👍 though, we'll need to update the `key` prop to another name then update the factories to apply our standard key prop name to the `key` prop.

layershifter:
> Indeed, not sure what fixed it in the rebase but it seems to be fine now?

Yep.

> We'll have to pick a different prop name as key won't work, see here: #390 (comment)

``` jsx
events: customPropTypes.every([
    childKey: PropTypes.string,
  })),
]),
```

`childKey` will be applied to `FeedEvent` as it might be, or not?

> we'll need to update the key prop to another name then update the factories to apply our standard key prop name to the key prop.

??? Not understand you there.

levithomason:
What you've done for FeedEvent is correct.  My comment was in regard to passing a `childKey` prop to the `image` prop:

``` jsx
// on an element
const image = <img childKey='...' src='...' />
<FeedLabel image={image} />

// or a props object (this is a new feature of our factories)

<FeedLabel image={{ childKey='...', src='...' }}
```

Our factories are responsible for merging those props by default.  Currently, they will simply pass the `childKey` prop since onto the final element.  We need to map `childKey` to the `key` prop for the final element in the `createFactory.js mergePropsAndClassName()`.

This work can be done on a separate PR though to keep this one moving.

levithomason:
Anything else to update here?  Looks great.  I'd like to merge and make any updates on future PRs.

layershifter:
No, we can go :car:

levithomason:
💥 Released in `stardust@0.34.0`

PR #371, opened 2016-08-04
Dropdown: resize search on change

### LINKED ISSUE #370: Dropdown (multiple) search box does not grow with input
Currently when you have a multi select dropdown, the search box does not expand when you start typing in it.

multi-select-search-bug

In Semantic UI, the `width` style property is set on the DOM element itself when the content of the search box changes. We could take a similar approach here?

levithomason:
Fixes #370.  The search input has a short fixed width that truncates the value.  This PR resizes the Dropdown search input on change.  SUI css allows for a `span.sizer` child to measure the search input width.  We use this to perform the [same resizing](https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/dropdown.js#L1597) as the SUI Dropdown module, but in vanilla JS.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1,6 +1,7 @@
 import _ from 'lodash'
 import cx from 'classnames'
 import React, { cloneElement, PropTypes } from 'react'
+import { findDOMNode } from 'react-dom'
 
 import META from '../../utils/Meta'
 import { customPropTypes, getUnhandledProps, useKeyOnly, useKeyOrValueAndKey } from '../../utils/propUtils'
@@ -460,9 +461,19 @@ export default class Dropdown extends Component {
     // open search dropdown on search query
     if (search && newQuery && !open) this.open()
 
+    // resize the search input, temporarily show the sizer so we can measure it
+    let searchWidth
+    if (newQuery) {
+      const sizer = findDOMNode(this.refs.sizer)
+      sizer.style.display = 'inline'
+      searchWidth = Math.ceil(sizer.getBoundingClientRect().width)
+      sizer.style.removeProperty('display')
+    }
+
     this.setState({
       selectedIndex: 0,
       searchQuery: newQuery,
+      searchWidth,
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -460,9 +461,19 @@ export default class Dropdown extends Component {
...
+      const sizer = findDOMNode(this).querySelector('.sizer')
+      sizer.style.display = 'inline'
+      searchWidth = Math.ceil(sizer.getBoundingClientRect().width)
+      sizer.style.removeProperty('display')
+    }
+
     this.setState({
       selectedIndex: 0,
       searchQuery: newQuery,
+      searchWidth,
```
levithomason, line 476: If there is no `newQuery` (search text) then this value will be `undefined` so the width removed instead of set to 0.

src/modules/Dropdown/Dropdown.js
```diff
@@ -678,11 +689,19 @@ export default class Dropdown extends Component {
...
       />
     )
   }
 
+  renderSearchSizer = () => {
+    const { search } = this.props
+    const { searchQuery } = this.state
+
+    return !search ? null : <span className='sizer'>{searchQuery}</span>
+  }
```
levithomason, line 703: Only render the sizer on search Dropdowns.

src/modules/Dropdown/Dropdown.js
```diff
@@ -460,9 +461,19 @@ export default class Dropdown extends Component {
     // open search dropdown on search query
     if (search && newQuery && !open) this.open()
 
+    // resize the search input, temporarily show the sizer so we can measure it
+    let searchWidth
+    if (newQuery) {
+      const sizer = findDOMNode(this).querySelector('.sizer')
```
dvdzkwsk: What about adding a ref to `sizer`? It may make your intentions more explicit.

levithomason: I was trying to avoid refs as much as possible.  I suppose we already have a search ref, and this is essentially the same thing.  Will update.

dvdzkwsk: Well you essentially have a ref by doing `findDOMNode(this)`, so it's just being more direct about the matter.

dvdzkwsk:
👻

PR #319, opened 2016-06-30
Update Menu to v1 API

awei01:
Semantic-UI allows menu items with the class `header` to be `<div>`. Change behavior to allow `header` property that will then render a `<div>`. If MenuItem has `header` and `href` prop, then, render `<a>` tag.

Issue #142, opened 2016-01-06
React Router `Links` and `MenuItems`


athurman:
ran into issues  with trying to wrap `MenuItem`s in react-router `Link`s and vice-versa. `Link` components do not play nice with `MenuItem` components, because they are both rendered as `a` tags in the dom, here is how it looks when elements are rendered:

``` html
<a class="sd-menu-item Sidebar-linkIcon item active" data-reactid=".0.0.$=1$/=010.$=11.$link-3/=1$link-3">
  ::before
</a>
<a class="" href="/users" data-reactid=".0.0.$=1$/=010.$=11.$link-3/=1$link-3.2">
  <i class="users icon" data-reactid=".0.0.$=1$/=010.$=11.$link-3/=1$link-3.2.0"></i>
  <span class="Sidebar-linkText" data-reactid=".0.0.$=1$/=010.$=11.$link-3/=1$link-3.2.1">
    USERS
  </span>
</a>
```

this causes issues with styling, and breaks behaviors with `MenuItem`

levithomason:
This should be handled in #319.

levithomason:
This was solved by component augmentation, #414.

``` jsx
<Menu.Item as={Link} to='/home' />
```

See the docs for more.

RobertShaw1:
This information was difficult to find.  A google search to landed me here.  Where can I find this information in the docs?

layershifter:
This example can be found on the [main page](http://react.semantic-ui.com/introduction) of docs 😞

RobertShaw1:
Thanks!  

**Suggestion**: Being that the concept of [Augmentation](https://react.semantic-ui.com/introduction#augmentation) "is essential for working with _MenuLinks_ and _react-router_" can we at least include a link to the [Augmentation documentation](https://react.semantic-ui.com/introduction#augmentation) within the "See" list located at the top of to the [Menu page](https://react.semantic-ui.com/collections/menu)? :grin:

hariDasu:
How can I do this in a card group  passing down props? It is clear where to put the `as {Link}` for a menu item, but  ` <Card.Group items={items} as={Link} to={items.map((item,i)=>(item.href))} /> ` doesn't seem to work. I know this is not stack overflow, so i will also post a formal question there. 😆

tylerjw:
I have the same question as @hariDasu.  Did you ever figure this out?  I can't find your stack overflow question.

layershifter:
[...]

PR #382, opened 2016-08-08
feat(Menu): Update to v1 API 

layershifter:
# WIP

Update [Menu](http://semantic-ui.com/collections/menu.html) to v1 API


levithomason:
Linking #142 for fixing in this PR as well.

levithomason:
Linking previous update PR #319.

src/collections/Menu/Menu.js
```diff
@@ -81,10 +81,11 @@ export default class Menu extends Component {
...
     return Children.map(children, (child, i) => {
       const isItem = child.type === MenuItem
+      const isLink = _.has(child, 'props.href') || _.has(child, 'props.link') || _.has(child, 'props.onClick')
 
       if (isItem) {
         const onClick = (e) => {
-          this.handleItemClick(e, i)
+          if (isLink) this.handleItemClick(e, i)
           if (child.props.onClick) child.props.onClick(e, i)
         }
```
layershifter: I'll take this part from `Accordion` component.

---

Can you review this fragment? It looks not good to me.

src/collections/Menu/MenuItem.js
```diff
@@ -8,16 +8,21 @@ import {
...
     'item',
   )
+  const handleClick = (e) => {
+    if (onClick) onClick(e)
+  }
   const rest = getUnhandledProps(MenuItem, props)
+  const ItemComponent = href || onClick ? 'a' : 'div'
 
-  return <a {...rest} className={classes}>{children}</a>
+  return <ItemComponent {...rest} className={classes} onClick={handleClick}>{children}</ItemComponent>
```
layershifter: As we pass `onClick` [here](https://github.com/TechnologyAdvice/stardust/pull/382/commits/f39a7463dd65e3590311519afe15724eb9ce8a9c#diff-151cbde1a0eb8a3f890a84401e62f45cR88), `ItemComponent` will be defined always as `a`. Ideas?

docs/app/Examples/collections/Menu/Content/SubMenu.js
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Menu vertical>
+      <Item>
+        <Input placeholder="Search..." />
+      </Item>
+
+      <Item>
+        Home
+
+        <Menu>
```
layershifter: Take a look there. `<Menu>` produces `ui menu`, but for correct look it might be only `menu`.

I'm creating `Menu.Menu`, `Menu.Sub`? :hankey:

layershifter:
_offtop_: How I can clean up these commits from PR?

image

src/collections/Menu/MenuItem.js
levithomason: This is not the first issue we've had with rendering `a` tags when there is an onClick prop.  I'm starting to think we only do so when there is an `href` or some explicit `link` prop.

Also, many other components and use cases have us considering the use of a library wide `as` prop:

``` jsx
<Segment as='section' />
//=> <section class="ui segment"></section>

<MenuItem as='a' />
//=> <a class="item></a>

<Dropdown as={Button} options={[...]} />
//=> <div class="ui button dropdown" />
```

Especially handy for `react-router` use with the MenuItem as it would allow passing props for both components:

``` jsx

<MenuItem link as={Link} to='/home' activeClass='active' />
//=> <a class="item active"></a>
```

The `MenuItem` would consume its own props (`link`) and spread the `rest`.  The `as` prop here means the MenuItem would render like:

``` jsx
render() {
  const ElementType = getElementType(props) // new util, like getUnhandledProps()

  return <ElementType {...rest} /* other menu item props */ />
}
```

I'm thinking of doing this today.  The `getElementType()` method would hold the logic for returning the value of the `as` prop, else if props.link return `a`, else `div`.  We could add more logic to how components are allowed to determine their element type as well.  This is good because all components would be consistent.

Feedback welcome here #403 

---

For now, just render a `div` unless there is a `link` or `href` prop.

layershifter: > I'm thinking of doing this today. The getElementType() method would hold the logic for returning the value of the as prop, else if props.link return a, else div.

Seems, it's not good idea, prop `link` might be passes to `class`, [this on `Step`](https://github.com/TechnologyAdvice/stardust/blob/20e9871b59dff33fa9149736d4b903683fa1ccfe/src/elements/Step/Step.js#L27).

image

src/collections/Menu/Menu.js
levithomason: I don't think we should add the click handler if the user is using subcomponents.  We did this initially elsewhere, but there is just no reliable way to determine which children we should add it to.

React.Children can only loop through the top level children.  If the user uses their own menu item component, then we have no reliable `type` or `displayName` to identify which components are menu items.  They may also use a higher order component or decorator on the MenuItem, in which case the menu item is no longer a top level child but a nested child wrapped in the decorator.

Because of this, I think the only reliable way to pass props to children is when using the `<Menu items={} />` prop.  In this case, we know we can add props to every object in the array.  It is the best way I know of to manage the parent-child coupling.

More here: https://facebook.github.io/react/docs/context.html#parent-child-coupling

**EDIT**

If the user uses subcomponents, they will just have to wire their own click handlers.  The same is true for our `<Dropdown options={} />` vs `<Dropdown><Dropdown.Item /></Dropdown>` markup.

src/collections/Menu/MenuItem.js
levithomason: Good catch.  It looks like links are perhaps not quite consistent enough to render the `a` tag across all components.  Though, having the `getElementType()` pattern would still be of value as it would allow us to standardize the `as` prop and render `a` tags for `href`.  Also, as we find any new logic for conditionally rendering element types, it would go here.

You are correct though, the `link` prop should be handled individually by each component.  Here is some more context on how it is used in SUI:

### Steps

``` html
<a class="step">

<div class="link step">
```

> http://semantic-ui.com/elements/step.html#link

### Card:

``` html
<a class="ui card" href="http://www.dog.com">
<div class="ui link card">
```

> http://semantic-ui.com/views/card.html#link-card

### Label:

``` html
<a class="ui label"></a>

<div class="ui label">
  <a class="detail">View Mail</a>
</div>
```

> http://semantic-ui.com/elements/label.html#link

### Image:

``` html
<a href="http://google.com" class="ui medium image">
  <img src="/images/wireframe/image-text.png">
</a>
```

> http://semantic-ui.com/elements/image.html#image-link

### List:

``` html
<div class="ui link list">
  <div class="active item">Home</div>
  <a class="item">About</a>
  <a class="item">Jobs</a>
  <a class="item">Team</a>
</div>
```

> http://semantic-ui.com/elements/list.html#link

levithomason:
Looks like a rebase was done from an out of date master.  Update your local master, and the rebase from there.  On the `fix/menu` branch I'd do:

```
$ git fetch origin master
$ git rebase origin/master
```

**EDIT**
If you're remote name is not `origin`, then use your remote name for the main Stardust repo.

src/collections/Menu/Menu.js
layershifter: Take a look on basic example of [Menu](http://semantic-ui.com/collections/menu.html). After click on `item` it switches `active` class to this item. I've implemented it like it is in `Accordion`, but as you say this solution has problem.

I'm correctly understand that you suggest to use `Content`? But, as you answered earlier:

> Context is an advanced and experimental feature. The API is likely to change in future releases.

docs/app/Examples/collections/Menu/Content/SubMenu.js
levithomason: Hm, how about `Menu.Submenu`?  Follows the `Subheader` naming pattern.

levithomason: Technically, it should be `Menu.Menu`.  Let me think about this one for a little.  I don't want to suddenly break the pattern of subcomponents matching SUI component part class names.

layershifter: ~~Not bad, better than my :smile:  I'll implement it.~~

EDIT: Will wait for naming deсision.

layershifter:
Seems, I'll left it for you on merge :cold_sweat: 

image

levithomason:
no problem, one moment while i merge and force push...

levithomason:
Give it a look and see if it is correct now.  I have a copy of the original branch just in case I hosed this one.

layershifter:
Looks okay, thanks :+1: Waiting for your reply [there](https://github.com/TechnologyAdvice/stardust/pull/382#discussion_r75188248).

src/collections/Menu/Menu.js
levithomason: I'm proposing two separate usage patterns.

## items={} prop

My suggestion is that this usage will automatically manage the active item state:

``` jsx
<Menu items={[{}, {}, ...]} />
```

With this usage we can add the onClick handlers our selves and manage the `active` prop on the `MenuItem`s internally.  We can do this easily since we must map over all the `items` to create the `MenuItems`.

## Subcomponents

This usage would not self manage the active item.  The user needs to implement their own `onClick` handlers and set the `active` prop:

``` jsx
class MenuExample extends Component {
  handleItemClick = (name) => this.setState({ activeItem: name })

  render() {
    const { activeItem } = this.state
    return (
      <Menu>
        <Menu.Item 
          active={activeItem === 'home'} 
          onClick={this.handleItemClick.bind('home')} 
        />
        <Menu.Item 
          active={activeItem === 'search'} 
          onClick={this.handleItemClick.bind('search')} 
        />
        <Menu.Item 
          active={activeItem === 'settings'} 
          onClick={this.handleItemClick.bind('settings')} 
        />
      </Menu>
    )
  }
}
```

---

These ideas can be improved on, but I think we should only auto manage the active state if you use the `items` prop.

layershifter: It's reasonable, but it produce [warning](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) of ESLint :disappointed:

levithomason: True, our eslint settings don't like `bind` in an event handler.  This can be solved easily like so:

``` js
onClick={() => this.handleItemClick('home')}
```

layershifter: image
image

It seems that ESLint is more intelligent that you think :smiley_cat:

docs/app/Examples/collections/Menu/Content/SubMenu.js
layershifter: I'm added it as `Menu.Menu`, but it's ugly.

layershifter:
`Menu` widely uses [`Popup`](https://github.com/TechnologyAdvice/stardust/issues/193) in docs. Are there any chance for its implementaion in near future?

layershifter:
`Fixed` examples using `<iframe>`, how I can correctly deal with this?

For finishing **Content** and **States** part of docs I need #414.

levithomason:
I paused the Form work to finish the augmentation.  I hope to ship that today.  I can do the popup next, or if you'd like to get a start on it we can tackle that one together.

docs/app/Examples/collections/Menu/Content/SubMenu.js
levithomason: I think that is the most correct for now.  👍

levithomason:
I've started the Tab component, #430.  It is made of a Menu and Segments.  I think if MenuItem's accept a name prop, it could help the Tab component.  It would also help our onClick binding as we can callback with the Menu.Item name:

``` jsx
export default class Menus extends Component {
  handleItemClick = (e, name) => this.setState({ activeItem: name })

  render() {
    const { activeItem } = this.state

    return (
      <Menu>
        <Menu.Item
          name='browse'
          active={activeItem === 'browse'}
        >
          Browse
        </Menu.Item>

        <Menu.Item
          name='submit'
          active={activeItem === 'submit'}
        >
          Submit
        </Menu.Item>

        <Menu.Menu position='right'>
          <Menu.Item
            name='signup'
            active={activeItem === 'signup'}
          >
            Sign Up
          </Menu.Item>

          <Menu.Item
            name='help'
            active={activeItem === 'help'}
          >
            Help
          </Menu.Item>
        </Menu.Menu>
      </Menu>
    )
  }
}
```

It can be used as shorthand props to generate the menu item text.  Same example as above but with shorthand MenuItem child text:

``` jsx
export default class Menus extends Component {
  handleItemClick = (e, name) => this.setState({ activeItem: name })

  render() {
    const { activeItem } = this.state

    return (
      <Menu>
        <Menu.Item
          name='browse'
          active={activeItem === 'browse'}
        />
        <Menu.Item
          name='submit'
          active={activeItem === 'submit'}
        />
        <Menu.Menu position='right'>
          <Menu.Item
            name='signup'
            active={activeItem === 'signup'}
          />
          <Menu.Item
            name='help'
            active={activeItem === 'help'}
          /></Menu.Menu>
      </Menu>
    )
  }
}
```

Lastly, it works well with a shorthand items array where the objects are spread on each item.  It could also be used to propagate the onClick to the Menu when using an array of items.

This would be an AutoControlledComponent now.  The active item can be self managed.  Optionally, a user could override the active item with an `activeItemName` prop:

``` jsx
const items = [
  { name: 'browse' },
  { name: 'submit' },
  { name: 'signup' },
  { name: 'help' },
]

export default class Menus extends Component {
  handleClick = (e, name) => {
    // do something with the clicked item
  }

  render() {
    return (
      <Menu onClick={this.handleClick} items={items} />
    )
  }
}
```

layershifter:
Looks pretty, take it work.

layershifter:
Added. I'm having problems with docs build on Windows, on `master` branch, too. And idea?

image

levithomason:
Gah, yep!  There are hardcoded `/`s in the paths.  Windows uses `\`s :/  I don't have a quick solve for this off the top of my head.

The `docgenInfo.json` is generated in node during the build.  The ComponentDoc components then import and parse the json in browser.  So, the browser code needs to know to use `/` instead of `\`.

One idea might be setting a global platform dependent `__PATH_SEP__` during the build.

``` js
// config.js
compiler_globals: {
  __PATH_SEP__: JSON.stringify(path.sep),
}
```

> https://nodejs.org/api/path.html#path_path_sep

Then we could use `__PATH_SEP__` in the doc build when building paths.  If you like, please open another issue and we'll work on this there.

levithomason:
Merged #435, give that a shot!

levithomason:
In addition to calling back with the menu item name onClick, let's also call back with the item index:

``` jsx
handleClick(e, { name, index}) {
  //... 
} 

<Menu.Item onClick={this.handleClick} />
```

This will help other components like Tab and Accordion use the Menu to control their activeIndex.

layershifter:
How I can get `index` when user passes children directly?

``` jsx
<Menu>
  <Menu.Item>A</Menu.Item>
  <Menu.Item>B</Menu.Item>
</Menu>
```

levithomason:
Hm.  We would not be able to do this reliably with Menu.Item children, only Menu items shorthand?

``` jsx
const items = [...]

handleClick(e, { name, index }) {
  // ...
}

<Menu items={items} onClick={this.handleClick}/>
```

layershifter:
> We would not be able to do this reliably with Menu.Item children, only Menu items shorthand?

Yes, only with `items` shorthand :cry:

levithomason:
Perhaps then we do not add the `index`.  I'd rather keep the arguments consistent in all cases.

layershifter:
**[Fixed](http://semantic-ui.com/collections/menu.html#fixed)** example uses `<iframe>`, what can I do there?

levithomason:
Hm, this is possible but we'd need to render an app in the iframe.  If we did this, I think we'd want to create a iframe component that abstracted all that away and allowed us to use children somehow.

In the meantime, why don't we just include a Message for the example that references the SUI docs.  Adding an iframe seems like a lot of work and complexity for a single example.

levithomason:
I am currently splitting the Feed examples and Feed prop examples into separate files.  Let's do this from now on.  You can reference this comment for more info and an example:

https://github.com/TechnologyAdvice/stardust/pull/344#discussion_r74088351

levithomason:
Rebasing this to the latest master...

levithomason:
I've updated this to the latest master.  I also fixed the existing tests and added missing sub component tests.  It should be all set for continuing with the docs.

Since I squashed and force pushed to clean it up a bit, you'll probably want to blow away your local branch in favor of this one.

I did not make any major updates to the functionality.  So, the Menu itself is still in progress.

layershifter:
Thanks much, I'll think I'll finish docs on this weekend.

docs/app/Examples/collections/Menu/Variations/ColoredInvertedMenus.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const ColoredInvertedMenus = () => {
+  const colors = Menu._meta.props.color
+  const menus = colors.map(color => <ExampleMenu color={color} key={color} />)
+
+  return <div>{menus}</div>
+}
+
+export default ColoredInvertedMenus
```
layershifter, line 38: @levithomason what you think about this?

docs/app/Examples/collections/Menu/Types/Attached.js
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <div>
+      <Menu attached='top'>
+        <Menu.Item as={Dropdown} icon='wrench'>
+          <Dropdown.Menu>
+            <Dropdown.Item>
+              <Icon name='dropdown icon' />
+              <span className='text'>New</span>
+
+              <Dropdown.Menu>
```
layershifter, line 17: Seems there is bug in `Dropdown` with handling sub-menus or I missed something?

docs/app/Examples/collections/Menu/Types/Attached.js
```diff
@@ -0,0 +1,49 @@
...
+                <Dropdown.Item>Document</Dropdown.Item>
+                <Dropdown.Item>Image</Dropdown.Item>
+              </Dropdown.Menu>
+            </Dropdown.Item>
+            <Dropdown.Item>Open</Dropdown.Item>
+            <Dropdown.Item>Save...</Dropdown.Item>
+            <Dropdown.Item>Edit Permissions</Dropdown.Item>
+            <Dropdown.Divider />
+            <div className='header'>Export</div>
```
layershifter, line 26: Seems there is missing `Dropdown.Header` component. Can I make PR for this?

docs/app/Examples/collections/Menu/Types/TabularOnLeft.js
```diff
@@ -0,0 +1,58 @@
...
+              name='links'
+              active={activeItem === 'links'}
+              onClick={this.handleItemClick}
+            >
+              Links
+            </Menu.Item>
+          </Menu>
+        </Grid.Column>
+
+        <Grid.Column stretched width={12}>
```
layershifter, line 23: There are no no mention about this property on `Column` is SUI docs. Will open PR for this.

layershifter:
### Docs

Unbelievable, but it is finished :v: We need review for them.

### Tests

Tests will be finished tomorrow. And final review will there soon :+1: 

### Shorthands

Discuss?

#### Menu.Menu

I think, we don't need provide `items` shorthand there because it's sub menu componen while shorthands cover only simplest cases.

### Menu
- Autocontroll of index prop is pretty idea, but we need provide item handler. What you think about this?

``` jsx
const items [
  {name: 'Open'},
  {name: 'Save'},
]

<Menu items={items} onItemClick={handler} />
```

``` jsx
const items [
  {name: 'Open', onClick: onOpenClick},
  {active: true, name: 'Save', onClick: onSaveClick},
]

<Menu items={items} />
```
- Skip `Menu.Header` and `Menu.Menu` for less complexity.

docs/app/Examples/collections/Menu/Types/Attached.js
levithomason: The Dropdown has very limited support for custom markup.  See #416 for more.  It is best to use a Dropdown with an `options` prop instead of sub components for now.

We can leave this PR as is since it is since you've already put so much into it.

docs/app/Examples/collections/Menu/Types/Attached.js
levithomason: Please! 👍

docs/app/Examples/collections/Menu/Types/TabularOnLeft.js
levithomason: Merged/released, thanks much.

levithomason:
**Docs** - Wow, great work!  Will get a review soon as I can.
**Tests** - LMK if/when these are done and I'll review as well.
**Shorthands** - I agree completely on Menu.Menu and the Menu items, onItemClick handler, and skipping header/sub menu.  Anything more complicated can use sub components.

docs/app/Examples/collections/Menu/Variations/ColoredInvertedMenus.js
levithomason: The loop is perfectly acceptable, but let's not show use of the private `_meta` prop.  I don't want users knowing/using that.  I think putting it in the docs will encourage use of it.  Instead, let's just list the colors in an array explicitly.

layershifter:
### Tests

There are here :ship: 

### Shorthands

I'll implement and add docs with tests soon :sleepy:

levithomason:
Will review this today, thanks for the massive amount of work here!

levithomason:
Heads up, doing a final review and some minor tweaks.  Hope to merge this today.

levithomason:
Latest updates:

### Start Case Name prop

Updated `<Menu.Item name='fooBar' />` to Start Case the `name` (i.e. "Foo Bar") when used as content.  Added a test for this and updated the examples as well.

### Dropdown Item Updated

Completed TODO for Dropdown header usage.  Also, reversed the augmentation from `<Menu.Item as={Dropdown} />` to `<Dropdown as={Menu.Item} />` as it seemed the Dropdown sub components made more sense this way.  Finally, used `icon` prop shorthand for Dropdown.Items.

### Colored Menus

Mapped over an array of colors to make examples a little more concise.  More importantly, to not expose the `_meta` prop to users in the docs.  I am sure it will lead to wider spread use of the `_meta` prop which is private.

### Misc

I made some trivial updates to the docs in terms of verbiage and/or organization.

levithomason:
🎉 This is great, thanks a million.  We can do shorthand separately since this PR was so massive.

levithomason:
Released in `stardust@0.42.0`

PR #387, opened 2016-08-11
Card: add component

athurman:
Adds Card component to v1 API

http://semantic-ui.com/views/card.html
# Types
## [Fluid Card](http://semantic-ui.com/views/card.html#fluid-card)

``` html
<div class="ui fluid card">
  ...
</div>
```

``` jsx
<Card fluid />
```
## [Centered Card](http://semantic-ui.com/views/card.html#centered-card)

``` html
<div class="ui centered card">
  ...
</div>
```

``` jsx
<Card centered />
```
## [Raised Card](http://semantic-ui.com/views/card.html#raised-card)

``` html
<div class="ui raised card">
  ...
</div>
```

``` jsx
<Card raised />
```
## [Link Card](http://semantic-ui.com/views/card.html#link-card)

``` html
<a class="ui card" href="http://www.dog.com">
  ...
</a>
```

``` jsx
<Card href='http://www.dog.com' />

<Card onClick={(e) => (this.onClick(e))} />
```
## [Colored Card](http://semantic-ui.com/views/card.html#colored)

``` html
<div class="ui red card">
  ...
</div>
```

``` jsx
<Card color='red' />
```
## [Cards](http://semantic-ui.com/views/card.html#cards)

A group of cards

``` html
<div class="ui cards">
  <div class="card">
    ...
  </div>
  <div class="card">
    ...
  </div>
</div
```

``` jsx
<Card.Group cards={[{}]} />

<Card.Group>
  <Card color='red' />
  <Card color='orange' />
</Card.Group>
```
## [Column Count](http://semantic-ui.com/views/card.html#column-count)

defines how many cards should exist in a row

``` html
<div class="ui four cards">
  <div class="card">
    ...
  </div>
  <div class="card">
    ...
  </div>
</div
```

``` jsx
<Card.Group itemsPerRow={4} />
```
## [Stackable](http://semantic-ui.com/views/card.html#stackable)

``` html
<div class="ui three stackable cards">
  <div class="card">
    ...
  </div>
  <div class="card">
    ...
  </div>
</div
```

``` jsx
<Card.Group itemsPerRow={3} stackable />
```
## [Doubling](http://semantic-ui.com/views/card.html#doubling)

``` html
<div class="ui six doubling cards">
  <div class="card">
    ...
  </div>
  <div class="card">
    ...
  </div>
</div
```

``` jsx
<Card.Group itemsPerRow={6} doubling />
```

### DIFF
--- a/docs/app/Examples/views/Card/Content/ExtraContent.js
+++ b/docs/app/Examples/views/Card/Content/ExtraContent.js
@@ -1,0 +1,24 @@
+import React from 'react'
+import { Card, Icon } from 'stardust'
+
+const description = [
+  'Amy is a violinist with 2 years experience in the wedding industry.',
+  'She enjoys the outdoors and currently resides in upstate New York.',
+].join(' ')
+
+const ExtraContent = () => (
+  <Card>
+    <Card.Content>
+      <Card.Header>
+        About Amy
+      </Card.Header>
+    </Card.Content>
+    <Card.Content description={description} />
+    <Card.Content extra>
+      <Icon name='user' />
+      4 Friends
+    </Card.Content>
+  </Card>
+)
+
+export default ExtraContent
--- a/docs/app/Examples/views/Card/Content/LinkCard.js
+++ b/docs/app/Examples/views/Card/Content/LinkCard.js
@@ -1,0 +1,13 @@
+import React from 'react'
+import { Card } from 'stardust'
+
+const LinkCard = () => (
+  <Card
+    href='#link'
+    header='Elliot Baker'
+    meta='Friend'
+    description='Elliot is a sound engineer living in Nashville who enjoys playing guitar and hanging with his cat.'
+  />
+)
+
+export default LinkCard
--- a/docs/app/Examples/views/Card/Types/IndividualCard.js
+++ b/docs/app/Examples/views/Card/Types/IndividualCard.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { Card, Icon, Image } from 'stardust'
+
+const IndividualCard = () => (
+  <Card>
+    <Image src='http://semantic-ui.com/images/avatar2/large/matthew.png' />
+    <Card.Content>
+      <Card.Header>
+        Matthew
+      </Card.Header>
+      <Card.Meta>
+        <span className='date'>
+          Joined in 2015
+        </span>
+      </Card.Meta>
+      <Card.Description>
+        Matthew is a musician living in Nashville.
+      </Card.Description>
+    </Card.Content>
+    <Card.Content extra>
+      <a>
+        <Icon name='user' />
+        22 Friends
+      </a>
+    </Card.Content>
+  </Card>
+)
+
+export default IndividualCard
--- a/docs/app/Examples/views/Card/Variations/Colored.js
+++ b/docs/app/Examples/views/Card/Variations/Colored.js
@@ -1,0 +1,23 @@
+import React from 'react'
+import { Card } from 'stardust'
+
+const src = 'http://semantic-ui.com/images/wireframe/white-image.png'
+
+const ColoredCards = () => (
+  <Card.Group itemsPerRow={4}>
+    <Card color='red' image={src} />
+    <Card color='orange' image={src} />
+    <Card color='yellow' image={src} />
+    <Card color='olive' image={src} />
+    <Card color='green' image={src} />
+    <Card color='teal' image={src} />
+    <Card color='blue' image={src} />
+    <Card color='violet' image={src} />
+    <Card color='purple' image={src} />
+    <Card color='pink' image={src} />
+    <Card color='brown' image={src} />
+    <Card color='grey' image={src} />
+  </Card.Group>
+)
+
+export default ColoredCards
--- a/docs/app/Examples/views/Card/Variations/ColumnCount.js
+++ b/docs/app/Examples/views/Card/Variations/ColumnCount.js
@@ -1,0 +1,23 @@
+import React from 'react'
+import { Card } from 'stardust'
+
+const src = 'http://semantic-ui.com/images/wireframe/image.png'
+
+const ColumnCountCards = () => (
+  <Card.Group itemsPerRow={6}>
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+    <Card raised image={src} />
+  </Card.Group>
+)
+
+export default ColumnCountCards
--- a/docs/app/Examples/views/Card/Variations/Fluid.js
+++ b/docs/app/Examples/views/Card/Variations/Fluid.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Card } from 'stardust'
+
+const FluidCards = () => (
+  <Card.Group>
+    <Card color='red' fluid>
+      <Card.Content header='Option 1' />
+    </Card>
+    <Card color='orange' fluid>
+      <Card.Content header='Option 2' />
+    </Card>
+    <Card color='yellow' fluid>
+      <Card.Content header='Option 3' />
+    </Card>
+  </Card.Group>
+)
+
+export default FluidCards
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -12,6 +12,8 @@ function Card(props) {
     className,
     color,
     fluid,
+    href,
+    onClick,
     raised,
     ...rest,
   } = props
@@ -25,10 +27,20 @@ function Card(props) {
     className,
   )
 
+  const handleClick = (e) => {
+    if (onClick) onClick(e)
+  }
+  const CardComponent = href || onClick ? 'a' : 'div'
+
   return (
-    <div {...rest} className={classes}>
+    <CardComponent
+      {...rest}
+      className={classes}
+      href={href}
+      onClick={handleClick}
+    >
       {children}
-    </div>
+    </CardComponent>
   )
 }
 
@@ -46,6 +58,8 @@ Card.propTypes = {
   className: PropTypes.string,
   color: PropTypes.oneOf(Card._meta.props.color),
   fluid: PropTypes.bool,
+  href: PropTypes.string,
+  onClick: PropTypes.func,
   raised: PropTypes.bool,
 }
 
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -4,6 +4,10 @@ import React, { PropTypes } from 'react'
 import META from '../../utils/Meta'
 import * as sui from '../../utils/semanticUtils'
 import { useKeyOnly } from '../../utils/propUtils'
+import CardContent from './CardContent'
+import CardDescription from './CardDescription'
+import CardHeader from './CardHeader'
+import CardMeta from './CardMeta'
 
 function Card(props) {
   const {
@@ -62,5 +66,10 @@ Card.propTypes = {
   onClick: PropTypes.func,
   raised: PropTypes.bool,
 }
+
+Card.Content = CardContent
+Card.Description = CardDescription
+Card.Header = CardHeader
+Card.Meta = CardMeta
 
 export default Card
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -1,0 +1,29 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import META from '../../utils/Meta'
+
+function Card(props) {
+  const { children, className, ...rest } = props
+  const classes = cx(
+    className,
+    'card',
+  )
+  return (
+    <div {...rest} className={classes}>
+      {children}
+    </div>
+  )
+}
+
+Card._meta = {
+  name: 'Card',
+  type: META.type.view,
+}
+
+Card.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+}
+
+export default Card
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -2,13 +2,29 @@ import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
 import META from '../../utils/Meta'
+import * as sui from '../../utils/semanticUtils'
+import { useKeyOnly } from '../../utils/propUtils'
 
 function Card(props) {
-  const { children, className, ...rest } = props
-  const classes = cx(
+  const {
+    centered,
+    children,
     className,
+    color,
+    fluid,
+    raised,
+    ...rest,
+  } = props
+
+  const classes = cx('ui',
+    useKeyOnly(centered, 'centered'),
+    useKeyOnly(fluid, 'fluid'),
+    useKeyOnly(raised, 'raised'),
+    color,
     'card',
+    className,
   )
+
   return (
     <div {...rest} className={classes}>
       {children}
@@ -19,11 +35,18 @@ function Card(props) {
 Card._meta = {
   name: 'Card',
   type: META.type.view,
+  props: {
+    color: sui.colors,
+  },
 }
 
 Card.propTypes = {
+  centered: PropTypes.bool,
   children: PropTypes.node,
   className: PropTypes.string,
+  color: PropTypes.oneOf(Card._meta.props.color),
+  fluid: PropTypes.bool,
+  raised: PropTypes.bool,
 }
 
 export default Card
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -4,6 +4,7 @@ import React, { PropTypes } from 'react'
 import {
   META,
   SUI,
+  getUnhandledProps,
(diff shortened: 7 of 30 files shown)


src/views/Card/Card.js
```diff
@@ -0,0 +1,29 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import META from '../../utils/Meta'
+
+function Card(props) {
+  const { children, className, ...rest } = props
+  const classes = cx(
+    className,
```
levithomason, line 22: Add `ui` class as the first class.  Then, put the user's className as the last class.

test/specs/views/Card/Card-test.js
```diff
@@ -0,0 +1,9 @@
+import faker from 'faker'
+import React from 'react'
+
+import * as common from 'test/specs/commonTests'
+import Card from 'src/views/Card/Card'
+
+describe('Card', () => {
+  common.isConformant(Card)
```
levithomason: Can add:

```
common.rendersChildren
common.hasUIClassName
```

levithomason:
Great, glad to have this started again.  Let's get the subcomponents in and a doc example going.  Reference the Rating docs for the most recent doc site pattern.

levithomason:
API looks like a good start.  Linking to the docs is helpful but you don't have to go through that if you don't want!

Still thinking about the absence of the `ui` class on the `card` when in `cards`.  Though, as we tested the markup is OK with it included.

We just need to add the subcomponents now:

### Content

``` html
<div class="content"></div>
```

``` jsx
<Card.Content />
```

#### extra

``` html
<div class="extra content"></div>
```

``` jsx
<Card.Content extra />
```

### Header

``` html
<div class="header"></div>
```

``` jsx
<Card.Header />
```

### Description

``` html
<div class="description"></div>
```

``` jsx
<Card.Description />
```

### Meta

``` html
<div class="meta"></div>
```

``` jsx
<Card.Meta />
```

src/views/Card/Card.js
```diff
@@ -2,13 +2,29 @@ import cx from 'classnames'
...
-  const { children, className, ...rest } = props
-  const classes = cx(
+  const {
+    centered,
+    children,
     className,
+    color,
+    fluid,
+    raised,
+    ...rest,
```
levithomason: We have another `propUtil` to get the `rest`:

``` jsx
import { getUnhandledProps } from '../../utils/propUtils'

const rest = getUnhandledProps(Card, props)
```

This is needed since we might not always destructure _all_ the props when using `...rest`.  Especially when there are multiple render methods.  The util creates an object of all the props that your component does not handle (i.e. the rest of the user's props).

levithomason:
I've updated this PR to use the newly refactored utils, see #388.  I also updated the shorthand props to comply with our latest RFC on this, #391.

docs/app/Examples/views/Card/Content/ExtraContent.js
```diff
@@ -0,0 +1,45 @@
...
+  'enjoys',
+  'the',
+  'outdoors',
+  'and',
+  'currently',
+  'resides',
+  'in',
+  'upstate',
+  'New',
+  'York.'].join(' ')
```
levithomason: This strikes me as strange, can we use a regular string here?  If we're fighting max column issues, we can put more content on a single line:

``` jsx
const description = [
  'Amy is a violinist with 2 years experience in the wedding industry.',
  'She enjoys the outdoors and currently resides in upstate New York.',
].join(' ')
```

levithomason:
Before I also forget, here's a reminder to mark the "Card" off in the README.md support table.

levithomason:
@athurman bump:

> Before I also forget, here's a reminder to mark the "Card" off in the README.md support table.

docs/app/Examples/views/Card/Content/LinkCard.js
```diff
@@ -0,0 +1,16 @@
...
+import { Card } from 'stardust'
+
+const LinkCard = () => (
+  <div>
+    <Card href='#link'>
+      <Card.Content
+        header='Elliot Baker'
+        meta='Friend'
+        description='Elliot is a sound engineer living in Nashville who enjoys playing guitar and hanging with his cat.'
+      />
```
levithomason: For shorthand examples where there are no nodes between the Card and Content, let's just put the props on the Card:

``` jsx
<Card 
  href='#link'
  header='Elliot Baker'
  meta='Friend'
  description='Elliot is a sound engineer living in Nashville who enjoys playing guitar and hanging with his cat.'
/>
```

docs/app/Examples/views/Card/Types/IndividualCard.js
```diff
@@ -0,0 +1,39 @@
...
+      </Card.Content>
+    </Card>
+
+    <Card>
+      <Card.Content
+        header='Elliot Baker'
+        meta='Friend'
+        description='Elliot is a sound engineer living in Nashville who enjoys playing guitar and hanging with his cat.'
+      />
+    </Card>
```
levithomason: To assist with the docs, let's separate the shorthand props examples into separate files.  This way there are two descriptions and two code blocks.  It also removes the need for the extra div wrapper.

See this thread for more context and an example: https://github.com/TechnologyAdvice/stardust/pull/344#discussion_r74117525

docs/app/Examples/views/Card/Variations/Colored.js
```diff
@@ -0,0 +1,47 @@
...
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+      <Card color='brown'>
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+      <Card color='grey'>
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+    </Card.Group>
+  </div>
```
levithomason: I think the `Card` should take an `image` shorthand prop.  This will make it more consistent with other component shorthand.  Also, it reduces this markup considerably:

``` jsx
import React from 'react'
import { Card } from 'stardust'

const src = 'http://semantic-ui.com/images/wireframe/white-image.png'

const ColoredCards = () => (
  <Card.Group itemsPerRow={4}>
    <Card color='red' image={src} />
    <Card color='orange' image={src} />
    <Card color='yellow' image={src} />
    <Card color='olive' image={src} />
    <Card color='green' image={src} />
    <Card color='teal' image={src} />
    <Card color='blue' image={src} />
    <Card color='violet' image={src} />
    <Card color='purple' image={src} />
    <Card color='pink' image={src} />
    <Card color='brown' image={src} />
    <Card color='grey' image={src} />
  </Card.Group>
)

export default ColoredCards
```

docs/app/Examples/views/Card/Variations/ColumnCount.js
```diff
@@ -0,0 +1,47 @@
...
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+      <Card raised href='#column-count'>
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+      <Card raised href='#column-count'>
+        <Image src='http://semantic-ui.com/images/wireframe/white-image.png' />
+      </Card>
+    </Card.Group>
+  </div>
```
levithomason: Let's remove the extra `div` wrapper in all examples that can.  React only requires a single element to be returned, the type does not matter.  In this case, it will be a `Card.Group`.

docs/app/Examples/views/Card/Variations/Fluid.js
```diff
@@ -0,0 +1,20 @@
...
+        <Card.Content header='Option 1' />
+      </Card>
+      <Card color='orange' fluid>
+        <Card.Content header='Option 2' />
+      </Card>
+      <Card color='yellow' fluid>
+        <Card.Content header='Option 3' />
+      </Card>
+    </Card.Group>
+  </div>
```
levithomason: Here's another extra `div`, I won't comment on the others.

test/specs/views/Card/Card-test.js
```diff
@@ -0,0 +1,53 @@
...
+  common.hasUIClassName(Card)
+  common.hasSubComponents(Card, [CardContent, CardDescription, CardGroup, CardHeader, CardMeta])
+  common.propKeyOnlyToClassName(Card, 'centered')
+  common.propKeyOnlyToClassName(Card, 'fluid')
+  common.propKeyOnlyToClassName(Card, 'raised')
+  common.rendersChildren(Card)
+
+  describe('renders different elements', () => {
```
levithomason: Prefer we reorganize these describe/it blocks a bit to be consistent with other components, reasoning below:

``` jsx
  it('renders a <div> by default', () => {
    ...
  })

  describe('href', () => {
    it('renders an <a> with an href attr', () => {
      ...
    })
  })

  describe('onClick', () => {
    it('can be omitted', () => {
      ...
    })

    it('renders <a>', () => {
      ...
    })

    it('is called with (event) on click', () => {
      ...
    })
  })
```

I split some of the `it()`s into multiple individual assertions.  Just like writing single purpose functions, targeted test assertions are preferred.  It makes them stronger and easier to maintain as a single test will only fail for a single specific reason.

I moved the default div tagName assertion to the root of the `describe('Card')`.  This is because that assertion is asserting something about the card itself.  It will help with the console output and also reads nicely.  I find it helpful to read the `describe()` strings and `it()` in succession to help figure out how to split tests, since that is how the test runner will also print their failures.  Here's how some of my changes will print:

``` diff
-Card renders different elements <div> by default
+Card renders a <div> by default

-Card renders different elements <a> with `href` prop
+Card href prop renders an <a> with an href attr
```

test/specs/views/Card/CardGroup-test.js
```diff
@@ -0,0 +1,43 @@
...
+        <CardGroup>
+          <Card>{firstText}</Card>
+          <Card>{secondText}</Card>
+        </CardGroup>
+      )
+        .find('Card')
+
+      wrapper.first().should.contain.text(firstText)
+      wrapper.last().should.contain.text(secondText)
+    })
```
levithomason: Let's replace this "with `children` prop" test with `common.rendersChildren(CardGroup)`.

test/specs/views/Card/CardGroup-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+    it('with `items` prop', () => {
```
levithomason, line 18: This would then be a describe.  Good rule of thumb is one describe() per prop.

If there is no prop to describe the behavior, it is probably a top level `it()` assertion.  In some rare cases, there are special scenarios that need a `describe()` outside of this.

test/specs/views/Card/CardHeader-test.js
```diff
@@ -0,0 +1,26 @@
...
+      shallow(<CardHeader content={text} />)
+        .should.contain.text(text)
+    })
+
+    it('renders child node', () => {
+      const child = <div data-child={faker.hacker.noun()} />
+
+      shallow(<CardHeader content={child} />)
+        .should.contain(child)
+    })
```
levithomason, line 24: This is a great example of good test organization 👍  One describe for the prop and multiple single purpose assertions about what it does.

docs/app/Examples/views/Card/Variations/Colored.js
athurman: I initially thought this too, but the issue I ran into with implementing this is, in placing the image within a prop, you lose control of where the image will be placed within the Card. Based on the semantic examples: http://semantic-ui.com/views/card.html#types

An image can go anywhere within the card, inbetween content or at the top, or the bottom - image does not have a particular position the same way it does within `<Feed.Label />` components.

levithomason: This is correct, though we allow shorthand props _or_ subcomponent children to cover all cases.  There are essentially two APIs for every component, shorthand props and subcomponents.

The shorthand props API makes assumptions about where the the values go.  When a component takes an `icon` or `image` prop, it is placed as a sibling immediately before the `content`.  In many components, the presence of the icon/image is what dictates the use of the `content`.

If this use case does not meet the user's needs, then they can use the sub component API and place the Image (along with it's own props) anywhere they choose.

Please let me know if I've overlooked some other issue at play.

docs/app/Examples/views/Card/Content/LinkCard.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Card } from 'stardust'
+
+const LinkCard = () => (
+  <Card
+    href='#link'
```
levithomason, line 6: ~~Whoops, no `#` when defining an `href`.~~

Oh, I see, this is going to link to the example of the same name.  In the future, prefer empty `#` to linking to actual anchors.  No need to update. 👍

docs/app/Examples/views/Card/Variations/ColumnCount.js
```diff
@@ -0,0 +1,23 @@
...
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
+    <Card raised href='#column-count' image={src} />
```
levithomason: Let's remove the `href` here.  Ideally, we show as little features as possible for the example at hand.  In this case, we are showing the `itemsPerRow` prop.  The `raised` and `image` are OK.

src/views/Card/Card.js
```diff
@@ -0,0 +1,129 @@
...
+
+  return (
+    <CardComponent
+      {...rest}
+      className={classes}
+      href={href}
+      onClick={handleClick}
+    >
+      {createImage(image)}
+      {children || <CardContent description={description} header={header} meta={meta} />}
```
levithomason: Content should only be rendered if there is description, header, or meta.  Otherwise, it renders an empty div creating extra whitespace:



Also, currently if a user passes children and an image, they'll get a warning but the component will still render an image and their children beside it.  We want to prevent the bad practice altogether.  We've been doing so with explicit if statements and multiple returns.

``` jsx
if (children) {
  return (
    <CardComponent {...rest} className={classes} href={href} onClick={handleClick}>
      {children}
    </CardComponent>
  )
}

return (
  <CardComponent {...rest} className={classes} href={href} onClick={handleClick}>
    {createImage(image)}
    {(description || header || meta) && (
      <CardContent description={description} header={header} meta={meta} />
    )}
    {extra && <CardContent extra>{extra}</CardContent>}
  </CardComponent>
)
```

Note that in this ^ code snippet, I've also added support for an `extra` prop on the Card which generates a `<CardContent extra />`.  I've also made it conditional so it doesn't render unless provided.  Let's include this and add it to the propTypes.

src/views/Card/CardContent.js
```diff
@@ -0,0 +1,72 @@
...
+  const classes = cx(
+    className,
+    useKeyOnly(extra, 'extra'),
+    'content',
+  )
+  const rest = getUnhandledProps(CardContent, props)
+
+  if (children) {
+    return <div {...rest} className={classes}>{children}</div>
+  }
```
levithomason, line 25: This is a good example of an early return to prevent improper use 👍

test/specs/views/Card/Card-test.js
```diff
@@ -0,0 +1,59 @@
...
+
+  describe('onClick prop', () => {
+    it('can be omitted', () => {
+      const click = () => mount(<Card>{faker.hacker.phrase()}</Card>).simulate('click')
+      expect(click).to.not.throw()
+    })
+
+    it('renders <a> and handles click', () => {
+      const handleClick = sandbox.spy()
+      const wrapper = mount(<Card onClick={handleClick} />)
```
levithomason: `tagName` assertion can be made from a shallow render.

Rendering shallow is good because your test is only rendering your own component (the Card) and not any nested components (CardContent, etc).  If all children are rendered (render(), mount()) then there is a greater chance of your test failing due to the render errors of nested components.

src/views/Card/CardGroup.js
```diff
@@ -0,0 +1,63 @@
...
+    customPropTypes.disallow(['description', 'header']),
+    PropTypes.arrayOf(PropTypes.shape({
+      description: PropTypes.node,
+      meta: PropTypes.node,
+      key: PropTypes.string,
+      header: PropTypes.node,
+    })),
+  ]),
+  itemsPerRow: PropTypes.oneOf(CardGroup._meta.props.width),
+  stackable: PropTypes.bool,
```
levithomason, line 70: All `propTypes` need a doc block.  This is what shows up on the doc site to document the props.  We use the SUI doc descriptions when available.  Here's a couple examples:

``` jsx
/** A group of cards can double its column width for mobile */
doubling: PropTypes.bool,
```

> http://semantic-ui.com/views/card.html#doubling

``` jsx
/** A group of cards can automatically stack rows to a single columns on mobile devices */
stackable: PropTypes.bool,
```

> http://semantic-ui.com/views/card.html#stackable

src/views/Card/Card.js
athurman: Nice catch! thanks for the explanation - I see where the previous code still caused the Image and children to render together.

levithomason:
Released in `v0.35.0`

PR #390, opened 2016-08-13
Statistic: enhance shorthand props and propTypes

layershifter:
### 1. Update README.md

I forget mark `<Step>` as implemented.
### 2. Update usage of `Message`

`<Message>` was updated to v1 API, I've updated its usage in docs.
### 3. Update Statistic's docs

I've updated naming of docs to the most recent pattern.
### 4. Update prop handling
1. Allow only string for shorthands.
2. Shorthands on `StatisticLabel` and `StatisticLabel`.
3. `StatisticGroup`:
   - Update `key` handling for children (antipattern with `index` usage);
   - Extend `prop` validation of `items` with `Proptypes.shape`.


docs/app/Examples/views/Statistic/Types/Groups.js
```diff
@@ -2,13 +2,14 @@ import React from 'react'
 import { Statistic } from 'stardust'
 
 const { Group, Label, Value } = Statistic
+
 const items = [
-  { label: 'Faves', value: 22 },
+  { label: 'Faves', value: '22' },
   { label: 'Views', value: '31,200' },
-  { label: 'Members', value: 22 },
+  { label: 'Members', value: '22' },
```
levithomason, line 9: FYI I think these would work as numbers as well.

src/views/Statistic/StatisticGroup.js
```diff
@@ -61,7 +60,12 @@ StatisticGroup.propTypes = {
   /** Array of props for Statistic. */
   items: customPropTypes.every([
     customPropTypes.disallow(['children']),
-    PropTypes.array,
+    PropTypes.arrayOf(PropTypes.shape({
+      key: PropTypes.string,
```
levithomason, line 64: Turns out the `key` prop is removed by React and is therefore undefined.  When used, these throw browser warnings to use a different prop name to pass the key down.  How about:

```
reactKey
childKey
???
```

src/views/Statistic/StatisticLabel.js
```diff
@@ -36,11 +31,11 @@ StatisticLabel.propTypes = {
   /** Classes that will be added to the StatisticLabel className. */
   className: PropTypes.string,
 
-  // /** Primary content of the StatisticLabel. Mutually exclusive with the children prop. */
-  // content: customPropTypes.every([
-  //   customPropTypes.disallow(['children']),
-  //   PropTypes.node,
-  // ]),
+  /** Primary content of the StatisticLabel. Mutually exclusive with the children prop. */
+  label: customPropTypes.every([
```
levithomason, line 35: Now that we've had some time to consider this, I think `content` is the best option here.  Many components have a `content` section which is the main body of the component.  So this make sense to have the prop called "content".

Also, many components have a `text` prop (Dropdown, Container, Statistic, etc) that is already taken for another purpose.  Given this, let's use `content` everywhere.  I'll open and issue to update all other components to use `content`.

src/views/Statistic/StatisticValue.js
```diff
@@ -37,14 +32,14 @@ StatisticValue.propTypes = {
...
-  // content: customPropTypes.every([
-  //   customPropTypes.disallow(['children']),
-  //   PropTypes.node,
-  // ]),
-
   /** Format the value with smaller font size to fit nicely beside number values. */
   text: PropTypes.bool,
+
+  /** Primary content of the StatisticValue. Mutually exclusive with the children prop. */
+  value: customPropTypes.every([
```
levithomason, line 39: See comment above https://github.com/TechnologyAdvice/stardust/pull/390/files#r74691117

levithomason:
Thanks much for this!  A few comments and we can merge.

docs/app/Examples/views/Statistic/Types/Groups.js
layershifter, line 9: Nope :smile: `22` is number, while `'22'` is string.

Ideas?

src/views/Statistic/StatisticGroup.js
layershifter, line 64: `childKey` looks fine :+1:

levithomason, line 64: Shall I update the PR or are you working on it?

layershifter, line 64: Already updated it.

docs/app/Examples/views/Statistic/Types/Groups.js
levithomason, line 9: I think we should just change the propType validation to string/number.  Since technically it _could_ support `node` as it ends up here:

``` jsx
<div {...rest} className={classes}>{children || value}</div>
```

layershifter, line 9: It cannot be `node`, as it's used for key on `StepGroup`.

---

``` js
Proptypes.oneOf([Proptypes.number, Proptypes.string])
```

I think that it's not good solution. This raises the question of the behavior of other properties.

levithomason, line 9: Good call.  OK, perhaps we only stick with string/number then.

layershifter, line 9: If we use string/number, we need allow it to all shorthand props.

levithomason:
~~Anything else on this guy or is it ready for merging?~~

Replied on shorthand string/number types.

docs/app/Examples/views/Statistic/Types/Groups.js
levithomason, line 9: Hm, I suppose you are right about this.  Feel free to update here or we can merge this and make that update next.

levithomason, line 9: I'll make a separate PR for this actually.  The PR will resolve #391.

layershifter:
I'll update it now with `oneOf`, so we will can merge it.

levithomason:
Great, I'll merge this tomorrow.  It's getting late here! 🌔

levithomason:
Released in `v0.34.2`

PR #393, opened 2016-08-14
Update keyboardKey util to latest specs

levithomason:
This PR updates OS/ContextMenu key names to Meta to meet the spec:
#### `Meta`

The Meta key. Allows issuing special command inputs. This is the <kbd>Windows</kbd> logo key, or the<kbd>Command</kbd> or <kbd>⌘</kbd> key on Mac keyboards.  

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key


levithomason:
This is an ugly rabbit hole that should be addressed later.  We don't have immediate need for the Meta key.  Refactoring means breaking tests and figuring out a way to store multiple Meta keys with 3 different codes.  This is not possible with the current hashes nor methods for converting back and forth.

PR #394, opened 2016-08-14
Image: add inline prop

layershifter:
`<Image>` component doesn't have `inline` prop while it [exists](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/image.less#L73). This PR fixes it.


levithomason:
Awesome, I just tried to use `inline` intuitively on the doc site.  Resorted to using `spaced='right'`.  Thanks for the fix!

levithomason:
Released in `v0.34.2`

PR #400, opened 2016-08-16
feat(Form): update to v1 API

### LINKED ISSUE #247: Remove jQuery dependency

### LINKED ISSUE #432: Hoist propTypes for all wrapper components
This work is instrumental in facilitating the doc site component explorer and sandbox, see #427.

Some of our components are simple wrappers around others.  When wrapping other components, all props are spread.  Because since they did not access properties of the props object, propTypes definitions were not required by linting.  See:
1. Radio
2. ~~Header / HeaderH[1-6]~~  _Made obsolete by augmentation, see #414._
3. Textarea
4. Select

We do however want to hoist the propTypes.  Not only for better stack traces, but also for docs.  Docs are generated from propTypes.  Without any propTypes defined on the wrapper components, there are no docs.

Where Radio composes Checkbox, propTypes should be hoisted like so:

``` jsx
Radio.propTypes = {
  ...Checkbox.propTypes,
}
```

levithomason:
Fixes #247 Remove jQuery Dependency
Fixes #432 Hoist propTypes for all wrapper components

This PR addresses the final item in removing jQuery.  Since the last component is the Form, this PR also updates the Form to the v1 API and handles validation.
## Reviewing

Try out the Form docs first and look at the code snippets to get a good feel for the API. Then, focus on src/collections/Form/*.js.  Most of the code here is just docs,  tests, and updating old Form usages. 

Also, have a look at the line comments I left below. 
### Form
- The entire component API has been replaced, see docs.
- Removed jQuery validation.  React friendly validation will be added soon.
### Checkbox

Confusion between radio, checkbox, slider, toggle, and various combinations thereof have been addressed.  See the docs for updated usage.
#### inputType & type

The `inputType` prop has been removed, set `type` instead.

The `type` prop is no longer used to set visual SUI type ("slider", "toggle", "checkbox", "radio").  It now controls the type of the html input.
#### inputType & type

See Checkbox
#### Exclusive Groups

Radio groups (multiple radios with the same name) must be controlled components.  Previously we relied on the browser to manage radio group state and used refs to read the `checked` state.  Reading from the DOM is an antipattern.


brsanthu:
@levithomason any idea when this might be merged?

levithomason:
I'm actively working on this, though I suspect it will  be a few days of work.  I'm considering shipping the API without Form validation first.  Then, shipping the validation later.  This could save a substantial amount of time.

This is because validation is something we're still tossing ideas around on.  Input highly encouraged!

brsanthu:
I think shipping without validation and adding that later is good idea. When validations are added, it would be progressive and (I hope) will not have any api impact.

levithomason:
The basic API should not change when validation is added.

levithomason:
Confirmed, validation will be done separately.  See #407 for that discussion.

src/collections/Form/Form.js
```diff
@@ -1,143 +1,261 @@
...
-    _.invoke(this, 'element.off')
+    return name
   }
+}
 
-  static _meta = {
-    name: 'Form',
-    type: META.TYPES.COLLECTION,
-  }
+function formSerializer(formNode) {
```
levithomason, line 52: Heads up, this serializer reads from the DOM.  This is a React antipattern, though short of the validation PR which will handle input binding, this is how we're achieving serialization.

src/lib/AutoControlledComponent.js
```diff
@@ -54,7 +54,7 @@ export default class AutoControlledComponent extends Component {
       } else if (prop === 'checked') {
         res[prop] = false
       } else if (prop === 'value') {
-        res[prop] = ''
+        res[prop] = this.props.multiple ? [] : '' // eslint-disable-line react/prop-types
```
levithomason, line 57: A select element value must be an array if `multiple` is true.  Dropdowns render a hidden select.  We need to use an empty array as the default value when multiple, instead of an empty string.

src/modules/Checkbox/Checkbox.js
```diff
@@ -107,57 +108,51 @@ export default class Checkbox extends Component {
...
   handleClick = (e) => {
     debug('handleClick()')
+    debug('firing change event')
+
     const { disabled, onChange, onClick, name, readOnly, value } = this.props
-    // using a ref here allows us to let the browser manage radio group state for us
-    // this is a special exception where we are reading state from the DOM
-    // otherwise, all radio groups would have to be controlled components
-    const refChecked = _.get(this.refs, 'input.checked')
+    const { checked } = this.state
```
levithomason, line 117: Did away with refs for Checkboxes.  This is one more antipattern off the list.  Ideally, we never read from the real DOM but store all state in the virtual DOM only.

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -23,7 +23,12 @@ export default class ComponentProps extends Component {
     }
   }
   defaultValueRenderer = (item) => {
-    const defaultValue = _.get(item, 'defaultValue.value')
+    let defaultValue = _.get(item, 'defaultValue.value')
+
+    if (_.startsWith(defaultValue, 'function ')) {
+      defaultValue = defaultValue.match(/^function(.*)\{/)[1].trim()
```
dvdzkwsk, line 29: Is this to get the function name? If I'm reading it correctly, seems like you could also do:

``` js
if (typeof defaultValue === 'function') {
  defaultValue = defaultValue.name
}
```

I could be totally reading this wrong, so ignore if that's the case.

docs/app/Examples/addons/Radio/Types/index.js
```diff
@@ -23,7 +25,14 @@ const RadioTypesExamples = () => (
       title='Radio Group'
       description='Radios can be part of a group.'
       examplePath='addons/Radio/Types/RadioGroup'
-    />
+    >
+      <Message warning>
+        Radios in a group must be
```
dvdzkwsk, line 30: `&nbsp;` should probably be after the `be`? Otherwise it doesn't achieve anything different than a `{' ' }`.

docs/app/Components/ComponentDoc/ComponentProps.js
levithomason: You are half correct, this _is_ to get the function name.  However, the defaultValue here is a json representation of the function from react-docgen.  The defaultValue is the actual function stringified.

See `"value": "function formSerializer(formNode) {\n  debug('formSerializer...`:

``` json
"serializer": {
  "type": {
    "name": "func"
  },
  "required": false,
  "defaultValue": {
    "value": "function formSerializer(formNode) {\n  debug('formSerializer()')\n  const json = {}\n  // handle empty formNode ref\n  if (!formNode) return json\n\n  // ----------------------------------------\n  // Checkboxes\n  // Single: { name: value|bool        }\n  // Group:  { name: [value|bool, ...] }\n\n  _.each(formNode.querySelectorAll('input[type=\"checkbox\"]'), (node, index, arr) => {\n    const name = getNodeName(node)\n    const checkboxesByName = _.filter(arr, { name })\n\n    // single: (value|checked)\n    if (checkboxesByName.length === 1) {\n      json[name] = node.checked && node.value !== 'on' ? node.value : node.checked\n      debugSerializedResult(json, name, node)\n      return\n    }\n\n    // groups (checked): [value, ...]\n    if (!Array.isArray(json[name])) json[name] = []\n    if (node.checked) json[name].push(node.value)\n    debugSerializedResult(json, name, node)\n\n    // in dev, warn about multiple checkboxes with a default browser value of \"on\"\n    if (process.NODE_ENV !== 'production' && node.value === 'on') {\n      const errorMessage = [\n        \"Encountered a checkbox in a group with the default browser value 'on'.\",\n        'Each checkbox in a group should have a unique value.',\n        \"Otherwise, the checkbox value will serialize as ['on', ...].\",\n      ].join(' ')\n      console.error(errorMessage, node, formNode) // eslint-disable-line no-console\n    }\n  })\n\n  // ----------------------------------------\n  // Radios\n  // checked: { name: checked value }\n  // none:    { name: null }\n\n  _.each(formNode.querySelectorAll('input[type=\"radio\"]'), (node, index, arr) => {\n    const name = getNodeName(node)\n    const checkedRadio = _.find(arr, { name, checked: true })\n\n    if (checkedRadio) {\n      json[name] = checkedRadio.value\n    } else {\n      json[name] = null\n    }\n\n    debugSerializedResult(json, name, node)\n\n    // in dev, warn about radios with a default browser value of \"on\"\n    if (process.NODE_ENV !== 'production' && node.value === 'on') {\n      const errorMessage = [\n        \"Encountered a radio with the default browser value 'on'.\",\n        'Each radio should have a unique value.',\n        \"Otherwise, the radio value will serialize as { [name]: 'on' }.\",\n      ].join(' ')\n      console.error(errorMessage, node, formNode) // eslint-disable-line no-console\n    }\n  })\n\n  // ----------------------------------------\n  // Other inputs\n  // { name: value }\n\n  _.each(formNode.querySelectorAll('input:not([type=\"radio\"]):not([type=\"checkbox\"])'), (node) => {\n    const name = getNodeName(node)\n    json[name] = node.value\n    debugSerializedResult(json, name, node)\n  })\n\n  // ----------------------------------------\n  // Other inputs and text areas\n  // { name: value }\n\n  _.each(formNode.querySelectorAll('textarea'), (node) => {\n    const name = getNodeName(node)\n    json[name] = node.value\n    debugSerializedResult(json, name, node)\n  })\n\n  // ----------------------------------------\n  // Selects\n  // single:   { name: value }\n  // multiple: { name: [value, ...] }\n\n  _.each(formNode.querySelectorAll('select'), (node) => {\n    const name = getNodeName(node)\n\n    if (node.multiple) {\n      json[name] = _.map(_.filter(node.querySelectorAll('option'), 'selected'), 'value')\n    } else {\n      json[name] = node.value\n    }\n\n    debugSerializedResult(json, name, node)\n  })\n\n  return json\n}",
    "computed": false
  },
  "docBlock": {
    "description": "Called onSubmit with the form node that returns the serialized form object",
    "tags": []
  }
}
```

src/addons/Radio/Radio.js
```diff
@@ -24,10 +27,22 @@ Radio.propTypes = {
...
+   * Display as a checkbox, radio, slider, or toggle.
+   * The input type is `checkbox` for both slider and toggle types.
+   * You can set `inputType` separately to mix and match appearance and behavior.
+   */
+  type: Checkbox.propTypes.type,
 }
 
 Radio.defaultProps = {
   as: Checkbox,
+  inputType: 'radio',
```
dvdzkwsk: Why would a `Radio` ever not be an input type of `radio`? The overloading of `type`, where `inputType` is _actually_ the input `type`, seems pretty confusing. I could imagine a counterargument, to my first question, where users of `Radio` will rarely change the input type, which would mean `type` could be overloaded for other purposes (the styling of the input, as it currently is).

Still, just seems super confusing that `type` is no longer the actual type of the input.

src/collections/Form/Form.js
```diff
@@ -1,143 +1,264 @@
...
+
+    // groups (checked): [value, ...]
+    if (!Array.isArray(json[name])) json[name] = []
+    if (node.checked) json[name].push(node.value)
+    debugSerializedResult(json, name, node)
+
+    // in dev, warn about multiple checkboxes with a default browser value of "on"
+    if (process.NODE_ENV !== 'production' && node.value === 'on') {
+      const errorMessage = [
+        "Encountered a checkbox in a group with the default browser value 'on'.",
```
dvdzkwsk, line 82: This message makes sense because you explained it to me, but I think a user seeing this would be pretty lost. Maybe a link to an issue or a simplified rewrite would help.

src/collections/Form/Form.js
```diff
@@ -1,143 +1,264 @@
...
+    } else {
+      json[name] = null
+    }
+
+    debugSerializedResult(json, name, node)
+
+    // in dev, warn about radios with a default browser value of "on"
+    if (process.NODE_ENV !== 'production' && node.value === 'on') {
+      const errorMessage = [
+        "Encountered a radio with the default browser value 'on'.",
```
dvdzkwsk, line 110: ^^

src/collections/Form/Form.js
```diff
@@ -1,143 +1,264 @@
...
+  })
+
+  // ----------------------------------------
+  // Other inputs
+  // { name: value }
+
+  _.each(formNode.querySelectorAll('input:not([type="radio"]):not([type="checkbox"])'), (node) => {
+    const name = getNodeName(node)
+    json[name] = node.value
+    debugSerializedResult(json, name, node)
```
dvdzkwsk, line 125: Could we eliminate these calls in production? I know they are noop, but this is a lot of calls each time a form input is changed.

src/collections/Form/Form.js
```diff
@@ -1,143 +1,264 @@
...
+  })
+
+  // ----------------------------------------
+  // Other inputs and text areas
+  // { name: value }
+
+  _.each(formNode.querySelectorAll('textarea'), (node) => {
+    const name = getNodeName(node)
+    json[name] = node.value
+    debugSerializedResult(json, name, node)
```
dvdzkwsk, line 135: ^

src/collections/Form/Form.js
```diff
@@ -1,143 +1,264 @@
...
-      'form'
-    )
-    const props = getUnhandledProps(Form, this.props)
-    const ElementType = getElementType(Form, this.props)
-    return (
-      <ElementType {...props} className={classes} ref='element'>
-        {this.props.children}
-      </ElementType>
-    )
+    debugSerializedResult(json, name, node)
```
dvdzkwsk, line 152: ^

src/collections/Form/Form.js
dvdzkwsk, line 82: Sorry, forgive me, this is only called `onSubmit` it seems. Still would be nice to eliminate the unneeded calls.

webpack.config.js
```diff
@@ -81,7 +81,6 @@ webpackConfig.plugins = [
     versions: {
       sui: require('semantic-ui-css/package.json').version,
       faker: require('faker/package.json').version,
-      jquery: require('jquery/package.json').version,
```
dvdzkwsk, line 84: 🎉

dvdzkwsk:
Looks awesome, that was a ton of work, so massive props on getting it so solid.

My only real hold up is the conflating of the `type` property for the `Radio` component. If that can be resolved you have my emoji.

docs/app/Components/ComponentDoc/ComponentProps.js
dvdzkwsk: Gotcha, makes sense. Thanks for the explanation. Comment resolved.

src/addons/Radio/Radio.js
levithomason: Though this PR does not refactor the Radio API, I agree completely on this.  I struggled with the naming there.

A SUI Checkbox can have an input type (radio|checkbox) and one of four "Checkbox types" (checkbox|radio|slider|toggle). The confusing thing to me about the current Checkbox API is the `type` prop, is it the input type or the SUI Checkbox type?

This is what spurred the Radio component, separation of input `type`.  We still have confusion of the "SUI Checkbox type".  Now that we a Radio, maybe we remove the ability to set the input type at all.  Intuition says the input type is already set by the fact that it is a checkbox or radio component imo.

If so, what then should we call the SUI type?  I think `style` or `type` are best suited for this prop's purpose, but both of those are seriously conflicting prop names.  It could be `appearance`, `kind`, ...something better:

``` jsx
// always input type radio
<Radio />
<Radio appearance='slider' />
<Radio appearance='toggle' />

// always input type checkbox
<Checkbox />
<Checkbox appearance='slider' />
<Checkbox appearance='toggle' />

// strange edge cases, if you must
<Radio appearance='checkbox' />
<Checkbox appearance='radio' />
```

Another option is to create a component for each visual type, and accept a `type` prop only for the input type.  IMO this is more intuitive regarding input type, but seems to unnecessarily fracture the top-level API where a prop could have handled the styling differences.

``` jsx
<Radio />                       // input radio
<Checkbox />                    // input checkbox

<Slider />                      // input checkbox
<Slider type='radio' />

<Toggle />                      // input checkbox
<Toggle type='radio' />

// strange edge cases, if you must
<Radio type='checkbox' />
<Checkbox type='radio' />
```

src/collections/Form/Form.js
levithomason: `formSerializer` is only called if you provide an `onSubmit` prop and did not provide your own `serializer` prop function.  The messages are only invoked if NODE_ENV is not production.

I can't see any other calls to remove.  Can you clarify what you mean by eliminating unneeded calls?

src/collections/Form/Form.js
levithomason: It's a noop only called onSubmit.  Though, we could wrap each call in a production check.  I don't think this will ever show up in a perf profile, but I can wrap them all if you feel very strongly about it still.

docs/app/Examples/addons/Radio/Types/index.js
levithomason: I'm not aware of the differences between those, though the spaced string did give issues elsewhere.

Since space is removed between components that are separated by a new line, and the component's content should always be one space away from adjacent text, I've opted to include `&nbsp;` inside of components.  This way, no matter where the components are moved the spacing is correct without having to add/remove spacing elements every time you move the component.

src/collections/Form/Form.js
levithomason: Saw the comment below regarding each call to the noop, see response there.

src/collections/Form/Form.js
dvdzkwsk: Yeah, not worth the effort of optimizing until deemed to actually matter. Just ignore.

src/addons/Radio/Radio.js
levithomason: Per offline discussion:

``` jsx
// Default type for each

<Radio slider />
<Radio toggle />

<Checkbox slider />
<Checkbox toggle />
```

levithomason:
Released in `v0.40.0`

/cc @brsanthu

brsanthu:
thanks for the great release @levithomason. This is the major piece. Is this end of Jquery in stardust?

levithomason:
This is indeed the end of jQuery in Stardust.  There are no traces of jQuery in this library any longer. See #247.

PR #409, opened 2016-08-20
refactor(TextArea): rename and use functional component

levithomason:
This is a cherry-pick from the Form update PR, #400.  This was previously `Textarea` and is now `TextArea`.  It was also made functional and a doc description added.


levithomason:
Released in `v0.36.0`

PR #411, opened 2016-08-20
fix(customPropTypes): false demand() errors

levithomason:


src/lib/customPropTypes.js
```diff
@@ -122,8 +122,12 @@ export const some = (validators) => {
  */
 export const givenProps = (propsShape, validator) => {
   return (props, propName, componentName, ...rest) => {
-    // only validate if all propsShape validators pass
-    if (!_.every(propsShape, (val, key) => !val(props, key, componentName, ...rest))) return
+    const shouldValidate = _.every(propsShape, (val, key) => {
+      // require propShape validators to pass or prop values to match
+      return _.isFunction(val) ? !val(props, key, componentName, ...rest) : val === props[propName]
+    })
```
levithomason, line 128: Support defining `giveProps` by value instead of by PropType function only:

``` js
control: customPropTypes.givenProps(
  { control: 'input' },
  customPropTypes.disallow(['children'])
),
```

src/lib/customPropTypes.js
```diff
@@ -170,9 +174,9 @@ export const demand = (requiredProps) => {
     }
 
     // do not require requiredProps if the prop does not exist in props
-    if (!_.has(props, propName)) return
+    if (_.isUndefined(props, propName)) return
 
-    const missingRequired = requiredProps.filter(required => !_.has(props, required))
+    const missingRequired = requiredProps.filter(required => _.isUndefined(props, required))
```
levithomason, line 179: Don't demand props when the prop at hand is `undefined`.  React treats `undefined` props as not present.  Previously, `_.has` here was demanding props when the prop implementing `demand()` was undefined.

levithomason:
Released in `v0.35.1`

PR #414, opened 2016-08-21
Breaking Changes: Add component augmentation

### LINKED ISSUE #403: RFC: Consider component augmentation
# Dropdown

> _Copied from #401 regarding the `icon` prop.  This conversation now applies to all Dropdown types_.
## Preliminary Findings

The Dropdown supports many variations, including rendering as a Button or a Menu Item.  Here is what I've gathered that helped clarify some things for me:
- There is only ever one icon
- Dropdowns can appear as other components like a _Button_ or a _Menu Item_
- The icon is either a Dropdown icon, Button icon, or Menu Item icon but only one at a time

---

image

image

image

image

image

image

image

image

image
## Brainstorming Solutions

Given the above, we may be able to use a single `icon` prop after all.  It would just be applied to as either a Dropdown, Button, or Menu Item icon.

Our offline work going on with Forms may introduce a pattern that is helpful to Dropdowns as well.  We may allow either component augmentation or specialized subcomponents in `Form.Field`s.  This will facilitate our shorthand props API for Forms.  Here are some of the ideas we are playing with:

``` jsx
<Form.Field control='input' type='password' label='Enter password:' />

<Form.Field control={Checkbox} label='I agree to the terms and conditions' />
// or
<Form.Checkbox label='I agree to the terms and conditions' />
```

``` html
<div class="field">
  <label>Enter password:</label>
  <input type='password' />
</div>

<div class="field">
  <Checkbox label='I agree to the terms and conditions' />
</div>
```

There is also a need library wide to explicitly set the element type for a component.  Example, you may want to render a `<Segment />` as an HTML `<section />`.  We are considering a library wide standardized prop such as:

``` jsx
<Segment />
<Segment elementType='section' />
```

``` html
<div class="ui segment"></div>
<section class="ui segment"></section>
```
## Applying it to the Dropdown

All of these ideas are ways of augmenting components.  I have not thought this idea through at all, but it seems like we might support something along the lines of:

``` jsx
<Dropdown icon='filter' options={...} />
<Dropdown.Button labeled icon='settings' options={...} />
<Dropdown.Item icon='dropdown' options={...} />

// or

<Dropdown icon='filter' options={...} />
<Button elementType={Dropdown} labeled icon='settings' options={...} />
<Menu.Item elementType={Dropdown} icon='dropdown' options={...} />
```

The top group `Dropdown.Item` is already taken, so we'd have to solve the name conflict.  

The latter group I think makes more sense.  Consider the Button, all the props are regular button props (labeled, icon, etc) then the extra props (options) are passed to the `Dropdown`.  This would result in a SUI dropdown being rendered with all the props/classes of both the Button and Dropdown.

As I write this, I'm thinking it might help users better understand the augmentation if we use the prop name `as`:

``` jsx
<Button as={Dropdown} labeled icon='settings' options={...} />
<Menu.Item as={Dropdown} icon='dropdown' options={...} />

// or

<Dropdown as={Button} labeled icon='settings' options={...} />
<Dropdown as={Menu.Item} icon='dropdown' options={...} />
```
# MenuItem

> Copied from https://github.com/TechnologyAdvice/stardust/pull/382#discussion_r75179898

Many other components and use cases have us considering the use of a library wide `as` prop:

``` jsx
<Segment as='section' />
//=> <section class="ui segment"></section>

<MenuItem as='a' />
//=> <a class="item></a>

<Dropdown as={Button} options={[...]} />
//=> <div class="ui button dropdown" />
```

Especially handy for `react-router` use with the MenuItem as it would allow passing props for both components:

``` jsx

<MenuItem link as={Link} to='/home' activeClass='active' />
//=> <a class="item active"></a>
```

The `MenuItem` would consume its own props (`link`) and spread the `rest`.  The `as` prop here means the MenuItem would render like:

``` jsx
render() {
  const ElementType = getElementType(props) // new util, like getUnhandledProps()

  return <ElementType {...rest} /* other menu item props */ />
}
```

The `getElementType()` method would hold the logic for returning the value of the `as` prop, else if `props.link` return `a`, else `div`.  We could add more logic to how components are allowed to determine their element type as well.  This is good because all components would be consistent.

levithomason:
Fixes #403.  This PR will add component augmentation.
- [x] Add augmentation to docs intro page
- [ ] ~~Add Dropdown as Button augmentation example~~ related but beyond the scope of this PR
- [X] ~~Investigate solving #291~~ related but beyond the scope of this PR
- [x] Update react-docgen once https://github.com/reactjs/react-docgen/pull/99 is released
# Testing Changes

https://github.com/TechnologyAdvice/stardust/pull/414/commits/415c87037a00f8e89e9d83b8386ca3580501661d We have well over 2k tests. This PR updated the watcher to **only run changed tests**.  If you change a non test file, all tests will re-run.  This makes it easier to iterate quickly on tests without waiting for thousands of unrelated tests to run.
# Breaking Changes
## Item.Items

Updated to `Items.Group`.  You'll see deprecation warnings for using `Item.Items`.  Reminder, all deprecated APIs will be removed in v1.
## Header

This PR overhauls the Header component to work with component augmentation.  It also fixed some out standing issues with the Header.

See the docs for examples.
### Added `sub` prop

SUI defines ["Sub Headers"](http://semantic-ui.com/elements/header.html#sub-headers), a type of Header. A Header can also have a ["Subheader"](http://semantic-ui.com/elements/header.html#subheader) as content (children).  Our previous API supported only the "Subheader" child.  Now we correctly support both "Sub Header" types and "Subheader" children.
### Removed `H*` Sub Components

We previously provided an API like so `<Header.H1 />`.  Now you should `<Header as='h1' />`.  
### Shorthand props

Shorthand props and `children` are now exclusive, as they should've been from the start.  The `content` prop has also been added to complete the shorthand prop API.


src/lib/getElementType.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {object} [typeMap] Maps props (keys) to element types (values).
+ * @returns {string|function} A ReactElement type
+ */
+function getElementType(Component, props, typeMap) {
+  const { defaultProps = {} } = Component
+
+  // ----------------------------------------
+  // user defined element type
+
+  if (props.as && props.as !== defaultProps.as) return props.as
```
dvdzkwsk, line 18: Should this just always return `props.as` if it exists?

test/specs/modules/Progress/Progress-test.js
```diff
@@ -113,10 +113,10 @@ describe('Progress', () => {
   })
 
   describe('percent', () => {
-    it('adds the percent to the data-percent prop', () => {
-      shallow(<Progress percent={10} />)
-        .should.have.prop('data-percent', 10)
-    })
+    // it('adds the percent to the data-percent prop', () => {
```
dvdzkwsk: cruft?

dvdzkwsk:
👻

src/lib/getElementType.js
levithomason: Let me go back and figure out why I did this, could be cruft.  Used to have `defaultAs` prop here.

levithomason: OK, this is still required.  The element type is chosen from this priority:
1. User defined an `as` prop
2. getElementType() typeMap arg mapped a prop to an element type:  `{ onClick: 'a' }`
3. Inferred type: `href => a'
4. defaultProp.as
5. `div`

When there is a defaultProp, the `props.as` will be set.  However, this _is not_ the user's `as` value.  It is the component default.  If we remove the `props.as !== defaultProps.as` comparison, then the defaultProp.as is always used as the first priority since it is copied to `props.as`.

So, we only want to use the `props.as` if the user defined one (i.e. it does not equal the default).

**Image.js**
- default - `img`
- maps - `wrapped => div`
- infers - `href => a`

If the props.as is used, it will always be an `img`.  Even if the user uses `wrapped`, which should be a div or `href` which should wrap the image in an `a` tag.

``` jsx
render() {
  const ElementType = getElementType(Image, props, {
    wrapped: 'div',
  })

  if (ElementType === 'img') {
    return <ElementType {...rootProps} {...imgTagProps} />
  }

  return (
    <ElementType {...rootProps} href={href}>
      <img {...imgTagProps} />
    </ElementType>
  )
}

Image.defaultProps = {
  as: 'img',
}
```

levithomason:
Released in `v0.37.0`

PR #415, opened 2016-08-21
fix(Input): remove extra curly brace

fuatsengul:
It's probably non-intentional typo mistake.


levithomason:
Thanks for the fix.  Note, this component is a legacy scheduled to be rewritten to our v1 API.  It will change soon.  See #269 for more.

levithomason:
Released in `v0.36.1`

Issue #416, opened 2016-08-22
Dropdown enhancements: Better remote API support, item customization


jeffcarbs:
_This wound up a little long, sorry in advance!_

I'm trying to use the dropdown with a remote data source. I'm running into a few interconnected issues so I'm just going to enumerate them here:
### 1. Disabling/customizing local filtering

I'm using the `onSearchChange` callback to send a request to the server with the search term and it's sending me back filtered results. The issue I'm running into is that the results are being filtered client-side as well:
https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/Dropdown.js#L527-L531

``` js
// filter by search query
if (search && searchQuery) {
  const re = new RegExp(_.escapeRegExp(searchQuery), 'i')
  filteredOptions = _.filter(filteredOptions, (opt) => re.test(opt.text))
}
```

In my case, the `text` label for each option is not the criteria that's used server-side, causing results to be filtered out incorrectly. I'd basically just like to skip this filtering since I'm relying on the server to do it.

**Some possible solutions:**
1. Allow disabling local filtering via a boolean
2. Allow someone to pass as a prop their own function that gets passed to `_.filter`.
   - In my case I could just pass `() => true`.
   - I could also see it being useful to perform a more complex match based on the option or match based on some criteria other than `text`.
3. Allow someone to pass as a prop their own filter function that takes the options and the search query and returns filtered options. Similar to 2, but 2 wouldn't allow you to pre-calc a regex like you're doing above for performance.

Option 2/3 seems to provide more flexibility. The default could be the current functionality since that's what one would expect for the simple text-option case.
### 1a. Hiding/showing the "Add item" option

There' also a filter function to determine whether the "Add item" option should be shown that's slightly different than the filter above:

``` js
// insert the "add" item
if (allowAdditions && search && searchQuery && !_.some(filteredOptions, { text: searchQuery })) {
```

That could also be a prop (e.g. `showAdditionOption` or some better name) that's a function that takes the options and the search query and returns a boolean. That would be necessary for my case, but you could imagine using that to doing things like validation or more complex logic.
### 2. Option customization

I want to display a more complex "dropdown item" with some custom styles. I think this could be done with two small (backwards compatible) tweaks:
1. Allow specifying the element type of `Dropdown.Item` as a prop so one could pass a custom component. This is already in progress: #414 
[...]

PR #418, opened 2016-08-22
Dropdown: Pass 'options' objects as props to Dropdown.Item

jeffcarbs:
When specifying options on a dropdown, previously only 'text' and 'value' were being passed to `<Dropdown.Item>`. This makes it so the entire object is passed as props.

cc #416

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -46,20 +46,11 @@ export default class Dropdown extends Component {
       PropTypes.string,
     ]),
 
-    /** Array of `{ text: '', value: '' }` options */
+    /** Array of Dropdown.Item props e.g. `{ text: '', value: '' }` */
     options: customPropTypes.every([
       customPropTypes.disallow(['children']),
       customPropTypes.demand(['selection']),
-      PropTypes.arrayOf(PropTypes.shape({
-        value: PropTypes.oneOfType([
-          PropTypes.string,
-          PropTypes.number,
-        ]).isRequired,
-        text: PropTypes.oneOfType([
-          PropTypes.string,
-          PropTypes.number,
-        ]),
-      })),
+      PropTypes.arrayOf(PropTypes.shape(DropdownItem.propTypes)),
     ]),
 
     /** Controls whether or not the dropdown menu is displayed. */
@@ -792,8 +783,7 @@ export default class Dropdown extends Component {
         active={isActive(opt.value)}
         onClick={this.handleItemClick}
         selected={selectedIndex === i}
-        text={opt.text}
-        value={opt.value}
+        {...opt}
       />
     ))
   }
--- a/src/modules/Dropdown/DropdownItem.js
+++ b/src/modules/Dropdown/DropdownItem.js
@@ -3,6 +3,7 @@ import cx from 'classnames'
 
 import {
   childrenUtils,
+  customPropTypes,
   META,
   useKeyOnly,
 } from '../../lib'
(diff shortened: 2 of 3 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -46,20 +46,11 @@ export default class Dropdown extends Component {
...
-    /** Array of `{ text: '', value: '' }` options */
+    /** Array of Dropdown.Item props e.g. `{ text: '', value: '' }` */
     options: customPropTypes.every([
       customPropTypes.disallow(['children']),
       customPropTypes.demand(['selection']),
-      PropTypes.arrayOf(PropTypes.shape({
-        value: PropTypes.oneOfType([
-          PropTypes.string,
-          PropTypes.number,
-        ]).isRequired,
```
jeffcarbs, line 57: Note: This `isRequired` got dropped in this change.

On the one hand, `PropTypes.arrayOf(PropTypes.shape(DropdownItem.propTypes))` is much DRYer and less brittle.

On the other, there's a slight difference in terms of what props are required that we're now ignoring.

Open to suggestions here.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -805,6 +805,17 @@ describe('Dropdown Component', () => {
       newItem.should.have.prop('text', 'bar')
       newItem.should.have.prop('value', 'bar')
     })
+
+    it('passes options as props', () => {
+      const customOptions = [
+        { text: 'abra', value: 'abra', className: 'customClass' },
+        { text: 'cadabra', value: 'cadabra', className: 'customClass' },
+        { text: 'bang', value: 'bang', className: 'customClass' },
```
levithomason: Since className is a likely handled prop, how about adding a `data-foo` prop in here to be completely sure unanticipated props are passed?

src/modules/Dropdown/DropdownItem.js
```diff
@@ -75,7 +75,10 @@ DropdownItem.propTypes = {
   selected: PropTypes.bool,
 
   /** Display text. */
-  text: PropTypes.string,
+  text: PropTypes.oneOfType([
```
jeffcarbs: I changed this to reflect the PropType definition for `options.text` in `Dropdown`. I can change back if specifying a string would be preferable.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -754,7 +754,7 @@ describe('Dropdown Component', () => {
   describe('options', () => {
     it('adds the onClick handler to all items', () => {
       wrapperShallow(<Dropdown options={options} selection />)
-        .children('DropdownItem')
+        .find('DropdownItem')
```
jeffcarbs, line 757: This spec was silently broken since this returned an empty list and `everyWhere` on an empty list is passing. I noticed because I used this as a starting point for the spec I added below.

src/modules/Dropdown/Dropdown.js
levithomason, line 57: Seems reasonable for now.  The text/value props can't reliably be required at the DropdownItem level as there is no guarantee they are being used in a selection Dropdown.

src/modules/Dropdown/DropdownItem.js
levithomason: I'm in full agreement on the update, thanks

levithomason: While we're updating these, would you mind updating the `text` and `children` to being exclusive.  `customPropTypes` lives in `lib`:

``` js
  /** Primary content. */
  children: customPropTypes.every([
    customPropTypes.disallow(['text']),
    PropTypes.node,
  ]),

  /** Display text. */
  text: customPropTypes.every([
    PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.number,
    ]),
  ]),
```

jeffcarbs:
Updated based on comments 👍

levithomason:
Released in `v0.36.3`, thanks!

PR #431, opened 2016-08-24
Item: Update to v1 API

layershifter:
# Finished

Update [Item](http://semantic-ui.com/views/item.html) component to v1 API.

### DIFF
--- a/src/elements/Image/Image.js
+++ b/src/elements/Image/Image.js
@@ -24,7 +24,7 @@ function Image(props) {
   } = props
 
   const classes = cx(
-    ui && 'ui',
+    useKeyOnly(ui, 'ui'),
     size,
     useVerticalAlignProp(verticalAlign, 'aligned'),
     useKeyOnly(avatar, 'avatar'),
--- a/src/elements/Image/Image.js
+++ b/src/elements/Image/Image.js
@@ -20,11 +20,11 @@ import ImageGroup from './ImageGroup'
 function Image(props) {
   const {
     verticalAlign, alt, avatar, bordered, centered, className, disabled, floated, fluid,
-    hidden, height, href, inline, shape, size, spaced, src, width,
+    hidden, height, href, inline, shape, size, spaced, src, width, ui,
   } = props
 
   const classes = cx(
-    'ui',
+    ui && 'ui',
     size,
     useVerticalAlignProp(verticalAlign, 'aligned'),
     useKeyOnly(avatar, 'avatar'),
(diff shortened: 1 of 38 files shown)


layershifter:
@levithomason `Item.Items` was removed, what I must do with [this](https://github.com/TechnologyAdvice/stardust/blob/master/src/deprecatedComponents.js#L45)?

src/views/Item/ItemImage.js
```diff
@@ -0,0 +1,15 @@
+import React, { PropTypes } from 'react'
+import { META } from '../../lib'
+import Image from '../../elements/Image/Image'
+
+function ItemImage(props) {
+  return <Image {...props} />
```
layershifter: @levithomason I have problem there.
1. I think I need use `createImage` factory, but will it pass props?
2. `ItemImage` doesn't have `ui` class by default. How we can deal with this?

layershifter:
@levithomason, finished :v: 

Can be merged after [issue will be solved](https://github.com/TechnologyAdvice/stardust/pull/431#discussion_r76516759) and code review :smiley_cat:

levithomason:
Taking a look...

levithomason:
I've added a `ui` prop (and tests) to Image to allow turning off the `ui` className:

``` jsx
<Image {...rest} as={ElementType} ui={false} />
```

This seems to solve the issue with the `ui` className simply.  We may add this feature for all components, though for now it will just be on the Image.

levithomason:
There are a couple warnings to solve, looks like we're spreading props we shouldn't be:

```
Warning: Failed prop type: Invalid prop `items[0].childKey` of type `number` supplied to `ItemGroup`, expected `string`.
```

```
Warning: Unknown prop `childKey` on <div> tag. Remove this prop from the element.
in ItemGroup
```

layershifter:
Seems, it fixed :+1:

src/elements/Image/Image.js
```diff
@@ -20,11 +20,11 @@ import ImageGroup from './ImageGroup'
 function Image(props) {
   const {
     verticalAlign, alt, avatar, bordered, centered, className, disabled, floated, fluid,
-    hidden, height, href, inline, shape, size, spaced, src, width,
+    hidden, height, href, inline, shape, size, spaced, src, width, ui,
   } = props
 
   const classes = cx(
-    'ui',
+    ui && 'ui',
```
layershifter: Stupid question. Why you used there '&&' instead of `useKeyOnly`? :rabbit:

levithomason: Stupid answer, whoops!

levithomason:
Will merge/release once tests pass

levithomason:
Released in `v0.38.0`, we're getting closer to v1 :D 🎉

Issue #434, opened 2016-08-24
Docs: Generation issue of docgenInfo.json on Windows


layershifter:
There are build error of docs on Windows.

![image](https://cloud.githubusercontent.com/assets/14183168/17939977/a04b9948-6a35-11e6-9293-03ff466eaf29.png)

As @levithomason  decided [there](https://github.com/TechnologyAdvice/stardust/pull/382#issuecomment-242136417), generated `docgenInfo.json` uses only unix-like path delimiter (`/`) while Windows requires `\`.

Extracted from #382.

levithomason:
I've got a branch going against this issue.

layershifter:
![image](https://cloud.githubusercontent.com/assets/14183168/17978584/720e5992-6aff-11e6-874e-dfb841ee1739.png)

I reopened issue, seems not fixed.

levithomason:
Can you try this in `ComponentDescription.js` @ line 16?

``` diff
-const pathSepRegEx = new RegExp(__PATH_SEP__, 'g')
+const pathSepRegEx = new RegExp(_.escapeRegExp(__PATH_SEP__), 'g')
```

layershifter:
It works :+1:

PR #438, opened 2016-08-25
Docs: Comparison width fix

### LINKED ISSUE #433: Docs: Comparison width error
On FHD screen I'm getting following:

image

Classes looks correctly:

image

But width taken another:

image

Looks like issue with SUI or I am missing something :thought_balloon:

layershifter:
Fixes #433

---

I can't test it because #434.


levithomason:
I think the divider will disappear on widescreen, but let's find out!

layershifter:
Works fine on current doc site :+1:

PR #439, opened 2016-08-25
fix(Checkbox): render required empty labels

levithomason:
Empty labels in Checkbox were removed from the render function recently.  They are required by SUI CSS for other checkbox types.

---

image


levithomason:
Released in `v0.37.1`

Issue #284, opened 2016-06-24
RFC: Delegate/Pool DOM Event Listeners


dvdzkwsk:
## Disclaimer:

This is just me thinking out loud in an effort to initiate some brainstorming. I am not proposing this to be considered/worked on until after V1 is released, since that is the main priority.
## Proposal

For components that need to reach outside of React land and bind listeners on the real DOM, such as the `Dropdown`, I think we should consider limiting how many event handlers are attached to the document, especially when this can occur in frequent succession. Here's an [example of what I'd like to discourage](https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/Dropdown.js#L265-L298). This could be achieved by creating a facade on top of the core DOM API that would perform pooling/event delegation behind the scenes. Basically, a pub/sub system.
## Potential benefits:

1) Eliminates event listeners being repeatedly added to and removed from the document. This is one aspect React has thankfully eliminated from the front end developer's life: having to decide if and when to [delegate events](https://encrypted.google.com/search?hl=en&q=event%20delegation). Stardust could just listen to these common events and act as a pub/sub system for pooled handlers.

2) Smarter event handlers could return an `unbind` function (or `off` method), to make cleanup easier:

``` js
// idea #1 =======================
const listenToKeydown = documentApi.on('click', () => { /* ... */ })
listenToKeydown() // or listenToKeydown.off()

// idea #2 =======================
// (could just be an extension of #1, too)
this._documentEvents = new eventApi.Pool('document') // could be global
this._documentEvents.on('click', () => { /* ... */ })
this._documentEvents.on('keydown', () => { /* ... */ })

// later, in unmount, perhaps
this._documentEvents.unbindAll()
```

3) By piping all attach/detach calls through a single place, we can easily do performance profiling or diagnose cases where too many handlers exist or leaked event handlers do not get detached. This is possible without our own API wrapper, but it would help.

4) Could improve event mock-ability in testing. This is debatable.

5) If the API was made to be intentionally cumbersome, i.e. with long method names for binding (something done frequently in React core), it would help to discourage widespread usage, since we should be doing this as little as possible.

6) Event pooling could be scoped to a component-level, to remove the chance for developer error in forgetting to unbind events.

7) Solutions may already exist for this. Have not looked, yet.
## Potential downsides:

[...]

PR #440, opened 2016-08-26
feat(Modal): add configurable close behaviors (#440)

### LINKED ISSUE #437: Modal: Support configurable close behaviors
Modal should support disabling hiding the modal when user clicks outside the modal. This is useful when we are dealing with complicated form and we don't want user to lose the work if they accidentally clicks outside the modal.

brsanthu:
This PR enhances Modal component to support configurable close behaviors (closes #437)

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -52,10 +52,10 @@ class Modal extends Component {
     /** A modal can reduce its complexity */
     basic: PropTypes.bool,
 
-    /** Closes the modal if Escape is pressed. Defaults to true */
+    /** Closes the modal if Escape is pressed */
     closeOnEscape: PropTypes.bool,
 
-    /** Closes the modal if user clicks anywhere outside the modal. Defaults to true */
+    /** Closes the modal if user clicks anywhere outside the modal */
     closeOnClickOutside: PropTypes.bool,
 
     /** A modal can appear in a dimmer */
@@ -206,7 +206,7 @@ class Modal extends Component {
   }
 
   render() {
-    const { active, basic, children, className, dimmer, size} = this.props
+    const { active, basic, children, className, dimmer, size } = this.props
     const { marginTop, scrolling } = this.state
     const classes = cx(
       'ui',
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -206,7 +206,7 @@ class Modal extends Component {
   }
 
   render() {
-    const { active, basic, children, className, dimmer, size, closeOnEscape, closeOnClickOutside } = this.props
+    const { active, basic, children, className, dimmer, size} = this.props
     const { marginTop, scrolling } = this.state
     const classes = cx(
       'ui',
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -52,6 +52,12 @@ class Modal extends Component {
     /** A modal can reduce its complexity */
     basic: PropTypes.bool,
 
+    /** Closes the modal if Escape is pressed. Defaults to true */
+    closeOnEscape: PropTypes.bool,
+
+    /** Closes the modal if user clicks anywhere outside the modal. Defaults to true */
+    closeOnClickOutside: PropTypes.bool,
+
     /** A modal can appear in a dimmer */
     dimmer: PropTypes.oneOfType([
       PropTypes.bool,
@@ -67,6 +73,8 @@ class Modal extends Component {
 
   static defaultProps = {
     dimmer: true,
+    closeOnEscape: true,
+    closeOnClickOutside: true,
   }
 
   static _meta = _meta
@@ -111,18 +119,25 @@ class Modal extends Component {
 
   handleHide = () => {
     debug('handleHide()')
+    const { closeOnEscape, closeOnClickOutside } = this.props
+
     // Always remove all dimmer classes.
     // If the dimmer value changes while the modal is open,
     //   then removing its current value could leave cruft classes previously added.
     document.body.classList.remove('blurring', 'dimmable', 'dimmed', 'scrollable')
 
-    document.removeEventListener('keydown', this.handleDocumentKeyDown)
-    document.removeEventListener('click', this.handleClickOutside)
+    if (closeOnEscape) {
+      document.removeEventListener('keydown', this.handleDocumentKeyDown)
+    }
+
+    if (closeOnClickOutside) {
+      document.removeEventListener('click', this.handleClickOutside)
+    }
   }
 
   handleShow = () => {
     debug('handleShow()')
-    const { dimmer } = this.props
+    const { dimmer, closeOnEscape, closeOnClickOutside } = this.props
     if (dimmer) {
       debug('adding dimmer')
       document.body.classList.add('dimmable', 'dimmed')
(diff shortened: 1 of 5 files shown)


brsanthu:
@levithomason Please take a look at the PR. Still working on tests.

src/modules/Modal/Modal.js
```diff
@@ -52,6 +52,12 @@ class Modal extends Component {
     /** A modal can reduce its complexity */
     basic: PropTypes.bool,
 
+    /** Closes the modal if Escape is pressed. Defaults to true */
```
levithomason: Let's leave the default value out of the descriptions as it is documented automatically from `defaultProps`:

image

levithomason:
Looking closer at the Modal, I think my original instructions were short sighted.  I don't think this will work as it should the way it is currently written.  What we want to do is "abort close on escape if !closeOnEscape" and "abort on click outside if !closeOnClickOutside || clicked inside".

Otherwise, all the closing logic should remain as is.  It is just that we only run it if the bool props permit.  Something more along these lines may be appropriate:

### Only close on escape if closeOnEscape

``` diff
handleDocumentKeyDown = (e) => {
+ const { closeOnEscape } = this.props
  const key = keyboardKey.getCode(e)
  debug('handleDocumentKeyDown()', key)

  switch (key) {
    case keyboardKey.Escape:
-     this.onHide()
+     if (closeOnEscape) this.onHide()
      break
    default:
      break
  }
}
```

### Only close on click outside if closeOnClickOutside && clicked inside

``` diff
handleClickOutside = (e) => {
-  // do nothing when clicking inside the modal
-  if (this._modalNode.contains(e.target)) return
+  const { closeOnClickOutside } = this.props
+  if (!closeOnClickOutside || this._modalNode.contains(e.target)) return

  debug('handleDimmerClick()')

  e.stopPropagation()
  this.onHide()
}
```

src/modules/Modal/Modal.js
```diff
@@ -133,8 +148,13 @@ class Modal extends Component {
       }
     }
 
-    document.addEventListener('keydown', this.handleDocumentKeyDown)
-    document.addEventListener('click', this.handleClickOutside)
+    if (closeOnEscape) {
+      document.addEventListener('keydown', this.handleDocumentKeyDown)
+    }
```
levithomason: Good call only adding the listeners when needed.  However, if the prop changes after the Modal is open, the listeners will not be added.

We'd need to move these to `componentDidUpdate` and add/remove them based on props.  That logic could get hairy though with lots of conditionals.  I think instead, it is acceptable to allow these two listeners to always exist and we just do nothing in the handlers if the `closeOn*` props are set to false.

Let me know your thoughts.

levithomason: To clarify, let's say we add listeners on open only if the `onClose*` props are true.

If a Modal opens and later, while open, then `closeOnEscape` becomes true. Then we need to add a condition to componentDidUpdate `if !prevProps.closeOnEscape && this.propscloseOnEscape then add listener`.  But, we don't really know if the listener was already added either :S We can assume so, if we put all the logic in place elsewhere to make it so.  But it is convoluted.

This is part of a larger issue, we have a pending RFC to implement a better way to handle events.  Handle pooling and delegation.  See #284.

levithomason:
All said, if you'd like to add the logic for ensuring the event listeners are added/removed only when needed and also handle the props changing while open... I'm all for it!  If this is a burden, you can use the proposed alternative.

I plan to clean these up when #284 is addressed.

src/modules/Modal/Modal.js
brsanthu: You have a valid point. I will change to add listener always but take action conditionally.

brsanthu:
I have made changes suggested by you and also added an example. 

How do I run the examples locally?

src/modules/Modal/Modal.js
```diff
@@ -144,16 +153,24 @@ class Modal extends Component {
     debug('handleDimmerClick()')
 
     e.stopPropagation()
-    this.onHide()
+
+    const { closeOnClickOutside } = this.props
+    if (closeOnClickOutside) {
+      this.onHide()
+    }
```
levithomason: This will still log a debug message and `e.stopPropagation()` even when `closeOnClickOutside` is false.  The negated condition should be added to short circuit early return at the top.  That way, nothing in the method runs if the condition is not met.

levithomason:
See CONTRIBUTING.md ["Documentation"](https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#documentation) for info on how to run/build docs.

brsanthu:
Tried but running into this error

```
npm start                                                                                                                                                

> stardust@0.37.1 start /Users/ossuser/stardust
> npm run docs


> stardust@0.37.1 docs /Users/ossuser/stardust
> gulp

[15:02:26] Requiring external module babel-register
/Users/ossuser/.nvm/versions/node/v6.4.0/lib/node_modules/gulp/node_modules/orchestrator/index.js:54
        this.tasks[name] = {
            ^

TypeError: Cannot read property 'tasks' of undefined
    at Orchestrator.add (/Users/ossuser/.nvm/versions/node/v6.4.0/lib/node_modules/gulp/node_modules/orchestrator/index.js:54:7)
    at Object.<anonymous> (dll.js:10:1)
    at Module._compile (module.js:556:32)
    at loader (/Users/ossuser/stardust/node_modules/babel-register/lib/node.js:148:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/ossuser/stardust/node_modules/babel-register/lib/node.js:158:7)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
```

levithomason:
Hm, what OS and what version of node and npm do you have?

levithomason:
I would also try `rm -rf node_modules && npm i`.  Will take a while, but it looks like gulp isn't defined for some reason.  I'm looking into it on master to see if there are any issues.

levithomason:
I see you're on a Mac with node 6.  I am as well, though, I am unable to reproduce the issue on master with a fresh install.  LMK if you have any more findings.

brsanthu:
Removed node modules, installed but still same error. Here is my info.

```
Darwin local 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64
Node Version
v6.4.0
Npm version
3.10.3
```

levithomason:
Very odd, I just updated to the same node/npm version as you but no issues.  Does it happen on master, or just this branch?

brsanthu:
sorry my bad. I was not on root but in src folder when I ran the commands. It is working fine now.

I fixed the comments and also linting errors.

levithomason:
Awesome, no worries.  I should have thought of that!

brsanthu:
Doc example is also added and tested in chrome. Do we need to test in other browsers?

levithomason:
Chrome is fine.  Our unit tests only test in PhantomJS.  We may add more browser testing later.

levithomason:
By the way, you can run `npm run lint:fix` to fix most errors.  Then you can run `npm run lint:watch` to fix any remaining errors and have the linter re-run as you make fixes.  The linter is a little slow so it takes a bit between changes.

brsanthu:
Good to know on those commands. Will come handy.

levithomason:
https://github.com/TechnologyAdvice/stardust/blob/master/CONTRIBUTING.md#commands

brsanthu:
@levithomason any idea when you be able to merge this?

levithomason:
Thanks for the bump, I'll release this later tonight.

levithomason:
Released in `stardust@0.37.4`, thanks for the PR!

brsanthu:
Awesome. Thanks for the guidance.

PR #443, opened 2016-08-26
fix(Dropdown): fix search width

### LINKED ISSUE #397: Dropdown: fix search type issues
### Search Width

The search input resizes on search change.  The search selection multiple dropdown type resizes OK.  The search selection type resizes, but not wide enough.
### Text

When focused, the current value should be shown as the placeholder.  This way, it is greyed out and clear to the user that they can type.  This is also how it works in SUI.  Currently, it remains solid black as if it is the value of the input.

levithomason:
Fixes #397 

Search type Dropdowns include a dynamically sized search input.  This PR fixes the width measurement.  It also fixes an issue where the Dropdown text was not muted when the search input is focused.  This was due to rendering the `sizer` when it was not needed.  It is only necessary on multiple type Dropdowns.


levithomason:
Released in `v0.37.3`

PR #444, opened 2016-08-26
feat(Header): support size prop

levithomason:
This PR adds the `size` prop to the Header. It was somehow overlooked.

We currently have docs for "Content Headers" but no component support.  A content header is a header sized using `size=''`.  It differs from h tag headers in that they are sized using `em` rather than `rem`.

http://semantic-ui.com/elements/header.html#content-headers

http://technologyadvice.github.io/stardust/elements/header#content-headers


levithomason:
Released in `v0.37.2`

PR #447, opened 2016-08-28
feat(Dimmer): Add component

### LINKED ISSUE #190: <Dimmer /> Component
http://semantic-ui.com/modules/dimmer.html

layershifter:
Fixes #190


layershifter:
ping @levithomason

levithomason:
## Responses

### Component

I think we first go with a stateless component here.  The `active` prop will be controlled from the parent component.  Once we get that, then we can consider if anything in the Dimmer warrants auto controlling its own state, but I don't think it will.

### Simple dimmer

There is actually no difference in markup between the "simple css dimmer" and a regular dimmer.  The only difference is in how you use it.  The jQuery dimmer simply adds the `dimmable dimmed` classes to the parent of the dimmer and the active classes to the dimmer itself.  A simple css dimmer just means the user added the classes to the parent and dimmer child manually.

### Dimmer events

Since the component will be controlled by parent state, I don't think we need events.  Let's leave those for future feature requests, if needed.

### Usage

Agreed, let's leave these out.

### Settings

`closable` - How about `closeOnClick`?  This way we can add other close settings like `closeOnEscape` if we want.

events - yes, let's skip and go stateless

## API Ideas

The parent component can be any component and also needs `dimmable dimmed` classes.  It must have a `<div class="ui dimmer" />` as an _immediate child_ for the CSS to work.

The dimmer itself can also have child content, which means we'll need an API for specifying the dimmed component's children and the dimmer's children.

The dimmer [centers content](http://semantic-ui.com/modules/dimmer.html#content-dimmer) horizontally and vertically.  Vertical centering requires two child divs. I think we include these two divs in all dimmer content.

> Content must be included inside `.content .center` to display centered correctly in the dimmer.

I'm proposing that the Dimmer have an API to control both the dimmed parent component as well as the dimmer child.

### Content Dimmer

Proposing the `dimmable`/`dimmed` props be used strictly for the parent component.

``` jsx
<Dimmer dimmable={Segment} dimmed>
  <Header>Overlayable Section</Header>
  <Dimmer active>
    <Header inverted icon='heart' content='Dimmed Message!' />
  </Dimmer>
</Dimmer>
```

``` html
<div class="ui segment dimmable dimmed">
  <h3 class="ui header">
    Overlayable Section
  </h3>
  <div class="ui dimmer transition visible active">
    <div class="content">
      <div class="center">
        <h2 class="ui inverted icon header">
          <i class="heart icon"></i>
          Dimmed Message!
        </h2>
      </div>
    </div>
  </div>
</div>
```

### Page Dimmer

In this case the `body` must receive the `dimmable dimmed` classes and the `ui dimmer` must an _immediate child_ to the `body`.  This would use a Portal as was done in the Modal dimmer.  The Portal provides a `div` as an immediate child to the `body`.

``` jsx
<Dimmer page active>
  <Header inverted icon='mail' content='Dimmer Message' subheader='Dimmer sub-header' />
</Dimmer>
```

``` html
<body>
  <div class="ui page dimmer transition visible active">
    <div class="content">
      <div class="center">
        <h2 class="ui inverted icon header">
          <i class="mail icon"></i>
          Dimmer Message
          <div class="sub header">Dimmer sub-header</div>
        </h2>
      </div>
    </div>
  </div>
</body>
```

layershifter:
I will start on it next week 🍰

layershifter:
### Component

After use in production, I can agree about stateless `Dimmer` :+1: 

I think that it will be class (will have `handleClick`, `handleEscape`), but will not have real state:

``` jsx
export default class Dimmer {
  handleClick = () => {}
  handleEscape = () => {}
  render () {}
}
```

### Dimmable

> In this case the body must receive the dimmable dimmed classes

It's real pain there, any ideas about `body`? I want following syntax, but it looks like impossible mission :open_mouth: 

``` jsx
<Segment>
  <Dimmer active={false}>...</Dimmer>
  ...
</Segment>
```

There is simple solution, but it's pretty shitty:

``` jsx
const active = true;

<Segment className={Dimmer.dimmable(active)}>
  <Dimmer active={active}>...</Dimmer>
  ...
</Segment>
```

@levithomason @jcarbo Please advise me something better :zipper_mouth_face:

levithomason:
> It's real pain there, any ideas about body? I want following syntax, but it looks like impossible mission 😮

Note, there are 2 types of Dimmers.  A "Content Dimmer" and a "Page Dimmer".

The "Page Dimmer" has the `body` as it's parent and needs to add classes to the body.  However, the "Content Dimmer" has _any_ component as it's parent and adds classes to it's parent (your use case above).

I've proposed an API for each use case in https://github.com/Semantic-Org/Semantic-UI-React/pull/447#issuecomment-246407676 above.  See above for more, specifics are here:

### Content Dimmer

Renders the `dimmable` component in place, much like a `trigger` for a Modal. You can then define a Dimmer as a child, which is shown when the `dimmed` prop is true.

``` jsx
<Dimmer dimmable={Segment} dimmed />
```

### Page Dimmer

Renders to body.

``` jsx
<Dimmer page active />
```

### Conclusion

I think we'll need to render with these two different approaches based on props:
1. `page` - renders to body as a page level dimmer
2. `dimmable` - renders the dimmable component in place, passing extra props to it
3. **otherwise** - renders the Dimmer markup

layershifter:
## Page

#### Dimmable/Dimmed

SUI adds `dimmable dimmed` classes to body, it makes window non-scrollable.



I've done this, any better solution there?

``` js
handlePageMount = () => {
    debug('handleMount()')
    document.body.classList.add('dimmed', 'dimmable')
  }

  handlePageUnmount = () => {
    debug('handleUnmount()')
    document.body.classList.remove('dimmed', 'dimmable')
  }
}
```

``` jsx
<Portal onMount={this.handlePageMount} onUnmount={this.handlePageUnmount} />
```

### Handle clickOutside

There is problem with handling outside click. I can't use there `Portal`'s `closeOnDocumentClick` because `Dimmer` has 100% width and 100% height.

So, I need handle it inside component, but I can't simply write `Dimmer onClick={()}` because this handler must omit `Dimmer`'s children (according to SUI examples). Any ideas there?

jeffcarbs:
@layershifter - `closeOnDocumentClick` actually includes the dimmer. It basically means "any click not within the portal content".

Also once #666 is merged you will have a little more control in that regard. You can either specify `closeOnDocumentClick` which is the default, or `closeOnRootNodeClick` which basically means "click on the portal wrapper/root node".

levithomason:
> `closeOnDocumentClick` actually includes the dimmer.

@jcarbo is this due to event bubbling?  If so, that makes sense because the Dimmer click bubbles to the `document` where the event handler is actually attached.  If however the portal node (div appended to the body) is receiving the click, I think we should go with something like `closeOnClickOutside`, if the actual `document` object is not where the event listener is attached.

Could you clarify?

P.S. I also left a comment on #666 regarding closing multiple Modals, LMK.

layershifter:
@levithomason I'll added `dimmable`, I've splitted it into separate component. Can you take a look there (`DimmerDimmable` and `blurring` example)?

I don't like that we need handle also `dimmed` prop, any ideas there?

docs/app/Examples/modules/Dimmer/States/DimmerExampleDisabled.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Dimmer, Segment } from 'semantic-ui-react'
+
+const DimmerExampleDisabled = () => (
+  <Segment>
+    <Dimmer disabled />
```
levithomason: Hm, do you know what this is supposed to do?  The dimmer still is made active when I add:

``` jsx
<Dimmer active disabled />
```

Though, the description says:

> A disabled dimmer cannot be activated

Perhaps this is only applicable to the jQuery plugin in SUI core?  Thoughts?

docs/app/Examples/modules/Dimmer/Variations/DimmerExampleBlurring.js
```diff
@@ -0,0 +1,33 @@
...
+
+  handleShow = () => this.setState({ active: true })
+  handleHide = () => this.setState({ active: false })
+
+  render() {
+    const { active } = this.state
+
+    return (
+      <div>
+        <Dimmer blurring dimmable={Segment} dimmed={active}>
```
levithomason: This currently does not pass extra props to the Segment.  Example, adding `color='red'` should pass `color` to the `Segment`.  What if we went with this API, it does pass extra props:

``` jsx
<Dimmer.Dimmable as={Segment} blurring dimmed={active} color='red'>
  <Dimmer active={active} />

  <p>
    <img src='http://semantic-ui.com/images/wireframe/short-paragraph.png' />
  </p>
  <p>
    <img src='http://semantic-ui.com/images/wireframe/short-paragraph.png' />
  </p>
</Dimmer.Dimmable>
```

I tested the above with only these changes required:
1. Expose `Dimmer.Dimmable = DimmerDimmable`
2. Rename `DimmmerDimmable` prop `Component` to `as` and use our standard `getElementType()` util and render `<ElementType />` instead.
3. I did not do this, but we should then remove all `dimmable` prop reference and logic from the main `Dimmer`.

This might be more in line with our existing patterns.

src/modules/Dimmer/Dimmer.js
```diff
@@ -0,0 +1,187 @@
...
+  static defaultProps = {
+    closeOnOutsideClick: true,
+  }
+
+  static _meta = _meta
+
+  handlePageMount = () => {
+    debug('handleMount()')
+
+    document.body.classList.add('dimmed', 'dimmable')
```
levithomason: I believe we only want to add `dimmed` if `active` is true.  Is that correct?

levithomason (changes requested):

levithomason:
Left some comments.  I think we must handle `dimmed` since it is on a child component that can appear anywhere in the child tree.

docs/app/Examples/modules/Dimmer/States/DimmerExampleDisabled.js
levithomason: Agreed, especially since it does not override the `active` prop.  Let's remove it.

src/modules/Dimmer/Dimmer.js
layershifter: Not exactly, it added on `Portal`s mount, while `Portal` will mount on `active`. This was plan.

layershifter:
I've made cleanup there, all examples where added, except example with [`Image`](http://semantic-ui.com/modules/dimmer#/examples). I was right when I thought that we will need `children` on `Image`.

Seems, we need initial review there, I'll add tests after it.

layershifter:
Updated to latest `master`, completed TODOs and added tests 😸

@levithomason waiting for review 😄 All is ready except, `common.rendersChildren()`, can yoo take a look there?

layershifter:
Added new example with `Image` and shorthands for it (some style refactor on `Image`, too).

src/modules/Dimmer/Dimmer.js
```diff
@@ -23,165 +21,105 @@ const _meta = {
...
-        blurring={blurring}
-        className={className}
-        Component={dimmable}
-        dimmed={dimmed}
-      >
-        {children}
-      </DimmerDimmable>
-    )
+    _.invoke(children, 'props.onClick', e)
+    e.stopPropagation()
```
levithomason: Do we need to stop propagation here?  Would prefer we not prevent handlers above this component from firing.

If we're trying to prevent our own handlers from firing, perhaps we should instead detect if the click was inside rather than stop the event.

test/specs/modules/Dimmer/Dimmer-test.js
```diff
@@ -0,0 +1,119 @@
...
+import * as common from 'test/specs/commonTests'
+import { sandbox } from 'test/utils'
+
+describe('Dimmer', () => {
+  common.isConformant(Dimmer)
+  common.hasSubComponents(Dimmer[DimmerDimmable])
+  common.hasUIClassName(Dimmer)
+
+  // TODO: Renders children / content
+  // common.rendersChildren()
```
levithomason: I saw your request to take a look here, but I'm not sure what the issue is.  Did this test not work for this component?

levithomason (commented):
Looks good, left one question regarding stopPropagation.  I'm also not sure what you'd like to review on rendersChildren(), please expound.

test/specs/modules/Dimmer/Dimmer-test.js
layershifter: Yes, this test is broken. I think problem in `cloneElement` that I use for handling click. I'll try rework `click` handling without `stopPropagation`.

layershifter:
`stopPropagation` replaced with `ref`, fixed example and corrected tests. Ready to go after review 😄

src/modules/Dimmer/Dimmer.js
```diff
@@ -0,0 +1,124 @@
...
+  handlePortalMount = () => document.body.classList.add('dimmed', 'dimmable')
+
+  handlePortalUnmount = () => document.body.classList.remove('dimmed', 'dimmable')
+
+  handleClick = (e) => {
+    const { onClick, onClickOutside } = this.props
+
+    if (onClick) onClick(e, this.props)
+    if (this.center && (this.center !== e.target && this.center.contains(e.target))) return
+    if (onClickOutside) onClickOutside(e, this.props)
```
layershifter, line 69: `onClickOutside` must be called **outside** children like in SUI.

layershifter:
@levithomason Any chance for review? :smiley_cat:

levithomason (approved):

levithomason:
Thanks much!

levithomason:
Released in `semantic-ui-react@0.58.1`

PR #451, opened 2016-08-29
feat(Table): update to v1 API

levithomason:
@layershifter https://github.com/TechnologyAdvice/stardust/issues/269#issuecomment-243243123

> It looks that `Table` can be stateless component, except one case, [sortable](http://semantic-ui.com/collections/table.html#sortable).
> 
> I think about making `Table.Sortable` that has state and allow it pass shorthands or only components `Table.Row`, `Table.Header`, etc. But, it is ugly. Any better ideas there?

I like striving to use stateless functional components, however, we also need to handle the selected row state. I think we just use AutoControlledComponent right out of the gate for the table.


levithomason:
I like striving to use stateless functional components, however, we also need to handle the selected row state.  I think we just use `AutoControlledComponent` right out of the gate for the table.

jeffcarbs:
Has there been any work done on this yet? Or was this just opened to capture that conversation from the referenced issue's comments?

levithomason:
It was opened to capture the conversation.  I believe @layershifter was planning on tackling this, not sure what his work load is.  Unless Alexander is planning to start work on it soon, I say it is fair game for anyone.

layershifter:
I'm planing to start on it today/tomorrow.

layershifter:
``` html
<table class="ui celled striped table">
  <thead>
    <tr>
       <th colspan="3">Git Repository</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td class="collapsing">
        <i class="folder icon"></i> node_modules
      </td>
      <td>Initial commit</td>
      <td class="right aligned collapsing">10 hours ago</td>
    </tr>
  </tbody>
</table>
```

``` jsx
<Table celled striped>
  <Table.Head>
    <Table.Row>
       <Table.Column as='th' colspan="3">Git Repository</Table.Column>
    </Table.Row>
  </Table.Head>
  <Table.Body>
    <Table.Row>
      <Table.Column collapsing>
        <Icon name="folder" /> node_modules
      </Table.Column>
      <Table.Column>Initial commit</Table.Column>
      <Table.Column collapsing textAlign="right">10 hours ago</Table.Column>
    </Table.Row>
  </Table.Body>
</Table>
```

@levithomason What you think about this, comrade? Is this right way?

levithomason:
~~Looks correct to me.~~  Suggest going with `Table.Header` since `thead` is short for table header and that would be more consistent with other components such as CardHeader, ModalHeader, etc.

**EDIT**

Since `td` translates to a [table cell](http://www.w3schools.com/tags/tag_td.asp), I think `Table.Column` should also be `Table.Cell`.

levithomason:
As I thought about this, I had hoped we could use only the `TableColumn` abstraction to create the various table variations.  Are there examples where you believe we'll need full sub component markup?

levithomason:
Also, we took inspiration from https://github.com/facebook/fixed-data-table.  I'd check that out if you haven't already.

layershifter:
> Are there examples where you believe we'll need full sub component markup?
- First SUI example:

``` html
<tfoot>
    <tr>
       <th colspan="3">
           <div class="ui right floated pagination menu">...</div>
       </th>
    </th>
  </tr>
</tfoot>
```
- [Definition](http://semantic-ui.com/collections/table#definition) example with checkboxes;
- [Structured](http://semantic-ui.com/collections/table#structured) example seems, too;

levithomason:
I suppose in either case it does make sense to start with full sub components and then build out the abstraction using those.

layershifter:
Added first docs.

levithomason:
Seeing the docs, how about mapping the sub components like so.  I've run this by our team as well, we couldn't find any further improvements.  Feedback welcomed:

``` jsx
<table>   <Table>
<thead>   <Table.Header>
<tfoot>   <Table.Footer>
<tbody>   <Table.Body>
<tr>      <Table.Row>
<td>      <Table.Cell>
<th>      <Table.HeaderCell>
```

I also think we should work out the shorthand API before proceeding further.  This way we don't box ourselves in.

jeffcarbs:
Some ideas off the top of my head for shorthand. Not really sure which I like best, but figured I'd just get some thoughts from you guys:

``` js
const columns = [
  {
    headerCell: 'Name'
    cell: (props) => `${props.firstName} ${props.lastName}`
    sortable: true
  },
  {
    headerCell: 'Occupation'
    cell: 'string'
  }
]

const rows = [
  { active: true, firstName: 'Bob', lastName: 'Smith' },
  { firstName: 'John', lastName: 'Doe' }
]

<Table columns={columns} rows={rows} />
<Table columnSchema={columns} rowData={rows} />

const headerCells = [
  { content: 'name', sortable: true },
  'occupation'
]

const bodyCells = [
  (props) => `${props.firstName} ${props.lastName}`,
  'n/a'
]

<Table headerCells={columns} bodyCells={bodyCells} rowData={rowData} />

const header = {
  className: 'my-thead',
  cells: [
    { content: 'name', sortable: true },
    'occupation'
  ]
}

const body = {
  className: 'my-tbody',
  cells: [
    (props) => `${props.firstName} ${props.lastName}`,
    'n/a'
  ]
}

<Table header={header} body={body} rowData={rowData} />
```

I kinda like the column approach, although how do we handle colspan? Maybe that's a case where you'd have to build your own table? Or maybe we'd allow passing an array of values for head or body to indicate there are two cells. E.g.

``` js
const columns = [
  {
    headerCell: {
      content: 'Name',
      colspan: 2
    }
    bodyCell: [
      (props) => `${props.firstName}`,
      (props) => `${props.lastName}`,
    ],
    sortable: true
  },
  {
    headerCell: 'Occupation'
    bodyCell: 'some-string'
  }
]
```

More work but more flexible:

``` js
const MyHeaderRow = (props) => (
  <Table.Row>
    <Table.HeaderCell content='Name' />
    <Table.HeaderCell content='Occupation' />
  </Table.Row>
)

const MyTableRow = (props) => (
  <Table.Row>
    <Table.Cell content={`${props.firstName} ${props.lastName}`} />
    <Table.Cell content='some-string' />
  </Table.Row>
)

<Table headerAs={MyHeaderRow} rowAs={MyTableRow} rowData={rowData} />
```

layershifter:
Docs finished, tests added. We need review there :dancers:

layershifter:
Note, there are no shorthands and sortable variation.

test/specs/collections/Table/TableHeader-test.js
```diff
@@ -0,0 +1,26 @@
...
+describe('TableHeader', () => {
+  common.isConformant(TableHeader)
+  common.rendersChildren(TableHeader)
+
+  it('renders as a thead by default', () => {
+    shallow(<TableHeader />)
+      .should.have.tagName('thead')
+  })
+
+  describe('fullWidth', () => {
```
jeffcarbs: Would be nice for this to just be `common.propKeyOnlyToClassName(TableHeader, 'fullWidth'), although I don't think it would work out of the box because that helper expects the prop name to be the same as the class name.

Maybe update the common function to be:

``` js
// Current
it('adds prop name to className', () => {
  shallow(createElement(Component, { ...requiredProps, [propKey]: true }))
    .should.have.className(propKey)
})

// Change to
it('adds prop name to className', () => {
  shallow(createElement(Component, { ...requiredProps, [propKey]: true }))
    .should.have.className(_.kebabCase(propKey))
})
```

layershifter: There are two props with custom tests `fullWidth` and `singleLine`. We cannot use `_.kebabCase` there, because it will convert `singleLine` to `single-line` while it must be `single line`.

BTW,  `full-width` looks like design mistake in SUI, but we have that we have.

As really solution I see updating all common tests to handle additional argument, but I don't like it because we have only two cases for this now.

``` diff
-export const propKeyOnlyToClassName = (Component, propKey, requiredProps = {}) => {
+export const propKeyOnlyToClassName = (Component, propKey, className, requiredProps = {}) => {
```

docs/app/Examples/collections/Table/States/TableError.js
```diff
@@ -0,0 +1,44 @@
...
+        <Table.Row>
+          <Table.HeaderCell>Name</Table.HeaderCell>
+          <Table.HeaderCell>Status</Table.HeaderCell>
+          <Table.HeaderCell>Notes</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+        <Table.Row>
+          <td>No Name Specified</td>
```
levithomason: Seems all `td` should be `Table.Cell` correct?

src/collections/Table/TableFooter.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { customPropTypes, META } from '../../lib'
+import TableHeader from './TableHeader'
+
+function TableFooter(props) {
+  const { as } = props
+
+  return <TableHeader {...props} as={as} />
```
levithomason: This can simply spread all props, `as` will be included:

``` js
<TableHeader {...props} />
```

src/collections/Table/TableHeaderCell.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { customPropTypes, META } from '../../lib'
+import TableCell from './TableCell'
+
+function TableHeaderCell(props) {
+  const { as } = props
+
+  return <TableCell {...props} as={as} />
```
levithomason: This can also spread all props without destructuring `as`.

levithomason:
Just updated to the latest master, reviewing...

docs/app/Examples/collections/Table/Variations/TableFixedLine.js
```diff
@@ -0,0 +1,40 @@
...
+          <Table.HeaderCell>Description</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+        <Table.Row>
+          <Table.Cell>John</Table.Cell>
+          <Table.Cell>Approved</Table.Cell>
+          <Table.Cell
+            title="This is much too long to fit I'm sorry about that"
```
levithomason: Possible cruft?  I don't see the `title` prop defined in the `TableCell`.

docs/app/Examples/collections/Table/Variations/TableFixedLine.js
```diff
@@ -0,0 +1,40 @@
...
+          >
+            This is much too long to fit I'm sorry about that
```
levithomason: Let's make this content a little longer so it wraps.  On a 15" MacBook it is not quite long enough for `...` to show up:

image

docs/app/Examples/collections/Table/Variations/TableFullWidth.js
```diff
@@ -0,0 +1,63 @@
...
+        <Table.HeaderCell>Registration Date</Table.HeaderCell>
+        <Table.HeaderCell>E-mail address</Table.HeaderCell>
+        <Table.HeaderCell>Premium Plan</Table.HeaderCell>
+      </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+      <Table.Row>
+        <Table.Cell collapsing>
+          <Checkbox fitted slider />
```
levithomason: FYI, [`fitted` is auto added](http://localhost:8080/modules/checkbox#fitted) when there is no `label`.

docs/app/Examples/collections/Table/Types/TableApprove.js
```diff
@@ -0,0 +1,63 @@
...
+          <Table.HeaderCell>Registration Date</Table.HeaderCell>
+          <Table.HeaderCell>E-mail address</Table.HeaderCell>
+          <Table.HeaderCell>Premium Plan</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+        <Table.Row>
+          <Table.Cell collapsing>
+            <Checkbox fitted slider />
```
levithomason: Noted in `TableFullWidth.js`, the [`fitted` prop is auto added](http://localhost:8080/modules/checkbox#fitted) when there is no `label` defined.

docs/app/Examples/collections/Table/Variations/TableLarge.js
```diff
@@ -0,0 +1,44 @@
...
+        <Table.Row>
+          <Table.HeaderCell>Name</Table.HeaderCell>
+          <Table.HeaderCell>Status</Table.HeaderCell>
+          <Table.HeaderCell>Notes</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+        <Table.Row>
+          <td>John</td>
```
levithomason: Looks like these should be `TableCell` as well.

docs/app/Examples/collections/Table/Variations/TableSmall.js
```diff
@@ -0,0 +1,44 @@
...
+        <Table.Row>
+          <Table.HeaderCell>Name</Table.HeaderCell>
+          <Table.HeaderCell>Status</Table.HeaderCell>
+          <Table.HeaderCell>Notes</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+
+      <Table.Body>
+        <Table.Row>
+          <td>John</td>
```
levithomason: Note, `TableCell`.

test/specs/collections/Table/TableHeader-test.js
levithomason: > As really solution I see updating all common tests to handle additional argument, but I don't like it because we have only two cases for this now.

Agreed, though we can easily default `className = propKey`. I've checked and there is only one test using `requiredProps`, so there is very little to update:

``` jsx
export const propKeyOnlyToClassName = (Component, propKey, className = propKey, requiredProps = {}) => {}
```

I'll make a PR that does this, then it can be overridden to work with `full-width` and `single line`.

levithomason (commented):
I rebased to the latest master, and made a trivial update to the examples mapping over colors.  I'll merge a separate PR right now for the commont tests className update so we can use common tests for `full-width` and `single line`.

Other than these, I left a few cleanup comments.

test/specs/collections/Table/TableHeader-test.js
levithomason: Done, including some other refactors that were pending. See https://github.com/TechnologyAdvice/stardust/pull/548.  In short, the final common test argument is now a config object instead of `requiredProps`:

``` jsx
common.propKeyOnlyToClassName(FormField, 'required', {
  requiredProps: { label: '' },
})

common.propKeyOnlyToClassName(Table, 'fullWidth', {
  className: 'full-width',
})

common.propKeyOnlyToClassName(Table, 'singleLine', {
  className: 'single line',
})
```

docs/app/Examples/collections/Table/Variations/TableFixedLine.js
layershifter: No, it's title for `Table.Cell` that you will be on hover.

docs/app/Examples/collections/Table/Variations/TableFixedLine.js
layershifter: Added some lorem :smile:

docs/app/Examples/collections/Table/Variations/TableFullWidth.js
layershifter: :+1:

docs/app/Examples/collections/Table/Types/TableApprove.js
layershifter: :+1:

docs/app/Examples/collections/Table/Variations/TableLarge.js
layershifter: Yep.

docs/app/Examples/collections/Table/Variations/TableSmall.js
layershifter: Yep.

layershifter:
Seems, fixed all remarks.

---

BTW, @levithomason seems I missed something in #545, can you take a look?

levithomason:
Looking, will also update to the latest factories fix in #550.  If all is well, will merge.

levithomason:
Looks good, merging.  We can fix/update on follow up PRs.  Thanks for the massive amount of work here.

layershifter:
🎉

levithomason:
Released in `stardust@0.47.0`

besh:
It looks like this doesn't support the `data` prop anymore. Is the new intended use to map your data to table cells directly in the component now?

``` js
data.map((value)=> {
  return <Table.Cell>value</Table.Cell>
});
```

If that's the case, are you planning to bring back support for the `data` prop?

EDIT: Looking again through here, it seems shorthanding was chatted about. I like some of the ideas that @jcarbo tossed out, mainly the 

``` js
<Table headerAs={MyHeaderRow} rowAs={MyTableRow} rowData={rowData} />
```

levithomason:
@hankthewhale and others looking, I've just updated the doc site usage of TableColumn to map over data manually.  You can see a comparison of what that looks like here: https://github.com/TechnologyAdvice/stardust/commit/99f0a2ab77c7b03f223ead48528a29c4bbf6b7d4

besh:
Thanks @levithomason for the update. Checking it out now.

PR #463, opened 2016-09-02
docs(Loader): update docs

layershifter:
- update Loader docs to v1 scheme;
- separate prop example;
- cleanup TODOs.


levithomason:
Thanks, I'll merge this.  Note:

I'm finding it tougher to navigate the codebase with the shorter filenames and export names:

`LoaderInvertedExample` => `Inverted`

As there are several `Inverted.js` examples.  The v1 docs only require the `index.js` file in each directory, but the filename and export name can remain long.  I'm thinking we leave them longer for easier navigation and only rename the `index.js` files.  Thoughts?

Also, I'll likely do all these files in one RegEx sweep once all examples are done.

layershifter:
Maybe `LoaderInverted`? Two word filename: first word is component name and second word is example name.

levithomason:
That is another valid possibility.

The primary concern for me is using "jump to file" or "jump to symbol" features in my editor and being able to quickly find files or symbols.  When I want to "jump to" a particular example or class I can do like so:

`loex` => **Lo**ader**Ex**ample.js
`loine` => **Lo**ader**I**nverted**Ex**ample.js

So, I think there is some merit to having `Example` in the filename.  Especially when there are lots of components, like with the form or feed.  There are enough src/ components (non examples) that the list of components matching `feed` would be tough to find.  Whereas, with the form, which includes `Example`, you can add `ex` to the search to find just the examples.

I think we leave current filenames as is for now, and allow trying different filenames moving forward.  Then when everything is complete, we can standardize on the best one.

PR #467, opened 2016-09-05
feat(Dropdown): close on blur

dylankiss:
Fixed the TODO that the Dropdown didn't close on blur.
I did not yet make it open on focus (other TODO), since I wanted to ask for your comments on this.
In my opinion that behavior is not always wanted, but it is how vanilla SUI works right now.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -280,6 +280,7 @@ export default class Dropdown extends Component {
       document.removeEventListener('keydown', this.moveSelectionOnKeyDown)
       document.removeEventListener('keydown', this.selectItemOnEnter)
       document.removeEventListener('keydown', this.removeItemOnBackspace)
+      this.close()
     }
 
     // opened / closed
@@ -299,6 +300,10 @@ export default class Dropdown extends Component {
       document.removeEventListener('keydown', this.selectItemOnEnter)
       document.removeEventListener('keydown', this.removeItemOnBackspace)
       document.removeEventListener('click', this.closeOnDocumentClick)
+      if (prevState.focus && this.state.focus) {
+        document.addEventListener('keydown', this.openOnArrow)
+        document.addEventListener('keydown', this.openOnSpace)
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -994,6 +1011,7 @@ describe('Dropdown Component', () => {
         .simulate('focus')
 
       domEvent.keyDown(document, { key: 'ArrowDown' })
+      domEvent.keyDown(document, { key: 'ArrowDown' })
```
levithomason, line 1014: Looks like a duplicate down arrow here.

levithomason, line 1014: Ah, never mind, I see the close on blur requires another arrow to open.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1008,6 +1026,7 @@ describe('Dropdown Component', () => {
         .simulate('blur')
         .simulate('focus')
 
+      domEvent.keyDown(document, { key: 'ArrowDown' })
```
levithomason, line 1029: There is a comment on 1024 that we should update or remove:

``` js
// blur, focus, move item selection up
```

Since we are now opening after the focus.

test/specs/modules/Dropdown/Dropdown-test.js
levithomason, line 1014: There is a comment on 1008 that we should update or remove:

``` js
// blur, focus, move item selection down
```

Since we are now opening after the focus.

src/modules/Dropdown/Dropdown.js
```diff
@@ -798,6 +798,7 @@ export default class Dropdown extends Component {
         active={isActive(opt.value)}
         onClick={this.handleItemClick}
         selected={selectedIndex === i}
+        onMouseDown={e => e.preventDefault()}
```
levithomason: Let's add a comment here something to effect of:

``` js
// prevent default to allow item select without closing on blur
```

This way, looking at this line of code later on gives us some inclination as to why we are preventing default.

levithomason:
Thanks for the solid work.  I've left a few requests for adding/updating comments only.  Then we're good to merge.

dylankiss:
@levithomason Changed the comments to your requests.

levithomason:
Released in `stardust@0.41.2`, thanks much for the PR.

PR #468, opened 2016-09-05
feat(Dropdown): Select highlighted item on blur

dylankiss:
Added an option to select the currently highlighted item on blur (defaults to `true` like vanilla SUI).

Depends on #467

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -125,6 +125,9 @@ export default class Dropdown extends Component {
     /** Message to display when there are no results. */
     noResultsMessage: PropTypes.string,
 
+    /** Define whether the highlighted item should be selected on blur */
+    selectOnBlur: PropTypes.bool,
+
     // ------------------------------------
     // Callbacks
     // ------------------------------------
@@ -203,6 +206,7 @@ export default class Dropdown extends Component {
     icon: 'dropdown',
     additionLabel: 'Add:',
     noResultsMessage: 'No results found.',
+    selectOnBlur: true,
   }
 
   static autoControlledProps = [
@@ -373,14 +377,9 @@ export default class Dropdown extends Component {
     this.trySetState({ open: true })
   }
 
-  selectItemOnEnter = (e) => {
-    debug('selectItemOnEnter()')
-    debug(keyboardKey.getName(e))
-    if (keyboardKey.getCode(e) !== keyboardKey.Enter) return
-    e.preventDefault()
-
+  selectHighlightedItem = (e) => {
     const { multiple, onAddItem, options } = this.props
-    const value = this.getSelectedItemValue()
+    const value = _.get(this.getSelectedItem(), 'value')
 
     // prevent selecting null if there was no selected item value
     if (!value) return
@@ -399,6 +398,15 @@ export default class Dropdown extends Component {
       this.onChange(e, value)
       this.close()
     }
+  }
+
+  selectItemOnEnter = (e) => {
+    debug('selectItemOnEnter()')
+    debug(keyboardKey.getName(e))
+    if (keyboardKey.getCode(e) !== keyboardKey.Enter) return
+    e.preventDefault()
+
+    this.selectHighlightedItem(e)
   }
 
   removeItemOnBackspace = (e) => {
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -686,10 +702,12 @@ export default class Dropdown extends Component {
...
     let _text = placeholder
-    if (text && !searchQuery) {
-      _text = text
-    } else if (searchQuery) {
+    if (searchQuery) {
       _text = null
+    } else if (text) {
+      _text = text
+    } else if (open && !multiple) {
+      _text = this.getSelectedItemText()
```
levithomason: This line is the same as 712 below:

``` jsx
/* 710 */ _text = this.getSelectedItemText()
/* 712 */ _text = _.get(this.getItemByValue(value), 'text')
```

Suggest removing the `getSelectedItemText` method, since it is only used once.  It is best to leave the code inline vs adding indirection.  Then, combine the if/else clauses:

``` js
} else if (hasValue || open && !multiple) {
}
```

levithomason:
Looks good so far.  I'll finish review after we first merge and release the close on blur PR.

src/modules/Dropdown/Dropdown.js
dylankiss: @levithomason The two lines are not exactly the same. Line 710 gets the text of the currently highlighted item (by going up and down with the arrow keys), while line 712 gets the current selected item (the `value` in state). In hindsight maybe `getHighlightedItemText()` might be more indicative?

Line 710 adds the behavior that the 'placeholder' value changes with what the value would become when you would blur the dropdown (as it is with the vanilla SUI component).

levithomason: **EDIT** see next comment.

~~I see, let's update the naming then.  The SUI `selected` class means it is highlighted, so all code usage of the term also means highlighted.  The class SUI `active` class means it the item is part of the Dropdown value.~~

~~So, a user "selects" items by arrow up/down, then they make one "active" by click, tab, or enter.~~

src/modules/Dropdown/Dropdown.js
```diff
@@ -481,7 +490,7 @@ export default class Dropdown extends Component {
     debug('handleBlur()')
     const { onBlur } = this.props
     if (onBlur) onBlur(e)
-    this.setState({ focus: false })
+    this.setState({ focus: false, blurEvent: e })
```
levithomason: Because of [React's event pooling](https://facebook.github.io/react/docs/events.html#event-pooling) there's a chance this event will be nullified by the time it is passed to `selectHighlightedItem`.  Instead of putting it on state, and passing it later, let's just pass it right here like `selectOnEnter` does:

``` js
handleBlur = (e) => {
  debug('handleBlur()')
  const { multiple, onBlur, selectOnBlur } = this.props
  if (onBlur) onBlur(e)
  if (selectOnBlur && !multiple) this.selectHighlightedItem(e)
  this.setState({ focus: false })
}
```

Then we can remove this from `componentDidUpdate`:

``` diff
-if (this.props.selectOnBlur && !this.props.multiple) this.selectHighlightedItem(this.state.blurEvent)
```

src/modules/Dropdown/Dropdown.js
levithomason: After looking at the code closer, I see that it is correctly named and performing a different task.  I think we should refactor these two methods into one, however.  There is a lot of code duplication between them:

``` js
getSelectedItemValue = () => {
  const { selectedIndex } = this.state

  return _.get(options, `[${selectedIndex}].value`)
}

getSelectedItemText = () => {
  const { selectedIndex } = this.state

  return _.get(options, `[${selectedIndex}].text`)
}
```

How about we replace it with:

``` js
getSelectedItem = () => {
  const { selectedIndex } = this.state

  return options[selectedIndex]
}
```

And the usages like so:

``` diff
-const value = this.getSelectedItemValue()
+const value = _.get(this.getSelectedItem(), 'value')

-_text = this.getSelectedItemText()
+_text = _.get(this.getSelectedItem(), 'text')
```

dylankiss: That is indeed better. I made a little change and used `_.get(options, `[${selectedIndex}]`)` instead of `options[selectedIndex]`, since the index could initially be `undefined`.

src/modules/Dropdown/Dropdown.js
dylankiss: You're right ;) Didn't take that into account. Changed it.

dylankiss:
@levithomason I adapted the PR according to the comments.

levithomason:
This is ready to merge, however, it has some conflicts that need resolved due to updates made to the close on blur updates.

dylankiss:
@levithomason I didn't see any conflicts. I already rebased after the 'close on blur' issue was merged. Just rebased again on the latest master commit without any conflicts.

levithomason:
Released in `stardust@0.41.4`, thanks for the PR!

PR #475, opened 2016-09-08
feat(Dropdown): editable search

dylankiss:
I added an extra property `editableSearch` that allows you to edit the search query of a Dropdown also after it previously lost focus.

**How to use**: 
1. Select an option in the dropdown. The dropdown will close.
2. Click in the dropdown search input.
3. Your search query is set to the current value and you can adapt the search query from here.

**Use case**: You have `allowAdditions` set to true. With `editableSearch` you can easily edit your custom value by clicking in the search input again and changing part of your value.

sep-08-2016 17-49-12

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Types/Selection.js
+++ b/docs/app/Examples/modules/Dropdown/Types/Selection.js
@@ -14,6 +14,7 @@ export default class DropdownSelectionExample extends Component {
       isFetching: false,
       multiple: true,
       search: true,
+      editableSearch: false,
       searchQuery: null,
       value: [],
       options: getOptions(),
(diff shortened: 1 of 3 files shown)


levithomason:
I'd like to wait for more community requests before implementing this into the core library, for two reasons.
1. My intuition says this is more of an edge case than a common use case.  So, it is probably best left to implement separately rather than in the core library.
2. It is not a SUI Dropdown feature.  Though we do add features beyond what SUI provides, I want to be cautious when deviating.

It's a bummer to write code and not have it implemented, sorry about that.  One way to help prevent that is to first open up an issue asking about the feature.  This way, the community and myself can weigh in before embarking.  You can reference #416 as an example of hashing out features to figure out what code we should write.  Hope this helps and thanks for all the contributions to date.

dylankiss:
@levithomason I understand it is more of an edge case than something that will often be used. It is a feature that works well for us, but we might not have the right control available for what we want to do.

We basically need an autosuggest input field, but as far as I've seen there is no such thing in Stardust or Semantic UI. ~~Would you allow a PR to add such functionality to `Input` or so? Or do you have other suggestions to achieve this? Thanks in advance!~~

**EDIT:** Looks like `<Search />` might be more appropriate for our use case. I see that this feature is on the way in #424. Hope it will be ready soon ;)

levithomason:
Yes sir, there have multiple Dropdown feature requests that are actually Search requests.  Once I clear the pending legacy component updates (#269), I'll focus on new components.  If Search isn't merged by then, I suspect that will be one of my first.

PR #476, opened 2016-09-08
feat(Header): Add subheader prop

### LINKED ISSUE #461: Header: add subheader shorthand
This is the current markup for Headers with subheader children:

``` jsx
<Header as='h2'>
  Sale!
  <Header.Subheader>
    $10.99
  </Header.Subheader>
</Header>
```

In keeping with other component shorthand, we ought to allow the `Subheader` to be configured via props:

``` jsx
<Header as='h2' content='Sale!' subheader='$10.99' />
```

As with all shorthand props, we should `disallow()` children when using the `subheader` shorthand.  There are already docs, examples, and tests for Header `icon` and `image` shorthand that can be referenced.

layershifter:
This PR fixes #461.

### DIFF
--- a/src/elements/Header/Header.js
+++ b/src/elements/Header/Header.js
@@ -64,9 +64,17 @@ function Header(props) {
     )
   }
 
+  if (children) {
+    return (
+      <ElementType {...rest} className={classes}>
+        {children}
+      </ElementType>
+    )
+  }
+
   return (
     <ElementType {...rest} className={classes}>
-      {children || content}
+      {content}
       {subheader && <HeaderSubheader content={subheader} />}
     </ElementType>
   )
--- a/src/elements/Header/Header.js
+++ b/src/elements/Header/Header.js
@@ -22,7 +22,7 @@ import HeaderContent from './HeaderContent'
 function Header(props) {
   const {
     color, content, dividing, block, attached, floated, inverted, disabled, sub, size, textAlign,
-    icon, image, children, className,
+    icon, image, children, className, subheader,
   } = props
 
   const classes = cx(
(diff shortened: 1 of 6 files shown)


src/elements/Header/Header.js
```diff
@@ -47,22 +47,29 @@ function Header(props) {
 
   if (icon && typeof icon !== 'boolean') {
     return (
-      <ElementType className={classes} {...rest}>
+      <ElementType {...rest} className={classes}>
         {createIcon(icon)}
         {content && <HeaderContent>{content}</HeaderContent>}
+        {subheader && <HeaderSubheader content={subheader} />}
```
layershifter, line 53: Seems, I missed test-case there.

src/elements/Header/Header.js
```diff
@@ -47,22 +47,29 @@ function Header(props) {
...
     )
   }
 
-  return <ElementType {...rest} className={classes}>{children || content}</ElementType>
+  return (
+    <ElementType {...rest} className={classes}>
+      {children || content}
+      {subheader && <HeaderSubheader content={subheader} />}
+    </ElementType>
+  )
```
levithomason, line 80: Currently, the Header will render children with shorthand even though there are prop type warnings.  While we're here, let's split this return with a conditional return to only render children if provided.  The goal in doing this in all components is to enforce proper usage (not rendering shorthand when children are present):

``` jsx
  if (children) {
    return (
        {children}
      </ElementType>
    )
  }

  return (
      {content}
      {subheader && <HeaderSubheader content={subheader} />}
    </ElementType>
  )
```

src/elements/Header/HeaderSubheader.js
```diff
@@ -1,16 +1,22 @@
...
-  const { children, className } = props
+  const { children, className, content } = props
   const classes = cx('sub header', className)
   const rest = getUnhandledProps(HeaderSubheader, props)
   const ElementType = getElementType(HeaderSubheader, props)
 
   return (
     <ElementType className={classes} {...rest}>
-      {children}
+      {children || content}
```
levithomason, line 19: In this case, having a single return is OK because either children or content will be rendered.  But, never will children _and_ shorthand be rendered.  👍

test/specs/elements/Header/Header-test.js
```diff
@@ -65,4 +67,28 @@ describe('Header', () => {
...
+      shallow(<Header subheader={text} />)
+        .find('HeaderSubheader')
+        .should.have.prop('content', text)
+    })
+    it('adds HeaderSubheader as child when given a name to icon prop', () => {
+      const text = faker.hacker.phrase()
+
+      shallow(<Header icon='user' subheader={text} />)
+        .find('HeaderSubheader')
+        .should.have.prop('content', text)
```
levithomason, line 84: Thanks for shallow rendering and only testing this prop was passed, instead of testing the Subheader's handling of the prop in the Header test 👍  This is a great test.

levithomason:
Will merge after the Header render is updated to strictly render children or shorthand, but not both.

levithomason:
Released in `stardust@0.42.2`, thanks!

PR #479, opened 2016-09-09
Dropdown: Custom trigger element

### LINKED ISSUE #477: Dropdown: Support arbitrary element for `text`
### My use-case

I have a menu that I want to be triggered by a custom element. It's not a `select`-type dropdown, just a menu of items that becomes visible after clicking this element.

_EDIT:_ To give a more concrete example, think of something like the github user menu with a custom `<Avatar />` component: 
screen shot 2016-09-08 at 5 55 15 pm

Currently, you can specify text for the dropdown:

``` jsx
<Dropdown text='something'>
  <Dropdown.Menu>
    {/* items */}
  </Dropdown.Menu>
</Dropdown>
```

``` html
<div tabindex="0" class="ui dropdown">
  <div class="text">
    something
  </div>
  <i class="dropdown icon">
  </i>
  <div class="menu transition">
  </div>
</div>
```

Ideally I'd want to be able to pass in a custom component that gets rendered where the `text` currently does. For example:

``` jsx
<Dropdown text={MyCustomComponent}>
  <Dropdown.Menu>
    {/* items */}
  </Dropdown.Menu>
</Dropdown>
```

``` html
<div tabindex="0" class="ui dropdown">
  <MyCustomComponent>
    <!--  whatever else -->
  </MyCustomComponent>
  <i class="dropdown icon">
  </i>
  <div class="menu transition">
  </div>
</div>
```

Note: This is different than using `as={MyCustomComponent}`, which would render the outer container as `MyCustomComponent`:

``` jsx
<Dropdown as={MyCustomComponent}>
  <Dropdown.Menu>
    {/* items */}
  </Dropdown.Menu>
</Dropdown>
```

``` html
<MyCustomComponent tabindex="0" class="ui dropdown">
  <div class="text">
  </div>
  <i class="dropdown icon">
  </i>
  <div class="menu transition">
  </div>
</MyCustomComponent>
```

I only want to render the `text` portion to be custom.

I could also see allowing composing the dropdown text manually via `Dropdown.Text` like:

``` jsx
<Dropdown>
  <Dropdown.Text as={MyCustomComponent} />
  <Dropdown.Menu>
    {/* items */}
  </Dropdown.Menu>
</Dropdown>
```

``` html
<div tabindex="0" class="ui dropdown">
  <MyCustomComponent class='text'>
    <!--  whatever else -->
  </MyCustomComponent>
  <i class="dropdown icon">
  </i>
  <div class="menu transition">
  </div>
</div>
```

Thoughts?

jeffcarbs:
Fixes #477

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Types/Trigger.js
+++ b/docs/app/Examples/modules/Dropdown/Types/Trigger.js
@@ -1,0 +1,31 @@
+import React from 'react'
+import { Dropdown, Icon } from 'stardust'
+
+const trigger = (
+  <span>
+    <Icon name='user' />
+    Hello, Bob
+  </span>
+)
+
+const DropdownExample = () => (
+  <Dropdown trigger={trigger}>
+    <Dropdown.Menu>
+      <Dropdown.Item disabled>
+        <div>Signed In as Bob Smith</div>
+        <strong>Bob Smith</strong>
+      </Dropdown.Item>
+      <Dropdown.Divider />
+      <Dropdown.Item>Your Profile</Dropdown.Item>
+      <Dropdown.Item>Your Stars</Dropdown.Item>
+      <Dropdown.Item>Explore</Dropdown.Item>
+      <Dropdown.Item>Integrations</Dropdown.Item>
+      <Dropdown.Item>Help</Dropdown.Item>
+      <Dropdown.Divider />
+      <Dropdown.Item>Settings</Dropdown.Item>
+      <Dropdown.Item>Sign Out</Dropdown.Item>
+    </Dropdown.Menu>
+  </Dropdown>
+)
+
+export default DropdownExample
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -104,6 +104,12 @@ export default class Dropdown extends Component {
     /** Name of the hidden input which holds the value. */
     name: PropTypes.string,
 
+    /** Custom element to trigger the menu to become visible. Takes place of 'text'. */
+    trigger: customPropTypes.every([
+      customPropTypes.disallow(['selection', 'text']),
+      PropTypes.node,
+    ]),
+
     /**
      * Allow user additions to the list of options (boolean).
      * Requires the use of `selection`, `options` and `search`.
@@ -861,6 +867,7 @@ export default class Dropdown extends Component {
       error,
       disabled,
       scrolling,
+      trigger,
     } = this.props
 
     // Classes
@@ -916,7 +923,7 @@ export default class Dropdown extends Component {
         {this.renderLabels()}
         {this.renderSearchInput()}
         {this.renderSearchSizer()}
-        {this.renderText()}
+        {trigger || this.renderText()}
         {createIcon(icon)}
         {this.renderMenu()}
       </ElementType>
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -370,6 +370,24 @@ describe('Dropdown Component', () => {
     })
   })
 
+  describe('trigger', () => {
+    it('displays the trigger', () => {
+      const text = 'Hey there'
+      const trigger = <div className='trigger'>{text}</div>
+
+      wrapperRender(<Dropdown options={options} trigger={trigger} />)
+        .find('.trigger')
+        .should.contain.text(text)
+    })
+    it('ignores the text prop', () => {
+      const text = faker.hacker.phrase()
+      const trigger = <div className='trigger'>{text}</div>
+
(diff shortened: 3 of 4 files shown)


docs/app/Examples/modules/Dropdown/Types/Trigger.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { Dropdown, Icon } from 'stardust'
+
+const trigger = (
+  <div style={{display: 'inline-block'}}>
```
levithomason: How about a `<span>` here to avoid the style prop?

src/modules/Dropdown/Dropdown.js
```diff
@@ -104,6 +104,12 @@ export default class Dropdown extends Component {
     /** Name of the hidden input which holds the value. */
     name: PropTypes.string,
 
+    /** Custom element to trigger the menu to become visible. Takes place of 'text'. */
+    trigger: customPropTypes.every([
+      customPropTypes.disallow(['selection']),
```
levithomason: Let's also disallow `text` since it replaces it.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -370,6 +370,25 @@ describe('Dropdown Component', () => {
     })
   })
 
+  describe('trigger', () => {
+    it('displays the trigger', () => {
+      const text = 'Hey there'
+      const trigger = (<div className='trigger'>{text}</div>)
```
levithomason: Can remove extra parens here.  We only require them on multi-line JSX.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -370,6 +370,25 @@ describe('Dropdown Component', () => {
...
+        .find('.trigger')
+        .should.contain.text(text)
+    })
+    it('ignores the text prop', () => {
+      const text = faker.hacker.phrase()
+      const trigger = (<div className='trigger'>{text}</div>)
+
+      wrapperRender(<Dropdown options={options} trigger={trigger} text={text} />)
+        .find('.text')
+        .should.have.lengthOf(0)
```
levithomason: For better assertion failure messages, let's use:

``` js
wrapperRender(<Dropdown options={options} trigger={trigger} text={text} />)
  .should.have.exactly(0).descendants('.text')
```

docs/app/Examples/modules/Dropdown/Types/Trigger.js
```diff
@@ -0,0 +1,29 @@
...
+    <Icon name='user' />
+    Hello, Bob
+  </div>
+)
+
+const DropdownExample = () => (
+  <Dropdown trigger={trigger}>
+    <Dropdown.Menu>
+      <Dropdown.Item text='New' />
```
levithomason: You may already be doing this, but how about shortening this Menu to only a couple common user menu options?  Like settings, logout, etc.

src/modules/Dropdown/Dropdown.js
```diff
@@ -687,6 +693,8 @@ export default class Dropdown extends Component {
   // ----------------------------------------
 
   renderText = () => {
+    if (this.props.trigger) return this.props.trigger
```
levithomason: I was thinking about this as well, we could/should probably just put this directly in the render() so it is more clear what is happening:

``` jsx
{trigger || this.renderText()}
```

jeffcarbs: Agreed. Also, many of the other renders (renderLabels, etc) will not be used when there's a `trigger` present since those are all `selection` related. I think it's probably OK for now but it would probably be good to think about how to split these two separate use-cases out.

levithomason: Ah, you're totally right.  Elsewhere, we've been using early conditional returns to only render what is necessary.  It also strictly prevents people from incorrectly using the component since it won't render incorrect usages.  What about this?

``` jsx
    if (trigger) {
      return (
        <ElementType
          {...rest}
          {...conditionalProps}
          onChange={this.onChange}
          onClick={this.handleClick}
          className={classes}
        >
          {trigger}
          {createIcon(icon)}
          {this.renderMenu()}
        </ElementType>
      )
    }

    return (
      <ElementType
        {...rest}
        {...conditionalProps}
        onChange={this.onChange}
        onClick={this.handleClick}
        className={classes}
      >
        {this.renderHiddenInput()}
        {this.renderLabels()}
        {this.renderSearchInput()}
        {this.renderSearchSizer()}
        {this.renderText()}
        {createIcon(icon)}
        {this.renderMenu()}
      </ElementType>
    )
```

jeffcarbs:
Feedback implemented 👍

levithomason:
Looks great, thanks!  Will merge/release once it passes.

levithomason:
Released in `stardust@0.42.3`, docs should be live in ~5m.

PR #480, opened 2016-09-09
fix(createFactory): do not create elements from `null` values

jeffcarbs:
Before this PR, all three of these specs failed:

```
    it('allows disabling icon via null', () => {
      Dropdown.defaultProps.icon.should.equal('dropdown')
      wrapperRender(<Dropdown icon={null} />)
        .should.not.contain.descendants('.icon')
    })
    it('allows disabling icon via undefined', () => {
      Dropdown.defaultProps.icon.should.equal('dropdown')
      wrapperRender(<Dropdown icon={undefined} />)
        .should.not.contain.descendants('.icon')
    })
    it("allows disabling icon via ''", () => {
      Dropdown.defaultProps.icon.should.equal('dropdown')
      wrapperRender(<Dropdown icon={''} />)
        .should.not.contain.descendants('.icon')
    })
```

...meaning there's no way to disable the icon on the `Dropdown` component. I'd say the third probably shouldn't disable it (it would have an icon element with an empty name), but one of the first two should.

This updates `createFactory` to do nothing for both `undefined` and `null`. Previously it only ignored `undefined`, which meant for components that had a default value (e.g. Dropdown icon defaults to `dropdown`) there was no way to disable it. The default prop will override `undefined` but it won't override `null`.

There may be wider implications to consider, but I figured it'd be easier/faster to discuss this over a PR with a potential fix vs an issue.

### DIFF
--- a/src/factories/createFactory.js
+++ b/src/factories/createFactory.js
@@ -19,7 +19,7 @@ const mergePropsAndClassName = (props, extraProps) => {
 /**
  * Return a function that produces ReactElements.  Similar to React.createFactory with some extras.
  * If the returned factory is passed a ReactElement it will be cloned.
- * If it passed undefined it will do nothing.
+ * If it passed null or undefined it will do nothing.
  * @param {function|string} Component A ReactClass or string
  * @param {function} mapValueToProps A function that maps a primitive value to the Component props
  * @param {function} [mergeExtraProps=mergePropsAndClassName]
@@ -38,7 +38,7 @@ const createFactory = (Component, mapValueToProps, mergeExtraProps = mergePropsA
     }
 
     // Map values to props and create an ReactElement
-    if (val !== undefined) {
+    if (!_.isNil(val)) {
       return <Component {...mergeExtraProps(mapValueToProps(val), extraProps)} />
     }
   }
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -514,13 +514,13 @@ export const implementsWidthProp = (Component, options, requiredProps = {}) => {
 
 export const implementsIconProp = (Component, requiredProps = {}) => {
   const iconName = faker.hacker.noun()
-  const assertValid = (element) => {
+  const assertValid = (element, expectedName = iconName) => {
     const wrapper = shallow(element)
     wrapper
       .should.have.descendants('Icon')
     wrapper
       .find('Icon')
-      .should.have.prop('name', iconName)
+      .should.have.prop('name', expectedName)
   }
 
   describe('icon (common)', () => {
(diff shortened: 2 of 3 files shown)


levithomason:
Thoughts:
1. Agree `''` should be passed as as empty string
2. React treats `undefined` as not present, so I don't think it should remove the Icon, since it is equivalent to not having an `icon` prop at all.
3. Agree that `null` should remove the icon, however, I'm seeing that this already works without any changes to the current factory.

http://technologyadvice.github.io/stardust/modules/dropdown#custom-trigger

image

jeffcarbs:
@levithomason - on master I ran those specs and logged the resulting HTML:

```
      ✖ allows disabling icon via null
LOG: '<div tabindex="0" class="ui dropdown"><div class="text"></div><i class="icon"></i><div class="menu transition"></div></div>'
      ✖ allows disabling icon via undefined
LOG: '<div tabindex="0" class="ui dropdown"><div class="text"></div><i class="dropdown icon"></i><div class="menu transition"></div></div>'
      ✖ allows disabling icon via ''
LOG: '<div tabindex="0" class="ui dropdown"><div class="text"></div><i class="icon"></i><div class="menu transition"></div></div>'
```

`null` is treated like `''`, so while the icon is technically not visible since there's no icon name class, there's an empty `<i class="icon"></i>`, which messes up the spacing. Here's a sceenshot with icon set to `null` (i manually made the icon's background red in chrome dev tools so you can see it):
screen shot 2016-09-08 at 10 22 21 pm

I think the following makes sense:
- `icon={undefined}` means use the default icon or don't output an icon if no default
- `icon={null}` means don't output an icon
- `icon=''` means output an icon with '' as the name
  - `icon='anyval` means output an icon with 'anyval' as the name

levithomason:
I see, it is still rendering an empty `<i>` which is not desirable.  I'll merge this.

I'd like to get more coverage than just the Dropdown.  We have a `common.implementsIconProp()`, however, testing that an icon _isn't_ present does not apply to every use case.  So, instead of asserting it there, I think we should instead add a `createFactory-test.js` and assert that the returned factory ignores null values.  This way, we know all components using the factory are covered.

jeffcarbs:
@levithomason - check that last commit. Some of tests within `implementsIconProp` weren't actually testing anything. The tests were checking for the existence of an `i` tag, but since `shallow` doesn't deeply render children, it will never render the contents of the `Icon` and thus never render an `i`.

In fixing the specs, I realized it wasn't testing for the case where the component had a default icon so I updated it to do so. I also temporarily rolled back the code change I made in this PR to ensure the specs I added were red before and green after.

levithomason:
Thanks, also fixed this and some other things in the Button PR #295.  Though, I think I like your default solution better than my decision to remove "does not exist" tests.

Thanks much!

levithomason:
Released in `stardust@0.42.4`

Issue #478, opened 2016-09-08
Dropdown: support `disabled` items


levithomason:
A SUI Dropdown item can be disabled individually.

![image](https://cloud.githubusercontent.com/assets/5067638/18370173/f91a22f0-75df-11e6-853c-e49bf6dc1add.png)

http://semantic-ui.com/modules/dropdown.html#disabled

Currently, the `Dropdown.Item` does not support the `disabled` prop/className.  The prop is passed directly to the underlying `div` as an HTML attribute instead.

levithomason:
Fixed in #482

PR #482, opened 2016-09-09
feat(Dropdown): support disabled items

jeffcarbs:
Addresses #478

I added code that should handle ignoring a click on the dropdown item and keeping the menu open. However, the onClick is never getting called so the event is bubbling to the `Dropdown` element itself which is toggling it's visibility.

Is there some magic happening somewhere that disables clicks on an element with `disabled` passed as props? As you can see, it's not being applied onto the element. It's just being used to add the `disabled` class.

_EDIT_ AHH it's `pointer-events: none;` 😞 . Thoughts on how to handle this? Two ideas:
- manually enable pointer-events via inline style
- in the dropdown `handleClick` function, check if the click was within the menu (can you even do that with synthetic events?) and ignore it.

I don't love either of those idea, tbh.

### DIFF
--- a/docs/app/Examples/modules/Dropdown/index.js
+++ b/docs/app/Examples/modules/Dropdown/index.js
@@ -43,8 +43,13 @@ const DropdownExamples = () => (
     <ExampleSection title='States'>
       <ComponentExample
         title='Disabled'
-        description='A disabled dropdown menu or item does not allow user interaction'
+        description='A disabled dropdown menu does not allow user interaction'
         examplePath='modules/Dropdown/States/Disabled'
+      />
+      <ComponentExample
+        title='Disabled Item'
+        description='A disabled dropdown item does not allow user interaction'
+        examplePath='modules/Dropdown/States/DisabledItem'
       />
     </ExampleSection>
   </div>
--- a/docs/app/Examples/modules/Dropdown/index.js
+++ b/docs/app/Examples/modules/Dropdown/index.js
@@ -43,12 +43,10 @@ const DropdownExamples = () => (
     <ExampleSection title='States'>
       <ComponentExample
         title='Disabled'
-        description='A disabled dropdown menu does not allow user interaction'
+        description='A disabled dropdown menu or item does not allow user interaction'
         examplePath='modules/Dropdown/States/Disabled'
       />
       <ComponentExample
-        title='Disabled Item'
-        description='A disabled dropdown item does not allow user interaction'
         examplePath='modules/Dropdown/States/DisabledItem'
       />
     </ExampleSection>
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -385,10 +385,10 @@ export default class Dropdown extends Component {
 
   selectHighlightedItem = (e) => {
     const { multiple, onAddItem, options } = this.props
-    const { value, disabled } = this.getSelectedItem() || {}
+    const value = _.get(this.getSelectedItem(), 'value')
 
-    // prevent selecting null (if there was no selected item value) or disabled options
-    if (!value || disabled) return
+    // prevent selecting null if there was no selected item value
+    if (!value) return
 
     // notify the onAddItem prop if this is a new value
     if (onAddItem && !_.some(options, { text: value })) onAddItem(value)
@@ -519,7 +519,7 @@ export default class Dropdown extends Component {
     if (search && newQuery && !open) this.open()
 
     this.setState({
-      selectedIndex: 0,
+      selectedIndex: this.getEnabledIndices()[0],
       searchQuery: newQuery,
     })
   }
@@ -528,9 +528,11 @@ export default class Dropdown extends Component {
   // Getters
   // ----------------------------------------
 
-  getMenuOptions = () => {
+  // There are times when we need to calculate the options based on a value
+  // that hasn't yet been persisted to state.
+  getMenuOptions = (value = this.state.value) => {
     const { multiple, search, allowAdditions, additionPosition, additionLabel, options } = this.props
-    const { searchQuery, value } = this.state
+    const { searchQuery } = this.state
 
     let filteredOptions = options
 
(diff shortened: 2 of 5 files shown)


dylankiss:
@jcarbo Your second option can be achieved by listening on the `<DropdownMenu />` component, I guess. It is the parent containing all `<DropdownItem />` components and would receive the click event first.

levithomason:
We can detect click inside/outside of a node, but this requires a `ref` and reading the DOM.  Also, there are several elements that could be inside of the item.  The event target could be any one of them.  We'd have to apply the `disabled` prop as a `div` attribute as well, since we wouldn't have access to the original virtual DOM item's props. Then we could check the event target and traverse its parents looking for `.item[disabled]`, which if found, stop immediate propagation.

As much as I don't like to do so, I think we add the inline style.  Then, we can simply add the style in `renderOptions()` and in `handleItemClick()` add `item.disabled` as a condition to stop immediate propagation:

``` js
// renderOptions() map() Dropdown.Item
style={{ ...opt.style, pointerEvents: 'all' }}

// handleItemClick()
const item = this.getItemByValue(value)

if (multiple || item.disabled) {
  e.nativeEvent.stopImmediatePropagation()
}
```

levithomason:
Inline comments are much appreciated anywhere there is odd or special case handling like this.  Thanks in advance.

jeffcarbs:
Thanks for the feedback.

I've implemented the inline-style option. As much as I hate inline styles, it was significantly simpler than dealing with the DOM. Clicking a disabled item now just does nothing and keeps the menu open.

Let me know if there's anything else this is missing.

docs/app/Examples/modules/Dropdown/index.js
```diff
@@ -43,9 +43,14 @@ const DropdownExamples = () => (
...
         title='Disabled'
-        description='A disabled dropdown menu or item does not allow user interaction'
+        description='A disabled dropdown menu does not allow user interaction'
         examplePath='modules/Dropdown/States/Disabled'
       />
+      <ComponentExample
+        title='Disabled Item'
+        description='A disabled dropdown item does not allow user interaction'
+        examplePath='modules/Dropdown/States/DisabledItem'
+      />
```
levithomason, line 51: Let's update the Disabled description to include "menu or item" and drop the title/description on the disabled item example.  This way they show as two examples under one heading:

image

src/modules/Dropdown/Dropdown.js
```diff
@@ -385,10 +385,10 @@ export default class Dropdown extends Component {
 
   selectHighlightedItem = (e) => {
     const { multiple, onAddItem, options } = this.props
-    const value = _.get(this.getSelectedItem(), 'value')
+    const { value, disabled } = this.getSelectedItem() || {}
```
levithomason: Am I correct in thinking you cannot actually highlight (select) a disabled item due to the [logic added](https://github.com/TechnologyAdvice/stardust/pull/482/files#diff-609214949026e594dbfd98481664c928R648) to the `moveSelectionBy`?  If so, this method should never be called with a disabled item as it is only called on enter and blur and requires first using keyboard nav to highlight the item.

---

It's beyond the scope of this PR, but It could be this method name needs updated.  `selected` means highlighted, `active` means it is part of the value.  `selectHighlightedItem` is a little redundant as the highlighted item _is_ already selected.  It might be more accurately named `makeSelectedItemActive` or similar.

src/modules/Dropdown/Dropdown.js
```diff
@@ -631,13 +634,19 @@ export default class Dropdown extends Component {
...
+
     // next is after last, wrap to beginning
     // next is before first, wrap to end
     let nextIndex = selectedIndex + offset
     if (nextIndex > lastIndex) nextIndex = 0
     else if (nextIndex < 0) nextIndex = lastIndex
 
     this.setState({ selectedIndex: nextIndex })
+
+    if (options[nextIndex].disabled) return this.moveSelectionBy(offset)
```
levithomason: This will set state every time it skips a disabled item.  It would be ideal to only set state once a valid next index was found.  What if we also allow passing a `startIndex`?

``` js
moveSelectionBy = (offset, startIndex = this.state.selectedIndex) => {
  // update selectedIndex usage
}
```

Then we can recurse before setting state:

``` js
if (options[nextIndex].disabled) return this.moveSelectionBy(offset, nextIndex)

this.setState({ selectedIndex: nextIndex })
```

src/modules/Dropdown/Dropdown.js
jeffcarbs: If the first item is disabled, it will be "selected" by default since the selectedIndex is set to 0. 

One option to fix this would be to have a `getFirstValidItem` function or something (open to a better name) that would give the index of the first non-disabled item. If none (e.g. all options disabled) the selected index could be `-1`.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -125,6 +125,22 @@ describe('Dropdown Component', () => {
...
+      const nativeEvent = { nativeEvent: { stopImmediatePropagation: _.noop } }
+
+      options[randomIndex].disabled = true
+
+      wrapperMount(<Dropdown options={options} selection />)
+        .simulate('click', nativeEvent)
+        .find('DropdownItem')
+        .at(randomIndex)
+        .simulate('click', nativeEvent)
+        .should.have.not.prop('selected', true)
```
levithomason: Could we reverse the `.have.not` to `.should.not.have.prop('selected', true)`?

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -260,6 +299,23 @@ describe('Dropdown Component', () => {
...
     })
+    it('does not become active on enter when disabled', () => {
+      const disabledOptions = _.map(options, (o) => ({ ...o, disabled: true }))
+      const item = wrapperMount(<Dropdown options={disabledOptions} selection />)
+        .simulate('click')
+        .find('DropdownItem')
+        .at(0)
+
+      // initial item props
+      item.should.have.prop('selected', true)
```
levithomason: Hm, it seems we should never be able to select a disabled item.

levithomason:
Left some inline comments.  If we could get a test that asserts disabled items can never be selected, then we can forgo the test asserting they aren't made active on Enter.  Since, we'd also need to ensure they aren't made active on blur.

jeffcarbs:
Implemented the remaining feedback, namely preventing disabled elements from being selected. If all elements are disabled, the `selectedIndex` is undefined which prevents things like selection.

Actually found a bug and another broken test: `keeps the selection within the range of remaining options`. The logic for keeping the selection in-bounds happens before the value is persisted to state, so the most recently selected item isn't filtered out of the options returned by `getMenuOptions`. I wound up fixing the bug and updating the logic to take the disabled state into account. So it will keep you in bounds at the next non-disabled item.

levithomason:
Thanks much, released in `stardust@0.42.6`.

Issue #469, opened 2016-09-06
Dropdown: support custom search function


levithomason:
Currently the Dropdown supports a boolean `search` prop.  This prop enables filtering the Dropdown menu items.  There is no way to control the filtering logic, which defaults to "item text contains query".

Per #416, the `search` prop should be extended to accept a filter function `search={myFilter}`.  The filter function would be used in the Dropdown `getMenuOptions()` in place of the currently hard coded filter function:

``` js
// Dropdown.js getMenuOptions()...
filteredOptions = _.filter(filteredOptions, (opt) => re.test(opt.text))
```

Instead, we would call the search prop function with the remaining options and the search query, allowing the user to filter the options themselves:

``` js
filteredOptions = this.props.search(filteredOptions, searchQuery)
```

levithomason:
Merged in #484

PR #484, opened 2016-09-09
feat(Dropdown): support custom search function

dylankiss:
Addresses issue #469

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -131,8 +131,14 @@ export default class Dropdown extends Component {
     /** Message to display when there are no results. */
     noResultsMessage: PropTypes.string,
 
-    /** Define whether the highlighted item should be selected on blur */
+    /** Define whether the highlighted item should be selected on blur. */
     selectOnBlur: PropTypes.bool,
+
+    /** Make the dropdown options searchable by substring matching (default) or with a custom search function. */
+    search: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.func,
+    ]),
 
     // ------------------------------------
     // Callbacks
@@ -184,9 +190,6 @@ export default class Dropdown extends Component {
       PropTypes.bool,
       PropTypes.oneOf(_meta.props.pointing),
     ]),
-
-    /** Make the dropdown options searchable. */
-    search: PropTypes.bool,
 
     /** The text displayed in the dropdown, usually for the active item. */
     text: PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -134,6 +134,12 @@ export default class Dropdown extends Component {
     /** Define whether the highlighted item should be selected on blur */
     selectOnBlur: PropTypes.bool,
 
+    /** Make the dropdown options searchable (`boolean` or `function(options, searchQuery) => filteredOptions`) */
```
levithomason: Let's keep the types out of the doc block descriptions since they are shown formally in the docs:

image

Function signatures are an issue still, though, I think we'll solve these with jsdoc tags at a later time.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1027,6 +1027,23 @@ describe('Dropdown Component', () => {
...
+      // search for value yields 2 results as per our custom search function
+      search.simulate('change', { target: { value: searchQuery } })
+
+      searchFunction.should.have.been.calledOnce()
+      searchFunction.should.have.been.calledWithMatch(options, searchQuery)
+
+      wrapper
+        .find('DropdownItem')
+        .should.have.lengthOf(2, 'Searching with custom search function did not yield 2 results.')
+    })
```
levithomason, line 1045: This is a great test, thanks 👍

src/modules/Dropdown/Dropdown.js
dylankiss: I see. The function signature could be handy though (as long as there is no other way to see this). Do I add that temporarily to the comment, or totally not?

Something like: `Make the dropdown options searchable by substring matching (default) or with a custom search function `(options, searchQuery) => filteredOptions``, or do I leave off the function signature?

test/specs/modules/Dropdown/Dropdown-test.js
dylankiss: It's easy if you have such a good test framework set up ;) 👍

src/modules/Dropdown/Dropdown.js
levithomason: Your proposed description without the signature is great.

I definitely agree a signature is better than no signature.  However, I want to have a consistent mechanism in place for adding them. I plan to extend the props table to handle function signatures via standardized jsdoc tags (i.e. `@param` etc).  I'll add an issue for this now!

dylankiss:
@levithomason I changed the comment. For consistency also added a period to the previous prop comment (forgot it before).

src/modules/Dropdown/Dropdown.js
levithomason: Added https://github.com/TechnologyAdvice/stardust/issues/487

levithomason:
Released in `stardust@0.42.5`, thanks for the PR!

Issue #474, opened 2016-09-07
refactor(Label): update props to the latest specs


levithomason:
**Shorthand**

Currently, the Label allows shorthand props `icon`, `image`, and `text` along with `children`.  These should be `disallow()`ed via `propTypes` and early returns put in place to render one or the other.

**`text`**

This prop should be renamed to `content`.  See #391.

**`link`**

This prop renders as an `a` tag.  It should go away as we now have augmentation `<Label as='a' />` instead.

**`empty`**

Currently, the Label auto applies the `empty` className when there are no children.  Magic is bad, unless it is magic application of something that is _always required_, like a [Checkbox fitted prop](http://technologyadvice.github.io/stardust/modules/checkbox#fitted).  A Label with no children does not _always require_ the `empty` class.  It is a slightly larger but valid Label without it:

![image](https://cloud.githubusercontent.com/assets/5067638/18392520/f0bfb200-7666-11e6-9124-be94038f1c6f.png)

We should remove the auto application of this className.

**Docs & Tests**

Docs and tests should then be updated for all usages.

jeffcarbs:
Somewhat related - I'm having an issue with this line which adds the `empty` class: https://github.com/TechnologyAdvice/stardust/blob/master/src/elements/Label/Label.js#L41.

I'm trying to set a circular label with the `text` shorthand prop and it's getting the `empty` class. This is because that logic is only based on children: 

``` js

// Currently
<Label text={count} circular />
<div class="ui empty circular label">11</div>

// To make it non-empty, needs children
<Label circular>{count}</Label>
<div class="ui circular label">11</div>
```

I could see `empty` being set when there's no content of any kind, although I think I'd prefer to just set it manually:

``` js
// Better (but still not completely intuitive):
<Label circular />
<div class="ui empty circular label"></div>

// Ideal:
<Label empty circular />
<div class="ui empty circular label"></div>
```

levithomason:
Yep, that's bad.  See https://github.com/TechnologyAdvice/stardust/issues/485

levithomason:
Actually, let's just do this all at once, updated description above.

levithomason:
Fixed in #486

PR #486, opened 2016-09-09
refactor(Label): update props to the latest specs

jeffcarbs:
Addresses #474

Some comments/questions:
- The issue description mentioned removing the `link` prop which set the element type to `a` but there was also logic so if `image` or `onClick` was passed it would be rendered as `a`. I removed that whole thing in favor of just using the `as` prop. I figured the less magic the better.
- There's similar logic for the `detail` [here](https://github.com/TechnologyAdvice/stardust/pull/486/files#diff-fde3c67bba82fd0efc2088a5fe1a4686R50). Thoughts on removing that logic and the `detailLink` prop in favor of a `detailAs` which would behave like `as`?

### DIFF
--- a/src/elements/Label/Label.js
+++ b/src/elements/Label/Label.js
@@ -2,6 +2,7 @@ import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
 import {
+  customPropTypes,
   getElementType,
   getUnhandledProps,
   META,
@@ -19,8 +20,8 @@ import { Icon, Image } from '../'
  */
 function Label(props) {
   const {
-    attached, basic, children, color, corner, className, circular, detail, detailLink, floating, horizontal,
-    icon, image, link, onClick, onDetailClick, onRemove, pointing, removable, ribbon, size, tag, text,
+    attached, basic, children, color, corner, className, circular, detail, detailAs, empty, floating, horizontal,
+    icon, image, onClick, onDetailClick, onRemove, pointing, removable, ribbon, size, tag, content,
   } = props
 
   const handleClick = e => onClick && onClick(e, props)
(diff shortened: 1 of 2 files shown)


levithomason:
I would suspect users would want pointer cursors (`a` tags) when onClick is provided, however, I'll hardly fight for keeping magic :P

+1 to `detailAs` updates.

jeffcarbs:
Made the `detail` changes and rebased onto the latest master 👍

src/elements/Label/Label.js
```diff
@@ -31,31 +32,30 @@ function Label(props) {
...
     (image || childrenUtils.someByType(children, Image) || childrenUtils.someByType(children, 'img')) && 'image',
     'label',
     className
   )
 
-  const DetailComponent = (detailLink || onDetailClick) && 'a' || 'div'
-  const ElementType = getElementType(Label, props, () => {
-    if (image || link || onClick) return 'a'
-  })
+  const DetailComponent = detailAs || 'div'
```
levithomason, line 50: Hm.  I think here we should be using `getElementType()` for consistent handling.  This makes me realize that this component is actually missing sub components.  It should have `Label.Detail` for instance.  That would also solve all the issues related to it.

If you like, we can merge this as is, and open an issue to create Label sub components.

levithomason:
Cool, will merge on pass.  LMK what you want to do regarding sub components mentioned [here](https://github.com/TechnologyAdvice/stardust/pull/486/files#r78247674)

src/elements/Label/Label.js
jeffcarbs: +1 for merging this as-is and opening a new issue to talk about sub components.

Quick thoughts:
There are a few sub components, but only `detail` can be semi-customized. The others are either there or not and are already covered by factories.

I think `Label.Detail` might make sense to break out into its own component with props:
- as
- children
- content
- onClick

It'd be really nice to keep a shorthand prop on `Label` for a vanilla detail, though. Can't think of an analogous component relationship off the top of my head.

levithomason: Definitely. See https://github.com/TechnologyAdvice/stardust/issues/488

levithomason:
Released in `stardust@0.43.0`, thanks!

PR #495, opened 2016-09-12
fix(dropdown): Label props for multi-select

jeffcarbs:
The labels for the `multiple` dropdown type were broken since they were using the old props. This fixes it.


levithomason:
Released in `stardust@0.44.1`, thanks!

PR #496, opened 2016-09-13
pass dataKey params to headerRenderer in Table.js

Edison-Hsu:
if data is initialized with [] we will get error because data[0] is undefined.
we should pass dataKey to get correct header content.

### DIFF
--- a/src/collections/Table/Table.js
+++ b/src/collections/Table/Table.js
@@ -98,7 +98,7 @@ export default class Table extends Component {
   }
 
   _getHeaders() {
-    const { children, data, sort } = this.props
+    const { children, sort } = this.props
 
     return Children.map(children, (column) => {
       const { dataKey, headerRenderer } = column.props
--- a/src/collections/Table/Table.js
+++ b/src/collections/Table/Table.js
@@ -102,7 +102,7 @@ export default class Table extends Component {
 
     return Children.map(children, (column) => {
       const { dataKey, headerRenderer } = column.props
-      const content = headerRenderer ? headerRenderer(data[0]) : _.startCase(dataKey)
+      const content = headerRenderer ? headerRenderer(dataKey) : _.startCase(dataKey)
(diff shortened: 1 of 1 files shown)


levithomason:
There shouldn't be an error, the argument will just be `undefined`.  This is correct because there is no data in the array.  The headerRender is passed the first object (row) in your data array.  This way, you can make headers from the initial values in your array.  I don't think passing the key only is correct.

Edison-Hsu:
Hi, thank you for your answer.

But I do not understand. If we pass first object(row) data to headerRender function, how do we filter corresponding header content (column name).
eg: 

```
data = [
  {
    a: 1,
    n: 'Mary',
    c: 2
  },
  {
    a:10,
    n: 'John',
    c: 20
]
```

We will get `{a: 1, n: 'Mary', c: 2}` in headerRender function, and I want make header content like: 

| name | age |
| --- | --- |
| Mary | 1 |
| John | 10 |

What should I do to convert n => name, a => ago and get correct order from first object.

But if we pass dataKey, we will easily get coresponding header content by a object like:

```
headerRender(dataKey) {
  return {a: 'age', n: 'name'}[dataKey]
}
```

levithomason:
Column order is preserved by child TableColumn order.  Specify a `dataKey` on each column, then set a header renderer to return what you want on the header.

``` jsx
import React from 'react'
import { Table } from 'stardust'

const data = [
  { a: 1, n: 'Mary', c: 2 },
  { a: 10, n: 'John', c: 20 },
]

const TableExample = () => (
  <Table data={data}>
    <Table.Column headerRenderer={() => 'Name'} dataKey='n' />
    <Table.Column headerRenderer={() => 'Age'} dataKey='a' />
  </Table>
)

export default TableExample
```

image

levithomason:
To be clear, you only need to use the row data if your headers are dynamically generated from the row data. If they are just strings, just return the string.

PR #498, opened 2016-09-13
refactor(deprecate): remove all deprecated

levithomason:
We've had a number of issues related to deprecated components.  While we're pre 1.0, we will not manage deprecated props and components. This PR removes all utils and deprecated components.

This PR also resolves the broken search on the docs by removing deprecated components.


levithomason:
Released in `stardust@0.44.0`

PR #500, opened 2016-09-15
fix(Dropdown): open/close on focus/blur

### LINKED ISSUE #493: Dropdown: <Dropdown.Item as={'a'} /> with href does not function appropriately
Going to post some thoughts below, but for context here's the actual issue I'm experiencing:

``` js
import React from 'react'
import { Dropdown, Icon } from 'stardust'

const stopDomEvent = (e) => {
  e.preventDefault()
  e.stopPropagation()
  e.nativeEvent.stopImmediatePropagation()
}

const itemProps = {
  as: 'a',
  href: '/something',
  onClick: (e) => {
    stopDomEvent(e)
    console.log('clicked!', e)
  }
}

const DropdownTriggerExample = () => (
  <Dropdown>
    <Dropdown.Menu>
      <Dropdown.Item disabled>
        Signed in as <strong>Bob Smith</strong>
      </Dropdown.Item>
      <Dropdown.Divider />
      <Dropdown.Item {...itemProps}>Your Profile</Dropdown.Item>
      <Dropdown.Item {...itemProps}>Your Stars</Dropdown.Item>
      <Dropdown.Item {...itemProps}>Explore</Dropdown.Item>
      <Dropdown.Item {...itemProps}>Integrations</Dropdown.Item>
      <Dropdown.Item {...itemProps}>Help</Dropdown.Item>
      <Dropdown.Divider />
      <Dropdown.Item {...itemProps}>Settings</Dropdown.Item>
      <Dropdown.Item {...itemProps}>Sign Out</Dropdown.Item>
    </Dropdown.Menu>
  </Dropdown>
)

export default DropdownTriggerExample
```

The `onClick` is never fired, here's the console output: 

```
stardust:dropdown handleBlur() +2s
stardust:dropdown render() +1ms
stardust:dropdown props +1ms Object {trigger: Object, children: Object, icon: "dropdown", additionLabel: "Add:", noResultsMessage: "No results found."…}
stardust:dropdown state +1ms Object {value: "", searchQuery: "", selectedIndex: -1, focus: false, open: true…}
stardust:dropdown renderHiddenInput() +1ms
stardust:dropdown name:      undefined +1ms
stardust:dropdown selection: undefined +0ms
stardust:dropdown value:      +1ms
stardust:dropdown renderLabels() +1ms
stardust:dropdown componentDidUpdate() +1ms
stardust:dropdown changed state: +1ms Object {focus: true}
stardust:dropdown close() +0ms
stardust:dropdown dropdown is not yet closed +1ms
stardust:dropdown render() +1ms
stardust:dropdown props +0ms Object {trigger: Object, children: Object, icon: "dropdown", additionLabel: "Add:", noResultsMessage: "No results found."…}
stardust:dropdown state +1ms Object {value: "", searchQuery: "", selectedIndex: -1, focus: false, open: false…}
stardust:dropdown renderHiddenInput() +1ms
stardust:dropdown name:      undefined +0ms
stardust:dropdown selection: undefined +1ms
stardust:dropdown value:      +1ms
stardust:dropdown renderLabels() +0ms
stardust:dropdown componentDidUpdate() +4ms
stardust:dropdown changed state: +1ms Object {open: true, dropdownClasses: "active visible", menuClasses: "visible"}
stardust:dropdown close() +1ms
stardust:dropdown dropdown is already closed +0ms
```

Removing the `href` attribute, it works. Here's the output:

```
clicked! Proxy {dispatchConfig: Object, _targetInst: ReactDOMComponent, _dispatchListeners: Array[2], _dispatchInstances: Array[2], nativeEvent: MouseEvent…}
```

Pretty sure it was introduced via related to this PR: https://github.com/TechnologyAdvice/stardust/pull/467

I kinda think I know what's happening here and why that PR works in other cases...

When you click on a regular item, or an anchor without an href, focus leaves the `Dropdown` onMouseDown of some other element, **non-focusable** (the key part) element. That PR prevents the default of the `onMouseDown` on a dropdown item, which prevents the blur. In fact, after clicking an item the menu closes from the handleClick callback and calling `document.activeElement` shows that the dropdown still has focus.

My hunch is that by adding an `href` to the `a` tag, you make it focusable. So the blur doesn't come from the `onMouseDown` but instead by the fact that the `Dropdown.Item` is now focused. Rolling back this one change (https://github.com/TechnologyAdvice/stardust/pull/467/files#diff-609214949026e594dbfd98481664c928R283), clicking an item with an href, and inspecting the `document.activeElement` shows that the `a` is in fact focused.

Somewhat wider gripe: The dropdown element (and now search, which was based on dropdown) is doing some event manipulation within click handlers to prevent the dropdown from closing. e.g. 

``` js
  handleItemClick = (e, value) => {
    // ...

    // prevent toggle() in handleClick()
    e.stopPropagation()
    // prevent closeOnDocumentClick() if multiple or item is disabled
    if (multiple || item.disabled) {
      e.nativeEvent.stopImmediatePropagation()
    }

    // ...
```

Thoughts on keeping a ref to the dropdown itself and just preventing things like `closeOnDocumentClick` via something like (which is actually already done in `<Modal>`):

``` js
if (this.refs.self.contains(e.target)) return
```

For the blur event, all browsers but FireFox support a property called `relatedTarget` which is the node that focus was moved to, so in the `handleBlur` we could have something like:

``` js
if (this.refs.self.contains(e.relatedTarget)) return
```

I can experiment with a PR to see what unforeseen issues this would create. Just wanted to get your thoughts first to see if this is wildly off base.

levithomason:
Fixes #493 

The Dropdown opens/closes on focus/blur.  Focus and blur are triggered by either keyboard (tab) or mouse events (click).  Focus and blur fire before on mouse down while click fires on mouse up.

This meant focus and blur would cause the menu to open/close before the click handler would also toggle the menu close.  Further, the blur event would close the menu before the click event could fire on a menu item.

This PR adds a `isMouseDown` flag.  Focus/blur functionality for open/close is only called if the mouse is not down.  This allows the user to tab to open/close, but the focus/blur fired from a click is ignored.  Allowing the user to select items without the Dropdown closing and click on the Dropdown to toggle its open state without double toggling the open state.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -385,6 +385,7 @@ export default class Dropdown extends Component {
     debug('openOnSpace()')
     if (keyboardKey.getCode(e) !== keyboardKey.Spacebar) return
     if (this.state.open) return
+    e.preventDefault()
     this.trySetState({ open: true })
   }
 
@@ -393,6 +394,7 @@ export default class Dropdown extends Component {
     debug('openOnArrow()')
     if (!_.includes([keyboardKey.ArrowDown, keyboardKey.ArrowUp], code)) return
     if (this.state.open) return
+    e.preventDefault()
     this.trySetState({ open: true })
   }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -27,7 +27,7 @@ const _meta = {
   name: 'Dropdown',
   type: META.TYPES.MODULE,
   props: {
-    pointing: ['bottom left', 'bottom right'],
+    pointing: ['left', 'right', 'top', 'top left', 'top right', 'bottom', 'bottom left', 'bottom right'],
     additionPosition: ['top', 'bottom'],
   },
 }
@@ -158,6 +158,9 @@ export default class Dropdown extends Component {
 
     /** Called with the React Synthetic Event on Dropdown focus. */
     onFocus: PropTypes.func,
+
+    /** Called with the React Synthetic Event on Dropdown mouse down. */
+    onMouseDown: PropTypes.func,
 
     // ------------------------------------
     // Style
@@ -379,19 +382,17 @@ export default class Dropdown extends Component {
   }
 
   openOnSpace = (e) => {
+    debug('openOnSpace()')
     if (keyboardKey.getCode(e) !== keyboardKey.Spacebar) return
     if (this.state.open) return
-    e.preventDefault()
-    // TODO open/close should only change state, open/closeMenu should be called on did update
     this.trySetState({ open: true })
   }
 
   openOnArrow = (e) => {
     const code = keyboardKey.getCode(e)
+    debug('openOnArrow()')
     if (!_.includes([keyboardKey.ArrowDown, keyboardKey.ArrowUp], code)) return
     if (this.state.open) return
-    e.preventDefault()
-    // TODO open/close should only change state, open/closeMenu should be called on did update
     this.trySetState({ open: true })
   }
 
@@ -456,8 +457,10 @@ export default class Dropdown extends Component {
   // Component Event Handlers
   // ----------------------------------------
 
-  handleMouseDown = () => {
+  handleMouseDown = (e) => {
     debug('handleMouseDown()')
+    const { onMouseDown } = this.props
+    if (onMouseDown) onMouseDown(e)
     this.isMouseDown = true
     document.addEventListener('mouseup', this.handleDocumentMouseUp)
   }
@@ -708,30 +711,15 @@ export default class Dropdown extends Component {
 
   open = () => {
     debug('open()')
-    debug(`dropdown is ${this.state.open ? 'already' : 'not yet'} open`)
-    if (this.state.open) return
     const { search } = this.props
     if (search) this._search.focus()
 
-    this.trySetState({
-      open: true,
-    }, {
-      dropdownClasses: 'active visible',
-      menuClasses: 'visible',
-    })
+    this.trySetState({ open: true })
   }
 
   close = () => {
     debug('close()')
-    debug(`dropdown is ${!this.state.open ? 'already' : 'not yet'} closed`)
-    if (!this.state.open) return
-
-    this.trySetState({
-      open: false,
-    }, {
-      dropdownClasses: '',
-      menuClasses: '',
-    })
+    this.trySetState({ open: false })
   }
 
   toggle = () => this.state.open ? this.close() : this.open()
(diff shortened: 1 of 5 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -849,9 +863,11 @@ export default class Dropdown extends Component {
         active={isActive(opt.value)}
         onClick={this.handleItemClick}
         selected={selectedIndex === i}
-        onMouseDown={e => e.preventDefault()} // prevent default to allow item select without closing on blur
+        // prevent default to allow item select without closing on blur
+        onMouseDown={e => e.preventDefault()}
```
jeffcarbs: I think you can get rid of this now since `preventDefault` here is meant to prevent closing on blur but you're preventing that more completely via this PR.

levithomason: I would have thought so, but it didn't seem to be the case.  I'll look at it again and see.

jeffcarbs:
I pulled this branch down to test it locally and it seems to have fixed the problems referenced in that issue, including interacting with focusable elements within the dropdown or items.

levithomason:
Awesome, thanks much.  Just need to get tests updated now.

src/modules/Dropdown/Dropdown.js
```diff
@@ -693,29 +711,15 @@ export default class Dropdown extends Component {
...
     const { search } = this.props
-    if (search) this.refs.search.focus()
+    if (search) this._search.focus()
 
-    this.trySetState({
-      open: true,
-    }, {
-      dropdownClasses: 'active visible',
-      menuClasses: 'visible',
-    })
```
levithomason, line 705: We should only be setting the visibility classes if the open state was successful.  I moved these to their respective render functions.

levithomason:
I've updated per the comments, thanks.  I also fixed another bug.

The visibility classes were being assigned every time open/close was called.  However, open/close only tries to set state, but defers to the `open` prop if it exists.  So, if the open prop was false (controlled component) then the state would not set to `open` but the classes would still be applied, showing the menu. Now, those classes are calculated in the render functions based on the `open` state.

levithomason:
Released in `stardust@0.44.2`

mclarentgp:
I am noticing an issue where I have a drop down with an OnChange event and after clicking on the drop down and selecting the value the OnChange fires off as expected, but if after doing this action and i click anywhere else on the screen it triggers the OnChange event again.  I was just wondering does this fix address this issue or is what im experiencing completely seperate from this fix?

levithomason:
This sounds like an unrelated issue.  Suggest testing against the latest version.  If it still occurs, please open a separate issue with explicit steps to reproduce.  Thanks for the report.

PR #501, opened 2016-09-15
Dropdown: Dropdown.Header via shorthand prop

jeffcarbs:
Adds support for a `header` shorthand prop for dropdown.

I was thinking of using a factory for `header` since it can have some other properties other than text (e.g. an icon). It would basically be:

``` js
const createDropdownHeader = createFactory(DropdownHeader, value => ({ content: value }))
```

However, there's no precedent in the codebase for using a factory on a subcomponent so I was unsure where to put that. Some options would be:
1. `src/factories/index.js` (seems weird since the factory is not really usable globally.)
2. `src/modules/Dropdown/factories.js`
3. `src/modules/Dropdown/Dropdown.js` (just define it at the top of the file)

Kinda leaning option 3, but would love some other opinions.

**Edit**: It seems the PR on standardizing childKey and shorthand props has some discussion related to this question about the factories. Perhaps we can hold off on the factory thing until and review this as-is until that's settled?

### DIFF
--- a/src/modules/Dropdown/DropdownHeader.js
+++ b/src/modules/Dropdown/DropdownHeader.js
@@ -1,15 +1,34 @@
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
-import { getElementType, getUnhandledProps, META } from '../../lib'
+import {
+  customPropTypes,
+  getElementType,
+  getUnhandledProps,
+  META,
+} from '../../lib'
+import { createIcon } from '../../factories'
 
 function DropdownHeader(props) {
-  const { className, children } = props
+  const { className, children, content, icon } = props
   const classes = cx('header', className)
   const rest = getUnhandledProps(DropdownHeader, props)
   const ElementType = getElementType(DropdownHeader, props)
 
-  return <ElementType className={classes} {...rest}>{children}</ElementType>
+  if (children) {
+    return (
+      <ElementType className={classes} {...rest}>
+        {children}
+      </ElementType>
+    )
+  }
+
+  return (
+    <ElementType className={classes} {...rest}>
+      {createIcon(icon)}
+      {content}
+    </ElementType>
+  )
 }
 
 DropdownHeader._meta = {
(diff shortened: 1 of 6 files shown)


src/modules/Dropdown/DropdownHeader.js
```diff
@@ -25,11 +37,23 @@ DropdownHeader.propTypes = {
...
   /** Additional classes */
   className: PropTypes.node,
+
+  /** Primary content of the header, same as children. */
+  content: PropTypes.node,
+
+  /** Add an icon by icon name or pass an <Icon /> */
+  icon: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
```
levithomason, line 62: This can also accept a `PropTypes.object`.

src/modules/Dropdown/DropdownHeader.js
```diff
@@ -1,15 +1,27 @@
...
-  const { className, children } = props
+  const { className, children, content, icon } = props
   const classes = cx('header', className)
   const rest = getUnhandledProps(DropdownHeader, props)
   const ElementType = getElementType(DropdownHeader, props)
 
-  return <ElementType className={classes} {...rest}>{children}</ElementType>
+  return (
+    <ElementType className={classes} {...rest}>
+      {createIcon(icon)}
```
levithomason, line 28: Let's also add `common.implementsIconProp()` to cover this usage.

levithomason (changes requested):

levithomason:
> **Edit:** It seems the PR on standardizing childKey and shorthand props has some discussion related to this question about the factories. Perhaps we can hold off on the factory thing until and review this as-is until that's settled?

Agreed, the PR exploring this is #345 "Reusable Component Parts".  The idea is to create standard factories for all component parts and reuse them between all components.  We can wait for that work to resolve this issue.  What you have here is OK as is. 👍

jeffcarbs:
@levithomason - Addressed your comments and rebased against the most recent master. Should be g2g 👍

levithomason (approved):

levithomason:
Released in `stardust@0.44.5`, thanks much!

Issue #502, opened 2016-09-16
fix(MenuItem): remove `down` prop


levithomason:
The "down" menu item is not supported.  It was added as it also appears in the SUI docs.  However, after opening a [bug report](https://github.com/Semantic-Org/Semantic-UI/issues/4518) it was confirmed that the SUI docs are simply incorrect.

We should remove the MenuItem `down` prop, tests, and docs.

PR #503, opened 2016-09-16
fix(MenuItem): remove `down` prop

layershifter:
Fix for #502


levithomason:
Thanks, released in `stardust@0.44.3`

PR #506, opened 2016-09-17
feat(Menu): Add `items` shorthand

layershifter:
This PR:
- Resolved old TODOs for `Menu` in docs;
- Adds `items` shorthand for `Menu`:

``` jsx
const items [
  {name: 'Open'},
  {name: 'Save'},
]

<Menu items={items} onItemClick={handler} />
```

``` jsx
const items [
  {name: 'Open', onClick: onOpenClick},
  {active: true, name: 'Save', onClick: onSaveClick},
]

<Menu items={items} />
```


levithomason:
Looks like tests crashed, `Maximum call stack exceeded`.  I've retried tests without cache to see if it resolves the issue.

There are also a number of new lines added without coverage.  I suspect codecov will fail unless tests are added for the red lines here:

image

FWIW, you can see coverage info right on the PR with the codecov browser extension (shown above).  Also, you can open `/coverage/lcov-report/index.html` and drill down into specific coverage info.

src/collections/Menu/Menu.js
```diff
@@ -128,6 +145,38 @@ class Menu extends Component {
   static Item = MenuItem
   static Menu = MenuMenu
 
+  renderItems() {
+    const { items, onItemClick } = this.props
+
+    this.trySetState({ activeIndex: _.findIndex(items, ['active', false]) || -1 })
```
jeffcarbs: I don't think there should be a `setState` call inside a render.

src/collections/Menu/Menu.js
```diff
@@ -128,6 +145,38 @@ class Menu extends Component {
...
+
+    return _.map(items, (index, item) => {
+      const { content, childKey, name, onClick, itemProps } = item
+
+      const active = this.state.activeIndex === index
+      const finalKey = childKey || [content, name].join('-')
+
+      const handleClick = (e, { itemName, itemIndex }) => {
```
jeffcarbs: Prop changes generally result in a component re-rendering. Since this function is being creating new for each item each time `renderItems` is called, these `MenuItem`s will always be receiving new props on each render. This should probably use a similar pattern of having `Menu` define a `handleItemClick` callback like the `Dropdown` does.

src/collections/Menu/Menu.js
layershifter: Agree, moved to `componentWillMount`

src/collections/Menu/Menu.js
```diff
@@ -128,6 +145,38 @@ class Menu extends Component {
...
+        const handler = onItemClick || onClick
```
levithomason: If the user passes both a Menu `onItemClick` and an Item `onClick`, their onClick will never be called.  Instead of choosing one or the other, I think we should call each separately:

``` js
if (onClick) onClick(e, { itemName, itemIndex })
if (onItemClick) onItemClick(e, { itemName, itemIndex })
```

I've also called them in order here, from the Item up to the Menu.

src/collections/Menu/Menu.js
```diff
@@ -145,28 +145,37 @@ class Menu extends Component {
...
 
-        if (handler) handler(e, { itemName, itemIndex })
+    return _.map(items, (item, index) => {
+      const { content, childKey, name, onClick, itemProps } = item
+      const finalKey = childKey || [content, name].join('-')
+
+      const handleClick = (e, data) => {
+        this.handleItemClick(e, data)
+        if (onClick) onClick(e, data)
       }
```
layershifter: > If the user passes both a Menu onItemClick and an Item onClick

Fixed there, but still

> Since this function is being creating new for each item each time renderItems is called, these MenuItems will always be receiving new props on each render.

What I can do there?

layershifter:
> There are also a number of new lines added without coverage.

I'll improve coverage and add docs soon.

src/collections/Menu/Menu.js
levithomason: Anything that doesn't recreate the handleClick function every loop would work.  You could for instance pass the `onClick` as an extra parameter to the `handleItemClick` method:

``` jsx
this.handleItemClick(e, data, onClick)
```

Then, in `handleItemClick` if `onClick` exists, call it with the same event and data.

layershifter: I'm moved condition to `handleItemClick`.

``` js
handleItemClick(e, { name, index }, onClick) {
  ...
  if (onClick) onClick(e, { name, index })
  if (onItemClick) onItemClick(e, { name, index })
}
```

But problem will be still there, `handleClick` will be created on each iteration, or not?

``` jsx
const handleClick = (e, data) => this.handleItemClick(e, data, onClick)
return <MenuItem onClick={handleClick} />
//
return <MenuItem onClick={(e, data) => this.handleItemClick(e, data, onClick)} />
```

levithomason: Sorry slow brain today, you are correct.  It looks like if you are going to call two functions, then we'll always have to create the callback function in every loop.  If we weren't calling both handleItemClick and onClick, then we could just pass `handleItemClick`.

As I write this, one other idea comes to mind.  You could just pass `onClick={this.handleItemClick}` which will get called with the item name/index.  Then, you could lookup the item in `items` and call its `onClick` with the same args if it exists:

``` js
handleItemClick = (e, { name, index }) => {
  const { items, onItemClick } = this.props
  if (_.get(items, `[${index}].onClick`)) items[index].onClick(e, { name, index })
  // ...
}
```

layershifter:
Updated to `master`, seems we ready to go.

levithomason:
Reviewing, will merge if I can

levithomason:
Just cleaned up one unused param, will merge when tests pass.

levithomason:
Released in `stardust@0.44.9`, thanks much!

PR #508, opened 2016-09-18
fix(Dropdown): skip select item on blur when closed

### LINKED ISSUE #504: Dropdown events triggering twice
I'm having an issue where I have a dropdown with an OnChange event that calls a function updateFilter that is being triggered twice.  My function updateFilter gets called once when i click the dropdown and select an item in the list and another after clicking elsewhere on the browser that is not the dropdown.

I have the following set up.

`const options = [
  { text: 'item1', value: '1' },
  { text: 'item2', value: '2' },
]`

`updateFilter = (e, value) => console.log(e)`

`<Dropdown 
selection 
options={options} 
onChange={this.updateFilter}  />`

after clicking the dropdown and selecting a value i see the following in the console
"Proxy {dispatchConfig: Object, _targetInst: ReactDOMComponent, _dispatchListeners: Array[2], _dispatchInstances: Array[2], nativeEvent: MouseEvent"

after performing the above actions i click anywhere else on the browser that is not the dropdown and see the following in the console
"Proxy {dispatchConfig: Object, _targetInst: ReactDOMComponent, _dispatchInstances: ReactDOMComponent, nativeEvent: FocusEvent, type: "blur"...}"

Is this a bug or am i doing something wrong?

levithomason:
Fixes #504 

The Dropdown method `selectHighlightedItem()` is called on blur.  This method should only continue if the Dropdown is open.

This PR adds the open logic to prevent duplicate calls to onChange.  Without this, onChange is called on item click and again when the Dropdown is blurred.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -401,11 +401,13 @@ export default class Dropdown extends Component {
   }
 
   selectHighlightedItem = (e) => {
+    const { open } = this.state
     const { multiple, onAddItem, options } = this.props
     const value = _.get(this.getSelectedItem(), 'value')
 
     // prevent selecting null if there was no selected item value
-    if (!value) return
+    // prevent selecting duplicate items when the dropdown is closed
+    if (!value || !open) return
 
     // notify the onAddItem prop if this is a new value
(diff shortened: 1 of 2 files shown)


levithomason:
Released in `stardust@0.44.4`

mclarentgp:
Thanks, the fix works great!

PR #512, opened 2016-09-20
fix(Header): properly wrap contents in HeaderContent

### LINKED ISSUE #511: fix(Header): subheader and image shorthand
The Header icon/image requires the `content` div.  The `content` div should contain all text.
## Image

The `image` shorthand currently does not wrap the `content` shorthand in HeaderContent.
## Subheader

Header subheaders should be rendered as HeaderContent children.  Currently, the `subheader` shorthand is rendered as a sibling when there is an icon/image.  Example here only shows an icon, but the issue exists for subheaders when there is an icon or an image.

``` js
<Header icon='user' content='foo' subheader='bar' />
```

Outputs:

``` html
<div class="ui header">
  <i class="user icon"></i>
  <div class="content">foo</div>
  <div class="sub header">bar</div>
</div>
```

It should output:

``` html
<div class="ui header">
  <i class="user icon"></i>
  <div class="content">
    foo
    <div class="sub header">bar</div>
  </div>
</div>
```

levithomason:
Fixes #511.

This PR wraps `subheader` shorthand in HeaderContent when there is an icon present.  It also wraps content in HeaderContent when there is an `image` provided.

### DIFF
--- a/src/elements/Header/Header.js
+++ b/src/elements/Header/Header.js
@@ -45,29 +45,24 @@ function Header(props) {
   const ElementType = getElementType(Header, props)
   const rest = getUnhandledProps(Header, props)
 
-  if (icon && typeof icon !== 'boolean') {
-    return (
-      <ElementType {...rest} className={classes}>
-        {createIcon(icon)}
-        {content && <HeaderContent>{content}</HeaderContent>}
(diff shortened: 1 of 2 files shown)


levithomason:
Released in `stardust@0.44.7`

PR #518, opened 2016-09-20
chore(build): add browser build

levithomason:
This PR adds a browser build.  It will be used in online editors (jsFiddle, Codepen, etc.) for bug reports.  We may also find use for this build in v2 of the doc site.

image


src/lib/debug.js
```diff
@@ -6,7 +6,7 @@ if (process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test') {
   // Safari private mode throws when writing to localStorage
   // https://github.com/TechnologyAdvice/stardust/issues/332
   try {
-    if (typeof localStorage === 'object') localStorage.debug = localStorage.debug || 'stardust:*'
+    if (typeof localStorage === 'object') localStorage.debug = localStorage.debug || null
```
levithomason: Made debug info opt-in as it is extremely noisy.

.babelrc
```diff
@@ -8,6 +8,9 @@
     "lodash"
   ],
   "env": {
+    "browser": {
+      "plugins": ["add-module-exports"]
+    },
```
levithomason, line 13: We need module exports for the window object, prevents us from having to do `stardust.default.Button`.

gulp/tasks/docs.js
```diff
@@ -31,9 +48,9 @@ task('generate-docs-json', () => {
     .pipe(dest(config.paths.docsSrc()))
 })
 
-task('webpack-docs', (cb) => {
-  const webpackConfig = require('../../webpack.config')
-  const compiler = webpack(webpackConfig)
+task('build:docs:dll', (cb) => {
```
levithomason, line 51: I properly named the gulp tasks.

src/index.js
```diff
@@ -1,5 +1,20 @@
...
+import * as modules from './modules'
+import * as views from './views'
+
+export default {
+  ...addons,
+  ...collections,
+  ...elements,
+  ...modules,
+  ...views,
+  version,
```
levithomason: I've added the package version to the exported object.  Available in the browser and in the commonjs build.

levithomason:
Released in `stardust@0.45.0`

PR #520, opened 2016-09-21
fix(src/index): use module exports

levithomason:
The browser build release broke the doc site, however, adding module exports breaks tests.  Opting to drop the plugin and use module exports explicitly.  This works since all other es modules remain in tact while the exported window variable for browser builds does not need to use `.default` after importing the main module.


levithomason:
Released in `stardust@0.45.1`

PR #522, opened 2016-09-21
fix(Label): Fix `ribbon` propType

layershifter:
### Small fix for `ribbon` type.

> Warning: Failed prop type: Invalid prop `ribbon` of value `true` supplied to `Label`, expected one of ["right"].

But, according to [SUI docs](http://semantic-ui.com/elements/label.html#ribbon) `ribbon` can be left, if position no defined. Component and tests accepts `bool` value, so bug is only in `propTypes`.


levithomason:
Released in `stardust@0.45.2`, thanks!

PR #523, opened 2016-09-21
feat(Menu): add `floated` and MenuItem `icon` props

layershifter:
This PR addes missed props:
- `Menu` - [`floated`](https://github.com/Semantic-Org/UI-Menu/blob/master/menu.css#L1194);
- `MenuItem` - [`icon`](https://github.com/Semantic-Org/UI-Menu/blob/master/menu.css#L1194).


levithomason:
Released in `stardust@0.45.3`

PR #526, opened 2016-09-22
feat(Label): Refactor, docs and new props

### LINKED ISSUE #488: fix(Label): add missing Detail sub component
All components with [component parts](http://semantic-ui.com/introduction/glossary.html#project-terminology) should have a corresponding sub component for that part.  Additionally, every part should have a shorthand configuration prop.  Example, Card has a `meta` "part".  Therefore:

``` jsx
<Card meta='More info here' />

<Card>
  <Card.Meta>More info here</Card.Meta>
</Card>
```

The only part a Label has is the [Detail](http://semantic-ui.com/elements/label.html#detail), therefore:

``` jsx
<Label detail='Friend' />

<Label>
  <Label.Detail>More info here</Label.Detail>
</Label>
```

Shorthand is already supported, however, we are missing the sub component.

layershifter:
This PR:
- fixes #488: adds `Detail` subcomponent;
- adds missing props [`above`](https://github.com/Semantic-Org/UI-Label/blob/master/label.css#L1154) and [`below`](https://github.com/Semantic-Org/UI-Label/blob/master/label.css#L1170);
- adds initial docs to `Label`.


src/elements/Label/Label.js
```diff
@@ -7,26 +7,25 @@ import {
...
 
 /**
  * A label displays content classification
  */
 function Label(props) {
   const {
-    attached, basic, children, color, corner, className, circular, detail, detailAs, empty, floating, horizontal,
-    icon, image, onClick, onDetailClick, onRemove, pointing, removable, ribbon, size, tag, content,
+    above, attached, basic, below, children, circular, className, color, content, corner, detail, empty, floating,
+    horizontal, icon, image, onClick, onRemove, pointing, removable, ribbon, size, tag,
```
layershifter: Resorted props

src/elements/Label/Label.js
```diff
@@ -35,31 +34,34 @@ function Label(props) {
...
     useKeyOnly(horizontal, 'horizontal'),
+    useKeyOnly(image, 'image'),
     useKeyOnly(empty, 'empty'),
     useKeyOnly(tag, 'tag'),
     useValueAndKey(attached, 'attached'),
     useKeyOrValueAndKey(corner, 'corner'),
     useKeyOrValueAndKey(pointing, 'pointing'),
+    /* Don't change order of `above` and `below`, they must be after `pointing` */
+    useKeyOnly(above, 'above'),
+    useKeyOnly(below, 'below'),
```
layershifter: Add missing props

src/elements/Label/Label.js
```diff
@@ -35,31 +34,34 @@ function Label(props) {
...
     useKeyOrValueAndKey(ribbon, 'ribbon'),
```
layershifter, line 47: Removed `image` shorthand because image is [prop](https://github.com/Semantic-Org/UI-Label/blob/master/label.css#L56)

src/elements/Label/Label.js
```diff
@@ -35,31 +34,34 @@ function Label(props) {
...
   )
 
-  const DetailComponent = detailAs || 'div'
   const ElementType = getElementType(Label, props)
   const rest = getUnhandledProps(Label, props)
 
+  if (children) {
+    return <ElementType {...rest} className={classes} onClick={handleClick}>{children}</ElementType>
+  }
+
```
layershifter, line 58: More correct render

src/elements/Label/Label.js
```diff
@@ -35,31 +34,34 @@ function Label(props) {
...
   return (
     <ElementType className={classes} onClick={handleClick} {...rest}>
       {createIcon(icon)}
       {createImage(image)}
       {content}
       {children}
-      {detail && (
-        <DetailComponent className='detail' onClick={handleDetailClick}>{detail}</DetailComponent>
-      )}
+      {detail && <LabelDetail content={detail} />}
```
layershifter: `Detail` subcomponent

layershifter:
Tests are failed on CI, but correctly runs locally :thought_balloon:

src/elements/Label/Label.js
```diff
@@ -114,12 +131,6 @@ Label.propTypes = {
   /** Additional text with less emphasis. */
   detail: PropTypes.string,
 
-  /** An element type to render the 'detail' as (string or function). */
-  detailAs: PropTypes.oneOfType([
```
jeffcarbs, line 115: Given we're removing the `detailAs`, should `detail` be a shorthand prop? So you can pass a string would be treated as `content` or you can pass an object as props or the element itself.

src/elements/Label/Label.js
```diff
@@ -90,21 +92,36 @@ Label.propTypes = {
...
   className: PropTypes.string,
 
+  /** Make the label circular, or a dot when used with 'empty' prop. */
+  circular: PropTypes.bool,
+
   /** Color of the label. */
   color: PropTypes.oneOf(Label._meta.props.color),
 
+  /** Primary content of the label, same as children. */
+  content: PropTypes.oneOf([
```
jeffcarbs: Should be `oneOfType`

src/elements/Label/Label.js
levithomason: I think these should be `_meta.props.pointing` values.  Since they are not valid on their own:

``` jsx
pointing
pointing='left'
pointing='right'
pointing='above'
pointing='below'
```

Then, we'll just have to write handling/tests to ensure the order is correctly implemented.  We won't be able to use common tests or lib use\* methods in this case.

src/elements/Label/Label.js
levithomason: Thanks!

src/elements/Label/Label.js
levithomason: Yes please

src/elements/Label/Label.js
layershifter: I've imploved propTypes in latest commit

src/elements/Label/Label.js
layershifter: I have simplified it while [standartization](https://github.com/TechnologyAdvice/stardust/pull/452#issuecomment-243614062) is pending

src/elements/Label/Label.js
```diff
@@ -35,31 +34,32 @@ function Label(props) {
...
     useKeyOnly(tag, 'tag'),
     useValueAndKey(attached, 'attached'),
     useKeyOrValueAndKey(corner, 'corner'),
     useKeyOrValueAndKey(pointing, 'pointing'),
+    /* Don't change order of `above` and `below`, they must be after `pointing` */
+    useKeyOnly(above, 'above'),
+    useKeyOnly(below, 'below'),
     useKeyOrValueAndKey(ribbon, 'ribbon'),
-    // TODO how to handle image child with no image class? there are two image style labels.
-    (image || childrenUtils.someByType(children, Image) || childrenUtils.someByType(children, 'img')) && 'image',
```
levithomason, line 45: I think we keep image shorthand and make the assumption that if you use it, you want the `image` class added.  This way, shorthand produces this type:

image

The user can define the markup without the image prop if they want this format:

image

### Compare to the Header

This is consistent with the assumptions we had to make in other components, like the icon Header.  In this case, we opted for the shorthand to _not_ generate the icon class (icon type Header):

image

image

levithomason (changes requested):

src/elements/Label/Label.js
layershifter: > I think these should be _meta.props.pointing values.

I think, too, but `useKeyOrValueAndKey(pointing, 'pointing')` will produce `above pointing` while it must be [`pointing above`](https://github.com/Semantic-Org/UI-Label/blob/master/label.css#L1157).

layershifter: So we need there custom handler?

src/elements/Label/Label.js
layershifter: So what we'll do there? I suggest leave as is, `detail` prop can be only `number` or `string`.

In #452 we can apply this usage:

``` jsx
<Label detail='1 day ago' />                           // string
<Label detail={{ as: 'a', content: '1 day ago' }} />   // props
<Label detail={<img src='1 day ago' />} />             // dom component
<Label detail={<Image src='1 day ago' />} />           // composite component
```

src/elements/Label/Label.js
layershifter: Some ugly code:

``` js
const prefixPointing = ['left', 'right']
const suffixPointing = ['above', `below`]

Label._meta = {
  props: {
    pointing: [
    ...prefixPointing,
    ...suffixPointing,
    ]
  }
}

const usePointingProp = (val, key) => prefixPointing.indexOf(val) ? `${val} ${key}` : `${key} ${val}`

function Label() {
  cx(
    usePointingProp(pointing, 'pointing')
  )
}
```

src/elements/Label/Label.js
layershifter: Reverted `image` remove, now it works like `Header` icon. It's enough?

src/elements/Label/Label.js
layershifter: Reverted `image` remove, now it works like `Header` icon. It's enough?

src/elements/Label/Label.js
```diff
@@ -159,8 +160,17 @@ Label.propTypes = {
...
-  /** A label can be formatted to emphasize an image. */
-  image: PropTypes.bool,
+  /** A label can be formatted to emphasize an image or prop can be used as shorthand for image. */
+  image: customPropTypes.every([
+    customPropTypes.givenProps(
+      { children: PropTypes.node.isRequired },
+      PropTypes.bool,
+    ),
+    customPropTypes.givenProps(
+      { icon: PropTypes.oneOfType([PropTypes.string, PropTypes.element, PropTypes.object]) },
```
levithomason: Looks like `icon` should be `image` here.

test/specs/elements/Label/Label-test.js
```diff
@@ -52,6 +51,25 @@ describe('Label Component', () => {
...
+  describe('image', () => {
+    it('adds an image class when true', () => {
+      shallow(<Label image />)
+        .should.have.className('image')
+    })
+    it('does not add an Image when true', () => {
+      shallow(<Label image />)
+        .should.not.have.descendants('Image')
+    })
+    it('adds an Image when given a name', () => {
```
levithomason: I believe the `common.implementsImageProp` should cover this and the next test.

levithomason (changes requested):

src/elements/Label/Label.js
levithomason: I think since this is a special case that is not reused, we should not create utilities for it.  We should just handle it directly in the component:

``` jsx
Label._meta = {
  props: {
    pointing: ['left', 'right', 'above', 'below']
  }
}

function Label() {
  const pointingClass = pointing === true && 'pointing'
    || (pointing === 'left' || pointing === 'right') && `${pointing} pointing`
    || (pointing === 'above' || pointing === 'below') && `pointing ${pointing}`

  cx(
    pointingClass
  )
}
```

**EDIT**

Removed lodash for vanilla JS

src/elements/Label/Label.js
layershifter: Agree, fixed.

test/specs/elements/Label/Label-test.js
layershifter: Yep 👍

src/elements/Label/Label.js
levithomason: Initially, I was going to suggest we make it a shorthand prop as well but the current method follows what was done for the [Dropdown.Header shorthand](https://github.com/TechnologyAdvice/stardust/pull/501/files).  Let's leave it as is and we'll update it later.

I'll try to focus on creating a generic shorthand factory right now. If it is memoized, we might be able to use it all over without creating multiple individual factories in `factories/index.js`.

src/elements/Label/Label.js
```diff
@@ -7,59 +7,61 @@ import {
...
     <ElementType className={classes} onClick={handleClick} {...rest}>
       {createIcon(icon)}
-      {createImage(image)}
+      {typeof image !== 'boolean' && createImage(image)}
       {content}
-      {children}
-      {detail && (
-        <DetailComponent className='detail' onClick={handleDetailClick}>{detail}</DetailComponent>
-      )}
+      {detail && <LabelDetail content={detail} />}
```
levithomason: OK, I've finished the shorthand factory update, see #529.  I also updated the Dropdown.Header shorthand I referenced above to use the new factory.

After rebasing, you can successfully use the new `createShorthand` factory.  It is meant for exactly this case, creating any arbitrary component from a shorthand value:

``` jsx
{createShorthand(LabelDetail, val => ({ content: val }), detail)}
```

You specify the component, a mapValueToProps function, and then the shorthand value.

levithomason (changes requested):

layershifter:
Updated to `master`, after #452 tests and propTypes for shorthand might be updated, but now we can merge if there are no remarks.

test/specs/elements/Label/Label-test.js
```diff
@@ -24,176 +25,106 @@ describe('Label Component', () => {
...
-        .childAt(0)
-        .should.match('i')
+      shallow(<Label content={text} />)
+        .should.contain.text(text)
     })
   })
 
-  describe('image', () => {
-    it('has no img when prop is not defined', () => {
+  describe('detail', () => {
```
levithomason: Let's replace this with `common.implementsShorthandProp()` for the`LabelDetail`.  Reference other tests for examples, or see #530.

levithomason (changes requested):
One more test update, then we're good to merge 👍

test/specs/elements/Label/Label-test.js
layershifter: Made :+1:

levithomason (approved):

levithomason:
Released in `stardust@0.46.0`, thanks!

Issue #407, opened 2016-08-19
RFC: Form Validation


levithomason:
This is a placeholder issue. Form validation will be implemented apart from the Form update PR #400.  We've had a few offline discussions surrounding the validation API.  Ideas will be posted here soon as we're able.

Have any input regarding form validation? We'd love to hear your ideas.

layershifter:
Don't forget about i18n there.

levithomason:
Could you expound a little?

layershifter:
SUI form validation has [default messages](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/form.js#L1108) for shorthand rules.

![image](https://cloud.githubusercontent.com/assets/14183168/17846914/ecb2dc4e-6853-11e6-81eb-e73d89413e00.png)

There is must be ability for redefine these messages :hammer:

levithomason:
Gotcha, agreed.

chuan137:
Hi,

I am trying to write validation for form fields myself, and find this threads. It would be nice to have a simplified API outside Form as indicated on the top of this thread. But I think it would also be nice to provide some straight forward way/api  to validate input fields manually. Maybe, like

```
<Button.Field label='email' name='email' validate={this.validateEmail} />
```

The validate method is called when the form is submit. When failed, the error message may be passed to its own state?

levithomason:
This was also my first inclination.  Having the ability to validate a single field seems useful.  I can see an equally valid argument that says, still validate the entire form model but only against a single field (e.g. do not specify validation for the other field names).  What I'm less sure of is what the API looks like for wiring the model, field values, and error messages into the UI.

@davezuko did you begin work on validation the other day?

chuan137:
>  What I'm less sure of is what the API looks like for wiring the model, field values, and error messages into the UI.

generally, serializer may pick up these information from states of sub-components? just my 2ct.

levithomason:
Yep, generally it is clear.  Specifics on the best way to do this is what I'm not clear about.  Would love to see some experiments on this though.  Goals being:
- loose coupling of model - form - validator - error messages
- no DOM dependency
- minimal boilerplate, but not at the expense of explicit, declarative, and decoupled code

jeffcarbs:
@levithomason - Not sure if this is beyond the scope of this issue, if so lmk and I can possibly open a new issue.

[...]

PR #527, opened 2016-09-22
Add example of advanced form

jeffcarbs:
Based on the comment I left of #407 regarding a more advanced form component, I put together an example. Right now the whole thing is in the example file but ideally we'd be able to move some/all of this logic into the `Form` itself.

The goals here are:
- being able to describe a form's structure as data and have it be rendered automatically
- have the form track it's own state outside of the DOM
- automatically handle advanced functionality like adding/removing items

The `schema` is basically just an array of form subcomponent props, which can be continually nested.

The following schema:

``` js
[
  {
    type: 'group',
    schema: [
      { type: 'text', name: 'firstName' },
      { type: 'text', name: 'lastName' }
    ]
  },
  { type: 'text', name: 'title' },
  { type: 'text', name: 'company' },
  {
    type: 'nested',
    name: 'emailAddresses',
    schema: [
      { type: 'text', name: 'label', label: null },
      { type: 'text', name: 'value', label: null }
    ]
  },
  {
    type: 'dropdown', name: 'gender', options: [
      { text: 'Male', value: 'male' },
      { text: 'Female', value: 'female' }
    ]
  },
  {
    type: 'checkbox', name: 'newsletter', label: 'Interested in receiving newletter'
  },
  { type: 'submit', name: 'Submit' }
]
```

produces:
screen shot 2016-09-22 at 2 09 20 pm

The component can almost certainly be cleaned up a little, I just whipped this together this morning. It also doesn't have provide a way for an outside component listen for onChange, but that would be pretty straightforward. I'd imagine in the `setValue` call we'd just pass those same arguments to a `onChange`.

### DIFF
--- a/docs/app/Examples/collections/Form/Types/FormAdvancedExample.js
+++ b/docs/app/Examples/collections/Form/Types/FormAdvancedExample.js
@@ -1,0 +1,243 @@
+import _ from 'lodash'
+import React from 'react'
+import { Form, Checkbox, Dropdown, Input, Button, Message } from 'stardust'
+
+const typeToControl = {
+  checkbox: Checkbox,
+  dropdown: Dropdown,
+  text: Input,
+  submit: Button
+}
+
+const labelFromNode = (node) =>
+  node.label || _.isUndefined(node.label) && _.startCase(node.name)
+
+const createKeyPath = (keys) => keys.join('.')
+
+class AdvancedForm extends React.Component {
+  static propTypes = {
+    formSchema: React.PropTypes.array,
+    onSubmit: React.PropTypes.func,
+    defaultValues: React.PropTypes.object
+  }
+
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      values: _.merge({}, props.defaultValues)
+    }
+
+    this.handleCheckboxChange = this.handleCheckboxChange.bind(this)
+    this.handleDropdownChange = this.handleDropdownChange.bind(this)
+    this.handleInputChange = this.handleInputChange.bind(this)
+    this.handleNestedAdd = this.handleNestedAdd.bind(this)
+    this.handleNestedRemove = this.handleNestedRemove.bind(this)
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  componentWillReceiveProps (props) {
+    this.setState({
+      values: _.merge({}, props.defaultValues)
+    })
+  }
+
+  setValue (keyPath, value) {
+    const { values } = this.state
+
+    _.set(values, keyPath, value)
+
+    this.setState({ values })
+  }
+
+  handleCheckboxChange (e, result) {
+    const { name: keyPath, checked } = result
+
+    this.setValue(keyPath, checked)
+  }
+
+  // TODO: Be able to infer name from result
+  handleDropdownChange (keyPath, e, value) {
+    this.setValue(keyPath, value)
+  }
+
+  handleInputChange (e) {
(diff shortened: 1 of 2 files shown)


levithomason:
Cool, will pull this and offer some feedback.

docs/app/Examples/collections/Form/Types/FormAdvancedExample.js
```diff
@@ -0,0 +1,243 @@
...
+        {this.renderFormSchema(formSchema)}
+        <Message>
+          <pre>state: {JSON.stringify(this.state, null, 2)}</pre>
+        </Message>
+      </Form>
+    )
+  }
+}
+
+const formSchema = [
```
levithomason, line 209: I like this, let's iterate on it and see where it goes!  Here's my first proposal:
- each object should be spread on the field
- explicitly define the field `control` prop, instead of reusing `type`
- use an array of `fields` where nested arrays would be groups
- default the `control` to `Input` and `type` to `text`
- capitalized `control` values will use the Stardust control, lower case for HTML controls (e.g. `input` vs `Input`)

This would make it more inline with our other shorthand.  It also gives the user more control, we could allow them to pass an actual `control` component and it's props, just like the Field does.  Finally, we're just spreading prop objects, a good thing.

I've left out the email `nested` part for now.  A refactor might look like this:

``` jsx
const formSchema = {
  fields: [
    [
      { name: 'firstName' },
      { name: 'lastName' },
    ],
    { name: 'title' },
    { name: 'company' },
    {
      control: 'Dropdown',
      name: 'gender',
      selection: true,
      options: [
        { text: 'Male', value: 'male' },
        { text: 'Female', value: 'female' },
      ],
    },
    { control: 'Checkbox', name: 'newsletter', label: 'Interested in receiving newletter' },
    { control: 'Button', type: 'submit', name: 'Submit' },
  ],
}
```

I could see an argument for allowing a simple string for the `name`.  Not sure how feel about adding more special case magic though:

``` jsx
const formSchema = {
  fields: [
    ['firstName', 'lastName'],
    'title',
    'company',
    // ...
  ],
}
```

levithomason (changes requested):

UnbrandedTech:
In the forms my company use, there are a lot of "duplicators" where an entry has an array of objects
```json
{
    "Supplies":[
	    {
	    	"Name":"Thing1",
	    	...
	    },
	    {
	    	"Name":"Thing2",
	    	...
	    },
    ]
}
```

Which looks like this when rendered
```JSX
<Grid>
	{fields.map((field, index) =>
	<Grid.Row key={index}>
		<Grid.Column>
			<Input name={`Supplies[${index}].Name`}>
		</Grid.Column>
		...
		<Grid.Column verticalAlign='middle'>
			<Button type="Button" icon='remove' onClick={() => fields.remove(index)}/>
		</Grid.Column>
	</Grid.Row>
	)}
	<Grid.Row>
		<Grid.Column>
			<Button type="button" content={`Add a Supply`} onClick={() => fields.push()} icon='plus' labelPosition='left' ></Button>
		</Grid.Column>
	</Grid.Row>
</Grid>
```
We use redux-form which handles this, but we'd like to use semantic's version what ever you decide to  implement.

levithomason:
Cool, good info.  Also @UnbrandedTech Grids/Rows/Columns can almost be replaced directly with Forms/Groups/Fields 👍 The layout APIs are somewhat similar.

levithomason:
Closing for housekeeping.

PR #528, opened 2016-09-22
feat(customPropTypes): add "as" prop type

levithomason:
This is a break out of #452.  It adds the `as` customPropType and updates all components to use it.

### DIFF
--- a/docs/app/Examples/elements/Icon/IconSet/IconCategoryExample.js
+++ b/docs/app/Examples/elements/Icon/IconSet/IconCategoryExample.js
@@ -16,12 +16,6 @@ const IconCategoryExample = ({ category }) => (
 )
 
 IconCategoryExample.propTypes = {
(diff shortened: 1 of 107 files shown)


docs/app/Examples/elements/Icon/IconSet/IconCategoryExample.js
```diff
@@ -16,12 +16,6 @@ const IconCategoryExample = ({ category }) => (
 )
 
 IconCategoryExample.propTypes = {
-  /** An element type to render as (string or function). */
-  as: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.func,
-  ]),
-
```
levithomason, line 24: The `as` prop was added to all components via regex.  Some received the as prop where it didn't make sense, like in doc components.  This PR also removes those.

PR #529, opened 2016-09-23
feat(factories): add shorthand factory

levithomason:
/cc @jcarbo @layershifter 

This PR refactors factories and updates all shorthand usages to use them.  We still have predefined factories as before:

``` js
createIcon()
createImage()
createHTMLImage()  // renamed from createImg
createLabel()
```

Our new `createShorthand` factory does the same thing as the previous `Factory` in `createFactory` did before.  It has just been refactored to allow standalone use.

``` jsx
createShorthand(Component, mapValueToProps, val, extraProps = {}) {...}
```

Our predefined factories are created by partially applying the new `createShorthand` factory:

``` js
function createShorthandFactory(Component, mapValueToProps) {
  return _.partial(createShorthand, Component, mapValueToProps)
}

const createIcon = createShorthandFactory(Icon, value => ({ name: value }))
// ...
```

This refactored factory makes it usable for previously non-reusable shorthand.  Specifically, where the component was not something that would be used elsewhere:

This...

``` jsx
{header && <DropdownHeader content={header} />}
```

Becomes...

``` jsx
{createShorthand(DropdownHeader, val => ({ content: val }), header)}
```

The Dropdown `header` shorthand now receives all the benefits of regular shorthand.  It can be created from a string, props object, or an element.

This shorthand factory can easily be extended later to handle arrays.  Then, it will also generate a key for each child.  If a key could not be generated, it will require the user to provide a `childKey` prop.  That is for another PR.


layershifter:
Looks awesome :+1:

jeffcarbs:
This is great! 😍

PR #530, opened 2016-09-23
feat(commonTests): add implementsShorthandProp

levithomason:
/cc @jcarbo @layershifter 

This PR adds a common test for shorthand props.  As with the factories, there is a base common test.  It takes all the same arguments as the factory.  The specific shorthand common tests simply call the base test with defaults.  Note, instead passing many arguments a config object is used to define the test params.
## Custom Shorthand Tests

When a component implements the `createShorthand` factory for a shorthand prop, you can define a test for that:

``` jsx
common.implementsShorthandProp(Header, {
  propKey: 'subheader',
  ShorthandComponent: HeaderSubheader,
  mapValueToProps: val => ({ content: val }),
})
```
## Pre-Defined Shorthand Tests

Basic Icon/Image/Label tests remain the same:

``` jsx
common.implementsIconProp(Header)
common.implementsImageProp(Header)
```

Since these are just implementations of `implementsShorthandProp` with default values set, you can customize them as well.  When a component passes additional params to the `create(Icon|Image|Label)` factory, you can configure the test to match:

``` jsx
common.implementsIconProp(BreadcrumbDivider, {
  requiredShorthandProps: {
    className: 'divider',
  },
})

common.implementsLabelProp(Button, {
  requiredShorthandProps: {
    basic: true,
    pointing: 'left',
  },
})
```

### DIFF
--- a/src/collections/Message/Message.js
+++ b/src/collections/Message/Message.js
@@ -77,9 +77,9 @@ function Message(props) {
         {createIcon(icon)}
         {(header || content || list) && (
           <MessageContent>
-            {header && <MessageHeader>{header}</MessageHeader>}
+            {createShorthand(MessageHeader, val => ({ children: val }), header)}
             {createShorthand(MessageList, val => ({ items: val }), list)}
-            {content && <p>{content}</p>}
+            {createShorthand('p', val => ({ children: val }), content)}
           </MessageContent>
         )}
       </ElementType>
--- a/src/elements/Step/StepContent.js
+++ b/src/elements/Step/StepContent.js
@@ -61,7 +61,7 @@ StepContent.propTypes = {
       PropTypes.string,
       PropTypes.number,
       PropTypes.object,
-    ])
+    ]),
   ]),
 }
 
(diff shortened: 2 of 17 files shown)


src/elements/Step/StepContent.js
```diff
@@ -57,7 +57,11 @@ StepContent.propTypes = {
   /** Primary content of the StepTitle. Mutually exclusive with children. */
   title: customPropTypes.every([
     customPropTypes.disallow(['children']),
-    PropTypes.node,
+    PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+      PropTypes.object,
+    ])
```
layershifter: Will be there something like `customPropTypes.shorthand()`? It looks it might be

``` jsx
PropTypes.oneOfType([
  PropTypes.element,
  PropTypes.number,
  PropTypes.object,
  PropTypes.string,
])
```

levithomason: Yep, I'm just making separate PRs from #452 since it got so large.  Also, the customPropType for shorthand needs more thought, per the the discussion in #452.  Not sure what the final PR will look like for that, but taking it one step at a time.

src/collections/Message/Message.js
```diff
@@ -77,9 +77,9 @@ function Message(props) {
         {createIcon(icon)}
         {(header || content || list) && (
           <MessageContent>
-            {header && <MessageHeader>{header}</MessageHeader>}
+            {createShorthand(MessageHeader, val => ({ children: val }), header)}
```
levithomason, line 80: Updated some previously missed shorthand prop renders.

PR #535, opened 2016-09-23
chore(commonTests): Standardize the assertions in common tests

jeffcarbs:
Saw some duplication in these specs doing the same thing slightly differently. Lmk if this makes sense. 

I left the assertions where they were, but any reason they're inside the "describe" block vs right inside the function definition. They're really not relevant to the tests themselves, more to the function that's generating the tests, so I'd imagine they should right at the top.


levithomason:
Cool, I almost crammed this into my other PR for shorthand tests but opted to postpone.  Looks good thanks!

PR #536, opened 2016-09-23
feat(DropdownMenu): add `active` prop

### LINKED ISSUE #533: Nested dropdown/menu does not expand on hover
## Steps to Reproduce

Check out the "Attached" example for dropdown in documentation
## Expected Behavior

"New" menu should expand and show additional options on hover
## Actual Behavior

Nothing opens
## Version

Stardust: master
Browser: Chrome
OS: Mac
## Test Case

Documentation

parisholley:
fixes #533

### DIFF
--- a/src/modules/Dropdown/DropdownItem.js
+++ b/src/modules/Dropdown/DropdownItem.js
@@ -75,6 +75,8 @@ export default class DropdownItem extends Component {
 
         return cloneElement(child, { opened: this.state.hovered })
       })
+    }else if(newChildren && newChildren.type === DropdownMenu){
+      newChildren = cloneElement(newChildren, { opened: this.state.hovered });
     }
 
     return (
--- a/src/modules/Dropdown/DropdownItem.js
+++ b/src/modules/Dropdown/DropdownItem.js
@@ -30,6 +30,8 @@ export default class DropdownItem extends Component {
       description,
       icon,
       onClick,
+      onMouseEnter,
+      onMouseLeave,
       selected,
       text,
       value,
@@ -39,11 +41,15 @@ export default class DropdownItem extends Component {
       if (onClick) onClick(e, value)
     }
 
-    const handleMouseEnter = () => {
+    const handleMouseEnter = (e) => {
+      if (onMouseEnter) onMouseEnter(e, value)
+
       this.setState({ hovered: true })
     }
 
-    const handleMouseLeave = () => {
+    const handleMouseLeave = (e) => {
+      if (onMouseLeave) onMouseLeave(e, value)
+
       this.setState({ hovered: false })
     }
 
@@ -75,7 +81,7 @@ export default class DropdownItem extends Component {
       <ElementType {...rest}
         className={classes}
         onClick={handleClick}
-        onMouseOver={handleMouseEnter}
+        onMouseEnter={handleMouseEnter}
         onMouseLeave={handleMouseLeave}
       >
         {createShorthand('span', val => ({ className: 'description', children: val }), description)}
--- a/src/modules/Dropdown/DropdownItem.js
+++ b/src/modules/Dropdown/DropdownItem.js
@@ -145,6 +145,12 @@ DropdownItem.propTypes = {
     PropTypes.string,
   ]),
 
-  /** Called on click with (event, value, text). */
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/DropdownItem.js
```diff
@@ -11,44 +12,85 @@ import {
...
+
+    let newChildren = children
+
+    if (Array.isArray(children)) {
+      newChildren = Children.map(children, (child) => {
+        if (child.type !== DropdownMenu) {
+          return child
+        }
+
+        return cloneElement(child, { opened: this.state.hovered })
```
levithomason, line 76: We're trying to get away from looping children.  Why? If a user passes their own menu component, this will fail since it will not be a `DropdownMenu`.  The same is true if they decorate the Menu, wrapping it in a HOC.  Also if it is augmented with the `as` prop.

Because of this, I don't want to add any features that deal with parsing children since they almost always cause bugs.

Let's remove this, which means we can remove the state `hovered`, and also allows us to keep this as a functional component.  The parent / implementing component can manage the state.  The mouse props will be spread in `rest` as well.  Assuming I haven't overlooked something, I don't believe there is a need to extend the MenuItem to accomplish this feature.

src/modules/Dropdown/DropdownMenu.js
```diff
@@ -32,6 +35,7 @@ DropdownMenu.propTypes = {
 
   /** Classes to add to the className. */
   className: PropTypes.string,
-}
 
-export default DropdownMenu
+  /** Will mark the menu as opened when nested in another menu */
+  opened: PropTypes.bool,
```
levithomason, line 40: We do need this, thanks.  Let's call the prop `active` though so it is inline with SUI core:

http://semantic-ui.com/modules/dropdown.html#active

src/modules/Dropdown/DropdownMenu.js
```diff
@@ -8,13 +8,16 @@ import {
...
 
-function DropdownMenu(props) {
-  const { children, className } = props
-  const classes = cx('menu transition', className)
-  const rest = getUnhandledProps(DropdownMenu, props)
-  const ElementType = getElementType(DropdownMenu, props)
+export default class DropdownMenu extends Component {
+  render() {
+    const { children, className } = this.props
+    const classes = cx('menu transition', className, { visible: this.props.opened })
```
levithomason, line 14: Please update this className buildup to be consistent with other components:
- destructure props
- utilize `useKeyOnly` for the `visible` class
- put the user's `className` last in the list

src/modules/Dropdown/DropdownMenu.js
```diff
@@ -8,13 +8,16 @@ import {
...
+    const rest = getUnhandledProps(DropdownMenu, this.props)
+    const ElementType = getElementType(DropdownMenu, this.props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+    return (<ElementType {...rest} className={classes}>
+      {children}</ElementType>)
```
levithomason, line 19: Not sure why the linter didn't catch this.  Please return all on one line without parens, or return like so:

``` jsx
return (
  // <ElementType...
)
```

levithomason (changes requested):
High level, I think we should fix the nested Dropdown issue by adding the [`simple`](http://semantic-ui.com/modules/dropdown.html#simple) prop to the Dropdown in the docs.  This wold also solve the hover issue.  This allows CSS to handle the visible state:

simple dropdown

Other than that, I've left inline comments.

src/modules/Dropdown/DropdownMenu.js
levithomason: Also, our props descriptions follow the SUI core descriptions when possible:

image

levithomason: Once this is updated, let's update the actual Dropdown usage of the DropdownMenu to use the prop instead of passing the `className`.

https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/Dropdown.js#L872

levithomason:
Note, I've shipped a fix for #533 via #541.  I think this PR should be about the DropdownMenu portion only.

parisholley:
going to close as simple solves my problem

PR #537, opened 2016-09-23
docs(Label): fix pointing docs

### LINKED ISSUE #534: fix(Label): pointing examples are incorrect
## Steps to Reproduce

Go the doc site and review the Label pointing examples.
## Expected Behavior

All the labels should be pointing in the correct direction.
## Actual Behavior

Two are pointing in the incorrect direction.
## Test Case

http://technologyadvice.github.io/stardust/elements/label#pointing

image

layershifter:
This PR fixes #534.


layershifter:
@levithomason I will work on `Label` docs after `Table`, it will be interesting :smile:

PR #539, opened 2016-09-23
fix(factories): explicitly handle each value type

levithomason:
Factories were returning empty components give boolean values.  This PR makes each return type explicitly list what it _includes_ rather than _excludes_.  The final else condition is null.


jeffcarbs:
Ran locally and fixes the dropdown icon issue. Lgtm 👍

levithomason:
Released in `stardust@0.46.1`

PR #540, opened 2016-09-23
docs(Menu): MenuItem icon and dropdown item example

levithomason:
This PR changes augmentation order so the Dropdown is augmented with a Menu.Item.  I also added the missing MenuItem icon shorthand.  The prop was there but not actually rendered.
# Before

image
# After

image


levithomason:
Released in `stardust@0.46.1`

PR #543, opened 2016-09-24
feat(common): Improve shorthand array handling

jeffcarbs:
_**Note:** Abstracted from #430_

---

This updates the createShorthand factory to take two special props as extraProps to generate a key if missing. This is necessary for shorthand props that are arrays. The props are:
- childKey - enables explicitly setting the key via props
- getChildKey - function that takes the given props and returns a key

This PR also fixes some issues I noticed using the menu `items` shorthand prop:
- Use the createShorthand factory to create a MenuItem for each item.
- Do not infer activeIndex from the items array on mount.

Regarding the second point, there were a few issues with it:
- The array items may not be objects - they could be primitives or elements which are valid shorthand.
- You should be able to reliably set the activeIndex or defaultActiveIndex via props, but this would overwrite that on mount.

---

I was also thinking about just using `childKey` that could either be a function or a string, so:

``` js
const mergeChildKey = ({ childKey, ...props }) => {
  if (!props.key) {
    props.key = _.isFunction(childKey) ? childKey(props) : childKey
  }

  return props
}
```

But then we'd have to make sure that if the `childKey` were passed in props, it's not overwritten by `extraProps`. Might be worth it to remove the need for that additional key.

### DIFF
--- a/src/factories.js
+++ b/src/factories.js
@@ -9,15 +9,19 @@ import Label from './elements/Label/Label'
 /**
  * Merges props and classNames.
  *
+ * @param {object} defaultProps A props object
  * @param {object} props A props object
- * @param {object} extraProps A props object
  * @returns {object} A new props object
  */
-const mergePropsAndClassName = (props, extraProps) => {
-  const newProps = { ...props, ...extraProps }
+const mergePropsAndClassName = (defaultProps, props) => {
+  const { childKey, ...newProps } = { ...defaultProps, ...props }
 
-  if (_.has(props, 'className') || _.has(extraProps.className)) {
-    newProps.className = cx(props.className, extraProps.className) // eslint-disable-line react/prop-types
+  if (_.has(props, 'className') || _.has(defaultProps.className)) {
+    newProps.className = cx(defaultProps.className, props.className) // eslint-disable-line react/prop-types
+  }
+
+  if (!newProps.key) {
+    newProps.key = _.isFunction(childKey) ? childKey(newProps) : childKey
   }
 
   return newProps
@@ -30,23 +34,23 @@ const mergePropsAndClassName = (props, extraProps) => {
  * @param {function|string} Component A ReactClass or string
  * @param {function} mapValueToProps A function that maps a primitive value to the Component props
  * @param {string|object|function} val The value to create a ReactElement from
- * @param {object} extraProps Additional props to add to the final ReactElement
+ * @param {object} defaultProps Default props to add to the final ReactElement
  * @returns {function|null}
  */
-export function createShorthand(Component, mapValueToProps, val, extraProps = {}) {
+export function createShorthand(Component, mapValueToProps, val, defaultProps = {}) {
   // Clone ReactElements
   if (isValidElement(val)) {
-    return React.cloneElement(val, mergePropsAndClassName(val.props, extraProps))
+    return React.cloneElement(val, mergePropsAndClassName(defaultProps, val.props))
   }
 
   // Create ReactElements from props objects
   if (_.isPlainObject(val)) {
-    return <Component {...mergePropsAndClassName(val, extraProps)} />
+    return <Component {...mergePropsAndClassName(defaultProps, val)} />
   }
 
   // Map values to props and create a ReactElement
   if (_.isString(val) || _.isNumber(val)) {
-    return <Component {...mergePropsAndClassName(mapValueToProps(val), extraProps)} />
+    return <Component {...mergePropsAndClassName(defaultProps, mapValueToProps(val))} />
   }
 
   // Otherwise null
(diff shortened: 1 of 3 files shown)


jeffcarbs:
Could also simplify to:

``` js
const mergePropsAndClassName = (props, extraProps) => {
  const { childKey, ...newProps } = { ...props, ...extraProps }

  if (_.has(props, 'className') || _.has(extraProps.className)) {
    newProps.className = cx(props.className, extraProps.className) // eslint-disable-line react/prop-types
  }

  if (!props.key) {
    props.key = _.isFunction(childKey) ? childKey(props) : childKey
  }

  return mergeChildKey(newProps)
}
```

This also makes me wonder, currently anything in `extraProps` will "win" out over `props` in the merge. Should this be the case? I know generally `extraProps` are things necessary for SUI functionality (like `active`), but I'd imagine we'd want to give the user the ultimate control. Thoughts?

levithomason:
> I was also thinking about just using childKey that could either be a function or a string, so:

Love the function idea.  Unrelated to this PR, I was thinking the `extraProps` should have been a function as well.  It is sometimes necessary to build up extra props based on the original factory value/props/returned element.

> This PR also fixes some issues I noticed using the menu items shorthand prop:

👍 

Regarding comments on `extraProps` overriding the `props`, I think the override is mixed up actually (changed my response as I wrote it!).  There are 3 levels of props at play:
1. `mapValueToProps`
   
   The base props object that the factory defines from a shorthand value.
2. `extraProps`
   
   Passed to the factory when called `createIcon('user', { color: 'red' })`.  I think of these sort of like default props.  They ~~are only~~ _should only be_ applicable if the user does not provide the same props.
3. User's props
   
   This is where it is mixed up.  The `extraProps` should only override when the factory value is a literal.  That makes sense because the user's props should always win, IMO.  So, if the user passes a props object or an element then we want to use their props object/element props as the `extraProps`.  This way, they can override the factory's "base props" and "extra props" defined in the factory call.

Simply flipping the args in those calls to `mergePropsAndClassName()` will ensure the user's props (if provided) always win.  It may then also make sense to rename them from `props` `extraProps` to `defaultProps` `props` so it is more clear why one is winning over the other.

jeffcarbs:
@levithomason - Updated based on the comments here.
- Just going to use `childKey` which could be either a literal or a function.
- Got rid of the separate method and just handle `childKey` as part of the `mergePropsAndClassName` function.
- Swapped and renamed `extraProps` to `defaultProps`.

I think should be g2g 👍

src/factories.js
```diff
@@ -30,23 +34,23 @@ const mergePropsAndClassName = (props, extraProps) => {
...
   // Create ReactElements from props objects
   if (_.isPlainObject(val)) {
-    return <Component {...mergePropsAndClassName(val, extraProps)} />
+    return <Component {...mergePropsAndClassName(defaultProps, val)} />
   }
 
   // Map values to props and create a ReactElement
   if (_.isString(val) || _.isNumber(val)) {
-    return <Component {...mergePropsAndClassName(mapValueToProps(val), extraProps)} />
+    return <Component {...mergePropsAndClassName(defaultProps, mapValueToProps(val))} />
```
levithomason, line 53: In this case, I believe the factory's mapValueToProps is the default.  The other props here are what ever the developer passed when calling `createFoo`.  Otherwise, you wouldn't be able to override the default mapping when using the component.

Feedback?

levithomason (changes requested):
Just one comment regarding defaultProps ordering

src/factories.js
jeffcarbs: The `defaultProps` will always be what we (the stardust authors) set as the default. The props, whether passed directly or derived from `mapValueToProps`, will be what the end-user wants. So those should always win.

Take this example component that has an image with a default src:

``` js
// Example:
const Foo = (props) => (
  <div>
    createShorthand(Image, val => ({ src: val }), props.image, { src: 'foo.png' })
  </div>
)
```

The different usages are:

React Element:

``` js
<Foo image={<MyImage src='bar.png'> />

// Within createShorthand:
// Clone ReactElements
if (isValidElement(val)) {
  return React.cloneElement(val, mergePropsAndClassName(defaultProps, val.props))
}
```

Props object

``` js
<Foo image={{ src: 'bar.png' }} />

// Within createShorthand:
// Create ReactElements from props objects
if (_.isPlainObject(val)) {
  return <Component {...mergePropsAndClassName(defaultProps, val)} />
}
```

Primitive:

``` js
<Foo image='bar.png' />

// Within createShorthand:
// Map values to props and create a ReactElement
if (_.isString(val) || _.isNumber(val)) {
  return <Component {...mergePropsAndClassName(defaultProps, mapValueToProps(val))} />
}
```

In all of these cases, if we swapped the argument order of `defaultProps` the src would not be what the user passed.

levithomason: Wrote out a detailed explanation of how I didn't actually think this would work for the string/number literal case only to come to see it correctly.  Thanks for the clarifications!

levithomason (approved):

levithomason:
Released in `stardust@0.46.3`, thanks!

Issue #378, opened 2016-08-06
Config: support i18n


tomitrescak:
Hi, I was wondering if you would consider having internationalisation Support out of the box? In my version of semanticui-react, i have approached it that i am passing translation function to the semanticui-react and then this function is called for any string. By default this function returns the original string. 

It makes the code very nicely readable. And less cluttered.

Since with this approach you know the list of elements and their possible internationalisable properties, I even wrote a simple extractor, that extracts all i18n strings to a JSON file for processing.

I would love to switch to your lib, maintaining a separate one makes no sense.

levithomason:
I think i18n makes good sense.  I have some questions and ideas but it would be helpful to see some code for this.  Could you point me to where this is used in your project for reference? Or, you could open a PR here and we can hash out the details there instead.

levithomason:
bump @tomitrescak

tomitrescak:
Sorry @levithomason I'm having quite busy days. I experimented with several approaches, including generating specific `i18n-capale` components. But the easiest and most functional solution was following:
1. Create a config point for your library, a simple exported object would do:
   
   ``` js
   export default {
     translate: (key) => key; // default translate function returns the key itself
   }
   ```
2. Each place, where component is rendering text is replaced by calling of this function. For example:
   
   ``` js
   import config from '../config';
   <Button text={config.i18n(text)} />
   ```

Yes, this requires modification in almost all components. Many times when component is rendering "children" I leave it as it is, so all i18n strings need to be specified via text props.

With this approach you leave complete freedom to choose i18n library on the user.
From the user perspective you only need to configure what to use to translate your strings (MessageFormat ...). Following is an example of a custom translate function.

``` js
import config from 'stardust';

let transStrings = { ... };

config.i18n = (key) => return transStrings(key); // you can be much more explicit here
```

In my translate function I allow overriding of translation by prefixing ! to display non translated value, I also allow translation parameters to support plurals and more. Or you just use message format and make sure you specify parameters to it as it needs.

Does this sound ok?

[EDIT] for the purposes of i18n I have added a `<Text text="value" />` element, which displays translated string.

levithomason:
I like the config idea.  I was hoping to see something like this.  We may have more configuration options later as well.  I would accept a PR that added a config object with an `i18n` key.

[...]

PR #544, opened 2016-09-24
feat(config): possibility to modify default link element

tomitrescak:
As per discussion in #378


levithomason:
Couple lint errors:

```
/home/ubuntu/stardust/src/lib/config.js
  3:19  error  Missing trailing comma                         comma-dangle
  4:2   error  Newline required at end of file but not found  eol-last
```

Did you plan on wiring this in anywhere, or just adding the config for now?  Thanks much!

levithomason:
@tomitrescak it looks like this PR includes typings also now.  We'll to merge only the config for this PR.  I suggest creating a branch off of your master that only has the config work on it so only that work will be merged.  Alternatively, you could keep this PR open and just temporarily remove the typings so only the config is being proposed.

tomitrescak:
I have removed typings from this PR and will create new one

jeffcarbs:
In it's current state, I'm not sure this is something we should add. For the use-case referenced in #378 (usage with `react-router`) this won't actually have any effect. The `Link` doesn't actually do anything unless there's a `to` prop (see: https://github.com/ReactTraining/react-router/blob/master/modules/Link.js#L116).

My issues:
- This is very a specific use (only elements where an `href` prop is passed) and it already doesn't work with one of the main React-based routing libraries. This is probably the biggest red flag.
- Wherever this would be working, you're already going to be passing a path as a prop (whether via `href` or `to` or whatever your library calls for). I don't think adding `as={Link}` adds much more complexity so I don't think this really saves a lot of work.

@levithomason - would love your thoughts here as well.

levithomason:
Although the original intent of the config was for an i18n function and I think we should allow that, I agree that users should use `as={Link}` instead of allowing an `anchorElement` config option.

Thanks @jcarbo for the clarification.  Apologies @tomitrescak as I had not fully realized the goal here.  If you'd like to use an element as a `Link`, you should do so like this:

``` jsx
<Menu>
  <Menu.Item as={Link} to='/home'>
    Home
  </Menu.Item>
</Menu>
```

See: http://react.semantic-ui.com/introduction#augmentation

PR #548, opened 2016-09-26
feat(commonTests): add options and expected className

### LINKED ISSUE #481: refactor(commonTests): rename for consistency and clarity
### Factory Props

Some components pass a prop value to a factory, giving it many supported values.  Other components might implement the same prop name but without a factory.  Our current `implements*Prop` tests are only applicable to props that are implemented with the corresponding factory.  These should be renamed:

`implementsIconProp` => `implementsIconFactoryProp`
`implementsLabelProp` => `implementsLabelFactoryProp`
`implementsImageProp` => `implementsImageFactoryProp`
### className Props

Every className builder (`src/lib/classNameBuilders.js`) has a corresponding common test with a similar name.  However, `useKeyOrValueAndKey` and `propKeyOrValueToClassName` are not consistent.

The tests doesn't assert key or value, it asserts key or value _and key_. Rename to `propKeyOrValueToClassName` => `propKeyOrValueAndKeyToClassName`.

levithomason:
Fixes #481 

Per [this thread](https://github.com/TechnologyAdvice/stardust/pull/451#discussion_r80383662), we need to sometimes specify the expected `className` in common prop-to-className tests.  This PR adds this feature.

Instead of adding more common test params, the final common test  param is now an options object.  This is consistent with the newest shorthand common tests.  It has served us well there.  It also allows us to add future specialized behavior to common tests without running out of param space and requiring large refactors.


.github/CONTRIBUTING.md
```diff
@@ -252,7 +252,7 @@ describe('Segment', () => {
   common.propValueOnlyToClassName(Segment, 'color')
   common.propKeyOnlyToClassName(Segment, 'basic')
   common.propKeyAndValueToClassName(Segment, 'floated')
-  common.propKeyOrValueToClassName(Segment, 'padded')
+  common.propKeyOrValueAndKeyToClassName(Segment, 'padded')
```
levithomason, line 255: Updated this test name to be consistent with `useKeyOrValueAndKey()`, fixed #481.

.github/CONTRIBUTING.md
```diff
@@ -407,18 +407,18 @@ describe('Menu', () => {
 })
 ```
 
-The last argument to a common test is always `requiredProps`.  If your component requires certain props to render, you can pass those last:
+The last argument to a common test is always `options`.  You can configure the test here.  Example, if your component requires certain props to render, you can pass in `requiredProps`:
```
levithomason, line 410: Updated the contributing guide.

test/specs/collections/Form/FormField-test.js
```diff
@@ -11,7 +11,9 @@ describe('FormField', () => {
   common.propKeyOnlyToClassName(FormField, 'error')
   common.propKeyOnlyToClassName(FormField, 'disabled')
   common.propKeyOnlyToClassName(FormField, 'inline')
-  common.propKeyOnlyToClassName(FormField, 'required', { label: '' })
+  common.propKeyOnlyToClassName(FormField, 'required', {
+    requiredProps: { label: '' },
+  })
```
levithomason, line 16: Updated all tests that were using the `requiredProps` last arg to now use the `options` object param.

test/specs/commonTests.js
```diff
@@ -455,77 +463,79 @@ const _definesPropOptions = (Component, propKey) => {
   })
 }
 
-const _noDefaultClassNameFromProp = (Component, propKey, requiredProps = {}) => {
+const _noDefaultClassNameFromProp = (Component, propKey, options = {}) => {
+  const { className = propKey, requiredProps = {} } = options
```
levithomason, line 468: The expected `className` always defaults to the `propKey`.

PR #551, opened 2016-09-26
fix(List): Update to v1 API

layershifter:
This PR updates [**List**](http://semantic-ui.com/elements/list.html) component to v1 API.


docs/app/Examples/elements/List/Types/ListDivided.js
```diff
@@ -0,0 +1,32 @@
...
+      </List.Item>
+      <List.Item>
+        <List.Icon name='github' size='large' verticalAlign='middle' />
+        <List.Content>
+          <List.Header as='a'>Semantic-Org/Semantic-UI-Docs</List.Header>
+          <List.Description as='a'>Updated 22 mins ago</List.Description>
+        </List.Content>
+      </List.Item>
+      <List.Item>
+        <List.Icon name='github' size='large' verticalAlign='middle' />
```
layershifter: We have there hateful situation, `Icon` doesn't have prop `verticalAlign` but `.list .icon` [has](https://github.com/Semantic-Org/UI-List/blob/master/list.css#L386). I don't like idea with `List.Icon`, may be something better?

levithomason: For clarity, we're talking about this:

image

Oh boy :/  The proper extrapolation of our API is to add `List.Icon`.  However, odds of users using this over `Icon` are slim.  Especially considering all other components use the `Icon` directly.

I think we add `verticalAlign` to the `Icon` itself.  I realize that prop would only be valid when used in a List, but it seems to make the most sense for users.

Any other wise ideas @jcarbo?

jeffcarbs: I think `List.Icon` might actually be the way to go. It would just be a wrapper around `Icon` that handles the `verticalAlign` prop:

``` js
function ListIcon(props) {
  const { verticalAlign, className } = props
  const classes = cx(
    useValueAndKey(verticalAlign, 'verticalAlign'),
    className
  )
  const rest = getUnhandledProps(ListIcon, props)
  return <Icon {...rest} className={classes} />
}
```

In all likelihood, if I were using a `List` I'd either be:
- using the `icon` prop as a shorthand, so it wouldn't matter whether it's `Icon` or `List.Icon` under the hood.
- looking at the example page for `List` to see how to structure it and would see the `List.Icon` subcomponent there.

I could also see just building it into the `Icon` class with a comment noting that it's a specific use-case, although I really wish there were at least one other use-case for that.

levithomason: Agreed.  I was on the fence as well, but after sleeping on it and hearing this I think we should stick with the API pattern we've chosen rather than make an exception.

@layershifter let's keep it consistent and go with `List.Icon`.  Thanks for the input guys.

src/elements/List/ListItem.js
```diff
@@ -1,54 +1,54 @@
...
+    useKeyOnly(active, 'active'),
+    className,
+    'item'
+  )
+
+  const ElementType = getElementType(ListItem, props)
+  const rest = getUnhandledProps(ListItem, props)
+  const valueProp = ElementType === 'li' ? { value } : { dataValue: value }
+
+  return <ElementType {...rest} {...valueProp} className={classes}>{children}</ElementType>
```
layershifter, line 32: Comrades, what you think about this (`valueProp`)? May be more elegant solution possible there?

I'm talking about this:

``` html
<li value="*">User Benefits</li>
```

> You can also manually specify a value for an ordered list using data-value on a div, or value on an li

levithomason: I don't see any React warnings when adding `value` to an `<li>`.  Any reason not to with this?

``` jsx
<List.Item value='*' />
//=> <li value="*"></li>
```

If there is an issue, one alternative could be to use `bullet`:

``` jsx
<List.Item bullet='#' />
//=> <li value="#"></li>
```

Though, I'd lean toward SUI parity if we can.

layershifter: I'm talking about this transform, it looks not very clean (`valueProp` spreading).

``` jsx
<List.Item value='*' />
//=> <div data-value="*"></div>

<List.Item as='li' value='*' />
//=> <li value="*"></li>
```

src/elements/List/ListList.js
```diff
@@ -0,0 +1,42 @@
...
+  useKeyOnly,
+} from '../../lib'
+
+function ListList(props) {
+  const { children, className } = props
+
+  const ElementType = getElementType(ListList, props)
+  const classes = cx(
+    className,
+    useKeyOnly(ElementType !== 'ul' && ElementType !== 'ol', 'list'),
```
layershifter, line 17: Another evil place with class handling, see SUI docs about `ul` &  `ol`.
Combo like `Menu.Menu` is there, too :-\

``` jsx
<List>
  <List.Item>
    Item
    <List.List>
      <List.Item>Sub item</List.Item>
    </List.List>
  </List.Item>
</List>
// ...
// <div class="ui list">
//   <div class="item">
//     <div class="list">...</div>
//   </div>
// </div>
// ...
```

``` jsx
<List as='ul'>
  <List.Item as='li'>
    Item
    <List.List as='ul'>
      <List.Item as='li'>Sub item</List.Item>
    </List.List>
  </List.Item>
</List>
// ...
// <ul class="ui list">
//   <li>
//     <ul>...</ul>
//   </li>
// </ul>
// ...
```

levithomason: Per the docs the "list" class may be omitted "as a convenience" when using an "ol"  or "ul" tag. Though, I don't think it is required to be omitted. We can likely always have the "list" class present. 

I suspect we can also have the "item" class on all items as well. Let's try always having the classes first. That way, the components are more simple.

layershifter: > Per the docs the "list" class may be omitted "as a convenience" when using an "ol" or "ul" tag.

Sad, but it must be omitted.

Without classes:
image
With classes:
image

layershifter:
Ahoy, docs are fully there :ship: 

Working on tests.

src/elements/List/ListList.js
levithomason: Oh boy, well then.  I think the logic you originally had there is just the way it has to be.

levithomason:
Awesome, pulling and trying them out.  Will let you know.

src/elements/List/List.js
```diff
@@ -1,40 +1,136 @@
...
-      'list'
-    )
-    const ElementType = getElementType(List, this.props)
-    return (
-      <ElementType {...this.props} className={classes}>
-        {this.props.children}
-      </ElementType>
-    )
-  }
+function List(props) {
```
levithomason, line 25: Let's add the component description from the SUI docs as well.

docs/app/Examples/elements/List/Types/index.js
```diff
@@ -0,0 +1,50 @@
...
+    <ComponentExample examplePath='elements/List/Types/ListExampleTree' />
+
+    <ComponentExample
+      title='Bulleted'
+      description='A list can mark items with a bullet'
+      examplePath='elements/List/Types/ListExampleBulleted'
+    />
+    <ComponentExample examplePath='elements/List/Types/ListExampleBulletedSimple'>
+      <Message info>
+        For convenience, a simple bulleted list can also use <code>ul</code> tag.
```
levithomason: I think this message is only applicable to SUI core where the user defines HTML/CSS.  Since, in Semantic-UI-React the classNames are hidden from the user.  Let's go ahead and remove this.

docs/app/Examples/elements/List/Types/index.js
```diff
@@ -0,0 +1,50 @@
...
+    <ComponentExample examplePath='elements/List/Types/ListExampleBulletedHorizontal' />
+
+    <ComponentExample
+      title='Ordered'
+      description='A list can be ordered numerically'
+      examplePath='elements/List/Types/ListExampleOrdered'
+    />
+    <ComponentExample examplePath='elements/List/Types/ListExampleOrderedSimple'>
+      <Message info>
+        You can also manually specify a value for an ordered list using <code>value</code>.
```
levithomason, line 34: This message belongs on the next example.

docs/app/Examples/elements/List/Variations/ListExampleHorizontalNumbered.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { Image, List } from 'stardust'
+
+const ListExampleHorizontalNumbered = () => (
```
levithomason: How about we rename this `ListExampleHorizontalOrdered` to keep with the prop name?

src/elements/List/List.js
```diff
@@ -1,40 +1,136 @@
...
+  const rest = getUnhandledProps(List, props)
+
+  return <ElementType {...rest} className={classes}>{children}</ElementType>
+}
+
+List._meta = {
+  name: 'List',
+  type: META.TYPES.ELEMENT,
+  props: {
+    floated: ['left', 'right'],
```
levithomason: Can use `SUI.FLOATS` here for the values.

src/elements/List/ListContent.js
```diff
@@ -0,0 +1,61 @@
...
+
+  return <ElementType {...rest} className={classes}>{children}</ElementType>
+}
+
+ListContent._meta = {
+  name: 'ListContent',
+  parent: 'List',
+  type: META.TYPES.ELEMENT,
+  props: {
+    floated: ['left', 'right'],
```
levithomason: Can use `SUI.FLOATS` here for the values.

src/elements/List/ListItem.js
```diff
@@ -1,54 +1,61 @@
...
+    disabled,
+    value,
+  } = props
+
+  const ElementType = getElementType(ListItem, props)
+  const classes = cx(
+    useKeyOnly(active, 'active'),
+    useKeyOnly(disabled, 'disabled'),
+    className,
+    useKeyOnly(ElementType !== 'li', 'item'),
```
levithomason, line 25: Can we move this up so the user's `className` is added very last?

src/elements/index.js
```diff
@@ -24,6 +24,8 @@ export { default as Input } from './Input/Input'
 export { default as Label } from './Label/Label'
 export { default as LabelDetail } from './Label/LabelDetail'
 
+// TODO: Export `List`
```
levithomason: Looks like this is done now, but we should export the rest of the List sub components as well.  I'll add a common test that ensures all components are exported at the top level as well.

levithomason (changes requested):
Once again, thanks for the massive amount of work here.  Especially for keeping up to date on all the latest decisions and incorporating them here.  It is very much appreciated!

I've left a few comments.  I'll update common tests as well, then we should update this branch to have the latest.

layershifter:
I've fixed comments and added tests.

There is one failed test on `Icon`, could you please fix it? I have to move away for a few hours :cry: 
After fixes, we're ready to go as I think :+1:

layershifter:
Updated to latest `master` and fixed test. @levithomason your move 👍

levithomason:
I'll release this later today, thanks much!

levithomason:
Released in `stardust@0.49.0`

PR #552, opened 2016-09-27
test(factories): add factory tests

levithomason:
This PR adds tests for `createShorthand`, it does not add tests for every shorthand factory, however, they are partially applied so it is _almost_ redundant.

### DIFF
--- a/src/lib/factories.js
+++ b/src/lib/factories.js
@@ -1,6 +1,6 @@
 import _ from 'lodash'
 import cx from 'classnames'
-import React, { isValidElement } from 'react'
+import React, { cloneElement, isValidElement } from 'react'
 
 /**
  * Merges props and classNames.
@@ -53,12 +53,10 @@ export function createShorthand(Component, mapValueToProps, val, defaultProps = 
 
   defaultProps = _.isFunction(defaultProps) ? defaultProps(usersProps) : defaultProps
   const props = mergePropsAndClassName(defaultProps, usersProps)
-  console.log('props', props)
-  console.log('defaultProps', defaultProps)
 
   // Clone ReactElements
   if (type === 'element') {
-    return React.cloneElement(val, props)
+    return cloneElement(val, props)
   }
 
   // Create ReactElements from props objects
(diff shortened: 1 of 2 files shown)


src/lib/factories.js
```diff
@@ -16,7 +16,7 @@ const mergePropsAndClassName = (defaultProps, props) => {
     newProps.className = cx(defaultProps.className, props.className) // eslint-disable-line react/prop-types
   }
 
-  if (!newProps.key) {
+  if (!newProps.key && childKey) {
```
levithomason, line 19: Prevent adding `childKey: undefined` due to props with no key and also not defining a `childKey`.  This issue was revealed by tests.

levithomason:
Just force pushed `from an array` tests

test/specs/lib/factories-test.js
```diff
@@ -0,0 +1,268 @@
...
+      })
+    })
+
+    describe('from null', () => {
+      itReturnsNull(null)
+      itReturnsNullGivenDefaultProps(null)
+      itReturnsNullGivenDefaultPropsFunc(null)
+    })
+
+    describe('from undefined', () => {
```
jeffcarbs: Add similar tests for booleans. Both `true` and `false` should return `null`

test/specs/lib/factories-test.js
```diff
@@ -0,0 +1,268 @@
...
+    const props = { 'data-foo': 'foo' }
+    const wrapper = shallow(getShorthand({ value, mapValueToProps: () => props }))
+
+    _.each(props, (val, key) => {
+      wrapper.should.not.have.prop(key, val)
+    })
+  })
+}
+
+const itMergesClassNames = (extraClassName, classNameSource, shorthandConfig) => {
```
jeffcarbs: Put `classNameSource` as first arg:
- matches `propsSource` being first in the functions below
- currently args 1 and 3 are used in the test and 2 is used to describe the test which is a little confusing.

jeffcarbs (approved):
Left a few suggestions but looks good otherwise so I'm approving 👍

levithomason:
Thanks much, updated both suggestions.  Will merge on pass.

PR #556, opened 2016-09-28
fix(Menu): add left/right `fixed` enums

levithomason:
This PR adds the missing fixed left/right options.

http://semantic-ui.com/collections/menu#fixed


levithomason:
Released in `stardust@0.47.2`

PR #567, opened 2016-09-29
feat(Table): add table row shorthand

### LINKED ISSUE #565: feat(Table): add shorthand for generating tables
The Table has been updated to our v1 API in #451.  This was done with the sub component API only.  We now need to add shorthand for generating tables.

Our previous `TableColumn` implementation was inspired by Facebook's [fixed-data-table](https://facebook.github.io/fixed-data-table).  @jcarbo has also put forth many great starter ideas in [this comment](https://github.com/TechnologyAdvice/stardust/pull/451#issuecomment-248774081).

Though it is more work, I am leaning toward the suggested row based approach (last example from jcarbo) because it offers more control.  Or at least providing this as _one_ of the shorthand options.  With the old `TableColumn` approach, here at TA we ran into issues by not being able to control the actual row or cell that was rendered.

I think providing "renderer" props that expect table elements to be returned is the way to go.  We could even provide multiple "renderer" props that would allow varying levels of control.  Even along side a `columns` prop that let you define columns like we used to.  This way, you can have several approaches to accomplish the most common goals.

jeffcarbs:
Fixes #565 

Add shorthand for Table and subcomponents:
- TableCell shorthand
- TableRow shorthand
- Table shorthand
- Add example to test table shorthand

---

I wound up implementing almost exactly like I had laid out in this comment: https://github.com/TechnologyAdvice/stardust/issues/565#issuecomment-250344807

I really like the way this came out. I think it allows you to be incredibly succinct in the simplest case but give you full control over the entire table (save for the `thead`, `tbody`, `tfoot`) elements. I think this is a good example of how powerful these shorthand factories are.

---

To give some more examples, assume this data:

``` js
const tableData = [
  { name: undefined, status: undefined, notes: undefined },
  { name: 'Jimmy', status: 'Requires Action', notes: undefined },
  { name: 'Jamie', status: undefined, notes: 'Hostile' },
  { name: 'Jill', status: undefined, notes: undefined },
]
```

The most basic table would be:

``` js
const headerRow = [
  'Name',
  'Status',
  'Notes',
]

const bodyRow = ({ name, status, notes }) => [name, status, notes]

<Table headerRow={headerRow} bodyRow={bodyRow} tableData={tableData} />
```

Now if one of those cells needs customizing, you can change that cell to any valid `TableCell` shorthand:

``` js
// Shorthand for TableRow
const headerRow = [
  'Name',
  { icon: 'attention', content: 'Status' },
  <MyCustomTableHeaderCell />,
]

// Function that returns shorthand for TableRow
const bodyRow = ({ name, status, notes }) => [
  name || 'No name specified',
  { icon: 'attention', content: status }
  <MyCustomElement someProp={notes} />
]
```

In these examples, the array of cells is just a shorthand for `TableRow`, so you can return `TableRow` props or a your own custom component:

``` js
const bodyRow = ({ name, status, notes }) => {
  someRowProp: true,
  items: [
    name || 'No name specified',
    { icon: 'attention', content: status }
    <MyCustomElement someProp={notes} />
  ]
]

// Or
const bodyRow = (data) => <MyCustomRow data={data} />
```

### DIFF
--- a/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
+++ b/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
@@ -1,0 +1,36 @@
+import React from 'react'
+import { Icon, Table } from 'stardust'
+
+const headerRow = [
+  'Name',
+  'Status',
+  'Notes',
+]
+
+const bodyRow = ({ name, status, notes }) => ({
+  warning: !!(status && status.match('Requires Action')),
+  items: [
+    name || 'No name specified',
+    status
+      ? { icon: 'attention', content: status }
+      : 'Unknown',
+    notes
+      ? { icon: 'attention', content: notes, warning: true }
+      : 'None',
+  ]
+})
+
+const tableData = [
+  { name: undefined, status: undefined, notes: undefined },
+  { name: 'Jimmy', status: 'Requires Action', notes: undefined },
+  { name: 'Jamie', status: undefined, notes: 'Hostile' },
+  { name: 'Jill', status: undefined, notes: undefined },
+]
+
+const TableWarningShorthand = () => {
+  return (
+    <Table celled headerRow={headerRow} bodyRow={bodyRow} tableData={tableData} />
+  )
+}
+
+export default TableWarningShorthand
--- a/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
+++ b/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Icon, Table } from 'stardust'
+import { Table } from 'stardust'
 
 const headerRow = [
   'Name',
@@ -17,7 +17,7 @@ const bodyRow = ({ name, status, notes }) => ({
     notes
       ? { icon: 'attention', content: notes, warning: true }
       : 'None',
-  ]
+  ],
 })
 
 const tableData = [
--- a/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
+++ b/docs/app/Examples/collections/Table/States/TableWarningShorthand.js
@@ -7,9 +7,9 @@ const headerRow = [
   'Notes',
 ]
 
-const bodyRow = ({ name, status, notes }) => ({
+const renderBodyRow = ({ name, status, notes }) => ({
   warning: !!(status && status.match('Requires Action')),
-  items: [
+  cells: [
     name || 'No name specified',
     status
       ? { icon: 'attention', content: status }
@@ -29,7 +29,7 @@ const tableData = [
 
 const TableWarningShorthand = () => {
   return (
-    <Table celled headerRow={headerRow} bodyRow={bodyRow} tableData={tableData} />
+    <Table celled headerRow={headerRow} renderBodyRow={renderBodyRow} tableData={tableData} />
   )
 }
 
--- a/src/collections/Table/TableRow.js
+++ b/src/collections/Table/TableRow.js
@@ -2,6 +2,7 @@ import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
 import {
+  createShorthandFactory,
   customPropTypes,
   getElementType,
   getUnhandledProps,
@@ -11,6 +12,7 @@ import {
   useTextAlignProp,
   useVerticalAlignProp,
 } from '../../lib'
+import TableCell from './TableCell'
 
 function TableRow(props) {
   const {
@@ -19,6 +21,8 @@ function TableRow(props) {
     className,
     disabled,
     error,
+    itemAs,
+    items,
     negative,
     positive,
     textAlign,
@@ -40,7 +44,15 @@ function TableRow(props) {
   const ElementType = getElementType(TableRow, props)
   const rest = getUnhandledProps(TableRow, props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+  if (children) {
+    return <ElementType {...rest} className={classes}>{children}</ElementType>
+  }
+
+  return (
+    <ElementType {...rest} className={classes}>
+      {items.map((item) => TableCell.create(item, { as: itemAs }))}
+    </ElementType>
+  )
 }
 
 TableRow._meta = {
@@ -55,6 +67,7 @@ TableRow._meta = {
 
 TableRow.defaultProps = {
   as: 'tr',
+  itemAs: 'td',
 }
 
 TableRow.propTypes = {
@@ -65,7 +78,10 @@ TableRow.propTypes = {
   active: PropTypes.bool,
 
   /** Primary content of the TableRow. */
-  children: PropTypes.node,
+  children: customPropTypes.every([
+    customPropTypes.disallow(['content', 'icon']),
+    PropTypes.node,
+  ]),
 
   /** Classes that will be added to the TableRow className. */
   className: PropTypes.string,
@@ -75,6 +91,20 @@ TableRow.propTypes = {
 
   /** A row may call attention to an error or a negative value. */
   error: PropTypes.bool,
+
+  /** An element type to render as (string or function). */
+  itemAs: customPropTypes.as,
+
+  /** Shorthand array of props for TableCell. Mutually exclusive with children. */
+  items: customPropTypes.every([
+    customPropTypes.disallow(['children']),
+    // Array of shorthands for MenuItem
+    PropTypes.arrayOf(PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.element,
+      PropTypes.object,
+    ])),
+  ]),
 
   /** A row may let a user know whether a value is bad. */
   negative: PropTypes.bool,
@@ -91,5 +121,7 @@ TableRow.propTypes = {
   /** A row may warn a user. */
   warning: PropTypes.bool,
 }
+
+TableRow.create = createShorthandFactory(TableRow, items => ({ items }))
 
 export default TableRow
--- a/src/collections/Table/TableRow.js
+++ b/src/collections/Table/TableRow.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
@@ -50,7 +51,7 @@ function TableRow(props) {
 
   return (
     <ElementType {...rest} className={classes}>
-      {items.map((item) => TableCell.create(item, { as: itemAs }))}
+      {_.map(items, (item) => TableCell.create(item, { as: itemAs }))}
     </ElementType>
   )
 }
--- a/src/collections/Table/TableRow.js
+++ b/src/collections/Table/TableRow.js
@@ -18,12 +18,12 @@ import TableCell from './TableCell'
 function TableRow(props) {
   const {
     active,
+    cellAs,
+    cells,
     children,
     className,
     disabled,
     error,
-    itemAs,
-    items,
     negative,
     positive,
     textAlign,
@@ -51,7 +51,7 @@ function TableRow(props) {
 
   return (
     <ElementType {...rest} className={classes}>
-      {_.map(items, (item) => TableCell.create(item, { as: itemAs }))}
+      {_.map(cells, (cell) => TableCell.create(cell, { as: cellAs }))}
     </ElementType>
   )
 }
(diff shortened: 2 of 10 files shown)


levithomason:
This looks really great, thanks.  I'll try to focus on a review tomorrow.

jeffcarbs:
Added specs for the shorthand rendering that should bump the coverage back to 100%.

src/lib/factories.js
```diff
@@ -46,7 +46,7 @@ export function createShorthand(Component, mapValueToProps, val, defaultProps =
   } else if (_.isPlainObject(val)) {
     type = 'props'
     usersProps = val
-  } else if (_.isString(val) || _.isNumber(val)) {
+  } else if (_.isString(val) || _.isNumber(val) || _.isArray(val)) {
```
levithomason, line 49: (...submitting a little half baked, ran out of time on this comment but want to get the convo started)

I'm not sure we should pass arrays on to be mapped to props.  It means every component might do something different with the array (or not handle it at all).  If the factory does handle arrays, IMHO it is most intuitive to expect it to always create an array of elements.

``` jsx
const button = Button.create('Submit')

// `key` derived from the string/number literal or `childKey` prop
const buttons = Button.create(['A', 123])
const buttons = Button.create([{ content: 'A' }, { content: 123 }])
const buttons = Button.create([<Other>A</Other>, <Other>123</Other>])
```

Then we can standardize shorthand propTypes since all components with shorthand would be capable of handling arrays.

If the factory passes it along, we'd need to make sure every `mapValueToProps` handles arrays.  Though, every component then might do something different with the array which isn't clear to the user.  "How does X handle arrays again?" Also, not all components / `mapValueToProps()` may be able support arrays.  It could make standard propTypes tough.

Lastly, when passing an array, you are creating an array of X components.  If there is a need to pass that array through the X's props to some Y's shorthand, then the first component is just a pass through.  My gut says there should be a "one level deep" design decision.  Meaning, we would forgo the middle man and pass the shorthand directly to the Y instead of through X to Y.

...totally open on this, just thinking out loud.

src/collections/Table/TableRow.js
```diff
@@ -76,6 +93,20 @@ TableRow.propTypes = {
   /** A row may call attention to an error or a negative value. */
   error: PropTypes.bool,
 
+  /** An element type to render as (string or function). */
+  itemAs: customPropTypes.as,
```
levithomason: Not sure of/if there is a solution yet, but I'd love to remove this prop somehow.  Seems an indicator of a larger pattern problem. We had this on the Label as well, but were able to remove it with sub components.

One idea would be that the implementing component would pass the `as` prop / key in each item in the array so it is explicit.  If that is too cumbersome, perhaps we need `TableHeaderRow` or something, similar to the two Table cells.

docs/app/Examples/collections/Table/States/TableWarningShorthand.js
```diff
@@ -0,0 +1,36 @@
...
+const tableData = [
+  { name: undefined, status: undefined, notes: undefined },
+  { name: 'Jimmy', status: 'Requires Action', notes: undefined },
+  { name: 'Jamie', status: undefined, notes: 'Hostile' },
+  { name: 'Jill', status: undefined, notes: undefined },
+]
+
+const TableWarningShorthand = () => {
+  return (
+    <Table celled headerRow={headerRow} bodyRow={bodyRow} tableData={tableData} />
```
levithomason: API suggestion:

``` jsx
<Table 
  renderHeaderRow={...}
  renderBodyRow={...}
  renderFooterRow={...}
/>
```

Eventually, I'd like to also add column/cell shorthand.  These would disallow use of the row shorthand, but I think it would give the user the ability to map through their data in what ever way made the most sense to them.

levithomason (commented):
Awesome, thanks so much for starting this.  Let's iterate, it's going in a great direction.  Shorthand is really paying off!

src/collections/Table/TableRow.js
jeffcarbs: The `Label` had `detail` and `detailAs` which we merged into one:

``` js
<Label detail={{ content: 'x', as: 'div' }} />
```

This is a little more tricky since `items` is an array of shorthands so it wouldn't be straightforward to add an "as" prop to each one since they may all be literals. For reference, this it where it would have to be done:

``` js
{headerRow && <TableHeader>{TableRow.create(headerRow, { itemAs: 'th' })}</TableHeader>}
```

In this case, `headerRow` is a shorthand for a `TableRow`.

docs/app/Examples/collections/Table/States/TableWarningShorthand.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { Table } from 'stardust'
+
+const headerRow = [
+  'Name',
```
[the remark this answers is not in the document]
jeffcarbs: The reason the body row renderer needs to be a function is that each row in the body is a function of an element in the `tableData` array. The header and footer rows are static as far as the `Table` is concerned so I'm not sure if it would make sense for them to be functions.

> I think each row renderer should get access to the row element.

Also to clarify, `headerRow` _is_ the row element, same with `footerRow`. `bodyRow` is a more of a "template" of what a row element looks like.

src/lib/factories.js
jeffcarbs: To give a little more context, I came up with these shorthands by building upwards and trying to collapse where possible. So:

``` js
// This
<Table.Cell>Name</Table.Cell>
// became
<Table.Cell content='Name' />


// This
<Table.Row>
  <Table.Cell content='Name' />
  <Table.Cell content='Status' />
  <Table.Cell content='Company' />
</Table.Row>

// became
<Table.Row cells=['Name', 'Status', 'Company'] />


// This
<Table>
  <Table.Header>
    <Table.Row cells=['Name', 'Status', 'Company'] />
  </Table.Header>
</Table>

// became
<Table headerRow={{ cells: ['Name', 'Status', 'Company'] }} />
```

If you think about our shorthand, generally it lets you pass a literal in place of a one-key object, e.g.

``` js
// One-key object
<Label detail={{ content: 'Some detail' }} />

// Literal
<Label detail={'Some detail'} />
```

So I figured if we can describe a Table.Row as a one-key object, we should be able to describe it as a "literal":

``` js
// One-key object
<Table headerRow={{ cells: ['Name', 'Status', 'Company'] }} />

// Literal
<Table headerRow={['Name', 'Status', 'Company']} />
```

So I'm not thinking about using arrays to let you build an array of some component (e.g. `const buttons = Button.create(['A', 123])`), but instead I'm thinking about it as a valid way of describing a component that has a list of things (e.g. `TableRow` has a list of cells that can completely describe that row).

Other examples:

``` js
// One-key object
Menu.create({ items: ['menu item 1', 'menu item 2', 'menu item 3'] })

// Literal
Menu.create(['menu item 1', 'menu item 2', 'menu item 3'])


// One-key object
List.create({ items: [<List.Item />, { content: 'item 2', description: 'item 2'}, 'list item 3'] })

// Literal
List.create([<List.Item />, { content: 'item 2', description: 'item 2'}, 'list item 3'])
```

Thoughts?

jeffcarbs: I had a sudden moment of clarity thinking about this haha. There are really two separate types of shorthands we're talking about:

### What is shorthand?

A way of describing component using a literal value, a props object, or an complete React.Element

### Item Shorthand

What we've dealt with thus far. Describes an individual component that may or may not have children. If it does have children, the children tend to be heterogeneous/different (e.g. a label can have multiple children but they're all different: icon, image, content, detail, delete-button)

#### Valid propTypes:
- string/number
- Object (props)
- Element

#### Example Components
- Label
- Icon
- TableCell

### Collection Shorthand

What this PR is introducing. Describes an individual component that has a set of homogenous children (e.g. a menu has 1+ items that are all the same).

#### Valid propTypes:
- array (of "item shorthand"s for set of children)
- Object (props)
- Element

#### Examples
- Menu
- List
- TableRow

---

I think almost everything is the same for both (e.g. they can use the same factory function). The differences are:
- Which validations we apply for the "literal" case. Could just have `customPropTypes.collectionShorthand` and `customPropTypes.itemShorthand` (if we like these names).
- How the `mapValueToProps` function maps a literal, but that's defined within the individual component itself so there shouldn't really be an issue there.

levithomason: Cool, I think we're seeing the same split in types of shorthand.  This is what I ran into [here](https://github.com/TechnologyAdvice/stardust/pull/452#issuecomment-243614062) where I reference `customPropTypes.shorthandSingle` and `customPropTypes.shorthandMultiple`.  I like collection a lot better than multiple though 😄.

So, I think we have 2 confirmed types of shorthand then.  Both handle a props objects and elements identically.  The difference is in what other values they accept and how they `mapValueToProps`:
1. Single item factory (Icon, etc.) also handles **string/number** mapping them to props.
2. Collection factory (Menu, etc.) also handles **arrays** mapping them to a single collection prop (items, cells, etc).

There is a 3rd group we can deal with later, since I think they need API redesign. Breadcrumb, Accordion, and possibly Feed, which take an array but do not map it to props.  Instead, they create children in a special way, usually due to markup constraints.  We'll save that for another day and time.

I'm working on the Feed updates and shorthand propTypes as I find time.  Won't hold up this PR for that work.  I can circle back and update when my work there is ready.

src/collections/Table/TableRow.js
levithomason: OK, let's save it for a later PR.

Here's an idea for the record.  I have this same issue with `FeedExtra`.  It has the need for two different shorthands, one that uses `images` and one that uses `text`.  I'm thinking of just adding:

``` jsx
FeedExtra.createImages()
FeedExtra.createText()
```

What if we added this?

``` jsx
TableRow.create()
TableRow.createHeader()
```

The latter would set `as` on each element in the array.  I think this touches on how collection shorthands are different than an item shorthand.  A string or number is not valid in a collection, `TableRow.create('not valid') //=> <Table items='not valid' />`.  So it seems we don't need a factory that only provides a `mapValueToProps` but instead/also a way to update each item. Perhaps with a default props object/function somehow.  Again, this comment is for later reference.

src/collections/Table/TableRow.js
```diff
@@ -76,6 +93,20 @@ TableRow.propTypes = {
...
+
+  /** Shorthand array of props for TableCell. Mutually exclusive with children. */
+  items: customPropTypes.every([
+    customPropTypes.disallow(['children']),
+    // Array of shorthands for MenuItem
```
levithomason: `MenuItem` => `TableRow`, or just pull this comment entirely.

levithomason (changes requested):

levithomason:
Ugh, still getting used to the new way GH groups/shows/hides review comments.  LMK if that last review is not clear. I did it on the files tab where I could see my review comments next to the past comments.

levithomason:
Just ran into the need to pass array values to `mapValueToProps` in the Feed, https://github.com/TechnologyAdvice/stardust/pull/575/commits/cb5cb0628227836066ab64a80e291ebaee58c172.  Just confirming it seems this is certainly the way to go after all.

docs/app/Examples/collections/Table/States/TableWarningShorthand.js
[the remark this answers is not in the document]
jeffcarbs: +1 for the rename of `bodyRow` => `renderBodyRow` to differentiate it more. Not completely opposed to making the other two functions in the future, but for now I'd prefer keeping it as-is. Perhaps when we add column/cell shorthands it will give a little more clarity.

jeffcarbs:
Made the requested changes, rebased and pushed up the latest. There are some further enhancements that I think can be addressed on subsequent PRs, so I think this is ready for final review 👍

docs/app/Examples/collections/Table/States/TableWarningShorthand.js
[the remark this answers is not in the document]
levithomason: I'm  on board :thumbsup:

levithomason (approved):

ralphievolt:
Just a question.. So populating my table with data from my props, I should use the short hand method as shown here 
`const tableData = [
  { name: undefined, status: undefined, notes: undefined },
  { name: 'Jimmy', status: 'Requires Action', notes: undefined },
  { name: 'Jamie', status: undefined, notes: 'Hostile' },
  { name: 'Jill', status: undefined, notes: undefined },
]`

as opposed to `{ this.props.data.map(item) => <Table.Row> <Table.Cell> ......`

Thanks.

levithomason:
Both are valid, whatever works best for you and your app 👍 

http://react.semantic-ui.com/collections/table#warning-shorthand

PR #569, opened 2016-09-29
feat(Segment): add `size` prop

### LINKED ISSUE #568: feat(Segment): add `size` prop
The Segment `size` prop is not documented in the primary SUI docs, however, it is supported.

https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/segment.css#L752

We should add the `size` prop to the Segment.

layershifter:
Fixes #568.

I've added prop and tests, do we need add docs?

### DIFF
--- a/src/elements/Segment/Segment.js
+++ b/src/elements/Segment/Segment.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
@@ -35,6 +36,7 @@ function Segment(props) {
     piled,
     raised,
     secondary,
+    size,
     stacked,
     tertiary,
     textAlign,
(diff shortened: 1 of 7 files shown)


src/elements/Segment/Segment.js
```diff
@@ -78,10 +81,11 @@ Segment._meta = {
   type: META.TYPES.ELEMENT,
   props: {
     attached: ['top', 'bottom'],
+    color: SUI.COLORS,
     floated: SUI.FLOATS,
-    textAlign: SUI.TEXT_ALIGNMENTS,
     padded: ['very'],
-    color: SUI.COLORS,
+    size: _.without(SUI.SIZES, 'medium'),
```
layershifter, line 87: There are no `medium` size.

levithomason:
Docs would be fantastic!

layershifter:
> Docs would be fantastic!

Added examples 👍

levithomason:
Awesome, thanks a million @layershifter.  Released in `stardust@0.48.3`.

Issue #564, opened 2016-09-28
refactor(docs): rename file name to be the same as class name, etc


dpkwhan:
Proposed changes:
1. Rename file name to be consistent with class name
2. Use stateless component whenever possible
3. Rename class. For example, rename ButtonButtonExample to ButtonExampleButton and ButtonBasicExample to ButtonExampleBasic
4. Use index.js whenever it is possible

---
### Elements
- [x] #756 Button
- [x] #713 Container
- [x] #715 Divider
- [x] #722 Flag
- [x] #725 Header
- [x] #685 Icon
- [x] #726 Image
- [x] #679 Input
- [x] #576 Label 
- [x] #551 List
- [x] #654 Loader
- [x] #658 Rail
- [x] #751 Segment
- [x] #668 Step
### Collections
- [x] #671 Breadcrumb
- [x] #767 Form
- [x] #656 Grid
- [x] #835 Menu
- [x] #789 Message
- [x] #771 Table
### Views
- [x] #667 Card
- [x] #575 Feed
- [x] #881 Item
- [x] #866 Statistic
### Modules
- [x] #838 Accordion
- [x] #839 Checkbox
- [x] #682 Dropdown
- [x] #800 Modal
- [x] #664 Progress
- [x] #640 Rating
- [x] #840 Search
### Addons
- [x] #630 Confirm
- [x] #662 Radio

levithomason:
Yes! Thank you very much.  This has been on my bucket list.  Note the List PR, ~~#311~~ #551 handles number 4 for the List examples. So we should wait for that PR to be merged before updating the List examples.

Lastly, the other goal is to allow the code snippets to be live edited and have the example update.  This may affect how we name things as well.  I'll start a trial of that right now and see what I find.  Either way, having consistent class/filenames will make any refactors far more sane.  So we can move forward on this immediately.

dpkwhan:
@levithomason I will start to work on this in the next few days.

levithomason:
Update, I'm able to handle the live example code editing with any export name.  So no issues or limitations there.

layershifter:
I'm renamed doc files to this naming pattern in #551, LGTM :+1: 

I think we need create dedicated section for `shorthand` examples in docs:
- Types
- Content
- Shorhands

levithomason:
> I think we need create dedicated section for shorthand examples in docs:

I agree this should be standardized.  I'm not sure what order/pattern is most optimal.  

[...]

Issue #561, opened 2016-09-28
docs(shorthand): add dedicated shorthand docs page


levithomason:
Shorthand props are at the heart of what makes our API so powerful and easy to work with.  With the recent updates and additions to our factories, the shorthand concept is becoming the crux of this library.  We're set to only further improve shorthand for arrays, child keys, and prop types.  Understanding how to use it is paramount to using Semantic-UI-React.

Shorthand is covered briefly half way down the introduction page.  However, it is incomplete and deserves more prominence.  We should add a dedicated "Shorthand" page under "Getting Started".

levithomason:
Include information and examples from #882.

levithomason:
Include an input example, see #1386.

brianespinosa:
@levithomason Assigning this old one over to myself. I'll see about putting something together to help with this.

brianespinosa:
I've officially blocked out a day to work on some of this over the holiday. 🎁 🎉

layershifter:
Unstale, please :cat

brianespinosa:
@layershifter I saw you mention elsewhere that you were thinking about changing to docs to be built with Gatsby. Is that something you are still considering?

If so, we might want to look into something like MDX files to generate our docs. I actually am putting something together with my team that does this right now internally.

layershifter:
@brianespinosa we using ReactStatic now.

We can adopt MDX files there, just add an additional loader: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/static.webpack.js#L41

I don't have there a strategy for this now. Feel free to promote it 👍

brianespinosa:
Ah got it. Let me see how things pan out with our internal work first and once I have some more feedback on pros/cons I can put a proposal together.

layershifter:
@brianespinosa MDX support is added, #3196 ✌️

Issue #524, opened 2016-09-21
RFC: remove production propTypes, allow small custom builds


layershifter:
I've builded `stardust` with webpack in my project.

There are two problems:
- [x] unused components in bundle that increases bundle size;
- [x] `PropTypes` are in production bundle, too.

### bundlers TODO

- [x] #1452 test that webpack 1 removes `propTypes` from production build, also check `SUI.js` usage
- [ ] #1497 test that webpack 2 treeshaking works
- [ ] test that rollup works

### propTypes update

- [x] #1216 Button
- [x] #1159 Container
- [x] #1159 Divider
- [x] #1155 Flag 
- [x] #1159 Header
- [x] #1159 Icon
- [x] #1155 Image
- [x] #1167 Input
- [x] #1265 Label
- [x] #1270 List
- [x] #1165 Loader
- [x] #731 Rail
- [x] #1165 Reveal
- [x] #1165 Segment
- [x] #1203 Step
- [x] #1209 Breadcrumb
- [x] #1320 Form
- [x] #1231 Grid
- [x] #1264 Menu
- [x] #1254 Message
- [x] #1200 Table
- [x] #1284 Card
- [x] #1293 Comment
- [x] #1285 Feed
- [x] #1294 Item
- [x] #1232 Statistic
- [x] Accordion
- [x] #1155 Checkbox
- [x] Dimmer
- [x] #1334 Dropdown
- [x] #1217 Embed
- [x] #1331 Modal
- [x] #1322 Popup
- [x] #1269 Progress
- [x] #1253 Rating
- [x] #1150 Search
- [x] #1296 Sidebar
- [x] #1282 Confirm
- [x] #1300 Portal
- [x] #1155 Radio
- [x] Select
- [x] Text Area

levithomason:
Agreed on both concerns.  Removing propTypes in production and allowing smaller builds.  Here are my thoughts.

### propTypes

There are transforms to remove propTypes in production builds, however, we're currently using them in `getUnhandledProps` to determine which props the component does not support.  Removing them will break things.  We'll need to remove library use of propTypes.

### Unused Components

In the optimization phase (after getting the bulk of work completed) I plan to publish individual components.  This way users can just install what they need.

We should also consider structuring our primary published project to users can do something similar to:

``` jsx
import Button from 'stardust/Button'
```

[...]

Issue #200, opened 2016-03-16
<Transition /> Component


levithomason:
http://semantic-ui.com/modules/transition.html

layershifter:
@levithomason Any ideas about this component? It's required by `<Dimmer>`, `<Message>`, `<Modal>` and many other components.

levithomason:
Correct, in thinking this will end up more of a util than a component. I'd like to reuse Semantic UI CSS transitions. I made a proof of concept of this for the Dropdown. See Dropdown/Todo.md for the notes. 

My thought is that we'll need some way to cycle through animation classes and add inline styles for "transition-duration".  You'll see in the [TODO.md](https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/TODO.md) that it takes 3 setState calls to complete an animation when reusing SUI CSS.  I  want to limit setting state and triggering renders for this as well. 

This has to work for many components. Perhaps even be generalized for any component. I do not want add react-motion or react-motion-ui-pack. This seems like overkill and we'd have to duplicate all the CSS animations that the user has already downloaded. 

Finally, I want to do this last. I don't think animations are more important than v1 APIs or component support. 

Those are my thoughts for now. Ideas and suggestions on how to pull this off are welcomed.

levithomason:
I have not tested it, but the proposal noted in the TODO above appears to have been successfully implemented by @asvetliakov in semantic-react here:

https://github.com/hallister/semantic-react/blob/master/src/components/animation

It looks like it is reusing the CSS classes which is exactly what we want.  Who ever implements this should consider that work.  Any input or learning experiences @asvetliakov could offer would also be much appreciated!

asvetliakov:
@levithomason Wow, you're so fast noticed it! I actually didn't read your proposal, but forced to need replace react-motion with CSS transitions due to very bad performance on mobile devices (even with careful rendering and pure components).

You can copy my implementation if you want, it's pretty similar to how ReactTransitionGroup/ReactCSSTransitionGroup works but with few noticiable differences (only one child).

There are few things though:
- Animations on portal-ed components (Modal/Popup) should be performed on mounting (componentWillAppear) (it's obvious though)
- Leave transitions for multiple components when component inside another component are so much trouble. Obvious example is Modal transitions inside Dimmer transitions. The semantic-ui markup defines modal as child of dimmer but it won't work in my version due to:

``` jsx
return (
            <SemanticCSSTransition
                enter={enter}
[...]

PR #570, opened 2016-09-30
feat(Popup): add component

### LINKED ISSUE #193: <Popup /> Component
http://semantic-ui.com/modules/popup.html

typehorror:
Port of SUI Popup module

Fixes #193 
- [x] sizing
- [x] content / header
- [x] positioning
- [x] events
  - [x] hover
  - [x] click
  - [x] focus
- [x] hoverable (will be fixed by #590)
- [x] widths
  - [x] use SUI enums
- [x] sizes
  - [x] use SUI enums
- [x] inverted
- [x] basic
- [x] offset
- [x] hideOnScroll (quick solution to scrollable sub-content like a sidebar) 
- [x] exclusive
- [x] custom class names
- [x] custom styles

To be done later:
- inline (popup as a sibling of trigger, will be implemented later)
- transition (animation will probably be for later)

### DIFF
--- a/docs/app/Examples/modules/Popup/Triggers/PopupClickExample.js
+++ b/docs/app/Examples/modules/Popup/Triggers/PopupClickExample.js
@@ -1,0 +1,13 @@
+import React from 'react'
+import { Button, Popup } from 'semantic-ui-react'
+
+const PopupClickExample = () => (
+  <Popup
+    trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
+    content={<Button color='green' content='Confirm the launch' />}
+    on='click'
+    positioning='top right'
+  />
+)
+
+export default PopupClickExample
--- a/docs/app/Examples/modules/Popup/Triggers/index.js
+++ b/docs/app/Examples/modules/Popup/Triggers/index.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+const PopupTriggersExamples = () => (
+  <ExampleSection title='Triggers'>
+    <ComponentExample
+      title='Hover'
+      description='A popup can be triggered on hover'
+      examplePath='modules/Popup/Triggers/PopupHoverExample'
+    />
+    <ComponentExample
+      title='Click'
+      description='A popup can be triggered on click'
+      examplePath='modules/Popup/Triggers/PopupClickExample'
+    />
+    <ComponentExample
+      title='Focus'
+      description='A popup can be triggered on focus'
+      examplePath='modules/Popup/Triggers/PopupFocusExample'
+    />
+  </ExampleSection>
+)
+
+export default PopupTriggersExamples
--- a/docs/app/Examples/modules/Popup/Types/PopupExample.js
+++ b/docs/app/Examples/modules/Popup/Types/PopupExample.js
@@ -1,0 +1,11 @@
+import React from 'react'
+import { Button, Popup } from 'semantic-ui-react'
+
+const PopupExample = () => (
+  <Popup
+    trigger={<Button icon='add' />}
+    content='Add users to your feed'
+  />
+)
+
+export default PopupExample
--- a/docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
+++ b/docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { Popup, Card, Rating, Image } from 'semantic-ui-react'
+
+const IndividualCard = (
+  <Card>
+    <Image src='http://semantic-ui.com/images/movies/totoro-horizontal.jpg' />
+    <Card.Content>
+      <Card.Header>
+        My Neighbor Totoro
+      </Card.Header>
+      <Card.Description>
+        Two sisters move to the country with their father in order to be
+        closer to their hospitalized mother, and discover the surrounding
+        trees are inhabited by magical spirits.
+      </Card.Description>
+    </Card.Content>
+  </Card>
+)
+
+const PopupHtmlExample = () => (
+  <Popup trigger={IndividualCard}>
+    <Popup.Header>User Rating</Popup.Header>
+    <Popup.Content>
+      <Rating icon='star' defaultRating={3} maxRating={4} />
+    </Popup.Content>
+  </Popup>
+)
+
+export default PopupHtmlExample
--- a/docs/app/Examples/modules/Popup/Variations/PopupInvertedExample.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupInvertedExample.js
@@ -1,0 +1,19 @@
+import React from 'react'
+import { Button, Icon, Popup } from 'semantic-ui-react'
+
+const PopupInvertedExample = () => (
+  <div>
+    <Popup
+      trigger={<Button icon='add' />}
+      content='Hello. This is an inverted popup'
+      inverted
+    />
+    <Popup
+      trigger={<Icon circular name='heart' />}
+      content='Hello. This is an inverted popup'
+      inverted
+    />
+  </div>
+)
+
+export default PopupInvertedExample
--- a/docs/app/Examples/modules/Popup/Variations/PopupOffsetExample.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupOffsetExample.js
@@ -1,0 +1,21 @@
+import React from 'react'
+import { Icon, Popup } from 'semantic-ui-react'
+
+const PopupOffsetExample = () => (
+  <div>
+    <Popup
+      trigger={<Icon size='large' name='heart' circular />}
+      content='Way off to the left'
+      offset={50}
+      positioning='left center'
+    />
+    <Popup
+      trigger={<Icon size='large' name='heart' circular />}
+      content='As expected this popup is way off to the right'
+      offset={50}
+      positioning='right center'
+    />
+  </div>
+)
+
+export default PopupOffsetExample
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -1,0 +1,306 @@
+import React, { Component, PropTypes } from 'react'
+import cx from 'classnames'
+import _ from 'lodash'
+import {
+  getElementType,
+  getUnhandledProps,
+  META,
+  SUI,
+  useKeyOnly,
+  useKeyOrValueAndKey,
+} from '../../lib'
+import Portal from '../../addons/Portal'
+import PopupContent from './PopupContent'
+import PopupHeader from './PopupHeader'
+
+const _meta = {
+  name: 'Popup',
+  type: META.TYPES.MODULE,
+  props: {
+    content: [PropTypes.string, PropTypes.node],
+    on: ['hover', 'click', 'focus'],
+    positioning: [
+      'top left',
+      'top right',
+      'bottom right',
+      'bottom left',
+      'right center',
+      'left center',
+      'top center',
+      'bottom center',
+    ],
+    size: _.without(SUI.SIZES, 'medium', 'big', 'massive'),
+    wide: [true, false, 'very'],
+  },
+}
+
+/**
+ * A Popup displays additional information on top of a page.
+ */
+export default class Popup extends Component {
+  static propTypes = {
+    /** Display the popup without the pointing arrow */
+    basic: PropTypes.bool,
+
+    /** You may pass a content as children of the Popup */
+    children: PropTypes.node,
+
+    /** Classes to add to the Popup className. */
+    className: PropTypes.string,
+
+    /** Simple text content for the popover */
+    content: PropTypes.oneOfType(_meta.props.content),
+
+    /** A Flowing popup have no maximum width and continue to flow to fit its content */
+    flowing: PropTypes.bool,
+
+    /** Takes up the entire width of its offset container */
+    // TODO: implement the Popup fluid layout
+    // fluid: PropTypes.bool,
+
+    /** Header displayed above the content in bold */
+    header: PropTypes.string,
+
+    /** Whether the popup should not close on hover */
+    hoverable: PropTypes.bool,
+
+    /** Invert the colors of the popup */
+    inverted: PropTypes.bool,
+
+    /** The node where the popup should mount.. */
+    hideOnScroll: PropTypes.bool,
+
+    /** Horizontal offset in pixels to be applied to the popup */
+    offset: PropTypes.number,
+
+    /** Event triggering the popup */
+    on: PropTypes.oneOf(_meta.props.on),
+
+    /** Positioning for the popover */
+    positioning: PropTypes.oneOf(_meta.props.positioning),
+
+    /** Popup size */
+    size: PropTypes.oneOf(_meta.props.size),
+
+    /** custom popup style */
+    style: PropTypes.object,
+
+    /** Element to be rendered in-place where the popup is defined. */
+    trigger: PropTypes.node,
+
+    /** Popup width */
+    wide: PropTypes.oneOf(_meta.props.wide),
+  }
+
+  static defaultProps = {
+    positioning: 'top left',
+    on: 'hover',
+  }
+
+  static _meta = _meta
+  static Content = PopupContent
+  static Header = PopupHeader
+
+  state = {}
+
+  computePopupStyle(positions) {
+    const style = { position: 'absolute' }
+    const { offset } = this.props
+    const { pageYOffset, pageXOffset } = window
+    const { clientWidth, clientHeight } = document.documentElement
+
+    if (_.includes(positions, 'right')) {
+      style.right = Math.round(clientWidth - (this.coords.right + pageXOffset))
+      style.left = 'auto'
+    } else if (_.includes(positions, 'left')) {
+      style.left = Math.round(this.coords.left + pageXOffset)
+      style.right = 'auto'
+    } else {  // if not left nor right, we are horizontally centering the element
+      const xOffset = (this.coords.width - this.popupCoords.width) / 2
+      style.left = Math.round(this.coords.left + xOffset + pageXOffset)
+      style.right = 'auto'
+    }
+
+    if (_.includes(positions, 'top')) {
+      style.bottom = Math.round(clientHeight - (this.coords.top + pageYOffset))
+      style.top = 'auto'
+    } else if (_.includes(positions, 'bottom')) {
+      style.top = Math.round(this.coords.bottom + pageYOffset)
+      style.bottom = 'auto'
+    } else {  // if not top nor bottom, we are vertically centering the element
+      const yOffset = (this.coords.height + this.popupCoords.height) / 2
+      style.top = Math.round(this.coords.bottom + pageYOffset - yOffset)
+      style.bottom = 'auto'
+
+      const xOffset = this.popupCoords.width + 8
+      if (_.includes(positions, 'right')) {
+        style.right -= xOffset
+      } else {
+        style.left -= xOffset
+      }
+    }
+
+    if (offset) {
+      if (_.isNumber(style.right)) {
+        style.right -= offset
+      } else {
+        style.left -= offset
+      }
+    }
+
+    return style
+  }
+
+  // check if the style would display
+  // the popup outside of the view port
+  isStyleInViewport(style) {
+    const { pageYOffset, pageXOffset } = window
+    const { clientWidth, clientHeight } = document.documentElement
+
+    const element = {
+      top: style.top,
+      left: style.left,
+      width: this.popupCoords.width,
+      height: this.popupCoords.height,
+    }
+    if (_.isNumber(style.right)) {
+      element.left = clientWidth - style.right - element.width
+    }
+    if (_.isNumber(style.bottom)) {
+      element.top = clientHeight - style.bottom - element.height
+    }
+
+    // hidden on top
+    if (element.top < pageYOffset) return false
+    // hidden on the bottom
+    if (element.top + element.height > pageYOffset + clientHeight) return false
+    // hidden the left
+    if (element.left < pageXOffset) return false
+    // hidden on the right
+    if (element.left + element.width > pageXOffset + clientWidth) return false
+
+    return true
+  }
+
+  setPopupStyle() {
+    if (!this.coords || !this.popupCoords) return
+    let positioning = this.props.positioning
+    let style = this.computePopupStyle(positioning)
+
+    // Lets detect if the popup is out of the viewport and adjust
+    // the position accordingly
+    const positions = _.without(_meta.props.positioning, positioning)
+    for (let i = 0; !this.isStyleInViewport(style) && i < positions.length; i++) {
+      style = this.computePopupStyle(positions[i])
+      positioning = positions[i]
+    }
+
+    // Append 'px' to every numerical values in the style
+    style = _.mapValues(style, value => _.isNumber(value) ? value + 'px' : value)
+    this.setState({ style, positioning })
+  }
+
+  getPortalProps() {
+    const portalProps = { onOpen: this.onOpen }
+    const { on, hoverable } = this.props
+
+    switch (on) {
+      case 'click':
+        portalProps.openOnTriggerClick = true
+        portalProps.closeOnTriggerClick = true
+        portalProps.closeOnDocumentClick = true
+        break
+
+      case 'focus':
+        portalProps.openOnTriggerFocus = true
+        portalProps.closeOnTriggerBlur = true
+        break
+
+      default:  // default to hover
+        portalProps.openOnTriggerMouseOver = true
+        portalProps.closeOnTriggerMouseLeave = true
+        // Taken from SUI: https://git.io/vPmCm
+        portalProps.mouseLeaveDelay = 70
+        portalProps.mouseOverDelay = 50
+        break
+    }
+
+    if (hoverable) {
+      portalProps.closeOnPortalMouseLeave = true
+      portalProps.mouseLeaveDelay = 300
+    }
+
+    return portalProps
+  }
+
+  hideOnScroll = (event) => {
+    this.setState({ closed: true })
+    window.removeEventListener('scroll', this.hideOnScroll)
+    setTimeout(() => this.setState({ closed: false }), 50)
+  }
+
+  onOpen = (event) => {
+    this.coords = event.currentTarget.getBoundingClientRect()
+    if (this.props.hideOnScroll) {
+      window.addEventListener('scroll', this.hideOnScroll)
+    }
+  }
+
+  popupMounted = (ref) => {
+    this.popupCoords = ref ? ref.getBoundingClientRect() : null
+    this.setPopupStyle()
+  }
+
+  render() {
+    const {
+      basic,
+      children,
+      className,
+      content,
+      flowing,
+      header,
+      inverted,
+      size,
+      trigger,
+      wide,
+    } = this.props
+
+    const { positioning, closed } = this.state
+    const style = _.assign({}, this.state.style, this.props.style)
+    const classes = cx(
+      'ui',
+      positioning,
+      size,
+      useKeyOrValueAndKey(wide, 'wide'),
+      useKeyOnly(basic, 'basic'),
+      useKeyOnly(flowing, 'flowing'),
+      useKeyOnly(inverted, 'inverted'),
+      'popup transition visible',
+      className,
+    )
+
+    if (closed) return trigger
+
+    const rest = getUnhandledProps(Popup, this.props)
+    const ElementType = getElementType(Popup, this.props)
+    const portalProps = _.pick(rest, _.keys(Portal.propTypes))
+
+    const popupJSX = (
+      <ElementType {...rest} className={classes} style={style} ref={this.popupMounted}>
+        {children}
+        {!children && PopupHeader.create(header)}
+        {!children && PopupContent.create(content)}
+      </ElementType>
+    )
+
+    return (
+      <Portal
+        {...portalProps}
+        trigger={trigger}
+        {...this.getPortalProps()}
+      >
+        {popupJSX}
+      </Portal>
+    )
+  }
+}
--- a/src/modules/Popup/PopupContent.js
+++ b/src/modules/Popup/PopupContent.js
@@ -1,0 +1,36 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
+import {
+  createShorthandFactory,
+  getElementType,
+  getUnhandledProps,
+  META,
+} from '../../lib'
+
+/**
+ * A PopupContent displays the content body of a Popover.
+ */
+export default function PopupContent(props) {
+  const { children, className } = props
+  const classes = cx('content', className)
+  const rest = getUnhandledProps(PopupContent, props)
+  const ElementType = getElementType(PopupContent, props)
+
+  return <ElementType {...rest} className={classes}>{children}</ElementType>
+}
+
+PopupContent.create = createShorthandFactory(PopupContent, value => ({ children: value }))
+
+PopupContent.propTypes = {
+  /** The content of the Popup */
+  children: PropTypes.node,
+
+  /** Classes to add to the Popup content className. */
+  className: PropTypes.string,
+}
+
+PopupContent._meta = {
+  name: 'PopupContent',
+  type: META.TYPES.MODULE,
+  parent: 'Popup',
+}
--- a/src/modules/Popup/PopupHeader.js
+++ b/src/modules/Popup/PopupHeader.js
@@ -1,0 +1,36 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
+import {
+  createShorthandFactory,
+  getElementType,
+  getUnhandledProps,
+  META,
+} from '../../lib'
+
+/**
+ * A PopupHeader displays a header in a Popover.
+ */
+export default function PopupHeader(props) {
+  const { children, className } = props
+  const classes = cx('header', className)
+  const rest = getUnhandledProps(PopupHeader, props)
+  const ElementType = getElementType(PopupHeader, props)
+
+  return <ElementType {...rest} className={classes}>{children}</ElementType>
+}
+
+PopupHeader.create = createShorthandFactory(PopupHeader, value => ({ children: value }))
+
+PopupHeader.propTypes = {
+  /** The header of the Popup */
+  children: PropTypes.node,
+
+  /** Classes to add to the Popup header className. */
+  className: PropTypes.string,
+}
+
+PopupHeader._meta = {
+  name: 'PopupHeader',
+  type: META.TYPES.MODULE,
+  parent: 'Popup',
+}
--- a/docs/app/Examples/modules/Popup/Triggers/PopupFocusExample.js
+++ b/docs/app/Examples/modules/Popup/Triggers/PopupFocusExample.js
@@ -1,0 +1,13 @@
+import React from 'react'
+import { Input, Popup } from 'semantic-ui-react'
+
+const PopupFocusExample = () => (
+  <Popup
+    trigger={<Input icon='search' placeholder='Search...' />}
+    header='Movie Search'
+    content='You may search by genre, header, year and actors'
+    on='focus'
+  />
+)
+
+export default PopupFocusExample
--- a/docs/app/Examples/modules/Popup/Triggers/PopupHoverExample.js
+++ b/docs/app/Examples/modules/Popup/Triggers/PopupHoverExample.js
@@ -1,0 +1,12 @@
+import React from 'react'
+import { Button, Popup } from 'semantic-ui-react'
+
+const PopupHoverExample = () => (
+  <Popup
+    trigger={<Button icon='add' content='Add a friend' />}
+    content='Sends an email invite to a friend.'
+    on='hover'
+  />
+)
+
+export default PopupHoverExample
--- a/docs/app/Examples/modules/Popup/Types/PopupTitledExample.js
+++ b/docs/app/Examples/modules/Popup/Types/PopupTitledExample.js
@@ -1,0 +1,33 @@
+import React from 'react'
+import { Image, Popup } from 'semantic-ui-react'
+
+const users = [
+  {
+    name: 'Elliot Fu',
+    bio: 'Elliot has been a member since July 2012',
+    avatar: 'http://semantic-ui.com/images/avatar/small/elliot.jpg',
+  },
+  {
+    name: 'Stevie Feliciano',
+    bio: 'Stevie has been a member since August 2013',
+    avatar: 'http://semantic-ui.com/images/avatar/small/stevie.jpg',
+  },
+  {
+    name: 'Matt',
+    bio: 'Matt has been a member since July 2014',
+    avatar: 'http://semantic-ui.com/images/avatar/small/matt.jpg',
+  },
+]
+
+export default function PopupTitledExample() {
+  const avatars = users.map((user, index) =>
+    <Popup
+      key={user.name}
+      trigger={<Image src={user.avatar} avatar />}
+      header={user.name}
+      content={user.bio}
+    />
+  )
+
+  return <div>{avatars}</div>
+}
--- a/docs/app/Examples/modules/Popup/Types/index.js
+++ b/docs/app/Examples/modules/Popup/Types/index.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+const PopupTypesExamples = () => (
+  <ExampleSection title='Types'>
+    <ComponentExample
+      title='Popup'
+      description='An element can specify popup content to appear'
+      examplePath='modules/Popup/Types/PopupExample'
+    />
+    <ComponentExample
+      title='Titled'
+      description='An element can specify popup content with a title'
+      examplePath='modules/Popup/Types/PopupTitledExample'
+    />
+    <ComponentExample
+      title='HTML'
+      description='An element can specify HTML for a popup'
+      examplePath='modules/Popup/Types/PopupHtmlExample'
+    />
+  </ExampleSection>
+)
+
+export default PopupTypesExamples
--- a/docs/app/Examples/modules/Popup/Variations/PopupBasicExample.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupBasicExample.js
@@ -1,0 +1,12 @@
+import React from 'react'
+import { Button, Popup } from 'semantic-ui-react'
+
+const PopupBasicExample = () => (
+  <Popup
+    trigger={<Button icon='add' />}
+    content="The default theme's basic popup removes the pointing arrow."
+    basic
+  />
+)
+
+export default PopupBasicExample
--- a/docs/app/Examples/modules/Popup/Variations/PopupFlowingExample.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupFlowingExample.js
@@ -1,0 +1,30 @@
+import React from 'react'
+import { Header, Button, Popup, Grid } from 'semantic-ui-react'
+
+const PopupFlowingExample = () => (
+  <Popup
+    trigger={<Button>Show flowing popup</Button>}
+    flowing
+    hoverable
+  >
+    <Grid centered divided columns={3}>
+      <Grid.Column textAlign='center'>
+        <Header as='h4'>Basic Plan</Header>
+        <p><b>2</b> projects, $10 a month</p>
+        <Button>Choose</Button>
+      </Grid.Column>
+      <Grid.Column textAlign='center'>
+        <Header as='h4'>Business Plan</Header>
+        <p><b>5</b> projects, $20 a month</p>
+        <Button>Choose</Button>
+      </Grid.Column>
+      <Grid.Column textAlign='center'>
+        <Header as='h4'>Premium Plan</Header>
+        <p><b>8</b> projects, $25 a month</p>
+        <Button>Choose</Button>
+      </Grid.Column>
+    </Grid>
+  </Popup>
+)
+
+export default PopupFlowingExample
--- a/docs/app/Examples/modules/Popup/Variations/PopupHideOnScrollExample.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupHideOnScrollExample.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import { Button, Popup } from 'semantic-ui-react'
+
+const PopupHideOnScrollExample = () => (
+  <div>
+    <Popup
+      trigger={<Button icon>Click me</Button>}
+      content='Hide the popup on any scroll event'
+      on='click'
+      hideOnScroll
+    />
+    <Popup
+      trigger={<Button icon>Hover me</Button>}
+      content='Hide the popup on any scroll event'
+      hideOnScroll
+    />
+  </div>
+)
+
+export default PopupHideOnScrollExample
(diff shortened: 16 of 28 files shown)


typehorror:
Hey @levithomason,

I know that's not your initial design but I feel like adding an anchor element and an open state would simplify the implementation and open the way to additional feature (like a click trigger instead of a hover). The counter argument would be that it would complicate usage, as the user would have to maintain a state for the open status. The anchor element could be transmitted using `event.currentTarget` by the event handler.

Something among those lines:

``` jsx
export default class PopupOnClick extends Component {
  buttonClick = (event) =>
    this.setState({
      open: true,
      target: event.currentTarget,
    })

  render() {
    return <div>
      <button onClick={this.buttonClick} >Click me</button>
      <Popup 
        content="foo bar"
        open={this.state.open}
        target={this.state.target}
      />
    </div>
  }
}
```

I don't want to impose my style, I came here to help and I'm totally open to your opinion.

docs/app/Examples/modules/Popup/Types/PopupExample.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react'
+import { Button, Popup } from 'stardust'
+
+export default class PopupExample extends Component {
```
layershifter: We can keep example stateless there.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,68 @@
...
+      <div
+        style={this.getPosition(coords)}
+        className='ui popup top left transition visible'
+      >
+        boom
+      </div>
+    )
+
+    this.setState({ popup })
+  }
```
layershifter, line 234: Popup will be rerendered on each show.

layershifter:
I think we can transform render part to this.

``` jsx
const { active } = this.state

return (
   <div>
     <Portal isOpened={active}>{popup}</Portal>
     {trigger}
  </div>
 )
```

Also, I like SUI ideas there, can we implement them there?
image

docs/app/Examples/modules/Popup/Types/PopupExample.js
typehorror: I'll keep in mind to use stateless function when no state is required :+1:

typehorror:
I've made a few updates to the handling part of the positioning without introducing `refs` but I had to stop having the popup rendering the trigger.

While having an `open` prop on the popover isn't exactly the SUI approach, I believe it's actually closer to the react way. @layershifter this also opens the door to custom event triggering for the popup as you suggested (see my examples demonstrating `onFocus`, `onMouseOver` and `onClick`)

jeffcarbs:
Hey so I wound up writing our own `Portal` component (based on `react-portal` which we're currently using). I explain in more detail why in the comment here https://github.com/TechnologyAdvice/stardust/pull/571, but I think the changes could be useful here:
- It takes a `trigger` element which let's you avoid the wrapping `div`.
- You can optionally configure to show up `onClick` or `onMouseOver`.
- Control it yourself or let the `open` state be handled by the component.
- Currently doesn't support closing on mouseleave but I figured we could add that type of functionality on this PR given the modal really doesn't have that use-case.

If you have any feedback on some of the open items on that issue I'd appreciate it. Assuming we like that approach, the sooner we can work out the kinks the sooner we can apply it here.

typehorror:
@jcarbo this could indeed resolve my initial problem. I would need an additional `onMouseLeave` on your implementation but it looks like it would be a trivial addition (unless I missed it). Lastly I think your Portal version would allow me to retrieve the event and therefore position the popover accurately. 

**PS**: I believe there could be value in having the popup on focus (help on an input field), so maybe adding onFocus and onBlur event could be nice to have.

jeffcarbs:
@debrice - thanks for the suggestions! Wound up implementing all three options (closing `onMouseLeave`/`onBlur` and opening `onFocus`), see https://github.com/TechnologyAdvice/stardust/pull/571/commits/92390a02c936e91b475ce751d082ff31e9b861bf and https://github.com/TechnologyAdvice/stardust/pull/571/commits/e92914a215bfba883084a587c9cd31936545222e.

typehorror:
This is great, thank you @jcarbo , I'll pull your branch in and start implementing today.

docs/app/Examples/modules/Popup/Triggers/PopupClickExample.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Button, Icon, Popup } from 'stardust'
+
+export default function () {
```
levithomason: In light of #564, let's name example classes/functions `PopupExampleClick`.  Both the filename and default export should have the same name as well.

docs/app/Examples/modules/Popup/Triggers/PopupClickExample.js
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    <Button.Group>
+      <Button color='red'>Activate doomsday device</Button>
+      <Popup
+        trigger={<Button color='black' icon><Icon name='question' /></Button>}
```
levithomason: The button can be simplified with shorthand syntax here.  Shorthand is very useful for cases exactly like this, passing components through props:

``` jsx
trigger={<Button color='black' icon='question' />}
```

Then, you can also remove the `Icon` import.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import { Popup, Card, Rating, Image } from 'stardust'
+
+const users = [
```
levithomason: Unused var here, `npm run lint` to see lint issues.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
```diff
@@ -0,0 +1,54 @@
...
+    <div className='content'>
+      <Rating icon='star' defaultRating={3} maxRating={4} />
+    </div>
+  </div>
+
+  return <Popup
+    trigger={IndividualCard}
+    content={content}
+    hover
+  />
```
levithomason: Few points on this one:
1. Let's use `children` for defining full markup
2. Return multi-line JSX inside of parens (`npm run lint` to see all issues)
3. Use sub components for SUI component parts. See [this section](https://github.com/TechnologyAdvice/stardust/blob/master/.github/CONTRIBUTING.md#sui-html-markup) in the CONTRIBUTING.md for more details on API design.

Given our design and style patterns noted here ^, this example would look more like this:

``` jsx
return (
  <Popup trigger={IndividualCard} hover>
    <Popup.Header>User Rating</Popup.Header>
    <Popup.Content>
      <Rating icon='star' defaultRating={3} maxRating={4} />
    </Popup.Content>
  </Popup>
)
```

Note, I also realize now that there is a name conflict between the `data-title` prop and the `class="header"` component part.  Let's go with the Popup class names (`header`, etc) for our prop names and sub component names.

docs/app/Examples/modules/Popup/Variations/PopupWidthExample.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { Icon, Popup } from 'stardust'
+
+export default function () {
+  return (
+    <div>
+      <Popup
+        trigger={<Icon circular name='heart' />}
+        content='Hello. This is a wide pop-up which allows for lots of content
```
levithomason: Single quotes can't handle line breaks.  Though we could use backticks to preserve the line breaks, I think we should instead use `children` here:

``` jsx
<Popup trigger={<Icon circular name='heart' />} width='wide' hover>
  Hello. This is a very wide pop-up which allows for lots of content with additional space.
  You can fit a lot of words here and the paragraphs will be pretty wide.
</Popup>
```

src/modules/Popup/Popup.js
```diff
@@ -1,7 +1,8 @@
 import React, { Component, PropTypes } from 'react'
 import { findDOMNode } from 'react-dom'
 import { META } from '../../lib'
-import Portal from 'react-portal'
+import Portal from '../../addons/Portal/Portal'
+import classnames from 'classnames'
```
levithomason: We're using `cx` as the import name for `classnames`.  Let's update for consistency.

src/modules/Popup/Popup.js
```diff
@@ -12,24 +13,54 @@ export default class Popup extends Component {
...
   }
 
   getPosition(coords) {
-    const positions = (this.props.position || 'top left').split(' ')
+    if (!coords) return
+
+    const positions = this.props.position.split(' ')
     const style = { position: 'absolute' }
 
     if (positions.includes('right')) {
```
levithomason: `includes` has limited browser support.  Let's use `_.includes(positions, 'right')` instead here.

levithomason (changes requested):
I know this is a WIP, here are few comments along the way.  Thanks much for getting this going!

levithomason:
Left an interim review for now.  LMK when it's stable enough for a full review and I'll pull it down, try it out, and offer more input.  Cheers!

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
typehorror: I use the linter (also in my editor), all those variables are being used.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
typehorror: sounds good, I was also confused about title and header and was about to ask for confirmation :+1:

docs/app/Examples/modules/Popup/Variations/PopupWidthExample.js
typehorror: I somehow believed you were against using `props.children`, my bad.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
levithomason: Ah, I believe I'm getting confused on the new GH review flow.  This comment thread does not show it in use for some reason.

https://github.com/debrice/stardust/blob/d51317361c15723d2796c85a1752ee4d5fc63584/docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
typehorror: Somehow the linter does not complain about those being multi-line.
Great, going to implement those sub rendered.

docs/app/Examples/modules/Popup/Variations/PopupWidthExample.js
levithomason: No worries.  Our goal is to provide two APIs 1) a "Shorthand Props API" for the majority simplified use cases and 2) a sub component API as an escape hatch when you need access and control over the full markup.

It is worth noting, the two are mutually exclusive.  So, you cannot mix a shorthand `header` prop while also having `children` defined.  More [here](https://github.com/TechnologyAdvice/stardust/blob/master/.github/CONTRIBUTING.md#component-part-props) in the CONTRIBUTING.md.  Also, I plan to add more info to the docs on this, #561.

Here are some equivalent examples:

``` jsx
<Popup content='The content' />
<Popup>
  The content
<Popup>
```

``` jsx
<Popup header='My Header' content='The content' />
<Popup>
  <Popup.Header>My header</Popup.Header>
  <Popup.Content>The content</Popup.Content>
<Popup>
```

Most components have shorthand capabilities you can reference for more examples.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
levithomason: Hm, I get this in my editor:

image

And this from the CLI:

image

Maybe double check the linter setup and that the latest modules are installed?  I know NPM has had some issues as of late, sometimes requiring multiple installs to get all the modules you should've gotten the first go.  LMK if you have further issues and we can investigate deeper.

docs/app/Examples/modules/Popup/Types/PopupHtmlExample.js
typehorror: And you are right, I'm so confused :disappointed: It was a left over from a copy/paste and indeed my linter catch it too. I'm trying to do too many things at the same time.

levithomason: Yep, been there :)  Onward!

typehorror:
@levithomason I'm at a place where I could use some feedback from a UI experience. I'm not familiar with animation for example and how you expect them to be handled. 

Also I'm using `findDOMNode` which I believe isn't a sustainable solution but I should be able to drop it once I get access to the trigger's event from Portal. 

Positioning will need to have the "center" attribute too, but I can't think of a way to read the size of the popup without using ref with a function (...or `react-dom`):

``` jsx
//...
  popupMounted = (ref) => {
    this.setState({'popupBoundingRect': ref.getBoundingClientRect()})
  }
  render(){
    return <Popover ref={this.popupMounted} />
  }
}
```

Thank you for your help and patience!

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
+import React, { Component, PropTypes } from 'react'
+import { findDOMNode } from 'react-dom'
+import { META } from '../../lib'
+import Portal from '../../addons/Portal/Portal'
+import cx from 'classnames'
+import isString from 'lodash/isString'
```
jeffcarbs: The standard pattern in this project has been to include lodash as `_` as the very first import within the file. We use the `babel-plugin-lodash` plugin which handles cherry-picking only the functions you use (see https://github.com/lodash/babel-plugin-lodash#example)

levithomason:
> I'm at a place where I could use some feedback from a UI experience. I'm not familiar with animation for example and how you expect them to be handled.

We're forgoing animations until we get all components and APIs done.  We are just adding/removing the visibility classes as needed for now.  Later, we'll add a utility that reuses SUI CSS transitions by adding these classes (and styles) in sequence to trigger animations. See #200 for more notes on that.

> Also I'm using findDOMNode which I believe isn't a sustainable solution but I should be able to drop it once I get access to the trigger's event from Portal.

👍 Looks like the Portal is now passing the necessary events, see https://github.com/TechnologyAdvice/stardust/pull/571#issuecomment-250891167

> Positioning will need to have the "center" attribute too, but I can't think of a way to read the size of the popup without using ref with a function (...or react-dom):

Measuring is one of those things (focus being the other) that you just have to use a `ref` for.  I think you're OK there.  Though perhaps we consider borrowing a pattern from something like [react-measure](https://github.com/souporserious/react-measure).  Where we have an abstract component that can measure other components.  I don't want to use `react-measure` directly since it has a dependency [element-resize-detector](https://github.com/wnr/element-resize-detector/blob/master/dist/element-resize-detector.js) that is >1600 LOC.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
...
+    fluid: PropTypes.bool,
+
+    /** Display the popup on trigger focus */
+    focus: PropTypes.bool,
+
+    /** Header displayed above the content in bold */
+    header: PropTypes.string,
+
+    /** Display the popup on mouse over the trigger */
+    hover: PropTypes.bool,
```
levithomason: When this prop is pulled into the `on` enum prop, could we add `hoverable` here?

image

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
...
+    this.setState({ portalProps })
+  }
+
+  renderContent() {
+    const children = []
+    let index = 0
+    const { header, content } = this.props
+
+    if (header) {
+      children.push(<PopupHeader key={index++}>{header}</PopupHeader>)
```
levithomason: This method can go away in favor of shorthand factories suggested in the main render function.  However, for future reference avoid using `index` in keys.  We have a few in the codebase still that we're removing.  It is an [anti pattern that breaks React's ability to track/update children](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318#.e5e216vax).

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
...
+      flowing,
+      fluid,
+      inverted,
+      position,
+      size,
+      trigger,
+      width,
+    } = this.props
+
+    const { coords } = this.state
```
levithomason: Let's add `portalProps` used below as well.  I should add some lint rules for this since we destructure props/state always.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
...
+        flowing,
+        fluid,
+        inverted,
+      }
+    )
+
+    return (
+      <Portal trigger={trigger} {...this.state.portalProps}>
+        <div
+          style={this.getPosition(coords)}
```
levithomason: Suggest renaming to `this.getStyle()` given the usage.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
...
+          className={classes}
+        >
+        { children || this.renderContent() }
```
levithomason: Pending adding shorthand `create()` methods to the sub components, `this.renderChildren()` here can be replaced with this:

``` jsx
render() {
  return (
    <Portal trigger={trigger} {...portalProps}>
      <div style={this.getStyle(coords)} className={classes}>
        {children}
        {!children && PopupHeader.create(header)}
        {!children && PopupContent.create(content)}
      </div>
    </Portal>
  )
}
```

The `create()` shorthand can be created by importing `createShorthandFactory()` and adding a static method to each component:

``` jsx
import { createShorthandFactory } from '../../lib'

// PopupHeader.js
PopupHeader.create = createShorthandFactory(PopupHeader, value => ({ children: value }))

// PopupContent.js
PopupContent.create = createShorthandFactory(PopupContent, value => ({ children: value }))
```

The second argument is the `mapValueToProps()` function.  When `PopupHeader.create('my header')` is called, that function will map the string to `children`.  Factories also know how to handle numbers, props objects, and elements as a value.  So, your props will be able to handle many types of data.  You can read more in `factories.js` and see it in other components.  Again, more docs on the way for this, #561!

src/modules/Popup/PopupContent.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react'
+import { META } from '../../lib'
+
+/**
+ * A <PopupContent /> displays the content body of a Popover.
```
levithomason: For consistency `<PopupContent />` => `PopupContent`

src/modules/Popup/PopupContent.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+export default function PopupContent(props) {
+  return <div className='content'>{ props.children }</div>
+}
+
+PopupContent.propTypes = {
+  /** The header of the Popup */
```
levithomason: Copy pasta?  Probably should be `content` instead of `header`.

src/modules/Popup/PopupHeader.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react'
+import { META } from '../../lib'
+
+/**
+ * A <PopupHeader /> displays a header in a Popover.
```
levithomason: `<PopupHeader />` => `PopupHeader`

levithomason (changes requested):
Left another round of comments for you.  Thanks for putting up with it!  I'll try to focus on getting the Portal merged before doing a final here.  That way this PR only includes the Popup work.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,196 @@
+import React, { Component, PropTypes } from 'react'
+import { findDOMNode } from 'react-dom'
+import { META } from '../../lib'
+import Portal from '../../addons/Portal/Portal'
```
levithomason: Doesn't require a change, but FYI you can import components as named exports from the category folder index if you prefer:

``` jsx
import { Portal } from '../../addons'
```

Especially handy when there are several imports:

``` jsx
import { Button, Icon, Label } from '../../elements'
// vs
import Button from '../../elements/Button/Button'
import Label from '../../elements/Label/Label'
```

src/modules/Popup/Popup.js
typehorror: Great, I got into the habit of cherry picking method of the any library for size concerns. I was wondering if the same applies to stardust element then? If someone only wanted to import the Popup functionality, then if this person was using: 

``` js
import Popup from 'stardust/modules/Popup'
```

the way we import Portal in Popup:

``` jsx
import Portal from '../../addons/Portal/Portal'
```

instead of 

``` jsx
import { Portal } from '../../addons'
```

wouldn't it impact the size of the import by loading the whole `addons` content ?

src/modules/Popup/Popup.js
typehorror: in the case of the second example vs the first, wouldn't that impact the size of the final bundle if someone was to cherry pick a single component for a project ?

``` jsx
import Button from 'stardust/elements/Button'
```

levithomason: Indeed you are correct. I hadn't thought that through... It would automatically bundle all "elements". We're targeting small custom builds as part of our goals, #524. So from now on let's stick with directly importing files. Thanks for pointing it out.

src/modules/Popup/Popup.js
levithomason: Indeed, let's use the direct filename import. I've commented in #524 to reflect this issue.

jeffcarbs:
@debrice - could you rebase this PR against the most recent `master`? The `Portal` code has been merged in so it'll make the changes easier to read.

typehorror:
@jcarbo I've never rebased any code in the past, so that was quite an adventure (100 commits now). 

Rebase like you never rebased in the past

typehorror:
@levithomason @jcarbo I'm a bit unfamiliar with rebasing/cleaning up a PR . I believe I have 2 options now:
- I can have a single commit on my fork master branch but I would have to create a new PR.
- I could squash the commits on my current branch and force push the single squash commit.

Any advice ?

jeffcarbs:
I'd squash down the current branch to one commit. So assuming 2c144d1 is the commit you're rebased off of you would do:
- `git reset 2c144d1` - would put you back at 2c144d1 with all of the changes from this branch unstaged
- `git add .` - stage your changes
- `git commit -m 'feat(Popup): add component'`
- `git push --force`

All commits on a feature branch wind up getting squashed down to one anyway when they're merged into master but this would just clean things up a bit for now.

typehorror:
@jcarbo thanks for the help, force push is done 🎊 

Regarding Portal, I need to build `hoverable` props, which would allow a user to hover from the trigger to the popup, without the popup going away by the mouseleave:

Hoverable allowing going from trigger to popup

Should that be a feature of the Portal or the Popup? For it to be on the Popup, I believe I would still need some changes on the Portal to grant more control on the event.

levithomason:
Regarding the hoverable prop, I'd think that would set the Portal `closeOnMouseLeave` to false but keep `openOnMouseEnter` set to true.

This would leave the Popup open with no way to close it, since blur (mouseLeave) doesn't. Perhaps we'd use some kind of timer that would close the Popup but reset on hover. This way the user can move the mouse off the trigger (timer starts), mouse enters the Popup (timer cancelled), mouse leaves Popup (timer starts again), finally after X seconds the timer expires and the Popup closes.

jeffcarbs:
I think we can handle it by modifying one prop and adding another:
- Change `closeOnTriggerMouseLeave` to just `closeOnMouseLeave`. Calls `onClose` when the mouse leaves the trigger OR the portal (currently it's just the trigger).
- Add `mouseLeaveDelay`. On mouseleave, we delay for `mouseLeaveDelay` milliseconds before calling `onClose` and if we receive a mouseenter on the trigger OR portal during the delay we clear the timer. This is actually what SUI proper does. We could default to 70ms like they do (https://github.com/Semantic-Org/UI-Popup/blob/master/popup.js#L1373-L1377).

I can also see adding two other props for symmetry:
- closeOnTriggerClick
- mouseEnterDelay - SUI defaults this to 50ms

I'll open up a quick PR with these additions unless there are any objections.

typehorror:
@levithomason I can't think of an elegant way to attach a mouseLeave listener on the trigger as those are being overwritten on Portal. 

SUI offers a delay on `show` and `hide` (could be `open` & `close`) which I believe could be helpful on the portal itself. 

image

For now, I can think of two solutions:
- Portal could accept events which would overwrite the existing ones (or be triggered in addition of the existing ones)
- Portal would handle some sort of delay and does not close when Focus, Click or Hover events are happening within the Portal (I believe this is what you've suggested @jcarbo )

jeffcarbs:
Just pushed up a PR (#590) that gives the `Portal` the flexibility to handle the popup's use-cases. I tested that commit against this branch and it all worked well. You'll just need a few updates once that PR is merged in:

``` diff
     if (this.props.on === 'click') {
       portalProps.openOnTriggerClick = true
+      portalProps.closeOnTriggerClick = true
       portalProps.closeOnDocumentClick = true
     }
     if (this.props.on === 'hover') {
       portalProps.openOnTriggerMouseOver = true
       portalProps.closeOnTriggerMouseLeave = true
+      portalProps.closeOnPortalMouseLeave = true
+      // Taken from SUI: https://git.io/vPmCm
+      portalProps.mouseLeaveDelay = 70
+      portalProps.mouseOverDelay = 50
     }
```

levithomason:
Sweet, I'll review the portal updates now.

> I can't think of an elegant way to attach a mouseLeave listener on the trigger as those are being overwritten on Portal.

Note, no component should ever overwrite handlers.  They should wrap them.  This way a component can have it's own handler do what it needs to do, then inside that handler, check and call any parent handlers of the same name.  I believe we are doing this, if not, there is a bug and we'll fix it 👍

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  onOpen = (event) => {
+    this.coords = event.currentTarget.getBoundingClientRect()
+  }
+
+  componentWillMount() {
+    const portalProps = {}
+
+    if (this.props.on === 'click') {
```
jeffcarbs: I would move this out of `componentWillMount` and out of state. This is only called on initial mount, so if the `on` props change then this will be stale. I'd do something like:

``` js
// At the top of the file
const triggerActionToPortalProps = {
  click: {
    openOnTriggerClick: true,
    closeOnDocumentClick: true,
  },
  hover: {
    openOnTriggerMouseOver: true,
    closeOnTriggerMouseLeave: true,
  }
  focus: {
    openOnTriggerFocus: true,
    closeOnTriggerBlur: true,
  }
}

// Within render:
const { on } = this.props

<Portal trigger={trigger} {...onToPortalProps[on]}>
```

typehorror: Looks much better indeed. I'll update the Popup with your code and the new event trigger once the Portal feature gets merged.

typehorror:
@levithomason should the popup be covering all functionalities of the SUI one for initial release or could we go with the most common features being covered ?

**PS**: Added a checklist to the description of this PR

src/modules/Popup/Popup.js
jeffcarbs: Cool. Make sure to update with the changes to these configs here: https://github.com/TechnologyAdvice/stardust/pull/570#issuecomment-251022954

Another thing related to the `hoverable` prop - given that `closeOnTriggerMouseLeave` and `closeOnPortalMouseLeave` are separate props we can actually get closer to the SUI functionality:
- `hoverable: false`  is `{ closeOnTriggerMouseLeave: true }`
- `hoverable: true`  is `{ closeOnTriggerMouseLeave: true, closeOnPortalMouseLeave: true }`

`closeOnPortalMouseLeave` has the added effect of preventing close when mousing over the gap from trigger -> portal

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,244 @@
...
+        break
+      }
+    }
+
+    if (this.props.hoverable) {
+      portalProps.closeOnPortalMouseLeave = true
+      portalProps.mouseLeaveDelay = 300
+    }
+
+    this.setState({ portalProps })
```
jeffcarbs: Need to `return portalProps` here.

typehorror: Indeed, this commit was a bit on the hasty side with no linter on... might be living a bit too dangerously right here 😄

typehorror:
@jcarbo The hoverable works nicely, but the Portal updates might have re-introduced the bug on the focus where click to focus is removing the popup

levithomason:
> should the popup be covering all functionalities of the SUI one for initial release or could we go with the most common features being covered ?

MVP components are definitely acceptable.  I prefer quick iterations over single large shipments.  As long as we've vetted the API enough as to reasonably not expect to have boxed ourselves in when expanding it later.

typehorror:
@levithomason & @jcarbo 

I believe I'm at the last main feature needed for the Popup which is allowing the popup to be within an independent scrollable container. As of right now, since the popup is positioned relative to the body, scrolling such an element would make the trigger move away and the popup style fixed on the page. I haven't thought of all the possibility but here are what I could think of so far: 
- My preferred option would be to offer a hideOnScroll prop as provided by SUI. It would listen to window scroll events and temporarily take control of the open state of the Portal or simply replace the portal by the trigger when a scroll occurs, putting back the portal when the scrolling stops using a debounce.
- The other solution I thought of is that I might be able to make it work with the `mountNode` feature of the Portal but it's a bit convoluted as a ref needs to be passed to the Portal through the popup, which pretty much looks like something like that and requires a different positioning logic:

``` jsx
export default class PopupMountExample extends Component {
  state = {}

  divMounted = (ref) => {
    this.setState({ mountPoint: ref })
  }

  render() {
    return (
      <div ref={ this.divMounted } >
        <Popup
          trigger={<Button icon><Icon name='add' /></Button>}
          content="A custom mounted popup."
          mountNode={this.state.mountPoint}
          on='click'
        />
      </div>
    )
  }
}
```

My concern is offering a `mountNode` props might not be a feature we want to keep as I would rather have an boolean `inline` option in the long run. The `inline` option would make the popup a sibling element of the the `trigger` and require a bit more kung-fu for positioning but would also be more intuitive and closer to the spirit of SUI. I'm not sure how long it would take me to build a clean inline feature so I would rather deliver the popup without it initially. 

Thanks for your help!

levithomason:
Hm, I'll have to pull this down and try it/review to give an appropriate response.  Will do soon as I can.

levithomason:
It looks like this branch needs updated to the latest master.  The `index.js` files in each component type folder (i.e. `modules`) was deleted.  We now export everything in the `src/index.js` file.  You'll need to add an `src/modules/Popup/index.js` as well that just exports the Popup.  That way, it can imported like so:

``` js
import Popup from 'src/modules/Popup'
```

Then, the `src/modules/index.js` can be deleted in this branch as it is in `master`.  You can reference the other component directories for more examples.

typehorror:
@levithomason I believe (unless I'm mistaking, which happens quite often) I've already migrated to the new `index.js` structure. Also, I just pulled and merged master in again

jeffcarbs:
@levithomason - would be great if you could give a checklist of things left to do here, or comment on the one above. My opinion would be to solidify the functionality currently provided in this PR and add specs so that we can merge. Save any enhancements beyond that for separate issues/PRs.

debrice:
@jcarbo I put the checklist on the first comment of this PR

typehorror:
@jcarbo can you think of an easy way to have the portal injected as a sibling to the trigger? That's might be the last thing that prevents me from implementing the `inline` feature

levithomason:
> I've already migrated to the new index.js structure

Indeed, my branch was old, apologies!

@jcarbo agreed, let's feature freeze this where it is, cleanup and merge.  Then any extras can come later.  I'll start a final review now.

typehorror:
> @jcarbo agreed, let's feature freeze this where it is, cleanup and merge. Then any extras can come later. I'll start a final review now.

@levithomason in that case I'll work on spec ASAP

docs/app/Examples/modules/Popup/Triggers/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+export default function () {
+  return (
```
levithomason: All files' default export should match the filename.  Since index.js files don't make sense in this pattern, we're using the "Component + Directory + Examples".  Also note the const vs function, let's update to use:

``` js
const PopupTriggersExamples = () => (
  // ...
)
export default PopupTriggersExamples
```

docs/app/Examples/modules/Popup/Types/PopupExample.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Button, Icon, Popup } from 'semantic-ui-react'
+
+export default function PopupExample() {
+  return (
+    <div>
+      <Popup
+        trigger={<Button icon><Icon name='add' /></Button>}
```
levithomason: Please use shorthand icon button:

``` jsx
<Button icon='add' />
```

docs/app/Examples/modules/Popup/Types/PopupTargetExample.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { Popup, Card, Rating, Image } from 'semantic-ui-react'
+
+export default function PopupHtmlExample() {
```
levithomason: Export name doesn't match filename.  Though, this file appears the same as `PopupHtmlExample.js` and I cannot find it referenced anywhere else in this PR.  Is it cruft?

docs/app/Examples/modules/Popup/Variations/PopupFluidExample.js
```diff
@@ -0,0 +1,22 @@
...
+import { Message, Button, Popup, Grid } from 'semantic-ui-react'
+
+export default function PopupFluidExample() {
+  return (
+    <div>
+      <Message
+        warning
+        header='Work in Progress'
+        content='This one needs to be resized according to the parrent offset element.'
+      />
```
levithomason: Suggest pulling this example and allowing a future PR to add it since it needs more work.  Note, when adding Messages, add them to the `index.js` file so that it doesn't show up in the example code editor in the docs:

**docs/app/Examples/modules/Popup/Variations/index.js**

``` jsx
<ComponentExample
  title='Fluid'
  description='A fluid popup will take up the entire width of its offset container'
  examplePath='modules/Popup/Variations/PopupFluidExample'
>
  <Message
    warning
    header='Work in Progress'
    content='This one needs to be resized according to the parrent offset element.'
  />
</ComponentExample>
```

docs/app/Examples/modules/Popup/Variations/PopupInvertedExample.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { Button, Icon, Popup } from 'semantic-ui-react'
+
+export default function PopupInvertedExample() {
+  return (
+    <div>
+      <Popup
+        trigger={<Button icon><Icon name='add' /></Button>}
```
levithomason: Let's use a shorthand icon button here as well.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+    /** The node where the popup should mount.. */
+    hideOnScroll: PropTypes.bool,
+
+    /** Horizontal offset in pixels to be applied to the popup */
+    offset: PropTypes.number,
+
+    /** Event triggering the popup */
+    on: PropTypes.oneOf(['hover', 'click', 'focus']),
+
+    /** Positionning for the popover */
```
levithomason: `Positionning` => `Positioning`

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+    inverted: PropTypes.bool,
+
+    /** The node where the popup should mount.. */
+    hideOnScroll: PropTypes.bool,
+
+    /** Horizontal offset in pixels to be applied to the popup */
+    offset: PropTypes.number,
+
+    /** Event triggering the popup */
+    on: PropTypes.oneOf(['hover', 'click', 'focus']),
```
levithomason: Enum props should be listed in `_meta.props`:

``` js
_meta = {
  props: {
    on: ['hover', 'click', 'focus'],
  },
}

on: PropTypes.oneOf(_meta.props.on),
```

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+
+    /** Positionning for the popover */
+    position: PropTypes.string,
```
levithomason: Since there are fixed positions supported, let's also define them in `_meta.props` and use an enum here:

``` js
_meta = {
  props: {
    positioning: [
      'top left',
      'top center',
      'top right',
      'right center',
      'bottom right',
      'bottom center',
      'bottom left',
      'left center',
    ]
  },
}

positioning: PropTypes.oneOf(_meta.props.positioning),
```

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+
+    /** Popup size */
+    size: PropTypes.string,
```
levithomason: There are fixed sizes:

``` jsx
_meta = {
  props: {
    size: _.without(SUI.sizes, 'medium', 'big', 'massive')
  },
}

size: PropTypes.oneOf(_meta.props.size),
```

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+    className: PropTypes.string,
+
+    /** Simple text content for the popover */
+    content: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+
+    /** A Flowing popup have no maximum width and continue to flow to fit its content */
+    flowing: PropTypes.bool,
+
+    /** Takes up the entire width of its offset container */
+    fluid: PropTypes.bool,
```
levithomason: Since this example is not working, let's comment this out with a TODO.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+    size: PropTypes.string,
+
+    /** custom popup style */
+    style: PropTypes.object,
+
+    /** Element to be rendered in-place where the popup is defined. */
+    trigger: PropTypes.node,
+
+    /** Popup width */
+    width: PropTypes.string,
```
levithomason: There are fixed widths, though I think we should update this to `wide` and `wide='very'` given the SUI classes at hand.

``` js
_meta = {
  props: {
    wide: ['very']
  },
}

wide: customPropTypes.some([
  PropTypes.bool,
  PropTypes.oneOf(_meta.props.wide)
]),
```

> Here's more context as to why the rename.  The `width` is primarily used for components that have grid column style widths, or a group component that defines the widths of its children.  It takes a number or a word (`width={4}` `width='four'`).  In this case, there it is more like the `padded` or `relaxed` prop, where it is either boolean or takes a `very` value.

docs/app/Examples/modules/Popup/Variations/PopupOffsetExample.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Button, Icon, Popup } from 'semantic-ui-react'
+
+export default function PopupOffsetExample() {
+  return (
+    <div>
+      <Popup
```
levithomason: We can remove the extra `div` wrapper here and just return the `<Popup />`.  As long as a single element is returned, React is happy.  React only requires an extra containing element (such as a `div`) if multiple siblings need to be returned.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+        style.left -= xOffset
+      }
+    }
+
+    if (offset) {
+      if (_.isNumber(style.right)) {
+        style.right -= offset
+      } else {
+        style.left += offset
+      }
```
levithomason, line 148: ## Expected Behavior

This logic seems to always offset to the right.  I'd expect the offset to be in the opposite direction of the position:

``` jsx
position='top center'       // offset upward
position='left center'      // offset to the left
position='bottom center'    // offset downward
```

Or, provide an offset object with +/- `horizontal` and `vertical` values that would be applied:

``` jsx
position='top center'       // offset upward
offset={{ vertical: -5 }}

position='top center'       // offset downward
offset={{ vertical: 5 }}
```

## Current Behavior

### `position='top center'`



### `position='left center'`



### `position='bottom center'`

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,270 @@
...
+        fluid,
+        inverted,
+      }
+    )
+
+    if (closed) return trigger
+
+    return (
+      <Portal trigger={trigger} {...this.getPortalProps()}>
+        <div ref={this.popupMounted} style={style} className={classes} >
```
levithomason: Extra space here before closing angle: `className={classes} >`

levithomason (changes requested):
If any of these comments seem like adding features, let's comment out the feature with a TODO and push them off to another PR.  This way we can break down this work into smaller pieces 👍

src/modules/Popup/Popup.js
layershifter: I think there might be `ElementType` from `getElementType` instead of `div`.

``` jsx
const ElementType = getElementType(Popup, props)

<ElementType {...rest} ...>...</ElementType>
```

src/modules/Popup/PopupContent.js
```diff
@@ -0,0 +1,22 @@
+import React, { PropTypes } from 'react'
+import { createShorthandFactory, META } from '../../lib'
+
+/**
+ * A PopupContent displays the content body of a Popover.
+ */
+export default function PopupContent(props) {
+  return <div className='content'>{ props.children }</div>
```
layershifter: Please use `ElementType` there.

src/modules/Popup/Popup.js
levithomason: Nice catch

src/modules/Popup/PopupContent.js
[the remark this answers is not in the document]
levithomason: Again, nice catch.  Thanks!

src/modules/Popup/Popup.js
typehorror: I thought of the same thing, but it seems like SUI offset is only acting on the X axis and from left to right. I can change to one of your suggestion if you want me to

typehorror:
@levithomason, @layershifter  I addressed the PR comments, will start working on the specs now.

Also @jcarbo I can't get the focus through click to behave on the trigger (the popup appears and disappear almost instantly). Would you mind checking that my Portal config https://github.com/TechnologyAdvice/stardust/pull/570/files#diff-daca312d5932d8a0a079b86773c0ee3cR206 ?

src/modules/Popup/PopupContent.js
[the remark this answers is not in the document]
typehorror: Thanks for the guidance guys, your reviews are super helpful! :+1:

jeffcarbs:
Hey @debrice that was "unfixed" when I added `closeOnTriggerClick`. Here's the diff that will fix it if you want to just add it on your branch:

``` diff
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -205,14 +205,14 @@ class Portal extends Component {
       e.stopPropagation()
       this.close(e)
     } else if (!open && openOnTriggerClick) {
-      // Prevents closeOnDocumentClick from closing the portal when
-      // openOnTriggerFocus is set. Focus shifts on mousedown so the portal opens
-      // before the click finishes so it may actually wind up on the document.
-      e.nativeEvent.stopImmediatePropagation()
-
       e.stopPropagation()
       this.open(e)
     }
+
+    // Prevents closeOnDocumentClick from closing the portal when
+    // openOnTriggerFocus is set. Focus shifts on mousedown so the portal opens
+    // before the click finishes so it may actually wind up on the document.
+    e.nativeEvent.stopImmediatePropagation()
   }

   handleTriggerFocus = (e) => {
```

---

Also, can you squash your commits back down again? The reason your `rebase` was difficult last time was because you had `merge`d a few times before and git had to try to parse through all the merged commits with duplicate code.

Once you squash again, any time you want to pull in the latest changes from master if you use rebase you shouldn't have any conflicts and your work will be back on top without any merge commits.

typehorror:
Great, thank you very much @jcarbo, this fixes the issue indeed! I will do the rebase very soon, once I'm done wrapping up the unit test.

typehorror:
@levithomason I think we're ready for the final review.

src/modules/Popup/Popup.js
```diff
@@ -0,0 +1,302 @@
...
+      size,
+      trigger,
+      wide,
+    } = this.props
+
+    const { positioning, closed } = this.state
+
+    const style = _.assign({}, this.state.style, this.props.style)
+
+    const classes = cx(
```
jeffcarbs, line 270: Update usage of `cx` to follow the standard pattern of this lib (see https://github.com/TechnologyAdvice/stardust/blob/master/src/collections/Grid/Grid.js for an example). There are helpers to set the appropriate class based on the prop.

For example, `wide` here can use the useKeyOrValueAndKey (see similar prop usage: https://github.com/TechnologyAdvice/stardust/blob/master/src/collections/Grid/Grid.js#L34-L35)

jeffcarbs:
There's still quite a few lines missing test coverage. We keep this package at 100% test coverage so we won't be able to merge until all lines are covered. You can see the coverage by clicking details here:
screen shot 2016-10-05 at 12 04 35 am

You can also install the codecov.io browser extension to see the coverage right in github:
screen shot 2016-10-05 at 12 03 16 am

typehorror:
@jcarbo sounds good, I'll work on getting 100% coverage. Thanks for the heads up

typehorror:
@jcarbo I'm almost done. I'm missing some positioning/sizing tests for which I haven't yet find a solution.

levithomason:
## Coverage Report

You can also get a killer coverage report by opening:

```
file:///<path_to_project>/coverage/lcov-report/index.html
```

> Where `<path_to_project>` is your absolute file path to the project root.

## Investigating Coverage

You can sort, drill down to directories/files, and see specific lines:

image

---

image

---

**TYPO: I meant, when positions does NOT include 'right' :)**

image

typehorror:
Thanks @levithomason, I found the coverage yesterday evening after @jcarbo comment but I really appreciate how all of you go above and beyond to guide and help. 

I know it's a bit late but, is there some guide I should read ? Also, which element would incarnate the project code style (including test and example) ?

For example, many example are using Components while being stateless which turns out to be something we don't want. If we could illustrate guidance through an exemplary implementation of a SUI element I believe a new contributor (like myself) could save you a lot of time.

Finally, testing the positioning logic turns out to be quite a puzzle, I really want to to cover it fully prior to release but it might take a few more days.

levithomason:
> I know it's a bit late but, is there some guide I should read ? Also, which element would incarnate the project code style (including test and example) ?

Not sure if you've read the [CONTRIBUTING.md](https://github.com/TechnologyAdvice/stardust/blob/master/.github/CONTRIBUTING.md) yet.  It is a walkthrough from cloning the repo, to documenting and testing your component.  So, it is meant to be read like a book, from top to bottom.  It is fairly comprehensive, but could always be better.  LMK if you feel something is missing from there.

typehorror:
@levithomason did the CONTRIBUTING.md get deleted?

levithomason:
Whoops, no it moved to `.github`.  I had an old link, updated, try it now 👍

typehorror:
@jcarbo and @levithomason, the branch finally got a 100% code coverage 🎉   

I believe I could test the library more but I need more research on phantomJS (`event.target.getBoundingClientRect` returns 0 for all coordinates).

typehorror:
@levithomason I think we are good for the final review (unless there is something else I didn't see) :smile:

levithomason:
Cool, this looks like a great first shipment for this component.  Thanks for hanging in there and pioneering this stuff!

levithomason:

typehorror:
Thank you guys for your patience and especially @jcarbo for his great work on porting and upgrading the Portal lib! I had a blast 👍

Issue #553, opened 2016-09-27
RFC: Could/should the modal rely more heavily on React.Portal?


jeffcarbs:
In `react-portal` there's a prop called `openByClickOn` (which stardust doesn't support currently) that lets you specify a trigger without requiring an empty div wrapper. So instead of:

``` js
<div>
  <Button onClick={this.show}>Basic Modal</Button>
  <Modal basic size='small' active={active} onHide={this.hide} />
</div>
```

You could do:

``` js
<Modal basic size='small' active={active} onHide={this.hide} openByClickOn={<Button>Basic Modal</Button>} />
```

It got me thinking, it seems there's quite a bit of functionality baked into `React.Portal` (e.g. tracking visible state, close on click outside, close on esc) that we've duplicated in the `Modal` component. Could we update the modal to pass portal-related props straight to the portal and remove duplicate functionality from the modal?

I recall reading a comment here recently about the desire to decouple the modal from the portal, so maybe this would take us in the wrong direction. However, this would give the end-user more control over the portal functionality and allow us to remove code that's already been written and tested elsewhere.

---

_Edit_ I'm happy to take a pass at doing this if there's interest.

levithomason:
IIRC at first I tried to use the portal props for open/close and close on key/click but had to implement the then within Stardust for some reason.  I think it had something to do with the testing, or special SUI cases that didn't fit the portal API well?  Can't recall exactly.

Anyhow, I'm game for removing code and delegating this to the portal if possible.  I just think we'd rename the props to fit SUI conventions better (ie `open` vs `isOpen`).

At the very minimum, mapping `trigger` to `openByClickOn` makes perfect sense to me.  I could even see using `Button.create()` shorthand there.  Odds are, most modals triggers will be buttons:

``` jsx
<Modal trigger='Open modal!' />
```

levithomason:
Regarding decoupling, it's a nice idea to have the Modal as a stateless component that is then wrapped up in the Portal.  However, I'm doubting there are enough use cases to warrant it.  We'll see how the Dimmer #447 component plays out.  Its API is going to affect how the Modal/Portal relation plays out since the Portal _is required_ to be the Modal Dimmer for SUI css to work.  This could end up being a `<Dimmer as={Portal} />` or something.

For now, let's keep Modal/Portal baked together.  Time will tell the rest.

jeffcarbs:
Fixed via #571

Issue #405, opened 2016-08-18
Remove or update use of refs


levithomason:
See update below:

> I'd like to remove as many use of refs as possible, and those that are 100% necessary should be callbacks.

~~Some components use a `ref` where they can likely use `findDOMNode(this)` instead.  Regarding nested components, we have reliable markup and classNames to use.  We should prefer `findDOMNode(this).querySelector('...')` to find nodes nested within ourselves.  We should be able to remove most, if not all, use of refs in this way.~~

chuan137:
Can you explain the downside of using ref? At least the syntax looks quite clean to me.

levithomason:
Facebook will likely deprecate `ref='string'` syntax in favor of the callback syntax.

> Although string refs are not deprecated, they are considered legacy, and will likely be deprecated at some point in the future. Callback refs are preferred.
> 
> https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute

However, this issue should actually be updated.  As [Dan Abramov noted](https://github.com/yannickcr/eslint-plugin-react/issues/678#issue-165177220) in eslint-plugin-react, `findDOMNode` hopes to deprecated as well.  Starting first with an eslint rule against it to help move the community.The replacement again is a ref callback.

In general, refs should be avoided.  React is a virtual DOM environment, when using a ref, you break out of that virtual DOM into the real DOM.  It makes testing more difficult and opens the door for issues.  I'd like to remove as many use of refs as possible, and those that are 100% necessary should be callbacks.

levithomason:
More context, `ref`s do not work with stateless functional components.  Which, most of our components are functional components.  React `ref`s can only be used with classes.  There are workarounds, such as passing the ref through a different prop name or wrapping the functional component in a class component.  Though, I think neither of these are preferred patterns.

Finally, I'd like to share this point of view.  When accessing a ref, you are accessing a regular DOM node.  You've left React and the Virtual DOM.  Odds are, you are doing so to manage focus or measure the size/position of the node in the DOM.  That is because React does not yet have a model for focus nor size/position (though [Sebastian Markbåge says they are thinking of ways to handle focus](https://github.com/facebook/react/issues/4936#issuecomment-179916667)).

I'm starting to think it makes more sense to just use `querySelector()` to locate your node.  My thought process on this is:
- you have left React to do something React doesn't support
- you are going to use regular DOM APIs to do this thing, not React (focus or measure)
[...]

PR #571, opened 2016-09-30
feat(Modal): Enable usage without wrapper, add custom Portal

jeffcarbs:
In an issue (#553) I had suggested we rely more heavily on `react-portal`. After digging into it a bit more, I realized a few things:
- It's not very actively maintained. The maintainers haven't commented on a PR in weeks if not months.
- Didn't seem like they had an interest in adding much more. In order to make use of this component for other components (e.g. `Popup`) we may need more flexibility.
- There's not a ton of code to get "Portal" functionality.

Therefore, I decided to build our own `Portal` loosely based on `react-portal`. I used their test file as a starting point to ensure we didn't lose much parity. I put it in the "addons" for now but let me know if there's a better spot for it. It has some cool features:
- configurable `closeOnDocumentClick` and `closeOnEscape` props
- `trigger` props that lets you render a node (e.g. a button) where you define the Portal so you don't need a wrapping div.
- configurable `openOnTriggerClick` and `openOnTriggerMouseOver` props. There's no `closeOnTriggerMouseLeave` but I figured something like that would be useful for `Popup`/`tooltips`.
- It's more declarative than `react-portal`. The `open`/`close` methods just set state and then the portal renders based on that.

---

I updated `Modal` to work with the new `Portal` component. There are still a few things that I need to solidify and/or figure out:
- `Modal` now takes a `portal` prop, which are props to be passed to the `Portal`. It's not a shorthand because the `Portal` is not rendered like a regular element. The other option here would be to pass the props directly to the `Modal` and use something like `_.pick(rest, _.keys(Portal.propTypes))` to figure out what to pass to `Portal`. Here's why I prefer the one prop:
  - We can theoretically allow passing null to not render the portal (e.g. if the use wants a modal but wants to handle hiding/showing it some other way).
  - It goes farther in decoupling the modal from the portal.
  - The negative is that it's slightly more verbose: `<Modal portal={{ trigger: <Button>Long Modal</Button> }}>` vs `<Modal trigger={<Button>Long Modal</Button>}>`
- `Portal` is an autocontrolled component which changes the existing usage of `Modal` slightly.
  - Uncontrolled - you can specify a trigger and have it open on click/hover and then configure the modal to close via clickaway or escape. You can pass `onOpen` and `onClose` to the modal which will be called when the modal is opened or closed. Issue:
    - Ideally you'd be able to close the modal by interacting with some content (e.g. a button) within the modal in the uncontrolled case. Can that be possible somehow? 
  - Controlled - you can specify open/close state but since `onOpen` and `onClose` are only called when the modal is actually opened/closed, there are no hooks to actually control them from the outside.
    - I'm thinking that if the modal would have closed in the uncontrolled state (e.g. press escape and have `closeOnEscape` set) we fire a `onCloseAttempt` (better name?) callback. If it would've opened, we fire the `onOpenAttempt`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "classnames": "^2.1.5",
     "debug": "^2.2.0",
     "lodash": "^4.6.1",
-    "react-dom": "^15.3.1",
     "semantic-ui-css": "^2.2.2"
   },
   "devDependencies": {
@@ -109,6 +108,7 @@
     "react-addons-test-utils": "^15.3.1",
     "react-docgen": "^2.10.0",
     "react-document-title": "^2.0.2",
+    "react-dom": "^15.3.1",
     "react-hot-loader": "^1.3.0",
     "react-router": "^2.7.0",
     "react-transform-catch-errors": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "classnames": "^2.1.5",
     "debug": "^2.2.0",
     "lodash": "^4.6.1",
-    "react-portal": "^2.2.1",
+    "react-dom": "^15.3.1",
     "semantic-ui-css": "^2.2.2"
   },
   "devDependencies": {
@@ -109,7 +109,6 @@
     "react-addons-test-utils": "^15.3.1",
     "react-docgen": "^2.10.0",
     "react-document-title": "^2.0.2",
-    "react-dom": "^15.3.1",
     "react-hot-loader": "^1.3.0",
     "react-router": "^2.7.0",
     "react-transform-catch-errors": "^1.0.2",
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -1,4 +1,5 @@
-import React, { Children, cloneElement, PropTypes } from 'react'
+import _ from 'lodash'
+import React, { Children, PropTypes } from 'react'
 import ReactDOM from 'react-dom'
 
 import {
@@ -85,7 +86,7 @@ class Portal extends Component {
     }
   }
 
-  componentDidUpdate(prevProps, prevState) { // eslint-disable-line complexity
+  componentDidUpdate(prevProps, prevState) {
     // NOTE: Ideally the portal rendering would happen in the render() function
     // but React gives a warning about not being pure and suggests doing it
     // within this method.
@@ -111,6 +112,7 @@ class Portal extends Component {
 
   closeOnDocumentClick = (e) => {
     if (!this.props.closeOnDocumentClick) return
+    if (this.portal.contains(e.target)) return
 
     debug('closeOnDocumentClick()')
 
@@ -132,22 +134,25 @@ class Portal extends Component {
   // Component Event Handlers
   // ----------------------------------------
 
-  handleChildClick = (e) => {
-    debug('handleChildClick()')
-
-    // Prevent closeOnDocumentClick
-    e.nativeEvent.stopImmediatePropagation()
-  }
-
   handleTriggerBlur = (e) => {
-    if (!this.props.closeOnTriggerBlur) return
+    const { trigger, closeOnTriggerBlur } = this.props
+
+    // Call original event handler
+    _.invoke(trigger, 'props.onBlur', e)
+
+    if (!closeOnTriggerBlur) return
 
     debug('handleTriggerBlur()')
     this.close()
   }
 
   handleTriggerClick = (e) => {
-    if (!this.props.openOnTriggerClick) return
+    const { trigger, openOnTriggerClick } = this.props
+
+    // Call original event handler
+    _.invoke(trigger, 'props.onClick', e)
+
+    if (!openOnTriggerClick) return
 
     debug('handleTriggerClick()')
 
@@ -156,21 +161,36 @@ class Portal extends Component {
   }
 
   handleTriggerFocus = (e) => {
-    if (!this.props.openOnTriggerFocus) return
+    const { trigger, openOnTriggerFocus } = this.props
+
+    // Call original event handler
+    _.invoke(trigger, 'props.onFocus', e)
+
+    if (!openOnTriggerFocus) return
 
     debug('handleTriggerFocus()')
     this.open()
   }
 
   handleTriggerMouseLeave = (e) => {
-    if (!this.props.closeOnTriggerMouseLeave) return
+    const { trigger, closeOnTriggerMouseLeave } = this.props
+
+    // Call original event handler
+    _.invoke(trigger, 'props.onMouseLeave', e)
+
+    if (!closeOnTriggerMouseLeave) return
 
     debug('handleTriggerMouseLeave()')
     this.close()
   }
 
   handleTriggerMouseOver = (e) => {
-    if (!this.props.openOnTriggerMouseOver) return
+    const { trigger, openOnTriggerMouseOver } = this.props
+
+    // Call original event handler
+    _.invoke(trigger, 'props.onMouseOver', e)
+
+    if (!openOnTriggerMouseOver) return
 
     debug('handleTriggerMouseOver()')
     this.open()
@@ -182,11 +202,19 @@ class Portal extends Component {
 
   open = () => {
     debug('open()')
+
+    const { onOpen } = this.props
+    if (onOpen) onOpen()
+
     this.trySetState({ open: true })
   }
 
   close = () => {
     debug('close()')
+
+    const { onClose } = this.props
+    if (onClose) onClose()
+
     this.trySetState({ open: false })
   }
 
@@ -197,13 +225,9 @@ class Portal extends Component {
 
     this.node.className = className
 
-    const child = cloneElement(Children.only(children), {
-      onClick: this.handleChildClick,
-    })
-
     this.portal = ReactDOM.unstable_renderSubtreeIntoContainer(
       this,
-      child,
+      Children.only(children),
       this.node
     )
   }
@@ -211,11 +235,10 @@ class Portal extends Component {
   mountPortal = () => {
     if (this.node) return
 
+    this.open()
+
     this.node = document.createElement('div')
     document.body.appendChild(this.node)
-
-    const { onOpen } = this.props
-    if (onOpen) onOpen()
 
     document.addEventListener('keydown', this.closeOnEscape)
     document.addEventListener('click', this.closeOnDocumentClick)
@@ -224,14 +247,13 @@ class Portal extends Component {
   unmountPortal = () => {
     if (!this.node) return
 
+    this.close()
+
     ReactDOM.unmountComponentAtNode(this.node)
     document.body.removeChild(this.node)
 
     this.node = null
     this.portal = null
-
-    const { onClose } = this.props
-    if (onClose) onClose()
 
     document.removeEventListener('keydown', this.closeOnEscape)
     document.removeEventListener('click', this.closeOnDocumentClick)
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -4,9 +4,6 @@ import ReactDOM from 'react-dom'
 
 import {
   AutoControlledComponent as Component,
-  // customPropTypes,
-  // getElementType,
-  // getUnhandledProps,
   keyboardKey,
   makeDebugger,
   META,
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -36,6 +36,9 @@ class Portal extends Component {
     /** Controls whether or not the portal should close when escape is pressed is displayed. */
     closeOnEscape: PropTypes.bool,
 
+    /** Controls whether or not the portal should close when when mousing out of the trigger. */
+    closeOnTriggerMouseLeave: PropTypes.bool,
+
     /** Initial value of open. */
     defaultOpen: PropTypes.bool,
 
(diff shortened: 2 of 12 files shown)


jeffcarbs:
@levithomason @layershifter - would appreciate some thoughts on the above, specifically the open issues 😄

typehorror:
Hey @jcarbo , I'm facing an issue when using the `openOnTriggerFocus` where the click that provides the focus might also be capture by the click listener in charge of closing the portal. We might want to make sure closing clicks on the trigger are ignored.

levithomason:
Once again, thanks much.  I wanted to do exactly this ever since adding the dep 👍 

## Modal portal prop

We could also consider providing the Modal separately from the Portal and requiring the usage to include both the Portal and the Modal.  This way, the Modal is a static component.  Same with the Popup.  This inspired by react-bootstrap's [react-overlays](https://react-bootstrap.github.io/react-overlays/).  It makes it possible to use any portal configuration with any component.

I think I prefer more succinct APIs, where the Modal "just works".  So, we could bundle it but provide `StaticModal` or something for the static one.  Thinking out loud...

## Open Issues

> Ideally you'd be able to close the modal by interacting with some content (e.g. a button) within the modal in the uncontrolled case. Can that be possible somehow?

I think we do this for specialized modals such as the Confirm.  Otherwise, it is minimal wiring for the user if they want more control.  I say this as it deals with modifying children (adding click handlers).  The only sane way I think we should do that is via props.  This then means if the user passes us some button, we can insert that into children, but they cannot then use children.  This is almost exactly what the Confirm modal does.  I think any use case beyond that and they can build their own Modal implementation, for now.

> I'm thinking that if the modal would have closed in the uncontrolled state (e.g. press escape and have closeOnEscape set) we fire a onCloseAttempt (better name?) callback. If it would've opened, we fire the onOpenAttempt.

I've seen `request`/`attempt` prop names around, but I think they are not inline with React.  Consider the `<input onChange={} />`, it is fired regardless of whether or not the value will actually update.  So, I think change handlers should always fire when _the user tries to change_ the state.  The value to the change handler should always be the new proposed value IMO.  Whether or not that value was applied to the value prop or not (controlled/uncontrolled) makes no difference.

We had to deal with this exact issue in the Dropdown via [`this.onChange()`](https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/Dropdown.js#L360).  We [call this every time](https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/Dropdown.js#L419) there is a user prompted change attempt such as a tab, enter, click, etc.  We manually call this as there is no guarantee `trySetState` will be successful.

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+
+  static _meta = _meta
+
+  componentDidMount() {
+    if (this.state.open) {
+      this.renderPortal()
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) { // eslint-disable-line complexity
```
levithomason: Very few branches in this method, is `// eslint-disable-line complexity` still needed?

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+
+  // ----------------------------------------
+  // Component Event Handlers
+  // ----------------------------------------
+
+  handleChildClick = (e) => {
+    debug('handleChildClick()')
+
+    // Prevent closeOnDocumentClick
+    e.nativeEvent.stopImmediatePropagation()
```
levithomason: I think instead of stopping all clicks, the closeOnDocumentClick should check if the click was inside of the Modal or not before closing.  This way we don't mess with the events, we just react to them appropriately.

We used `node.contains` before, which has some support issues.  There's this better idea from react-bootstrap, [isNodeInRoot](https://github.com/bbc/react-bootstrap/blob/0cfbf3be2a7b53c1a581b26ef7831388264cd2b4/src/DropdownStateMixin.js#L5).  Might be more functional (compose with different roots) and make more sense to reorder the args and rename it though:

``` js
if (rootContainsNode(findDOMNode(this), e.target) return

// otherwise, the click target was outside `this` component's node
```

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+    this.open()
+  }
+
+  // ----------------------------------------
+  // Behavior
+  // ----------------------------------------
+
+  open = () => {
+    debug('open()')
+    this.trySetState({ open: true })
```
levithomason, line 220: Propose we call `onOpen` just before `trySetState` here, so the user has opportunity to control this component.

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  close = () => {
+    debug('close()')
+    this.trySetState({ open: false })
```
levithomason, line 229: Sam, propose we call `onClose` just before `trySetState` here, so the user has opportunity to update any controlled `open` state.

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+      child,
+      this.node
+    )
+  }
+
+  mountPortal = () => {
+    if (this.node) return
+
+    this.node = document.createElement('div')
+    document.body.appendChild(this.node)
```
levithomason: This was one of the issues I had with the portal, no way to configure where it renders to.  How about making `document.body` configurable and defaulting it to `document.body`?.  Perhaps:

``` jsx
<Portal mountNode={/* a DOM node */} />
```

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+
+  handleChildClick = (e) => {
+    debug('handleChildClick()')
+
+    // Prevent closeOnDocumentClick
+    e.nativeEvent.stopImmediatePropagation()
+  }
+
+  handleTriggerBlur = (e) => {
+    if (!this.props.closeOnTriggerBlur) return
```
levithomason: Since these are all overwriting the `trigger` props when cloned in the render, we need to also call the trigger handler if it exists, before we exit.  Probably add some tests that ensure this is the case.

Lodash's [`_.invoke`](https://lodash.com/docs/4.16.2#invoke) handles this gracefully:

``` jsx
handleTriggerBlur = (e) => {
  const { trigger, closeOnTriggerBlur } = this.props

  _.invoke(trigger, 'onBlur', e)
  if (!closeOnTriggerBlur) return
  // ...
```

src/addons/Portal/Portal.js
```diff
@@ -0,0 +1,255 @@
...
+    if (!this.node) return
+
+    ReactDOM.unmountComponentAtNode(this.node)
+    document.body.removeChild(this.node)
+
+    this.node = null
+    this.portal = null
+
+    const { onClose } = this.props
+    if (onClose) onClose()
```
levithomason: I don't think we should only call event handlers if they are triggered by direct user interactions, but not by programmatic changes.  This is consistent with vanilla HTML and vanilla React handling of change events:

http://codepen.io/levithomason/full/YGAXEb/

src/modules/Modal/Modal.js
```diff
@@ -84,52 +78,31 @@ class Modal extends Component {
...
 
-    document.removeEventListener('keydown', this.handleDocumentKeyDown)
-    document.removeEventListener('click', this.handleClickOutside)
+    const { onClose } = this.props
+    if (onClose) onClose()
   }
 
-  handleShow = () => {
-    debug('handleShow()')
+  handleOpen = () => {
```
levithomason: 👍 thanks for the rename, open/close makes much more sense.

src/addons/Portal/Portal.js
typehorror: I second this request, the popup might need to appear on a different component than body (for Fluid popup). Something to consider is how to handle the case where the container where to get dynamically destroyed.

src/addons/Portal/Portal.js
typehorror: Is there anyway we could consistently pass the event (if any) responsible for closing and opening to onClose and onOpen callback ?

levithomason: Yes sir, this is another reason to only use user events to trigger handlers.  Programmatic changes do not have reliable events to pass back.

We should call the corresponding `trigger` event handler in `handleTriggerClick` `handleTriggerFocus` `handleTriggerMouseLeave` and `handleTriggerMouseOver` and pass the event.  Noted above in [this comment](https://github.com/TechnologyAdvice/stardust/pull/571/files#r81428230).

We should pass the event from each of those to the `open`/`close` method as well so that it can be passed to the `onOpen` and `onClose` callbacks.  This way, all callbacks get all the events possible.

package.json
```diff
@@ -49,7 +49,7 @@
     "classnames": "^2.1.5",
     "debug": "^2.2.0",
     "lodash": "^4.6.1",
-    "react-portal": "^2.2.1",
+    "react-dom": "^15.3.1",
```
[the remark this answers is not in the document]
jeffcarbs: `ReactDOM` is required by the `Portal` component:

``` js
    this.portal = ReactDOM.unstable_renderSubtreeIntoContainer(
      this,
      child,
      this.node
    )
```

Is there a way we can get around that?

src/addons/Portal/Portal.js
jeffcarbs: `findDOMNode` would also require us to include `react-dom` in the production build. It also seems there's momentum against using it (see https://github.com/yannickcr/eslint-plugin-react/issues/678). Thoughts on just keeping a ref instead?

So, `if (rootContainsNode(this.portalChild, e.target) return`

package.json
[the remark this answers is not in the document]
levithomason: Hm, thinking it through I don't believe we can get away without it.  Though, we can at least remove it from deps and just import it.  It is listed in `peerDependencies` with a range of supported versions.  This way, users can bring their own react / react-dom and we just import it, rather than bundling it (deps).

I'm not versed on the unstable API, but I wonder why the original portal didn't use `ReactDOM.render()` and just treat it as it's own app.  I should probably watch his talk referenced in the README, https://www.youtube.com/watch?v=z5e7kWSHWTg. :/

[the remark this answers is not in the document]
jeffcarbs: Ah, didn't realize it was already in peer-dependencies! I'll move this back.

src/addons/Portal/Portal.js
levithomason: Definitely, thanks for the reminder.  I'm on board with dropping findDOMNode.  You probably already know to use the callback form, see #405.

jeffcarbs: Ah actually, the DOM node is the return value of `renderSubtreeIntoContainer`, which we're storing as `this.portal`. So we already have a reference to it.

I'm using `this.portal.contains(e.target)` to check if the click is within, which is working just fine. Looks like support is pretty extensive: https://developer.mozilla.org/en-US/docs/Web/API/Node/contains. but lmk if you think we should use that other function you referenced above though.

levithomason: I prefer `contains` if possible, let's keep on!  We can always change based on bug reports, if ever any.

jeffcarbs:
Ok made some updates:
- `onOpen`/`onClose` now fire when the portal is initially opened/closed AND whenever anything responsible for
- fixed an issue with `openOnTriggerFocus` where clicking the trigger would immediately close it (cc @debrice)
- invoke the trigger's event handlers (e.g. `onClick`) within the callbacks we're hijacking.

Still a few left to go:
- [x] Pass the triggering event as an argument to `onOpen`/`onClose`
- [x] Custom `mountNode` prop.
- [x] Clean up portal prop handling (see below)
- [x] Deal with `Modal` `handleOpen`/`handleClose` (see below)

Questions:
- For `mountNode`, should that be a DOM node? If so, is there a propType for that? Otherwise, should it be a selector? I'm kinda leaning DOM node to keep things more sane.
- Where did we land with `portal` props? Just pass them directly to the `Modal` and have a method to pick off the relevant ones (kinda like we do for `inputProps` inside of the `Input` component)?
- Related to the portal props, I think coupling the modal + portal is the way to go for now. We can decide on some sort of `Modal.Static`concept on a separate PR if that's cool. Esp given the Popup is relying on this, I don't want this to get held up too much.
- We're now calling `onOpen` and `onClose` always, even when nothing happened (in the controlled scenario). There's still code within `Modal` that needs to be called to add/remove the `dimmer`-related classed to the body (see https://github.com/TechnologyAdvice/stardust/pull/571/files#diff-e762b5cc3074c8997386b6de83cac7fbR102). That's currently relying on those callbacks as well, so I'm not sure how to have those only run when the modal is shown/removed. One thought is `onMount`/`onUnmount` callbacks that are called within `mountPortal` and `unmountPortal` since those are only called when the portal is actually opened/closed.

Thoughts?

jeffcarbs:
Ok, wound up going with the `onMount` and `onUnmount` which wound up making more sense I think:
- `onOpen` is called on every event that would trigger an open based on your props (e.g. `openOnTriggerFocus`)
- `onClose` is called on every event that would trigger a close based on your props (e.g. `closeOnEscape`)
- `onMount` is called when the portal mounts onto the DOM
- `onUnmount` is called when the portal unmounts from the DOM

The `Modal` needs to listen for the `mount`/`unmount` to handle adding/removing dimmer classes.

The one thing left to do is figure out how to handle portal props. I think I'm going to just remove the `portal` prop and just pass-through any portal-related props from the modal to the portal.

jeffcarbs:
I fixed up all the specs and examples and the API feels pretty good. Passing the props directly to `Modal` and having the props passed-through to `Portal` is definitely the way to go. So much cleaner than passing an object as a `portal` prop.

@levithomason - I think this is ready for final review 🙌

levithomason:
Made a few trivial cleanup commits and rebased/fix conflicts with latest master.  Will merge on pass.  I'll be out, so will release later tonight!

levithomason (approved):

levithomason:
... Or I'll just do it now. Released in `stardust@0.50.0`

PR #575, opened 2016-10-01
fix(Feed): update propTypes, shorthand and docs

levithomason:
This PR pulls the Feed updates from #452.  Many updates were made:
- use `content` prop for main content of Feed Date/Meta/Like/etc.
- fixed many prop type validations
- split shorthand examples into separate files, added missing shorthand examples
- renamed example files to `FeedExampleFoo` and `FeedExampleFooShorthand`


levithomason:
Just need coverage for one line, then we're good here!

image

levithomason:
Released in `stardust@0.51.0`

PR #576, opened 2016-10-01
feat(Label): add LabelGroup and missing docs

layershifter:
This PR:
- adds all doc examples from [SUI](http://semantic-ui.com/elements/label.html);
- adds `Group` component;
- cleanups component and tests;
- improves props handling.

Requires merge of #551.


layershifter:
@levithomason @jcarbo I need review there

docs/app/Examples/elements/Label/Types/LabelExampleAttached.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { Grid, Image, Label, Segment } from 'stardust'
+
+const imageStyles = {
```
levithomason: Are these style required? I'd like to avoid any styles in the docs. It would seem that SUI alone should handle all styling.

docs/app/Examples/elements/Label/Types/LabelExampleRibbon.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { Grid, Image, Label, Segment } from 'stardust'
+
+const imageStyles = {
```
levithomason: Same regarding styles here, any way to remove them?

docs/app/Examples/elements/Label/Variations/LabelExampleCircular.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Label } from 'stardust'
+
+const LabelExampleCircular = () => {
+  const colors = [
```
levithomason: When we have data for examples, let's put it up top outside the scope of the example itself.

docs/app/Examples/elements/Label/Variations/LabelExampleSize.js
```diff
@@ -0,0 +1,19 @@
+import _ from 'lodash'
+import React from 'react'
+import { Label } from 'stardust'
+
+const LabelExampleSize = () => {
+  const sizes = ['mini', 'tiny', 'small', 'medium', 'large', 'big', 'huge', 'massive']
```
levithomason: More data here, let's move it up top. There are several other examples with data to update so I won't comment on all of them.

levithomason (changes requested):
Looks great, left a few minor comments.

docs/app/Examples/elements/Label/Types/LabelExampleAttached.js
layershifter: If I remove them, example will look ugly 😿

levithomason: I don't think we should ever include styles in the examples.  I think it gives the impression that we require inline styles to be used for the example to work.  If we truly _must_ include styles, there are doc site css rules in the `index.ejs` template for this, but let's not put them in the JSX.

In this case, I think there is little enough difference that we can just leave them out:



Alternatively, we could use `<Segment padded>` to give whitespace similar to what the inline image styles did:

docs/app/Examples/elements/Label/Variations/LabelExampleSize.js
jeffcarbs: Actually, I think better would be to just use `Label._meta.props.size`, this way the sizes we give examples for are all the sizes the component supports. There are a few other places in the docs we should update to do this as well.

docs/app/Examples/elements/Label/Variations/LabelExampleCircular.js
jeffcarbs: Same comment re: sizes, should use `Label._meta.props.color`

docs/app/Examples/elements/Label/Variations/LabelExampleSize.js
levithomason: I don't want to expose the private _meta property. It may change or completely go away in the future.

levithomason: More info, meta used to have more keys and have more purpose. Over time, it keeps shrinking. Considering the parent, type, and name keys are only for documentation, I hope to pull them out of the code and into the doc block. We already parse the docblock for docs so grabbing type, parent, and name from there makes sense. 

This leaves only the props in the meta. In farther out, less defined goals, I'd like to better formalize how we define props to support the missing abilities of the component explorer #427.

layershifter:
Updated to `master` and fixed comments.

levithomason (approved):

levithomason:
Released in `stardust@0.51.1`

Issue #449, opened 2016-08-29
Standardize childKey and shorthand props


layershifter:
It seems, that [here](https://github.com/TechnologyAdvice/stardust/blob/master/src/views/Card/CardGroup.js#L30) is bug.

``` jsx
const content = !items ? children : items.map(item => {
  const key = item.key || [item.header, item.description].join('-')
  return <Card key={key} {...item} />
})
```
- `key` prop will be always overwritten with `item` contents;
- component uses `key` prop instead of `childKey`.

It can be rewritten as:

```
if(!children) {
  return <ElementType {...rest} className={classes}>{children}</ElementType>
}

const content = items.map(item => {
  const {childKey, ...itemProps} = item
  const finalKey = childKey || [item.header, item.description].join('-')

  return <Card {...item} key={childKey} />
})
```
- `key` cannot be `number`, I think we need allow this:

``` jsx
childKey: PropTypes.oneOfType([
  PropTypes.number,
  PropTypes.string,
]),
```

---

Also, I think there is problem with keys in all the components and current solution with `childKey` is little evil.

levithomason:
> key prop will be always overwritten with item contents;

It is only overridden if it exists, but the assignment already assigns it if exists.  So, if there is no `key` the header/description is used, if there is a `key` then that is used.

``` js
const key = item.key || ...
```

> component uses key prop instead of childKey.

Agreed, we should update this.

> It can be rewritten as:

For clarity and consistency, it should have likely been written:

Did you mean `if (children)` instead of `if (!children)`?  Otherwise, it will render the markup that uses children when there are no children.

Also, `items.map()` will fail if there are not items and no children.  Suggest using `_.map(items)` which handles the maybe case and always returns an array.

> key cannot be number, I think we need allow this:

Why not?  Index can be passed which is a number.  Also, if a user is using some kind of numerical id key mechanism, we would want to allow that.  Let me know if there is a bug in doing so or some documentation I am not aware of on this.

> Also, I think there is problem with keys in all the components and current solution with childKey is little evil.

I'd like to hear the evil parts.  If you have any suggestions, those would be great as well!

layershifter:
> Did you mean if (children) instead of if (!children)?
> Suggest using _.map(items) which handles the maybe case and always returns an array.

`!items`, so, it will not fail :smile: 

> Why not?

[...]

PR #579, opened 2016-10-01
fix(customPropTypes): fix logic and warning messages

levithomason:
This is a breakout of #449.

This PR fixes a number of issues with customPropTypes.  Typos, bad messages, issues with logic, etc.  I also reworked them to use `lodash/fp` which cleaned things up a bit.

Later, we'll add the additional propTypes for shorthand, icon, image, children, etc.


levithomason:
Released in `stardust@0.49.0`

PR #580, opened 2016-10-02
fix(package): remove semantic-ui-css dependency

levithomason:
I believe this was left over from when we used to bundle jquery plugins. We don't bundle any Semantic packages with ours.


levithomason:
Released in `stardust@0.51.1`

PR #581, opened 2016-10-02
feat(Dropdown): Pass object with name to onChange

### LINKED ISSUE #218: All event handlers should return the event data first

jeffcarbs:
Fixes #218 

This is a breaking change but I think it's worth it. There's currently no easy way to get the `name` of the `Dropdown` in the `onChange` event. Normally in an `onChange` the `event.target` will be the input that received the change. So in a form, for example, you can attach the same change handler to multiple elements and use the event to determine the appropriate field/value to update.

For `Dropdown` that's not the case - it will likely be a `DropdownItem`.

The `Checkbox` is an example of a component that faces a similar issue regarding the target. In that component, the `onClick` and `onChange` are both called with `{ name, value, checked }`. This PR updates `Dropdown` to be called with `{ name, value }`.

I also cleaned up the naming of our internal event handler to be `handleChange`. The pattern of this project seems to be:
- `handle<event>` for our internal handlers
- `on<event>` is a prop that the user passes in and is called by our internal handlers if set.

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Types/AllowAdditions.js
+++ b/docs/app/Examples/modules/Dropdown/Types/AllowAdditions.js
@@ -12,23 +12,23 @@ const options = [
 export default class AllowAdditionsExample extends Component {
   state = { optionsSingle: options, optionsMultiple: options }
 
-  handleAdditionSingle = value => {
+  handleAdditionSingle = (e, { value }) => {
     // validate, save to server, show a message to the user ...
     this.setState({
       optionsSingle: [{ text: value, value }, ...this.state.optionsSingle],
     })
   }
 
-  handleAdditionMultiple = value => {
+  handleAdditionMultiple = (e, { value }) => {
     // validate, save to server, show a message to the user ...
     this.setState({
       optionsMultiple: [{ text: value, value }, ...this.state.optionsMultiple],
     })
   }
 
-  handleChangeSingle = (e, value) => this.setState({ currentValue: value })
+  handleChangeSingle = (e, { value }) => this.setState({ currentValue: value })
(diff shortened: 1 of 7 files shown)


jeffcarbs:
This also uncomments the test referenced in https://github.com/TechnologyAdvice/stardust/issues/218. I grepped for all of our event handlers and it seems that they are all called with the event first so I think that can be closed with this PR.

jeffcarbs:
Actually realized that `onAddItem` wasn't receiving an event first. This updates the arguments so the event is passed first then `{ name, value }`.

So two breaking changes in this PR, both in `Dropdown`:

``` diff
-onChange(e, value)
+onChange(e, { name, value })

-onAddItem(value)
+onAddItem(e, { name, value })
```

levithomason:
Cool, looks good. I think we should always pass `(e, data)` where data is an object.  Just to keep it consistent and extensible.

I'll try to review/merge/ship this tonight.

levithomason (approved):

PR #583, opened 2016-10-02
fix(tests): Add coverage to missing lines

jeffcarbs:
It was really bothering me that our coverage was at ~99.6% and not just 100%. I went through codecov.io and found the lines that were missing coverage and added specs. Let's not let it dip back below 100!

On a related note, I realized recently that the one-liner style for conditionals actually masks missing coverage. For example:

``` js
open = (e) => {
  debug('open()')

  const { onOpen } = this.props
  if (onOpen) onOpen(e)

  this.trySetState({ open: true })
}
```

Codecov won't catch if we're missing a test to ensure `onOpen` gets called.

You can enable an eslint rule to force multi-line blocks (http://eslint.org/docs/rules/curly), although that would increase the verbosity of the codebase. Not necessarily advocating for doing anything, just wanted to bring this up.

### DIFF
--- a/test/specs/collections/Form/Form-test.js
+++ b/test/specs/collections/Form/Form-test.js
@@ -51,6 +51,30 @@ describe('Form', () => {
       serializer.should.have.been.calledWithMatch(formNode)
     })
 
+    it('logs an error if an input is missing a name', () => {
+      consoleUtil.disableOnce()
+      const spy = sandbox.spy(console, 'error')
+
+      const wrapper = mount(
+        <Form onSubmit={() => null}>
+          <input type='checkbox' data-has-no-name />
+        </Form>
+      )
+        .simulate('submit')
+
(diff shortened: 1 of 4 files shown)


levithomason (approved):
Very cool, been waiting a long while for this. 

FYI you can also open coverage/lcov/index.js to peruse coverage more quickly. It shows coverage per directory with links to drill down per file. Lots of other features, worth checking out! 

Also, all PRs fail if they do not maintain or increase coverage both for the project as a whole and also for the specific diff the PR touches. So, since adding codecov we've only ever had increasing coverage. Now, it will never be below 100% because you literally cannot merge a PR to lower it :D

I'm all for having true coverage reporting. I think we should definitely add the eslint rule.  Let's plan that for separate PRs as each refactor may require more test coverage.

PR #584, opened 2016-10-02
feat(Comment): Add component

### LINKED ISSUE #187: <Comment /> Component
http://semantic-ui.com/views/comment.html

layershifter:
Adds [**Comment**](http://semantic-ui.com/views/comment.html) component.

Fixes #187.


layershifter:
@levithomason I added first doc example, LGTM.

docs/app/Examples/views/Comment/Content/CommentExampleReplyForm.js
```diff
@@ -0,0 +1,29 @@
...
+      <Comment.Content>
+        <Comment.Author as='a'>Steve Jobes</Comment.Author>
+        <Comment.Metadata>
+          <div>2 days ago</div>
+        </Comment.Metadata>
+        <Comment.Text>Revolutionary!</Comment.Text>
+        <Comment.Actions>
+          <Comment.Action active>Reply</Comment.Action>
+        </Comment.Actions>
+        <Comment.ReplyForm>
```
levithomason: I see the `reply` class is required for the form when used in a comment or after all comments.  I think we should explore some other API options for adding the class.

This pattern reminds me of the `label` and `button` class being required for Inputs that are "labeled` or have an "action".  In the Input case the label/action is passed through props, so we have an opportunity to add the className without requiring additional components or props to add the class.

Ideas:

### Form `reply` prop

We could simply add a `reply` prop to the Form.  It would clean this up and clarify things.  However, this prop is not applicable anywhere except in comments, so it seems sub optimal.

### Comment `replyForm` prop

I think this one makes more sense since you'll want the same form in all comments.  Rather than pasting a reply form in every comment and one at the end of all comments, we could create a single comment reply form and pass it as a prop to each comment:

``` jsx
const replyForm = <Form />

<Comment.Group replyForm={replyForm} />
// and/or
<Comment replyForm={replyForm} />
```

### Looping Comments

This brings me to another realization, I think the actual React usage of the Comment will be to create a single comment component, query a server for a list of comments, then loop over the comments mapping the single comment component.  Because of this, I don't think most users will ever build out a comment thread in a hard coded sense.

That would put the Form `reply` prop back on the table since there would likely only be one comment component in your project (reply form included) that you would map over your comment data.

Feedback?

layershifter: We have same situation with `List.Icon`, if we add `reply` prop on `Form`, seems we'll need to remove `List.Icon`.

@jcarbo ?

levithomason: I'm really torn on this one.  While I agree entirely on the need for consistency with List.Icon, at the same time I really want to avoid this API:

``` jsx
<Comment.ReplyForm>
  <Form.TextArea />
  <Button labeled icon primary>
    <Icon name='edit' />
    Add Reply
  </Button>
</Comment.ReplyForm>
```

I think it comes off very confusing IMHO.  It would be great to get a single top level interface to creating a Comment form some how.

layershifter: In this way, `reply` prop looks more reliable solution.

layershifter:
I've finished docs, tomorrow I'll add tests and finalize component.

What is the decision for `Comment.ReplyForm`? We add `reply` prop to form?

levithomason:
Let's go with the `reply` prop on Form for now.  We can revisit later if necessary.

layershifter:
Added test, seems we need review there :+1: 
Pay attention to `CommentAvatar`, not sure that there is a good solution.

Also, there are no shorthands, I think we can do them in sepatate PR.

levithomason:
Agreed on separate PR for shorthand, I'll get a review here soon as I can.

docs/app/Examples/views/Comment/Variations/CommentExampleThreaded.js
```diff
@@ -0,0 +1,75 @@
...
+          <span>5 days ago</span>
+        </Comment.Metadata>
+        <Comment.Text>Dude, this is awesome. Thanks so much</Comment.Text>
+        <Comment.Actions>
+          <a>Reply</a>
+        </Comment.Actions>
+      </Comment.Content>
+    </Comment>
+
+    <Comment.ReplyForm>
```
levithomason: This component no longer exists, docs fail to load at present:

```
Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `CommentExampleThreaded`.

CommentExampleThreaded.js?42c9:180
```

levithomason (changes requested):
See comment on CommentExampleThreaded.js, will finish review after that.

layershifter:
Docs again with us.

docs/app/Examples/views/Comment/Content/CommentExampleReplyForm.js
```diff
@@ -0,0 +1,26 @@
...
+      <Comment.Content>
+        <Comment.Author as='a'>Steve Jobes</Comment.Author>
+        <Comment.Metadata>
+          <div>2 days ago</div>
+        </Comment.Metadata>
+        <Comment.Text>Revolutionary!</Comment.Text>
+        <Comment.Actions>
+          <Comment.Action active>Reply</Comment.Action>
+        </Comment.Actions>
+        <Form reply>
```
levithomason: To prevent page refresh on submit, let's add:

``` jsx
onSubmit={e => e.preventDefault()}
```

docs/app/Examples/views/Comment/Content/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+    <ComponentExample
+      title='Actions'
+      description='A comment can contain an list of actions a user may perform related to this comment'
+      examplePath='views/Comment/Content/CommentExampleActions'
+    />
+
+    <ComponentExample
+      title='Reply Form'
+      description='A comment can contain a form to reply to a comment. This may have arbitrary content'
```
levithomason: Needs a `.` "This may have arbitrary content."

docs/app/Examples/views/Comment/Content/index.js
```diff
@@ -0,0 +1,35 @@
...
+      examplePath='views/Comment/Content/CommentExampleReplyForm'
+    />
```
levithomason: Suggest we also add this `info` Message from the SUI docs:

> If a comment form is located inside a comment it will be formatted as a nested reply form. If the comment form is included after all comments, it will be formatted as a normal reply form.

docs/app/Examples/views/Comment/States/CommentExampleCollapsed.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component } from 'react'
+import { Checkbox, Comment } from 'semantic-ui-react'
+
+export default class CommentExampleCollapsed extends Component {
```
levithomason, line 4: This is a cool example, really like the simple interactivity to show how a prop works.

levithomason (changes requested):
Only a few comments, then I think we can merge this guy.

docs/app/Examples/views/Comment/Content/index.js
layershifter: Stupid question there. Why we need add trailing dot there, but don't add it in other places? :thinking:

docs/app/Examples/views/Comment/Content/index.js
layershifter: Stupid question. Why we need add trailing dot there, but don't add it in other places? :thinking:

layershifter:
Updated to `master`, fixed review comments.

layershifter:
Hm... @levithomason can you restart build? Strange failure there.

levithomason:
Intermittent failures associated with the Search tests, I've seen this a couple times now:

```
  Search Component
    active item
      ✖ moves up on arrow up when open
        PhantomJS 2.1.1 (Linux 0.0.0)
        expected the node in <Search /> to have a 'active' propHTML: Not available due to: null is not an object (evaluating 'this.el.outerHTML')
```

I restarted the build.

I believe since you have access to the repo, you should also have access to the builds at https://circleci.com/gh/Semantic-Org/Semantic-UI-React.

docs/app/Examples/views/Comment/Content/index.js
layershifter: LGTM, I'll add trailing dot to all example descriptions tomorrow, revolutionary improvement :)

layershifter:
Added dots :laughing: Seems, we ready to go

levithomason:
Another one down, thanks!

levithomason:

PR #585, opened 2016-10-02
fix(MenuItem): do not add `icon` class given `name` or `content`

levithomason:
The `icon` class collapses the whitespace.  It should only be added if there is an `icon` only, without any adjacent text.
### Before

image
### After

image


levithomason:
Released in `stardust@0.51.1`

PR #587, opened 2016-10-03
docs(elements): refactor the all components under elements category

dpkwhan:
Rename class name and file name and use functional stateless components when it is possible.


levithomason:
I would prefer to create a PR refactoring one component at a time. This way the PRs are not so large. Smaller PRs are quicker to review and often receive better feedback.

I also notice there is other clean up we should make along the way like removing extra divs and removing eslint overrides. However, making those changes in this PR would be over bearing. Smaller more focused PRs would make this more manageable. 

Would you mind submitting a PR for just one of these components?

dpkwhan:
@levithomason I will do that.

levithomason:
Thanks, much appreciated.

levithomason:
In light of splitting things up, I'm going to close this PR.

PR #588, opened 2016-10-03
docs(Button): rename filename and class name

dpkwhan:


levithomason (approved):
Looks good thanks

levithomason:
In order to keep the changes separate (Button and Container), you'll need to create separate branches with each change, or wait until the Button is merged before committing the Container changes.

I need to wait for tests before merging, but each commit will trigger tests to rerun.

dpkwhan:
@levithomason I will create separate branch for each element. Thanks for the advice.

PR #590, opened 2016-10-03
feat(Portal): More flexible configuration

jeffcarbs:
- Add `closeOnTriggerClick` prop
- Update `closeOnMouseLeave` to handle mouse leave of either trigger or portal
- Add configurable delay for open/close on hover

---

This addresses an open issue of #570 which is not being able to mouse-over from the trigger of a `Popup`.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -34,8 +34,18 @@ class Portal extends Component {
     /** Controls whether or not the portal should close when escape is pressed is displayed. */
     closeOnEscape: PropTypes.bool,
 
+    /**
+     * Controls whether or not the portal should close when mousing out of the portal.
+     * NOTE: This will prevent `closeOnTriggerMouseLeave` when mousing over the
+     * gap from the trigger to the portal.
+     */
+    closeOnPortalMouseLeave: PropTypes.bool,
+
     /** Controls whether or not the portal should close on blur of the trigger. */
     closeOnTriggerBlur: PropTypes.bool,
+
+    /** Controls whether or not the portal should close on click of the trigger. */
+    closeOnTriggerClick: PropTypes.bool,
 
     /** Controls whether or not the portal should close when mousing out of the trigger. */
     closeOnTriggerMouseLeave: PropTypes.bool,
@@ -45,6 +55,12 @@ class Portal extends Component {
 
     /** The node where the portal should mount.. */
     mountNode: PropTypes.any,
+
+    /** Milliseconds to wait before closing on mouse leave */
+    mouseLeaveDelay: PropTypes.number,
+
+    /** Milliseconds to wait before opening on mouse over */
+    mouseOverDelay: PropTypes.number,
 
     /** Called when a close event happens */
     onClose: PropTypes.func,
@@ -110,6 +126,10 @@ class Portal extends Component {
 
   componentWillUnmount() {
     this.unmountPortal()
+
+    // Clean up timers
+    clearTimeout(this.mouseOverTimer)
+    clearTimeout(this.mouseLeaveTimer)
   }
 
   // ----------------------------------------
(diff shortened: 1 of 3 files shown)


jeffcarbs:
@levithomason - this is blocking the `Popup` PR so if you could take a look I'd appreciate it 👍

levithomason:
On it, thanks!

src/addons/Portal/Portal.js
```diff
@@ -37,15 +37,27 @@ class Portal extends Component {
     /** Controls whether or not the portal should close on blur of the trigger. */
     closeOnTriggerBlur: PropTypes.bool,
 
-    /** Controls whether or not the portal should close when mousing out of the trigger. */
-    closeOnTriggerMouseLeave: PropTypes.bool,
+    /** Controls whether or not the portal should close on blur of the trigger. */
```
levithomason: `close on blur` != `closeOnTriggerClick`

src/addons/Portal/Portal.js
```diff
@@ -37,15 +37,27 @@ class Portal extends Component {
...
+    closeOnTriggerClick: PropTypes.bool,
+
+    /**
+     * Controls whether or not the portal should close when mousing out of the
+     * trigger OR the portal content.
+     */
+    closeOnMouseLeave: PropTypes.bool,
```
levithomason: Hm, seems like we should separate these.  I know the names are hideous, but I can see wanting one or the other:

`closeOnTriggerMouseLeave`
`closeOnPortalMouseLeave`

Thoughts?

src/addons/Portal/Portal.js
```diff
@@ -229,6 +273,12 @@ class Portal extends Component {
     this.trySetState({ open: false })
   }
 
+  closeWithTimeout = (e, delay) => {
+    // React wipes certain props (e.g. currentTarget) so we need to clone.
+    const eventClone = { ...e }
```
levithomason, line 292: Heads up, React wipes the entire event object, nullifying all values.  That's because React reuses a single event object.  In order to persist the event object for async access, use `e.persist()`:

https://facebook.github.io/react/docs/events.html#event-pooling

I think what is here will technically work, but it still seems as though we should be using `e.persist()`, even if we're copying the persisted object.

typehorror:
@jcarbo, I pulled your branch into the Popup and noticed that the first time the Popup shows up the timer on leave isn't respected. The bug goes away every time I hover the trigger after that.

levithomason (changes requested):

src/addons/Portal/Portal.js
jeffcarbs: Tried `e.persist` initially but doesn't preserve the currentTarget, see https://github.com/facebook/react/issues/2857. This is kinda gross but it's the only way to make a copy of the event as-is.

typehorror:
@jcarbo Nevermind the issue I reported earlier (aka. timer not respected the 1st time), I can't reproduce it anymore.

src/addons/Portal/Portal.js
jeffcarbs: That was my initial thought, but then I was wondering how do you configure allowing mousing from the trigger to the portal? I think I just figured it out though:
- If only `closeOnTriggerMouseLeave` is set, then mouseover/leave on the portal won't do anything
- If `closeOnPortalMouseLeave` is set, then mouseover of the portal clears the timer, mouseleave restarts the timer.

So `closeOnPortalMouseLeave` has the added effect of preventing close when mousing over the gap from trigger -> portal

This actually lets us keep more closely with SUI's hoverable:
- `hoverable => false`  is `{ closeOnTriggerMouseLeave: true }`
- `hoverable => true`  is `{ closeOnTriggerMouseLeave: true, closeOnPortalMouseLeave: true }`

levithomason: I'm moving quickly here, but I think this makes sense.

jeffcarbs:
Ok made the changes discussed here and checks came back green. I think this is  👍

levithomason:
Released in `stardust@0.52.1`

Issue #596, opened 2016-10-03
labeled icon <Button /> not rendered as <button>


Chris-R3:
### Steps to Reproduce

I'd like to render a `labeled  icon button` with `type='submit'` with following jsx code:

``` jsx
<Button type='submit' icon='save' labeled content='save' />
```
### Expected Behavior

``` html
<button class="ui labeled icon button" type="submit">
  <i class="save icon"></i>
  save
</button>
```
### Actual Behavior

``` html
<div type="submit" class="ui labeled icon button" tabindex="0">
  <i class="save icon"></i>
  save
</div>
```
### Test Case

http://codepen.io/Chlris-R3/pen/xEXBzK?editors=0010

Is this a bug or do I need to change my code somehow? I also tried adding `as='button'` with the same result.

levithomason:
Confirmed bug, thanks much.  @yesmin is preparing the fix in #597.

Chris-R3:
much appreciated
keep up the good work :heart:

levithomason:
Fixed in #597 and released in `stardust@0.52.5`

PR #597, opened 2016-10-03
fix(Button): render `labeled` as a <button>

x-shd:
Fix switch button to div 
#596

### DIFF
--- a/test/specs/elements/Button/Button-test.js
+++ b/test/specs/elements/Button/Button-test.js
@@ -129,7 +129,7 @@ describe('Button', () => {
 
   describe('labeled', () => {
     it('renders as a button', () => {
-      const wrapper = shallow(<Button labeled/>)
+      const wrapper = shallow(<Button labeled />)
       wrapper.should.have.tagName('button')
     })
   })
--- a/test/specs/elements/Button/Button-test.js
+++ b/test/specs/elements/Button/Button-test.js
@@ -126,4 +126,13 @@ describe('Button', () => {
       wrapper.children().at(1).should.match('.ui.label')
     })
   })
+
+  describe('labeled', () => {
+    it('is before the button and icon="plus" when labeled="left"', () => {
+      const wrapper = shallow(<Button labeled='left' icon='plus' />)
+      wrapper.should.have.tagName('button')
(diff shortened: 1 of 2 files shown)


levithomason:
Awesome, just need a test for this.  See `test/specs/elements/Button`.  Should be a very simply copy/paste update of one of the existing tests.

levithomason (changes requested):
Awesome, just need a test for this. See `test/specs/elements/Button`. Should be a very simply copy/paste update of one of the existing tests.
Add your review

x-shd:
Should work in production this trick fixed the problem.

test/specs/elements/Button/Button-test.js
```diff
@@ -126,4 +126,13 @@ describe('Button', () => {
...
     })
   })
+
+  describe('labeled', () => {
+    it('is before the button and icon="plus" when labeled="left"', () => {
+      const wrapper = shallow(<Button labeled='left' icon='plus' />)
+      wrapper.should.have.tagName('button')
+      wrapper.should.have.className('labeled')
+      wrapper.should.have.exactly(1).descendants('Icon')
+    })
```
levithomason, line 134: Thanks for the test, let's simplify it a bit.  It is best to write tests that are as minimal as possible and assert one thing.  This makes them stronger and more maintainable.

In this case, most of the assertions made here are covered by other existing tests, all we should need to assert is that a `labeled` button renders a `button` element:

``` js
it('renders as a button', () => {
  shallow(<Button labeled />)
    .should.have.tagName('button')
})
```

levithomason (changes requested):

test/specs/elements/Button/Button-test.js
x-shd: Made

levithomason (approved):

levithomason:
One linter issue:

image

levithomason:
Also, if the description of this PR were updated slightly, it will auto close the related issue on merge:

```
Fix switch button to div 
#596
```

```
Fixes #596 switch button to div 
```

https://github.com/blog/1506-closing-issues-via-pull-requests

PR #598, opened 2016-10-03
docs(Table): add `attached` example 

### LINKED ISSUE #595: docs(Table): add `attached` example
This is not documented in the core SUI Table definition, however, it is shown in the examples.  We've added support but need to show a similar example.

Supported added in #593.
SUI example here: http://semantic-ui.com/examples/attached.html

I've also added https://github.com/Semantic-Org/Semantic-UI/issues/4615 requesting it be added to the core SUI Table definition docs.

layershifter:
This PR:
- fixes #595;
- adds test for `attached` prop to `Table`.

### DIFF
--- a/docs/app/Examples/collections/Table/Variations/TableExampleAttached.js
+++ b/docs/app/Examples/collections/Table/Variations/TableExampleAttached.js
@@ -1,0 +1,104 @@
+import React from 'react'
+import { Table } from 'stardust'
+
+const TableCelled = () => {
+  return (
+    <div>
+      <Table attached='top' basic>
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell>Header</Table.HeaderCell>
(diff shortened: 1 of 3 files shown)


levithomason:
Thanks much, reviewing.

---

P.S. Based on how the changelog is categorizing labels, I'm going to start labeling issues/PRs like so:
- `docs` - for doc site changes (whether new docs, docs bug fix, or doc enhancement)
- `enhancement` - new library features
- `bug` - library bugs

levithomason:
I changed the export from:

``` jsx
const TableCelled = () => {
  return (
```

To:

``` jsx
const TableExampleAttached = () => (
```

I also reduced the LOC from 104 to 42.  I'd like to keep examples as small as possible for easier doc site usage.

Will ship when tests pass.

levithomason (approved):

PR #601, opened 2016-10-03
fix(Checkbox): no onChange/onClick when disabled

### LINKED ISSUE #600: fix(Checkbox): prevent onChange and onClick when disabled
**Version**
0.52.2

**Steps to Reproduce**
1. Create checkbox with prop disabled set to true.
2. Try to toggle and check if the `onChange` function is called

**Expected**
The Checkbox should not trigger the `onChange` function when disabled, since it doesn't change. 

**Result**
The Checkbox does call the onChange function.

**Testcase**
http://codepen.io/sprmn/pen/JROPjO

sprmn:
Check if `disabled` before calling `onChange` function. 

Fixes #600

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -110,22 +110,22 @@ export default class Checkbox extends Component {
 
   handleClick = (e) => {
     debug('handleClick()')
-    const { onChange, onClick, disabled, name, value } = this.props
+    const { onChange, onClick, name, value } = this.props
     const { checked } = this.state
     debug(`  name:       ${name}`)
     debug(`  value:      ${value}`)
     debug(`  checked:    ${checked}`)
 
-    if (onClick) onClick(e, { name, value, checked: !!checked })
-    if (onChange && !disabled) onChange(e, { name, value, checked: !checked })
-
     if (this.canToggle()) {
+      if (onClick) onClick(e, { name, value, checked: !!checked })
+      if (onChange) onChange(e, { name, value, checked: !checked })
+
       this.trySetState({ checked: !checked })
     }
   }
 
   render() {
-    const { className, label, name, radio, slider, toggle, type, value } = this.props
+    const { className, label, name, radio, slider, toggle, type, value, disabled, readOnly } = this.props
     const { checked } = this.state
     const classes = cx(
       'ui',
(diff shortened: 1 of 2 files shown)


layershifter:
@sprmn Thanks for PR! 

Your addition failed because you don't spreaded `disabled`, you need do this:

``` diff
- const { onChange, onClick, name, value } = this.props
+ const { onChange, onClick, disabled, name, value } = this.props
```

Could you add a test-case for this? In `Checkbox-test.js` in `describe('onChange'', () => {`

``` jsx
it('omitted when has disabled prop', () => {
  const click = () => mount(<Checkbox disabled onChange={spy} />).simulate('click')
  expect(click).to.not.throw()
})
```

sprmn:
Oops, thought that it was already spread, but that was in the `canToggle` function, my bad. 
I'll fix it and add a test case.

layershifter:
You can run tests with `npm run test:watch` and linter `npm run lint:fix` :+1:

jeffcarbs:
A few things:
- There are props for `disabled` and `readOnly` but they're not being applied as classes. They should add the `disabled` and `read-only` classes respectively. Let's also add specs for those:

``` js
  common.propKeyOnlyToClassName(Checkbox, 'disabled')
  common.propKeyOnlyToClassName(Checkbox, 'readOnly', {
    className: 'read-only',
  })
```
- There's already a method to determine if the checkbox can be toggled - `canToggle`. We're currently only using it to determine if we should try to update the checked state, but I think we should also use that to prevent the click/change handlers. So:

``` diff
-    if (onClick) onClick(e, { name, value, checked: !!checked })
-    if (onChange) onChange(e, { name, value, checked: !checked })
-
     if (this.canToggle()) {
+      if (onClick) onClick(e, { name, value, checked: !!checked })
+      if (onChange) onChange(e, { name, value, checked: !checked })
+
       this.trySetState({ checked: !checked })
     }
```

sprmn:
Even better! Should be fixed now.

jeffcarbs (approved):
Checked this out locally and works as expected, 👍  🚢 💨

jeffcarbs:
@levithomason - this is ready to merge when you are 👍

levithomason:
Thanks for taking care of this guys!  Released in `semantic-ui-react@0.54.1`.

PR #602, opened 2016-10-03
feat(GridRow): add missing `divided` prop

### LINKED ISSUE #560: feat(GridRow): add missing `divided` prop
### Steps to Reproduce

Create a `<Grid.Row divided='vertically' />`
### Expected Behavior

The `vertically divided` className should be added to the row.
### Actual Behavior

The className is not added, but the prop is spread on the underlying `div`.  It is treated as a `rest` prop, because it is unhandled.
### Version

Stardust: `0.47.2`
### Test Case

http://codepen.io/levithomason/pen/PGjBRZ

layershifter:
Fixes #560.

### DIFF
--- a/src/collections/Grid/GridRow.js
+++ b/src/collections/Grid/GridRow.js
@@ -34,7 +34,7 @@ function GridRow(props) {
     color,
     useKeyOnly(centered, 'centered'),
     useWidthProp(columns, 'column', true),
-    useKeyOrValueAndKey(divided, 'divided'),
+    useKeyOnly(divided, 'divided'),
     useValueAndKey(only, 'only'),
     useValueAndKey(reversed, 'reversed'),
     useKeyOnly(stretched, 'stretched'),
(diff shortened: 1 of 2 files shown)


src/collections/Grid/GridRow.js
```diff
@@ -41,6 +55,7 @@ GridRow._meta = {
   props: {
     color: SUI.COLORS,
     columns: SUI.WIDTHS,
+    divided: ['vertically'],
```
levithomason: It appears my initial bug report was only half correct.  The row only supports `divided` as bool, not `vertically`.

https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/grid.css#L1083

levithomason (changes requested):

src/collections/Grid/GridRow.js
layershifter: Looks, you are right, will update soon.

levithomason (approved):

levithomason:
Released in `stardust@0.52.4`

PR #603, opened 2016-10-03
feat(Image): add `label` shorthand

### LINKED ISSUE #538: feat(Image): add `label` shorthand
There is interesting [example](http://semantic-ui.com/elements/label.html#corner) in SUI docs. 

``` html
<div class="ui fluid image">
  <a class="ui left corner label">
    <i class="heart icon"></i>
  </a>
  <img src="/images/wireframe/image.png">
</div>
```

It sends `Label` to `Image` as child while we don't handle this in `Image` component. While I'll work on `Label` docs I'll check all examples, but now my initial proposal is add `label` shorthand for `Image`.

layershifter:
Fixes #538.

### DIFF
--- a/docs/app/Examples/elements/Label/Types/LabelExampleCorner.js
+++ b/docs/app/Examples/elements/Label/Types/LabelExampleCorner.js
@@ -1,26 +1,22 @@
 import React from 'react'
-import { Grid, Icon, Label } from 'stardust'
-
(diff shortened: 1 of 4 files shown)


levithomason (approved):
Added `number` as valid shorthand, will merge on pass.

levithomason:
Released in `stardust@0.52.3`

PR #610, opened 2016-10-03
chore(imports): Direct internal imports

jeffcarbs:
As per discussions https://github.com/TechnologyAdvice/stardust/pull/570#discussion_r81445116 and https://github.com/TechnologyAdvice/stardust/issues/524#issuecomment-250935483:
- updates all internal imports to be directly to the file instead of using the `index.js` in each "type" folder
- remove the index files in "type" folders.

Open questions:
- Should each top-level component (e.g. `Button` get its own index file? Would make it possible to do `import Button from 'stardust/elements/Button` instead of needing the double `/Button/Button`. Sub-components would still be referenced at the nested path, which I think makes sense `stardust/elements/Button/ButtonContent`. Thoughts?


besh:
I'm for the index files. If you didn't want to bring in index files, you could toss a package.json file in each dir and set the `main` key to `Button.js.` Not saying this is the best thing to do but it is another option.

levithomason:
Agree on not duplicating the component name in import paths.  My first thought is an index.js file as well.

levithomason:
Heads up, regarding conformance tests (near the top of the common tests):

> https://github.com/TechnologyAdvice/stardust/issues/524#issuecomment-250935483 Lastly, the common tests for exported components needs updated. The failure message instructs users to export their components in those index files. Once we remove them, that message is no longer valid. It should instead instruct users to export their components in the main index file.

jeffcarbs:
Just force-pushed a small change. I was going to update the tests to use the index-based imports but decided against it. I kinda like having the test import directly from the file they're testing.

Otherwise this is g2g I think 👍

levithomason:
:thumbsup: Great work, will review and merge soon as I can.

levithomason (approved):

PR #614, opened 2016-10-04
chore(style): Make some common patterns more consistent

jeffcarbs:
There were some things that we did in a lot of files but slightly differently. These aren't really lintable so there's no way to enforce them moving forward, but figured I take a pass through so at least it's more likely someone will copy/paste the "right" way 😄 . I tried to grep for some of this stuff when possible and chose the most common variation of the pattern:
- Added className prop to `MessageItem`
- Swapped order of `className={classes} {...rest}` to `{...rest} className={classes}` (the latter was way more common and makes more sense).
- Use `ElementType` in a few places we weren't.
- Always import 'classnames' as 'cx'
- Always order setting variables: props, classes, rest, ElementType
- Newline between setting variables and returning `<ElementType>`
- When returning early because children, render a one-liner
- Order children before className in props

### DIFF
--- a/src/views/Item/ItemImage.js
+++ b/src/views/Item/ItemImage.js
@@ -1,19 +1,14 @@
 import React from 'react'
 import {
-  getElementType,
-  getUnhandledProps,
   META,
 } from '../../lib'
 import Image from '../../elements/Image'
 
 /**
  * An item can contain an image
- * */
+ **/
 function ItemImage(props) {
-  const ElementType = getElementType(ItemImage, props)
-  const rest = getUnhandledProps(ItemImage, props)
-
-  return <Image {...rest} as={ElementType} ui={false} />
+  return <Image {...props} ui={false} />
 }
 
 ItemImage._meta = {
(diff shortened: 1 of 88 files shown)


src/views/Item/ItemImage.js
```diff
@@ -10,10 +8,7 @@ import Image from '../../elements/Image'
  * An item can contain an image
  * */
```
levithomason: While we're at it, there are a number of doc blocks with `* */` as a closing.  This should just be `*/`.  The copy pasta there has been killing me a little :)

levithomason (changes requested):

levithomason:
Also, there is technically a `style(...): msg` [commit type](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#type) we could use for this PR.

levithomason:
Github is back, the outage killed tests.  I just retried them, we'll see.

src/views/Item/ItemImage.js
jeffcarbs: Fixed 👍

jeffcarbs:
Was missing an import. Fixed and pushed up.

levithomason:
Ready to go here then?

levithomason (approved):

levithomason:
Going!

PR #616, opened 2016-10-04
BREAKING CHANGE: refactor `labeled` Button API

### LINKED ISSUE #510: refactor(Button): rework `labeled` for consistency with Input
At the time of writing the Input API is in progress at #281.  The button API is already complete.  This issue was started to keep consistency between the `labeled` APIs of these two components.
## Issue

The Input component can accept a label or an action on the left or right of the `<input />`.  If children are used for configuring the label or action (as in the Button) then the children must be parsed and the  `<input />` must be inserted into the children at the correct place.  Identifying and parsing children is not a practical solution.
## Solution

The best way to deal with this relationship is via props.  We'll leverage shorthand prop values (literal, props object, element instance) to construct labeled markup.
### Refactors

I've taken some doc examples and refactored them to show the new API.  Notable changes:
- `<Button labeled />` goes away.  It currently renders a `ui labeled button` wrapper.
- Button children will always be the primary content of the button
- `labeled='left|right'` is now `labelPosition`

I've shown the shortest syntax for the examples below.  Though, keep in mind that all shorthand props can take a primitive value, a props object, or an element instance.  You can of course organize (unusually rare) complicated label/icon markup into a separate render function and call it there as well.  This retains full control over the markup for the user.
#### `ButtonLabeledExample.js`

``` jsx
<Button labeled>
  <Button>
    <Icon name='heart' /> Like
  </Button>
  <Label as='a' basic>2,048</Label>
</Button>

<Button labeled='left'>
  <Label as='a' basic pointing='right'>2,048</Label>
  <Button>
    <Icon name='heart' /> Like
  </Button>
</Button>

<Button labeled='left'>
  <Label as='a' basic>2,048</Label>
  <Button icon>
    <Icon name='fork' />
  </Button>
</Button>
```

``` jsx
<Button
  content='Like'
  icon='heart'
  label={{ as: 'a', basic: true, content: '2,048' }}
/>

<Button
  content='Like'
  icon='heart'
  label={{ as: 'a', basic: true, pointing: 'right', content: '2,048' }}
  labelPosition='left'
/>

<Button
  icon='fork'
  label={{ as: 'a', basic: true, content: '2,048' }}
  labelPosition='left'
/>
```
#### `ButtonLabeledIconExample.js`

``` jsx
<Button labeled icon>
  <Icon name='pause' /> Pause
</Button>
<Button labeled='right' icon>
  <Icon name='right arrow' /> Next
</Button>
```

``` jsx
<Button icon='pause' labelPosition='left' content='Pause' />
<Button icon='right arrow' labelPosition='right' content='Next' />
```
## Conclusion

This API is necessary for the Input.  An Input should never have children, as the current labeled Button API precedent would suggest.  We think consistency across all components is extremely important.  So, the Button should also use the same `labeled` API.

After several offline chats, we also think it just makes more sense this way.  A Button should always render a Button.  Its children should always be the button content.  Instead of the current state, where it sometimes renders a `ui labeled button` container that accepts a Button as a child, which renders an actual button.  This is strange and unintuitive.

### LINKED ISSUE #612: fix(Button): revert rendering `labeled` buttons as <button>
This should be fixed by fixing #510, not by creating another patch PR. This bug was introduced in #597.

A `labeled` button has children that can include a `<button>` element.  Therefore, the container element cannot be a `button` and also have `<button>` children.  This is not valid markup.

http://technologyadvice.github.io/stardust/elements/button#labeled

image

image

levithomason:
Fixes #510 

This PR makes the Button `labeled` API consistent with the Input `labeled` API.  Previously, a Button could be used as a `labeled button` wrapper with another Button/Label child.  Now:
- `labeled` is now `labelPosition`
- Button children are _always_ the content of the Button.  A Button can no longer contain another Button as a child.
- Since Labels are not Button children, you can no longer pass them as such. You must use the shorthand `label` prop to pass the Label.

Fixes #612

~~It also fixes a recent regression, rendering `labeled` Buttons as `<button>` tags.  You cannot nest a `<button>` in a `<button>`, the labeled button must be a `div` so it can contain a Label and a Button (`<button>`) component.~~

This did not account for `labelPosition` with an `icon`.  In that case, a `<button>` can contain an `<i />`.  A `div` is only required when there is a `label`.  A fix commit has been made to master, https://github.com/TechnologyAdvice/stardust/commit/3379aa5b174a44e9113318032eb92feedf178a16, and will be included in this release.


docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -144,7 +144,7 @@ export default class ComponentExample extends Component {
       .map(l => {
         const defaultImport = _.get(/import\s+(\w+)/.exec(l), '[1]')
         const destructuredImports = _.get(/import.*({[\s\w,}]+)\s+from/.exec(l), '[1]')
-        const module = _.snakeCase(_.get(/import.*from.*'(.*)'/.exec(l), '[1]', '')).toUpperCase()
+        const module = _.snakeCase(_.get(/import.*from\s+['"]([\w\-_]+)/.exec(l), '[1]', '')).toUpperCase()
```
levithomason, line 147: Would like tests for this someday.  This fixes the module name detection RegEx, the previous was too permissive and captured too much.

PR #618, opened 2016-10-04
feat(shorthand): Custom shorthand proptypes

jeffcarbs:
This adds a few new customPropTypes (took some code and ideas from #452, thanks @levithomason :+1:). It's very much WIP but wanted to get it up and get some feedback before I go too crazy here. The new propTypes are:
- `itemShorthand`
  - disallows children
  - one of: string, number, object, element
- `arrayOfShorthand` (better name?)
  - disallows children
  - arrayOf(itemShorthand)
- `collectionShorthand`
  - disallows children
  - one of: arrayOfShorthand, object, element
- `nodeShorthand`
  - disallows children
  - node

The large majority of our components have a `content` prop that can be rendered in place of the children. This was validated differently in various components as strings/numbers/elements, but really it can be any renderable thing. This is where the `nodeShorthand` came from. We could theoretically just use `node` but I wanted to be able to disallow children.

I also wound up simplifying/standardizing the doc blocks for common propTypes (there was like 10 ways we described the `className` prop haha). I did it via regex so it's super easy to change so let me know if you think the descriptions could be improved.

Open questions:
- We double-disallow right now, e.g. content disallows children, children disallows content. Is there anything wrong with only disallowing on one side? Assuming all shorthands disallow children, children could just be `PropTypes.node` which is definitely simpler. Currently it's a customPropType that disallows any shorthands (inferring which are shorthand props from Component.propTypes)

cc @levithomason @layershifter

### DIFF
--- a/src/lib/customPropTypes.js
+++ b/src/lib/customPropTypes.js
@@ -181,6 +181,49 @@ export const demand = (requiredProps) => {
 }
 
 /**
+ * Ensure a component can render as a node passed as a prop value in place of children.
+ */
+export const content = (...args) => every([
+  disallow(['children']),
+  PropTypes.node,
+])(...args)
+
+/**
+ * Item shorthand is a description of a component that can be a literal,
+ * a props object, or an element.
+ */
+export const item = (...args) => every([
+  disallow(['children']),
+  PropTypes.oneOfType([
+    PropTypes.element,
+    PropTypes.number,
+    PropTypes.object,
+    PropTypes.string,
+  ]),
+])(...args)
+
+/**
+ * ArrayOf shorthand ensures a prop is an array of item shorthand
+ */
+export const items = (...args) => every([
+  disallow(['children']),
+  PropTypes.arrayOf(item),
+])(...args)
+
+/**
+ * Collection shorthand is a description of a component that can be an array of
+ * shorthand for the underlying items, a props object, or an element.
+ */
+export const collection = (...args) => every([
+  disallow(['children']),
+  PropTypes.oneOfType([
+    items,
+    PropTypes.element,
+    PropTypes.object,
+  ]),
+])(...args)
+
+/**
  * Show a deprecated warning for component props with a help message and optional validator.
  * @param {string} help A help message to display with the deprecation warning.
  * @param {function} [validator] A propType function.
--- a/src/lib/customPropTypes.js
+++ b/src/lib/customPropTypes.js
@@ -183,7 +183,7 @@ export const demand = (requiredProps) => {
 /**
  * Ensure a component can render as a node passed as a prop value in place of children.
  */
-export const content = (...args) => every([
+export const contentShorthand = (...args) => every([
   disallow(['children']),
   PropTypes.node,
 ])(...args)
@@ -192,35 +192,20 @@ export const content = (...args) => every([
  * Item shorthand is a description of a component that can be a literal,
  * a props object, or an element.
  */
-export const item = (...args) => every([
+export const itemShorthand = (...args) => every([
   disallow(['children']),
   PropTypes.oneOfType([
-    PropTypes.element,
-    PropTypes.number,
+    PropTypes.node,
     PropTypes.object,
-    PropTypes.string,
   ]),
 ])(...args)
 
 /**
- * ArrayOf shorthand ensures a prop is an array of item shorthand
+ * Collection shorthand ensures a prop is an array of item shorthand.
  */
-export const items = (...args) => every([
+export const itemsShorthand = (...args) => every([
   disallow(['children']),
-  PropTypes.arrayOf(item),
-])(...args)
-
-/**
- * Collection shorthand is a description of a component that can be an array of
- * shorthand for the underlying items, a props object, or an element.
- */
-export const collection = (...args) => every([
-  disallow(['children']),
-  PropTypes.oneOfType([
-    items,
-    PropTypes.element,
-    PropTypes.object,
-  ]),
+  PropTypes.arrayOf(itemShorthand),
 ])(...args)
 
 /**
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -490,10 +490,10 @@ Label.propTypes = {
   /** A label can reduce its complexity. */
   basic: PropTypes.bool,
 
-  /** Primary content of the label, same as text. */
+  /** Primary content. */
   children: PropTypes.node,
 
-  /** Classes to add to the label className. */
+  /** Additional classes. */
   className: PropTypes.string,
 
   /** Color of the label. */
--- a/src/addons/Confirm/Confirm.js
+++ b/src/addons/Confirm/Confirm.js
@@ -42,7 +42,7 @@ Confirm.propTypes = {
   /** The ModalHeader text */
   header: PropTypes.string,
 
-  /** The ModalContent text. Mutually exclusive with children. */
+  /** The ModalContent text. */
   content: PropTypes.string,
 
   /** Called when the OK button is clicked */
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -22,10 +22,10 @@ const _meta = {
  */
 class Portal extends Component {
   static propTypes = {
-    /** Primary content */
+    /** Primary content. */
     children: PropTypes.node.isRequired,
 
-    /** Classes to be added to the portal node element. */
+    /** Additional classes. */
     className: PropTypes.string,
 
     /** Controls whether or not the portal should close on a click outside. */
--- a/src/addons/Select/Select.js
+++ b/src/addons/Select/Select.js
@@ -1,6 +1,6 @@
-import React, { PropTypes } from 'react'
+import React from 'react'
 
-import { getUnhandledProps, META } from '../../lib'
+import { META } from '../../lib'
 import Dropdown from '../../modules/Dropdown'
 
 /**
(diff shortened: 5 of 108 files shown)


levithomason:
Ah!  So pumped to see this :D  I'm in agreement with everything noted in this PR.  These exact items were on my bucket list.

> Is there anything wrong with only disallowing on one side? Assuming all shorthands disallow children, children could just be PropTypes.node which is definitely simpler.

Yes please!  One way disallow is perfect, children can just be node.

### Icon/Image

I also had an `icon` and `image` shorthand prop type that disallowed the other, #452.  That way you can only add an icon or an image but not both.  This is applicable in places like Label and Header.  Can't recall where else if any.

layershifter:
May be some shorter names?

> itemShorthand => item
> arrayOfShorthand => items
> collectionShorthand => collection
> nodeShorthand => content (nodeShorthand only used on content prop as I see)

It will look good with `icon` and `image`.

> I also wound up simplifying/standardizing the doc blocks for common propTypes

:+1:

src/lib/customPropTypes.js
```diff
@@ -181,6 +181,71 @@ export const demand = (requiredProps) => {
...
+  itemShorthand,
+  nodeShorthand,
+]
+
+/**
+ * Ensure a prop conforms to children prop standards.
+ */
+export const children = (Component) => (...args) => {
+  const disallowedProps = _.flow(
+    _.pickBy(checker => _.some(shorthandPropTypes, (propType) => propType === checker)),
```
levithomason: Is this working correctly?  This file uses `lodash/fp` which takes data last, so I'd expect the `shorthandPropTypes` (data) to be provided after the `some` function.

Also, might be slightly cleaner to use `_.includes`:

``` jsx
_.pickBy(checker => _.includes(checker, shorthandPropTypes))
```

I believe this is correct, data last again.  Didn't test it though.

levithomason (changes requested):

levithomason:
Can we also close #452 with this PR?  It pretty much covers the rest of that one.  The only remaining prop of concern is childKey, and that will get handled when do the collection factory I'm sure.

src/lib/customPropTypes.js
jeffcarbs: As per https://github.com/TechnologyAdvice/stardust/pull/618#issuecomment-251318050 I'm just going to scrap this whole custom propType and make all the children `propType.node`. The "disallow" will happen on the shorthand side.

jeffcarbs:
Apologies for the long PR, but I don't think there's a great way to break it up and most of the changed lines are just comments. FWIW, I ran specs before/after and there were the same number of warnings, so I'm pretty confident this didn't introduce any regressions in prop validations.

This is ready for review now, but let's hold off on merging until #619 since there are going to be a ton of merge conflicts.

TODO on separate PRs:
- There are some components that were too complicated to try to unwind so I didn't apply the custom prop types there. `Dropdown` and `Button` are the main two. I think there's work to be done to try to simplify the logic there, if only to make it more clear to us what's going on.
- Audit components that are rendering children + shorthand (e.g. `DropdownItem`, `Message`, `FeedEvent`) and see if we can refactor to avoid it.
- We currently have a factory function that converts`customPropTypes.item` => `Component`. It would be nice to also have one that converts `customPropTypes.items` => `[Component, Component, ...]`. In some cases this is as simple as `_.map(items, SubComponent.create)` but many times the props depend on the index (e.g. active = `selectedIndex === index`), so TBD there.

levithomason:
Bump, this is ready for conflict resolutions. Good luck my friend 🎱

jeffcarbs:
@levithomason - Already rebased and handled the conflicts 😄 This is ready for final review 👍

levithomason:
Oh snap!  I'm on it, wanna trade me reviews?  #552 factory tests :)

src/lib/customPropTypes.js
```diff
@@ -181,6 +181,49 @@ export const demand = (requiredProps) => {
...
+    PropTypes.object,
+    PropTypes.string,
+  ]),
+])(...args)
+
+/**
+ * ArrayOf shorthand ensures a prop is an array of item shorthand
+ */
+export const items = (...args) => every([
+  disallow(['children']),
```
levithomason, line 207: This is duplicate disallowing since the item disallows.  Suggest removing that.  I'm also not sure we should have this checker.  Seems we could just do `PropTypes.arrayOf(item)` in the component and `collection` below to simplify.  Will let you make the call on that.

src/lib/customPropTypes.js
```diff
@@ -181,6 +181,49 @@ export const demand = (requiredProps) => {
 }
 
 /**
+ * Ensure a component can render as a node passed as a prop value in place of children.
+ */
+export const content = (...args) => every([
```
levithomason: Apologies for the lack of response when you asked about this previously, but I really liked having the longer names here.  I think it is more clear and more in line with the React community to be explicit.  Especially being `collection` is a legitimate data type and the collection propType refers to something different, shorthand.

``` jsx
contentShorthand
itemShorthand
collectionShorthand
```

Given all the back and forth you suffered here, if you wanna push this off to another PR, I get that and am OK doing so. :)

levithomason (approved):
Accepting as I only have 2 suggestions that I'm deferring to you on.  Will merge on your word, cheers!

src/lib/customPropTypes.js
jeffcarbs: No worries, I don't think there's been much back and forth here.

I'm torn because I like the terseness of the shorter names but I'm also kinda leaning these longer names. Ultimately I think clarity wins out, I'll update the names now.

src/lib/customPropTypes.js
jeffcarbs: I'm trying to unwind this a little bit because it's gets a little complicated with the nested `every` calls. Turns out this `disallow` is actually necessary, I'll try to explain. Consider the following:

``` js
const cells = ['Name', 'Status', 'Notes']
<TableRow cells={cells}>some child</TableRow>
```

If cells were `PropTypes.arrayOf(item)` this doesn't give a warning because `arrayOf` will use the given validation (`item` in this case) to validate each item in the `cells` within the context of `cells`, not within the context of `TableRow`.

Nested `disallow` doesn't actually seem to work as expected since it operates on the parent context of the given prop. So when checking `cells` with:

``` js
const items = (...args) => every([
  disallow(['children']),
  PropTypes.arrayOf(item),
])(...args)
```

It's evaluating `cells` within the context of the TableRow props and asking **"is cells an array of items AND is the children key undefined _in TableRow props_"**. When it's evaluating each item via `arrayOf(item)`, it's asking **"is this item one of these valid types AND is the children key undefined _in this item's array_"**  (i.e. `array['children']`) which doesn't throw but doesn't really make sense 

I hope that makes sense but it's pretty complicated haha.

jeffcarbs:
See that last comment regarding the need for the four types. I'm going to rename the others, but I'm struggling to think of a clear naming distinction between the `items` and `collection`. In this case, a "collection" is an item/entity that has a set of "items" and "items" is the array of items. However, arrays are commonly referred to as "collections" so it's kinda confusing.

What are your thoughts on adding `array` as a valid type of `itemShorthand`? An array is technically a valid "node" in that it's renderable wherever a string/number/element is renderable. I could see a user wanting to pass sibling elements as shorthand without a wrapping div (I actually did that in search: https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Search/SearchResult.js#L18-L27)

It could be up to the the component's `create` function how it wanted to handle a given literal. For example when passing non-arrays to `TableRow.create` it could either cast it to an array or throw. This also lines up with the `createShorthand` factory which treats string/number/array the same. It's also possibly more consistent since `TableRow.create(items)` is technically creating an item, not a collection/array.

This would mean that there'd be the following prop validations:
- contentShorthand - node/element
- itemShorthand - array/number/string/props-object/element
- collectionShorthand - arrayOf(itemShorthand)

Hate to throw a curveball like this after a PR has gone through review, but my gut is telling me maybe just those three would be sufficient and we can split out a `itemWithCollection` proptype later.

Thoughts?

jeffcarbs:
Appended `Shorthand` to all of the current shorthand prop types for now. LMK what you think about my last comment re: collapsing what is now `collectionShorthand` into `itemShorthand` and using `collectionShorthand` for the `arrayOf(itemShorthand)` like `items` and `cells`.

src/lib/customPropTypes.js
levithomason: Got it, perhaps that is one of the differences with the `arrayOf` checker.  I know if has a different signature and operates uniquely compared to all other checkers.  Either way, it makes sense and I think we leave it as is for now.

levithomason:
> This would mean that there'd be the following prop validations:
> - contentShorthand - node/element
> - itemShorthand - array/number/string/props-object/element
> - collectionShorthand - arrayOf(itemShorthand)

This seems correct to me, since it mirrors the `createShorthand` factory's supported types.

I also note that since we've expanded it's ability and definition quite a bit, `itemShorthand` is actually `node/object` now.  Since [`node` includes](https://facebook.github.io/react/docs/reusable-components.html) numbers, strings, elements and array (of any of these).  That being the case, `contentShorthand` is also just `node` since `node` includes elements.  So, I think we're left with:
- `contentShorthand` - node
- `itemShorthand` - node/object
- `collectionShorthand` - arrayOf(itemShorthand)

Where we're actually just adding `object` to the already valid renderable React types.

jeffcarbs:
👍  just made the updates. Once the specs pass I think we're cool to merge

levithomason:
Sounds good, will do.

PR #622, opened 2016-10-04
feat(Dropdown): add render icon to result select

x-shd:
Fixes: #620

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -755,17 +755,24 @@ export default class Dropdown extends Component {
       search && searchQuery && 'filtered'
     )
     let _text = placeholder
+    let _icon = null
+    let _icon_name = null
     if (searchQuery) {
       _text = null
     } else if (text) {
       _text = text
     } else if (open && !multiple) {
       _text = _.get(this.getSelectedItem(), 'text')
+      _icon_name = _.get(this.getSelectedItem(), 'icon')
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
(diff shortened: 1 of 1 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -755,17 +755,26 @@ export default class Dropdown extends Component {
...
     } else if (open && !multiple) {
       _text = _.get(this.getSelectedItem(), 'text')
+      iconName = _.get(this.getSelectedItem(), 'icon')
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
+      iconName = _.get(this.getItemByValue(value), 'icon')
+    }
+
+    if (iconName) {
+      _icon = <Icon name={iconName} />
```
levithomason, line 774: Since we allow Icons to be created from shorthand (by name, props object, or an Icon element), let's use the shorthand create factory.  It should be passed the value of the `icon` prop, whatever that may be.  It will handle creating the Icon from many values.  It will also not create the Icon if the value is null or undefined.

This removes the need to have both `_icon` and `_iconName`.  It also removes the need for this `if` statement.  You can just extract the value of the `icon` prop, and pass it to the shorthand create method in the return statement.  Here are the relevant lines:

``` jsx
// can initialize undefined
let _icon

// then update the values if available
_icon = _.get(this.getSelectedItem(), 'icon')
_icon = _.get(this.getItemByValue(value), 'icon')

// finally, pass it to the factory
return <div className={classes}>{Icon.create(_icon)}{_text}</div>
```

levithomason (changes requested):

Issue #599, opened 2016-10-03
docs(Example): Fix warnings


layershifter:
We have warnings in docs and test, they throw warnings in prop validation.
### Menu
- [ ] Complete

> Warning: Failed prop type: Invalid prop `link` of type `string` supplied to `Icon`, expected `boolean`.
>    in Icon (created by Input)
>    in Input (created by TabularOnTop)
>    in div (created by MenuItem)
>    in MenuItem (created by TabularOnTop)
>    in div (created by MenuMenu)
>    in MenuMenu (created by TabularOnTop)
>    in div (created by Menu)
>    in Menu (created by TabularOnTop)
>    in div (created by TabularOnTop)
>    in TabularOnTop
### Table
- [ ] Complete

> Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Table`. See https://fb.me/react-warning-keys for more information.
>    in TableRow (created by Table)
>    in Table (created by TableWarningShorthand)
>    in TableWarningShorthand
> 
> Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `TableRow`. See https://fb.me/react-warning-keys for more information.
>    in TableCell (created by TableRow)
>    in TableRow (created by Table)
>    in thead (created by TableHeader)
>    in TableHeader (created by Table)
>    in table (created by Table)
>    in Table (created by TableWarningShorthand)
>    in TableWarningShorthand
### Feed
- [ ] Complete

> Warning: Failed prop type: Prop `children` in `FeedContent` conflicts with props: `date`. They cannot be defined together, choose one or the other.
>    in FeedContent (created by ContentBlock)
>    in ContentBlock
> 
> Warning: Failed prop type: Prop `date` in `FeedContent` conflicts with props: `children`. They cannot be defined together, choose one or the other.
>    in FeedContent (created by ContentBlock)
>    in ContentBlock
### Dropdown
- [ ] Complete

> Warning: Failed prop type: `children` prop in `Dropdown` requires props: `text`.
>    in Dropdown (created by DropdownTriggerExample)
>    in DropdownTriggerExample
> 
> Warning: Failed prop type: `options` prop in `Dropdown` requires props: `selection`.
>    in Dropdown (created by DropdownHeaderExample)
>    in DropdownHeaderExample
### Portal
- [ ] Complete

> ```
> ✔ should call props.onUnmount() only once when portal closes and then is unmounted
> ```
> 
[...]

PR #626, opened 2016-10-05
chore(warnings): Fix warnings in Form and Modal

jeffcarbs:
Fixes warnings in two of the components referenced here: #599

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -88,7 +88,13 @@ function Input(props) {
   )
   const rest = getUnhandledProps(Input, props)
   const ElementType = getElementType(Input, props)
-  const inputProps = _.pick(props, htmlInputPropNames)
+  const inputProps = {
+    ..._.pick(props, htmlInputPropNames),
+    // React gives a warning if no onChange is given to a controlled input. We
+    // handle the onChange on the wrapping element, so we don't actually need
+    // onChange on the input itself but let's pass a noop to suppress the warning.
+    onChange: _.noop,
+  }
 
   if (children) {
     return <ElementType {...rest} className={classes}>{children}</ElementType>
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -32,6 +32,7 @@ export const htmlInputPropNames = [
   'maxLength',
   'min',
   'name',
+  'onChange',
   'pattern',
   'placeholder',
   'readOnly',
@@ -86,15 +87,11 @@ function Input(props) {
     className,
     'input',
   )
-  const rest = getUnhandledProps(Input, props)
+  const unhandled = getUnhandledProps(Input, props)
+
+  const rest = _.omit(unhandled, htmlInputPropNames)
+  const htmlInputProps = _.pick(props, htmlInputPropNames)
   const ElementType = getElementType(Input, props)
-  const inputProps = {
-    ..._.pick(props, htmlInputPropNames),
-    // React gives a warning if no onChange is given to a controlled input. We
-    // handle the onChange on the wrapping element, so we don't actually need
-    // onChange on the input itself but let's pass a noop to suppress the warning.
-    onChange: _.noop,
-  }
 
   if (children) {
     return <ElementType {...rest} className={classes}>{children}</ElementType>
@@ -121,7 +118,7 @@ function Input(props) {
       {actionPosition === 'left' && actionElement}
       {iconPosition === 'left' && iconElement}
       {labelPosition !== 'right' && labelElement}
-      {createHTMLInput(input || type, inputProps)}
+      {createHTMLInput(input || type, htmlInputProps)}
       {actionPosition !== 'left' && actionElement}
       {iconPosition !== 'left' && iconElement}
       {labelPosition === 'right' && labelElement}
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -157,7 +157,16 @@ class Modal extends Component {
     )
     const rest = getUnhandledProps(Modal, this.props)
     const ElementType = getElementType(Modal, this.props)
-    const portalProps = _.pick(rest, _.keys(Portal.propTypes))
+
+    // Remove portal-handled props from 'rest' and pass them through to Portal
+    const portalProps = _.reduce(rest, (memo, value, key) => {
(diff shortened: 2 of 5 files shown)


layershifter:
Note, I added tests for `Form`'s `bool` props in #584.

src/modules/Modal/Modal.js
```diff
@@ -157,7 +157,16 @@ class Modal extends Component {
     )
     const rest = getUnhandledProps(Modal, this.props)
     const ElementType = getElementType(Modal, this.props)
-    const portalProps = _.pick(rest, _.keys(Portal.propTypes))
+
+    // Remove portal-handled props from 'rest' and pass them through to Portal
+    const portalProps = _.reduce(rest, (memo, value, key) => {
```
levithomason: Just a suggestion here.  This adds another loop, but is a little clearer (at least to my eyes) and does not mutate any objects:

``` jsx
const unhandled = getUnhandledProps(Modal, this.props)

const portalPropKeys = _.keys(Portal.propTypes)
const portalProps = _.pick(unhandled, portalPropKeys)
const rest = _.omit(unhandled, portalPropKeys)
```

test/specs/elements/Input/Input-test.js
```diff
@@ -62,4 +63,23 @@ describe('Input', () => {
       })
     })
   })
+
+  describe('onChange', () => {
+    it('handles onChange', () => {
```
levithomason: This should be covered by the [conformance test for events](https://github.com/TechnologyAdvice/stardust/blob/master/test/specs/commonTests.js#L259).  It runs every component through every event handler and asserts that it was called, and was called with the appropriate event shape as the first arg.

src/elements/Input/Input.js
```diff
@@ -88,7 +88,13 @@ function Input(props) {
   )
   const rest = getUnhandledProps(Input, props)
   const ElementType = getElementType(Input, props)
-  const inputProps = _.pick(props, htmlInputPropNames)
+  const inputProps = {
+    ..._.pick(props, htmlInputPropNames),
+    // React gives a warning if no onChange is given to a controlled input. We
+    // handle the onChange on the wrapping element, so we don't actually need
+    // onChange on the input itself but let's pass a noop to suppress the warning.
+    onChange: _.noop,
```
levithomason: Perhaps we instead add `onChange` to the `htmlInputPropNames`?  This way, it is applied to the correct place and suppresses the warning only when it should be suppressed (user passed value/onChange).  My concern is that we're swallowing the warning.  If the user adds `<Input value='' />`, they should see the warning but they won't with the current change here.

test/specs/modules/Modal/Modal-test.js
```diff
@@ -104,6 +104,12 @@ describe('Modal', () => {
       shallow(<Modal open={false} />)
         .find('Portal')
         .should.have.prop('open', false)
+
+      // Does not pass this prop to the portal child
```
levithomason: Suggest pulling this into it's own `it()` block to keep assertions isolated.

src/elements/Input/Input.js
jeffcarbs: That makes sense, although adding it to the `htmlInputPropNames` actually causes it to be called twice. I think we need to do a similar thing here as the `portalProps`, where any prop we pass through to the input isn't included in the `rest`. e.g.

``` js
// Currently:
const rest = getUnhandledProps(Input, props)
const inputProps = _.pick(props, htmlInputPropNames)

// Change to:
const unhandled = getUnhandledProps(Input, props)

const htmlInputProps = _.pick(unhandled, htmlInputPropNames)
const rest = _.omit(unhandled, htmlInputPropNames)
```

levithomason: _I think_ that is true?  Just make sure all the examples work/look how they should.  SUI relies partially on the classNames _and_ partially on the input props to do what it does for the Input component.

There are gonna be some exceptions for things like `disabled`, which adds opacity to the input.  It can't be both a class and an attribute or it is doubly transparent.  Also, `readOnly` I believe must be on the input to prevent entries, but not required as a className.

jeffcarbs: Actually, this causes the `handles events transparently` test to fail. I think that test tries to call the event (`onChange` in this case) on the top-level component (the `Input` in this case). Since the `onChange` is now actually being passed as a prop to the `htmlInput`, it's not being called when simulated on the `Input`.

Thoughts on how to handle this? I agree it would be better to not suppress the warning, so ideally we can improve the test, assuming I'm understanding the issue correctly.

jeffcarbs:
Regarding my comment about the "handles events transparently" above, I updated the test slightly to account for this situation. I think it makes sense but let me know if not.

The idea behind the test (imho) is to ensure that if a user passes an event handler as a prop to the top-level component it is fired when that event takes place within that component. In this case, we're binding the event to a more appropriate sub-component that is the only element that will actually cause the event to be fired in an actual browser. So this test now is able to simulate a more real-world scenario. 

The only negative is that within other tests if you did `.find(Input).simulate('change')` it wouldn't trigger the `onChange` for the component as one might expect because the event needs to be simulated on the html `input` within.

Thoughts?

test/specs/commonTests.js
```diff
@@ -295,7 +298,7 @@ export const isConformant = (Component, options = {}) => {
         //                   ^ was not called on "blur"
         const leftPad = ' '.repeat(constructorName.length + listenerName.length + 3)
 
-        handlerSpy.called.should.equal(true,
+        handlerSpy.calledOnce.should.equal(true,
           `<${constructorName} ${listenerName}={${handlerName}} />\n` +
           `${leftPad} ^ was not called on "${eventName}".` +
```
levithomason: Update required: "was not called ~~on~~ once"

levithomason:
Test changes look good to me!  I think if someone is writing tests for a component, they should be responsible for `find()`ing the correct component to fire the event on.  Seems sane.

I added one last request to update the event assertion error to match the updated `calledOnce` assertion.  Other than this, I think we're there!

test/specs/commonTests.js
jeffcarbs: Fixed 👍

levithomason (approved):

levithomason:
Looks good, will merge on pass

PR #630, opened 2016-10-06
docs(Confirm): refactor and use index.js

dpkwhan:


levithomason:
Looks like this PR snuck in when docs were down, code looks good so merging.  There is an issue with the Confirm examples themselves (using old Modal API), but I'll update that on another PR.

PR #632, opened 2016-10-06
fix(customPropTypes): itemsShorthand -> collectionShorthand

levithomason:
This prop type checker was renamed in all components, but the customPropType method itself somehow remained with the old name.  Fixes these issues:

image

Where:

``` jsx
    /** Shorthand array of props for Menu. */
    items: customPropTypes.collectionShorthand,
```


jeffcarbs:
You're right. Not sure how I missed that.

levithomason:
No worries, you and I both ;)

PR #634, opened 2016-10-06
BREAKING CHANGE: update Confirm to Modal v1 API

levithomason:
This PR updates the Confirm addon to use the Modal v1 API.  It is currently broken.

`active` => `open`
`onHide` => `onClose`


jeffcarbs:
Feels like this should've been caught with a test. Can you add just a basic "pass open, confirm is visible on page" test, similar to the `Modal`? I try to avoid testing duplicate functionality, but `open` is actually a prop on `Confirm` so it makes sense that a test would cover its usage.

levithomason:
Absolutely agreed, on my list. IMO no public API ought to be untested.

levithomason:
Released in `semantic-ui-react@0.55.0`

PR #637, opened 2016-10-06
feat(List): Shorthand for list and list items

### LINKED ISSUE #589: feat(ListItem): add shorthand
The `List` update broke the Introduction page:

image

There may be other legacy usages as well.  Suggest searching the codebase and updating all usages of `List` to the v1 API.

jeffcarbs:
Fixes #589

This adds shorthand for list and list item. Still somewhat WIP:
- [x] Add missing test coverage
- [x] ~~Add pass-through props to item-content~~ Solved a different way

Only open question is regarding documentation. How do we want to handle shorthand documentation? I added a few examples here but there's a lot of them for List so I wanted to clarify our approach before going much further. Some thoughts:
- I definitely think we should write shorthand for all examples.
- I think the best idea would be to just show the example once and show the shorthand side-by-side with the regular code. 
  - Showing the shorthand in separate section makes it less discoverable.
  - Showing the same exact example twice is redundant and makes the docs hard to read.

This (more advanced shorthand documentation) may be something we leave for a separate PR, just wanted to get thoughts.

### DIFF
--- a/src/elements/Header/Header.js
+++ b/src/elements/Header/Header.js
@@ -52,13 +52,13 @@ function Header(props) {
     return <ElementType {...rest} className={classes}>{children}</ElementType>
   }
 
-  const iconNode = Icon.create(icon)
-  const imageNode = Image.create(image)
+  const iconElement = Icon.create(icon)
+  const imageElement = Image.create(image)
 
-  if (iconNode || imageNode) {
+  if (iconElement || imageElement) {
     return (
       <ElementType {...rest} className={classes}>
-        {iconNode || imageNode}
+        {iconElement || imageElement}
         {(content || subheader) && (
           <HeaderContent>
             {content}
--- a/src/elements/Header/Header.js
+++ b/src/elements/Header/Header.js
@@ -52,10 +52,13 @@ function Header(props) {
     return <ElementType {...rest} className={classes}>{children}</ElementType>
   }
 
-  if ((image && typeof image !== 'boolean') || (icon && typeof icon !== 'boolean')) {
+  const iconNode = Icon.create(icon)
+  const imageNode = Image.create(image)
+
+  if (iconNode || imageNode) {
     return (
       <ElementType {...rest} className={classes}>
-        {Icon.create(icon) || Image.create(image)}
+        {iconNode || imageNode}
         {(content || subheader) && (
           <HeaderContent>
             {content}
--- a/src/elements/List/List.js
+++ b/src/elements/List/List.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
@@ -33,6 +34,7 @@ function List(props) {
     floated,
     horizontal,
     inverted,
+    items,
     link,
     ordered,
     relaxed,
@@ -62,7 +64,15 @@ function List(props) {
   const rest = getUnhandledProps(List, props)
   const ElementType = getElementType(List, props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+  if (children) {
+    return <ElementType {...rest} className={classes}>{children}</ElementType>
+  }
+
+  return (
+    <ElementType {...rest} className={classes}>
+      {_.map(items, (item) => ListItem.create(item))}
+    </ElementType>
+  )
 }
 
 List._meta = {
@@ -106,6 +116,9 @@ List.propTypes = {
 
   /** A list can be inverted to appear on a dark background. */
   inverted: PropTypes.bool,
+
+  /** Shorthand array of props for ListItem. */
+  items: customPropTypes.collectionShorthand,
 
   /** A list can be specially formatted for navigation links. */
   link: PropTypes.bool,
--- a/src/elements/List/ListItem.js
+++ b/src/elements/List/ListItem.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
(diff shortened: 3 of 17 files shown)


src/elements/Header/Header.js
```diff
@@ -52,10 +52,13 @@ function Header(props) {
     return <ElementType {...rest} className={classes}>{children}</ElementType>
   }
 
-  if ((image && typeof image !== 'boolean') || (icon && typeof icon !== 'boolean')) {
```
jeffcarbs, line 55: These checks weren't entirely accurate. `image` or `icon` could have been passed as empty strings (valid itemShorthand) but still wouldn't be rendered. Also, the `!== boolean` is captured by the shorthand factory, which ignores booleans.

src/elements/Header/Header.js
```diff
@@ -52,10 +52,13 @@ function Header(props) {
...
+  const iconNode = Icon.create(icon)
+  const imageNode = Image.create(image)
```
levithomason: I've been moving these names to `fooElement`.  We had a number of `fooJSX` previously.  I think element is most accurate since factories return a ReactElement.  LMK if you disagree on the name, but I do want to make them consistent.

src/elements/Header/Header.js
levithomason: Cool, I'm on board with using the return of the factory to determine the markup.  Seems stronger to isolate that log in the factory's handling of the shorthand values.

src/elements/List/List.js
```diff
@@ -62,7 +64,15 @@ function List(props) {
...
   const ElementType = getElementType(List, props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+  if (children) {
+    return <ElementType {...rest} className={classes}>{children}</ElementType>
+  }
+
+  return (
+    <ElementType {...rest} className={classes}>
+      {_.map(items, (item) => ListItem.create(item))}
```
levithomason, line 73: Missing coverage here.  Codecov is still not updating PR status for some reason.  Killing me...

src/elements/List/ListItem.js
```diff
@@ -28,7 +40,38 @@ function ListItem(props) {
...
+
+  const headerNode = ListHeader.create(header)
+  const descriptionNode = ListDescription.create(description)
+
+  if (iconNode || imageNode) {
+    return (
+      <ElementType {...rest} className={classes} {...valueProp}>
+        {iconNode || imageNode}
+        {(content || headerNode || descriptionNode) && (
+          <ListContent>
```
levithomason, line 69: Coverage: 

image

src/elements/List/ListItem.js
```diff
@@ -50,11 +93,34 @@ ListItem.propTypes = {
...
+  description: customPropTypes.itemShorthand,
+
   /** A list item can disabled. */
   disabled: PropTypes.bool,
 
+  /** Shorthand for ListHeader. */
+  header: customPropTypes.itemShorthand,
+
+  /** Shorthand for ListIcon. */
+  icon: customPropTypes.every([
```
levithomason, line 134: What do you say we add `iconShorthand` and `imageShorthand` that can handle disallowing the each other?  This is also needed in the header, perhaps elsewhere.

levithomason (changes requested):

levithomason:
Agree on shorthand examples, I think there should be a shorthand example following every regular example.  See our resolution on this here https://github.com/Semantic-Org/Semantic-UI-React/issues/564#issuecomment-250924428.

You can also reference the [Feed docs](http://react.semantic-ui.com/collections/feed) which I think are a good example of shorthand following the regular example.

src/elements/List/ListItem.js
jeffcarbs: I think that's more of an exception, tbh. A quick search shows it's only applicable in the header and list while icon is a prop in over 20 component/subcomponents (although maybe we're not applying the `disallow` in some places we should be?).

In any case, the `*Shorthand` props includes a `disallow(['children'])` because that's required in _all_ cases. This is only applicable in some cases so I think it'll make it more clear to apply the `disallow` within the component's prop definitions that need it.

jeffcarbs:
@levithomason - added missing specs. Looks like the build notification is broken for codecov: https://codecov.io/gh/Semantic-Org/Semantic-UI-React/commit/edc973113439d5a74414f62203407b711b5baea9/builds.

>   2016-10-07T01:25:38-04:00 notify  POST /repos/Semantic-Org/Semantic-UI-React/statuses/edc973113439d5a74414f62203407b711b5baea9 bot=matheuspoleza code=404  
>       {"message":"Not Found","documentation_url":"https://developer.github.com/v3"}

Perhaps you need to update that key in the url?

levithomason:
> Perhaps you need to update that key in the url?

The key is the commit sha:

image

For some reason, making a POST to that endpoint gives a `404` saying it can't be found.  I'm betting this is due to renaming the original `Semantic-UI-React` repo to `Semantic-UI-React-Old`, then transferring `stardust` to `Semantic-UI-React`, and then deleting the original `Semantic-UI-React-Old` repo.  GitHub sets up redirects to handle all the moving around. I'm thinking the transfer/rename/delete has confused GH or one of these services.  Especially being I went back to TechnologyAdvice and created another `stardust` so I could redirect the old docs to the new docs.  This means there is a real repo there again, so links to it should not redirect any more.

The coverage reports are uploading to Codecov from CircleCI OK, and Codecov shows the correct coverage.  It also comments on the PR correctly.  It just seems the `/statuses` endpoint is not finding the repo.  I've reached out to Codecov support to see if they can help as well.

src/elements/List/ListItem.js
```diff
@@ -50,11 +104,48 @@ ListItem.propTypes = {
...
+   * This is handled slightly differently than the typical `content` prop since
+   * the wrapping ListContent is not used when there's no icon or image.
+   *
+   * If you pass content as:
+   * - an element/literal, it's treated as the sibling node to
+   * header/description (whether wrapped in Item.Content or not).
+   * - a props object, it forces the presence of Item.Content and passes those
+   * props to it. If you pass a content prop within that props object, it
+   * will be treated as the sibling node to header/description.
+   */
```
levithomason, line 121: I pushed an update to the `ComponentProps` component to preserve line breaks.  That way this doesn't render as a single ran together line in the prop table.

levithomason:
Made a fix to the simple `bulleted` example, it doesn't require the `bulleted` prop since the elements `ul` and `li` are provided.  Then, I added some descriptions to the other examples.  Will merge on pass.

levithomason (approved):

levithomason:

PR #640, opened 2016-10-06
refactor(Rating): add rating icon component

layershifter:
This PR:
- splites `Rating`'s icon to separate internal component;
- updates docs to unified format;
- minor refactors.

### DIFF
--- a/src/modules/Rating/Rating.js
+++ b/src/modules/Rating/Rating.js
@@ -25,7 +25,7 @@ const _meta = {
 
 /**
  * A rating indicates user interest in content
- * */
+ */
 export default class Rating extends Component {
   static autoControlledProps = [
     'rating',
@@ -53,6 +53,15 @@ export default class Rating extends Component {
       PropTypes.bool,
     ]),
 
+    /** The initial rating value. */
+    defaultRating: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+    ]),
+
+    /** You can disable or enable interactive rating.  Makes a read-only rating. */
+    disabled: PropTypes.bool,
+
     /** A rating can use a set of star or heart icons. */
     icon: PropTypes.oneOf(_meta.props.icon),
 
@@ -62,26 +71,17 @@ export default class Rating extends Component {
       PropTypes.number,
     ]),
 
+    /** Called with (event, { rating, maxRating }) after user selects a new rating. */
+    onRate: PropTypes.func,
+
     /** The current number of active icons. */
     rating: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number,
     ]),
 
-    /** The initial rating value. */
-    defaultRating: PropTypes.oneOfType([
-      PropTypes.string,
-      PropTypes.number,
-    ]),
-
     /** A progress bar can vary in size. */
     size: PropTypes.oneOf(_meta.props.size),
-
-    /** You can disable or enable interactive rating.  Makes a read-only rating. */
-    disabled: PropTypes.bool,
-
-    /** Called with (event, { rating, maxRating }) after user selects a new rating. */
-    onRate: PropTypes.func,
   }
 
   static _meta = _meta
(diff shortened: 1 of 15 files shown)


src/modules/Rating/Rating.js
```diff
@@ -21,6 +23,9 @@ const _meta = {
   },
 }
 
+/**
+ * A rating indicates user interest in content
+ * */
```
levithomason: `* */` => `*/`

levithomason (approved):
LGTM!

layershifter:
I'll update tests tomorrow, so we can merge it soon 👍

layershifter:
I added tests for `RatingIcon`, it was easy. 

I started updating of `Rating` tests for use `RatingIcon` instead of `i.icon`, but get stuck there. I added todos where test might be updated.

I don't understand how I can convert this to component test:

``` js
icons.find('.active').last().simulate('click')
icons.every('.icon.active')
  .should.equal(true, 'Some icon lost its "active" class')
```

If I correcty undestand [enzyme](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/find.md) docs, I can convert it to this?

``` js
icons.findWhere(n => prop('active', true)).last().simulate('click')
icons.everyWhere(n => prop('active', true))
  .should.equal(true, 'Some icon lost its "active" class')
```

levithomason:
You can also use the [object property selector](https://github.com/airbnb/enzyme/blob/master/docs/api/selector.md#4-object-property-selector):

``` jsx
icons.find({ active: true })
```

layershifter:
> You can also use the object property selector:

According to [airbnb/enzyme/#582](https://github.com/airbnb/enzyme/issues/582) - no, in `mount` mode, so I used `findWhere` there.

All tests are working with `RatingIcon` now, we're ready for final review there. Please, pay attension to `Rating` tests.

test/specs/modules/Rating/Rating-test.js
```diff
@@ -1,190 +1,199 @@
...
         .simulate('mouseEnter')
         .simulate('click')
 
-      icons.every('.icon.selected')
-        .should.equal(false, 'Some icon did not remove its "selected" class')
-
       wrapper.should.not.have.className('selected')
+
+      icons.findWhere((i) => i.prop('selected', true))
+        .should.have.length(0)
```
levithomason: Would prefer to keep the custom failure messages in these assertions.  Otherwise, a failed test will just say `Expected array to have length of 0, but it was X`.  Whereas, the custom message is more helpful `Some icon did not remove its "selected" prop`

test/specs/modules/Rating/Rating-test.js
```diff
@@ -1,190 +1,199 @@
...
-      const icons = wrapper.find('.icon')
+      const icons = wrapper.find('RatingIcon')
 
       icons.last().simulate('mouseEnter')
       wrapper.simulate('mouseLeave')
 
-      icons.every('.icon.selected')
-        .should.equal(false, 'Some icon did not remove its "selected" class')
+      icons.findWhere(i => i.prop('selected', true))
+        .should.have.length(0)
```
levithomason: Let's restore this failure message as well.

test/specs/modules/Rating/Rating-test.js
```diff
@@ -1,190 +1,199 @@
...
-        .find('.icon')
+        .find('RatingIcon')
 
-      icons.find('.active').last().simulate('click')
+      icons.last().simulate('click')
 
-      icons.every('.icon.active')
-        .should.equal(true, 'Some icon did not retain its "active" class')
+      icons.findWhere((i) => i.prop('active', true))
+        .should.have.length(5)
```
levithomason: Another failure message to restore.

test/specs/modules/Rating/Rating-test.js
```diff
@@ -1,190 +1,199 @@
...
-        .find('.icon')
+        .find('RatingIcon')
 
-      icons.find('.active').last().simulate('click')
+      icons.at(3).simulate('click')
 
-      icons.every('.icon.active')
-        .should.equal(false, 'Some icon did not remove its "active" class')
+      icons.findWhere((i) => i.prop('active', true))
+        .should.have.length(0)
```
levithomason: Looks like there are a lot of these so I'll skip commenting on the rest!  Though, let's update them all.  Note, the last argument to any assertion is the custom failure message.

levithomason (changes requested):
Looks good, let's just restore the test failure messages.

layershifter:
Restored messages to asserts and added some new. It's enought? Or I need add message to _each_ assert?

levithomason:
Released in `semantic-ui-react@0.55.1`, thanks once again!

PR #644, opened 2016-10-08
fix(Label): fix shorthand create method

levithomason:
This PR fixes 2 issues:
- left over deprecated `labeled` Button usages in the docs
- `Label.create` returned undefined

Heads up, we cannot put a `static create = createShorthandFactory()` call in a class, as was the Label.  The `Label` class is not yet defined, so the factory returns `undefined`.

I've updated the factories to throw if the Component is not a string or function.  I've also added tests for this.

### DIFF
--- a/docs/app/Examples/collections/Table/Types/TableApprove.js
+++ b/docs/app/Examples/collections/Table/Types/TableApprove.js
@@ -48,7 +48,7 @@ const TableApprove = () => {
         <Table.Row>
           <Table.HeaderCell />
           <Table.HeaderCell colSpan='4'>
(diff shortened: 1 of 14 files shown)


levithomason:
Still working on figuring out what is going on with PR build statuses.  This PR is passing and fully covered:

Tests passed: https://circleci.com/gh/Semantic-Org/Semantic-UI-React/56
100% covered: https://codecov.io/gh/Semantic-Org/Semantic-UI-React/pull/644

levithomason:
Released in `semantic-ui-react@0.54.4`

PR #645, opened 2016-10-08
test(Examples): assert no console errors

### LINKED ISSUE #599: docs(Example): Fix warnings

### LINKED ISSUE #604: RFC: Open questions around "Collection Shorthand"
See https://github.com/TechnologyAdvice/stardust/pull/567#discussion_r81444584 for a more detailed description of what I'm calling "Collection Shorthand".

---

I've been thinking about collection shorthand a little bit and have some open questions that I wanted to bring up so we can start figuring it out.
### ChildKey

When rendering the `items` it's necessary to generate a unique key for each item unless one is passed. In some cases it may be possible to do this given the props (see the rendering of each `MenuItem` [here](https://github.com/TechnologyAdvice/stardust/blob/master/src/collections/Menu/Menu.js#L166). However, there are issues even with this type of approach
- If the items are string/props shorthand, generally we'll be able to imply a key from those props. However, if the items are elements with only children, this wouldn't work (e.g. in the `MenuItem` case the content/name props would be blank).
- Certain types of data may not even have uniqueness amongst content. `TableRow > TableCell` comes to mind. Imagine a row where each cell is the string "n/a" - we'd be unable to generate a key just from the props.

Should we do any of the following:
- Fallback to index-based keys in cases where we can't guarantee uniqueness. This is an anti-pattern but it would work. Perhaps we do this and display a warning in dev like "no `key` or `childKey` was passed so we're falling back to index-based keys, which can yield unreliable results. You should specify keys yourself to guarantee uniqueness".
- Don't set a default childKey unless we're certain we guarantee uniqueness and just have the user see the `Warning: Each child in an array or iterator should have a unique "key" prop`.

When you're missing keys, React needs to re-render the whole collection because it can't be sure. When you generate duplicate keys for different elements React squashes the duplicates and only renders one. So getting it wrong is much worse than just not doing anything.
### Interacting with child props

In some components, it's necessary to be able to use the props for each item created by the shorthand. For example, there's an open issue to allow specifying string elements for `Dropdown` options (#470). In this case, I think it makes sense to make to use shorthand. However, we need to do things like filtering/searching, which rely on turning a literal into a props object or being able to use the props of a passed React element.

I'm thinking we can do something like this:
- whenever props change (e.g. componentWillMount, componentWillReceiveProps), if the `items` have changed we run something like (e.g. for `Dropdown`):

``` js
this.setState({ items: _.map(this.props.items, (item, index) => DropdownItem.create(item, defaultProps) })
```
- now whenever we do anything with the items, they will be always be fully "hydrated" react elements so we can interact with them in a consistent way (e.g. `item.props.value`) instead of needing to do something like `_.isObject(item) ? item.value : item`.

---

There may be some other issues so feel free to add them here for discussion.

Thoughts?

levithomason:
Fixes #604 
Fixes #599 

This PR adds a test that renders all doc site examples and asserts that no console errors were logged.  I'll fix the errors as well before merging.  I may also extend this to render all doc site routes instead, that way every page is covered and we know there are no console errors in the docs.

```
FAILED TESTS:
  examples
    ✖ render without errors
      PhantomJS 2.1.1 (Mac OS X 0.0.0)
      Example File: ./collections/Menu/Types/Attached.js
      Console Error: Warning: Failed prop type: `children` prop in `Dropdown` requires props: `text`.
          in Dropdown

      Example File: ./elements/Input/Variations/InputActionDropdown.js
      Console Error: Warning: Failed prop type: `options` prop in `Dropdown` requires props: `selection`.
          in Dropdown

      Example File: ./views/Card/Content/ContentBlock.js
      Console Error: Warning: Failed prop type: Prop `date` in `FeedContent` conflicts with props: `children`. They cannot be defined together, choose one or the other.
          in FeedContent

      : expected [ Array(3) ] to be empty

      Stack:
      test/specs/docs/examples-test.js:27:0
```


jeffcarbs:
This is great!

While writing examples for `List` shorthand I discovered a bug in how the shorthand was being rendered. Got me thinking - could we automate tests for full markup and shorthand to ensure they produce the same html? This would expose errors in our use of the shorthand or errors within the components themselves - both things we should definitely fix! For example:

``` js
exampleContext.keys().forEach(path) => {
  if (!path.match(/Shorthand$/)) return

  // Makes use of the naming convention of examples
  // Full example: ListExampleBasic
  // Shorthand example: ListExampleBasicShorthand
  const fullExamplePath = path.replace(/Shorthand$/, '')

  const FullExample = createElement(exampleContext(fullExamplePath).default)
  const ShorthandExample = createElement(exampleContext(path).default)

  render(ShorthandExample).should.equal(render(FullExample))
});
```

levithomason:
That sounds pretty sane.  I think we'd need to call [`html()`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/html.html) on the wrapper to compare the html strings, but the point still stands.

I also added `Foo.create()` common tests in https://github.com/Semantic-Org/Semantic-UI-React/pull/644.  The Label create method returned undefined since it was passing `Label` inside of its own class.  The common tests `implementsCreateMethod` just asserts it exists and creates elements for every supported shorthand value.

levithomason:
I updated the test to `mount()` each example, which revealed far more errors.  I've fixed them all except for those with `key` errors.  We need to solve how factories will generate keys, or require the user to pass a `key`, in order to solve this.

Will pick this up later.

```
FAILED TESTS:
  examples
    ✖ renders without errors: ./collections/Table/States/TableWarningShorthand.js
      PhantomJS 2.1.1 (Mac OS X 0.0.0)
      Console error: Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Table`. See https://fb.me/react-warning-keys for more information.
          in TableRow (created by Table)
          in Table (created by Unknown)
          in Unknown (created by Unknown)
          in Unknown,Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `TableRow`. See https://fb.me/react-warning-keys for more information.
          in TableCell (created by TableRow)
          in TableRow (created by Table)
          in thead (created by TableHeader)
          in TableHeader (created by Table)
          in table (created by Table)
          in Table (created by Unknown)
          in Unknown (created by Unknown)
          in Unknown: expected error to not have been called

      Stack:
      test/specs/docs/examples-test.js:21:0

    ✖ renders without errors: ./elements/List/Types/ListExampleBasicShorthand.js
      PhantomJS 2.1.1 (Mac OS X 0.0.0)
      Console error: Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `List`. See https://fb.me/react-warning-keys for more information.
          in ListItem (created by List)
          in List (created by Unknown)
          in Unknown (created by Unknown)
          in Unknown: expected error to not have been called

      Stack:
      test/specs/docs/examples-test.js:21:0
```

levithomason:
I've pushed some work in progress for creating `key` props for children.  I think we should consider a `createCollection` method.  This way, we know when to create keys.

The `create` method has no way of telling if it is being called due to mapping over a shorthand collection.  Therefore, it has no way of knowing whether or not it should generate a key if none were provided.

If we allow a `createCollection` method, then we can reliably attempt to generate keys and also warn users in development mode if we so choose.  Currently, I'm experimenting with generating a hash code key based on a shallow stringified final props object.  So it's as unique as the full shallow prop signature.  This is only if there were no `key` nor `childKey` provided.  If we only do this when creating a collection (ie `createCollection`) then I think this could be the way to go.

levithomason:
I've split `createShorthandItem` and `createShorthandCollection`.  The latter handles retrieving child keys or generating them.

I just need to finish updating the TableRow.create to utilize the TableCell factory.  It seems collection factories should be responsible for passing the item factory that handles their array values.  Or, use it in the map value to props method.

src/collections/Table/Table.js
```diff
@@ -84,7 +84,7 @@ function Table(props) {
     <ElementType {...rest} className={classes}>
       {headerRow && <TableHeader>{TableRow.create(headerRow, { cellAs: 'th' })}</TableHeader>}
       <TableBody>
-        {renderBodyRow && _.map(tableData, (data, index) => TableRow.create(renderBodyRow(data, index)))}
+        {renderBodyRow && TableRow.create(_.map(tableData, (data, index) => renderBodyRow(data, index)))}
```
jeffcarbs: I think this change is incorrect. This should map over each item in the `tableData` array and create a `TableRow` from each one. This change would map over each item in the `tableData` array and create one giant `TableRow`.

levithomason:
Interestingly in React v15.4 there are no longer warnings for missing keys nor duplicate keys.  It still appears missing or duplicate keys results in overridden child nodes when rendered to the DOM.

I believe I solved all the remaining issues.  I'll come back and clean up and confirm this when I get time.

levithomason:
Haha, there are no warnings because the docs started using React min via CDN...

docs/app/Examples/collections/Menu/Types/MenuExampleAttached.js
```diff
@@ -10,7 +10,7 @@ const MenuExampleAttached = () => {
         <Dropdown as={Menu.Item} icon='wrench' simple>
           <Dropdown.Menu>
             <Dropdown.Item>
-              <Icon name='dropdown icon' />
+              <Icon name='dropdown' />
```
levithomason, line 13: You'll see several fixes to the docs like this one.

docs/app/Examples/elements/List/Content/ListExampleDescription.js
```diff
@@ -4,14 +4,14 @@ import { List } from 'semantic-ui-react'
 const ListExampleDescription = () => (
   <List>
     <List.Item>
-      <List.Icon name='map marker' />
+      <List.Icon name='marker' />
```
levithomason, line 7: This was wrong in core docs as well.  There is on a `map` or `marker` icon, not a `map marker`.  The prop type checker is sooo helpful for these :D 

image

docs/app/utils.js
```diff
@@ -18,7 +18,7 @@ export const parentComponents = _.flow(
 /**
  * Get the Webpack Context for all doc site examples.
  */
-export const exampleContext = require.context('docs/app/Examples/', true, /\.js$/)
+export const exampleContext = require.context('docs/app/Examples/', true, /(\w+Example\w+|index)\.js$/)
```
levithomason, line 21: This now excludes non-renderable other files, like `common.js`.  It is used in a few places, but only for rendering the example view components themselves.

>P.S.  The `common.js` files actually need to go as we cannot truly import any files into the examples since we're in a browser.  That is for another PR.

src/collections/Breadcrumb/BreadcrumbSection.js
```diff
@@ -93,4 +93,4 @@ export default class BreadcrumbSection extends Component {
   }
 }
 
-BreadcrumbSection.create = createShorthandFactory(BreadcrumbSection, content => ({ content, link: true }))
+BreadcrumbSection.create = createShorthandFactory(BreadcrumbSection, content => ({ content, link: true }), true)
```
levithomason, line 96: This is how we now flag a factory to `generateKey`.  It will generate a key as a hash of shallow props.

src/lib/customPropTypes.js
```diff
@@ -251,6 +251,7 @@ export const contentShorthand = (...args) => every([
 export const itemShorthand = (...args) => every([
   disallow(['children']),
   PropTypes.oneOfType([
+    PropTypes.array,
```
levithomason, line 254: Fixes warnings with `tableFooter`.  This prop is shorthand for a Row but the array literal case is passed to the row cells via `mapValueToProps`.  Since the shorthand item factory supports arrays, the shorthand item type should as well.

src/lib/factories.js
```diff
@@ -2,86 +2,146 @@ import _ from 'lodash'
...
-    newProps.className = cx(defaultProps.className, props.className) // eslint-disable-line react/prop-types
+const getHashCode = function hashCode(str) {
+  const { length } = str
+  if (!str || length === 0) return 0
+  let hash = 0
+  for (let i = 0; i < length; i++) {
+    hash = hash * 31 + str.charCodeAt(i) | 0
   }
+  return hash
+}
```
levithomason: When we generate keys we make a prop string, then we make a hash of that.  The only added benefit of the hash is having very short keys.  The downside is more processing at render time.  We could just leave the shallow stringified props as the key.  DOM weight is not an issue here as the key is in memory only.

I'm on the brink of dropping it already, LMK your thoughts.

src/lib/getUnhandledProps.js
```diff
@@ -37,6 +37,7 @@ const getUnhandledProps = (Component, props) => {
   // Return _unhandled_ props
   // ----------------------------------------
   return Object.keys(props).reduce((acc, prop) => {
+    if (prop === 'childKey') return acc
```
levithomason, line 40: Always "handle" the childKey prop so we don't apply it in factories.

test/specs/collections/Table/Table-test.js
```diff
@@ -59,7 +59,11 @@ describe('Table', () => {
 
     const headerRow = ['Name', 'Status', 'Notes']
 
-    const renderBodyRow = ({ name, status, notes }) => [name || '', status || '', notes || '']
+    const renderBodyRow = ({ name, status, notes }) => [
+      name || { key: 0 },
+      status || { key: 1 },
+      notes || { key: 2 },
```
levithomason, line 65: Since these are cells and in the event that there are multiple missing cells in the `tableData` (which there are) then the cells will all have identical props objects and therefore identical generated keys.

I've instead defaulted to objects with a key defined.

test/specs/docs/examples-test.js
```diff
@@ -0,0 +1,37 @@
...
+    console.log.restore()
+    console.warn.restore()
+    console.error.restore()
+  })
+  exampleContext.keys().forEach(path => {
+    // don't test index files, they repeat errors of individual files
+    if (/index\.js$/.test(path)) return
+    const filename = path.replace(/^.*\/(\w+\.js)$/, '$1')
+
+    it(`${filename} renders without console messages`, () => {
```
levithomason, line 27: 🎉

levithomason (commented):
Woop! I believe this ready to go! I've self-reviewed and add some inline notes.  Would be great to get some eyes on this from one of @jcarbo, @davezuko, or @layershifter.

Thanks much!

levithomason:
🎉 🎉 🎉

PR #646, opened 2016-10-08
fix(ItemImage): wrap image to prevent vertical strech

typehorror:
Address vertical stretch of images within item view

screen shot 2016-10-08 at 7 44 48 am


levithomason:
Oh boy, thanks!  Can we update the assertion in `ItemImage-test.js` to include assertions for `ui={false}` and `wrapped`?

typehorror:
@levithomason done!

levithomason:
Released in `semantic-ui-react@0.55.1`, thanks much!

PR #647, opened 2016-10-08
fix(confirm): use the new modal props, deprecate active for open

typehorror:
I'm not sure how you feel about adding the warning library ( just 62 LoC) into the repository.

This PR updates the deprecated lib function with a simpler one (dropping the validation method) for a simpler use case. 

Warning is only triggered on DEV: https://github.com/BerkeleyTrue/warning/blob/master/warning.js#L19

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
   "dependencies": {
     "classnames": "^2.1.5",
     "debug": "^2.2.0",
-    "lodash": "^4.6.1"
+    "lodash": "^4.6.1",
+    "warning": "^3.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.5.1",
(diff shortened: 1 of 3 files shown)


levithomason:
We used to keep deprecation warnings for components and props, however, with all the pre 1.0 churn it is a lot to manage.  Until we hit v1, there will be no deprecation warnings.  We're doing everything we can to trim overhead until we hit v1.

Per the readme:

image

Once we hit v1, I'm definitely in agreement 👍

typehorror:
Hey @levithomason ,
I understand the decision and sorry for not seeing the note. I see 2 options here:
- If you think we should switch from active to open for the confirm addons, I'm willing to create a PR for that (which updates the docs too). 
- if not, I can fix confirm to just associate `active` props with `open` which won't generate any deprecation.

levithomason:
Fixed Confirm props in `semantic-ui-react@0.55.0`

PR #648, opened 2016-10-08
feat(Confirm): switch from active prop to open

typehorror:
- Switch from `Confirm.props.active` to `Confirm.props.open`
- Use `createShorthandFactory` on Modal shorthands
- Use the shortHand create method for Confirm content
- Use children as modal content **or** props shorthands


levithomason:
This work is ongoing in #634

levithomason:
Sorry, just saw your comment on #647 about making this. Had I seen it early, I would've pointed you to the work in #634 to avoid the duplicate!

typehorror:
No worries, that's what I guessed 👍

levithomason:
Fixed in `semantic-ui-react@0.55.0`

Issue #517, opened 2016-09-20
feat(Dropdown): add `openOnFocus` and `closeOnBlur`


njnash:
Right now it seems as though the content you can add to a Dropdown Module is a Header, and then a list of items. Is it possible to support other types of content, such as a Form?

levithomason:
Can you provide some example code and expected output?  Then I can help guide you here.

njnash:
Here is an example:

``` js
<Dropdown trigger={trigger} pointing='top right' icon={null}>
   <Dropdown.Menu>
     <Form>
       <Header>
          New
       </Header>
     <Form.Field>
        <label>Name</label>
        <input placeholder='enter name' />
     </Form.Field>
     <Form.Select label='Select from drop down' options={List} placeholder='select one...' />
     <Button className="basic">Cancel</Button>
     <Button className="basic">Submit</Button>
     </Form>
  </Dropdown.Menu>
</Dropdown>
```

I expected that the form inside the dropdown would be functional. However when I go to type in the Form Field or use the Form Select the dropdown disappears and I cannot use the form.

levithomason:
I believe in this case you are looking for the functionality of a Popup (see #193, not implemented yet) rather than a Dropdown.  A Popup is meant to show additional info over a page.  A Dropdown is meant to allow a user to select a value from a series of options, hence the Dropdown.Menu.

Here's an example Popup that would work well for your case:

![http://g.recordit.co/1aSlDjwIix.gif](http://g.recordit.co/1aSlDjwIix.gif)

What we might be able to do in the meantime is provide two Dropdown props I've been wanting to add, `closeOnBlur` and `openOnFocus`.  If these were available, you could use the Dropdown as a workaround for now by setting `closeOnBlur={false}`.

I'd accept a PR that added these two props.

levithomason:
I've update this issue to request those props be added.  This should support your use case temporarily while waiting for the Popup.  It would be fantastic if you could contribute either to the Dropdown, or ideally, the Popup.

njnash:
Great, thank you so much for the detailed response! It looks like the Popup component is exactly what I want. Any guess on the timeframe for that?

levithomason:
See the issue comments for details.

njnash:
[...]

Issue #509, opened 2016-09-19
RFC: Abstract base component to handle similarities b/w Dropdown and Search (and possibly Select as a unique component)


jeffcarbs:
### Background

When building the `Search` component (#424) I wound up using the `Dropdown` as a guide because it was very similar. User interacts with an element (in the case of `Search`, the search input) and a menu opens which you can page through using arrow keys and select an option. Things like handling focus, blur, clickaway, etc were already handled in `Dropdown` so I mostly copy/pasted the relevant implementations.
#### Search vs Dropdown/Select

There are a lot of similarities between these two elements, but I think they fall into two overlapping-but-different chunks of functionality:
- Floating menu that appears based on some user interaction (focus/blur/click/typing)
- List of options that you can page through via arrow up/down and select

These components have a basic structure of:

```
<wrapping element>
  <some content />
  <menu (hidden/visible depending on state/props) />
</wrapping element>

// Search
<Search>
  <Input />
  <Results />
</Search>

// Dropdown
<Dropdown>
  <Trigger />
  <Menu />
</Dropdown>

// Select
<Select>
  <Text (hidden input, labels, search, search sizer, text) />
  <Menu />
</Select>
```

It might be possible to have an abstract component that each of components wrapped.
- The abstract component would be responsible for maintaining the open/focus states and could be configured to open/close based on subset of interactions (focus, blur, etc) via props.
- The wrapping component would pass the "content" and "menu" as props for the underlying component to render.
#### Dropdown vs Select

In addition to similarities with search, I think the `Dropdown` currently has a two very different use-cases that are both built into the single component (I mentioned this on a previous thread):
- Take the place of a select element and have a "value".
- A hidden menu that's shown after clicking a "trigger" element.

I know SUI only has the one `Dropdown` component, but if you look at the SUI examples page you can see that there are TONS of variations. I think it would make these components simpler if we were able to break the two use-cases apart. There's already a `Select` component in this repo but it's just a light wrapper that adds one prop to `Dropdown`.

If we were to pull these apart, we could have:
- `Dropdown`
  - Takes a list of options and renders them
  - Keeps state for: "open", "focus", "selectedIndex" (for arrowing up/down)
  - No inherent notion of "value" - "active" could be passed as a prop on the relevant option.
- `Select`
  - Wrapper around `Dropdown`.
  - Passes the hidden input, search box, text, as the "trigger"
  - Handles filtering of options.
  - Keeps "value" state

---

[...]

PR #653, opened 2016-10-10
feat(Dropdown): Refactor dropdown

jeffcarbs:
⚠️ **NOTE: This was closed in favor of #672**

---

This PR intends to clean up the `Dropdown` component and address some open issues (e.g. #517 and #509.

The rough goals:
- Separate `Dropdown` and `Select` functionality:
  - `Dropdown` is renders a `.ui.dropdown` with a "trigger" and "menu". Meant to be used for static-ish dropdown menus.
  - `Select` will render a dropdown under the hood, but will maintain a value and add search functionality. Meant to be used in place of an html `select`.
- Defer the show/hide functionality to `Portal` (or some other shared component).

The plan is to break out the `Dropdown` to the simplest form, then add back functionality/shorthand as it makes sense.

TODO:
- [x] Use portal for dropdown show/hide
- [x] Remove select-related functionality
- [ ] Add select-related functionality via the `Select` component

Open questions:
- I'm reusing the `Portal` class and adding an `inline` class which renders the "portal" as a wrapping element (e.g. `<div class='ui dropdown' />`). The main reason is that there's already a bunch of open/close configuration. However, in this case it's no longer really a "portal" and the rendering is pretty different which leads to some `if (inline)` flow control. I'm on the fence about making this a different component with a similar interface. Thoughts?

### DIFF
--- a/docs/app/Examples/modules/Dropdown/index.js
+++ b/docs/app/Examples/modules/Dropdown/index.js
@@ -4,6 +4,75 @@ import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
 
 import { Message } from 'semantic-ui-react'
 
+// const DropdownExamples = () => (
+//   <div>
+//     <ExampleSection title='Types'>
+//       <ComponentExample
+//         title='Dropdown'
+//         description='A dropdown menu'
+//         examplePath='modules/Dropdown/Types/Dropdown'
+//       />
+//       <ComponentExample
+//         title='Custom Trigger'
+//         description='A dropdown can use a custom trigger element'
+//         examplePath='modules/Dropdown/Types/Trigger'
+//       />
+//       <ComponentExample
+//         examplePath='modules/Dropdown/Types/ImageTrigger'
+//       />
+//       <ComponentExample
+//         title='Uncontrolled'
+//         description='A dropdown works as an uncontrolled component'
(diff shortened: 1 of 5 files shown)


layershifter:
> I'm on the fence about making this a different component with a similar interface.

Agree, I don't think that it's `Portal` use case.

jeffcarbs:
That's where I'm leaning as well. I wanted to be able to pull out that functionality since it's shared by some other components (e.g. `Search`) but for now I think we may just be better off keeping that functionality into the `Dropdown` itself.

levithomason:
I'm hesitant to use the Portal here.  We don't need to transport any components to new containers AFAICT.  The code for handling close on click outside / escape is very minimal.  I'm not sure we'd gain much else by adding the Portal.  Though, it does add some complexity due to how it renders.

We also intend to reuse SUI CSS for animations, that means we need to the SUI visibility class transitions.  You can see the POC Dropdown animation work that was done [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/TODO.md).  We won't do it exactly as shown in the POC, but it does show we're able to use component props (ie SUI classNames) to handle animations.  I'm not sure what issues the Portal would bring to this.

jeffcarbs:
Points taken re: the `Portal`. Going to close this in favor of #672

Issue #642, opened 2016-10-06
feat(Dropdown): support MenuItem and Button types


Chris-R3:
### Steps

I'm using a `Dropdown as Menu.Item` and noticed that the width of the menu item seems off. I checked the [Semantic-UI](http://semantic-ui.com/collections/menu.html#menu) docs (3rd example) and they are using an additional `icon` class for the menu item if the `Dropdown` only contains an `Icon`.
### Expected

`class="ui dropdown icon item"`
### Result

`class="ui dropdown item"`
### Workaround

adding `className='icon'`as additional `Dropdown`prop
### Testcase

http://codepen.io/Chlris-R3/pen/PGQbqZ
- left: without icon class
- right: using `className='icon'` as additional `Dropdown`prop

Is this a bug / missing feature / working as intended?

levithomason:
Confirmed issue.  We're lacking feature support here.  This one is going to need some thought.  The `icon` class is required only in limited circumstances.  There are multiple Dropdown usages to consider here:

## Menu Item

When a Dropdown is a Menu Item:
- only the `item` className is required
- the `icon` child must appear _after_ the `text` child (regular Dropdowns have `text` _then_ `icon`)
- if there is an `icon` and no `text` nor `placeholder`, then the `icon` className is required

``` html
<div class="ui dropdown icon item">
```

![image](https://cloud.githubusercontent.com/assets/5067638/19201334/28dfa75e-8c82-11e6-80f5-fd17a659844e.png)

## Button

When the Dropdown is a Button
- the `icon` child must appear _after_ the `text` child (regular Dropdowns have `text` _then_ `icon`)
- the `icon` className is required if there is an an `icon` with no `text` nor `placeholder`
- the `icon` className is required if there is an an `icon` with text and it is `labeled`
- the `icon` className is _not_ required if the Dropdown button does not also have an icon

### Labeled Icon Button Dropdown

``` html
<div class="ui floating labeled icon dropdown button">
```

![image](https://cloud.githubusercontent.com/assets/5067638/19199117/436ac73a-8c77-11e6-8769-25745f55cf3e.png)

### Icon Button Dropdown

``` html
<div class="ui icon top left pointing dropdown button">
```

![image](https://cloud.githubusercontent.com/assets/5067638/19199185/a2754502-8c77-11e6-946e-af94e7d360f0.png)

## Conclusion

[...]

PR #659, opened 2016-10-11
fix(Dropdown): use `item` instead of `as={Menu.Item}`

levithomason:
# Update

Augmenting the Dropdown with a `Menu.Item` is no longer supported.  Instead, use the `item` prop when using a Dropdown as a menu item.  See the changes in this PR for more:

### Before
```jsx
<Dropdown as={Menu.Item} />
```
### After
```jsx
<Dropdown item />
```

# Original Post
Per https://github.com/Semantic-Org/Semantic-UI-React/pull/628#issuecomment-252592636, this PR checks for `this._dropdown.blur()` before calling it.  When augmented, the `ref` may be a composite component ref instead of an actual DOM node.  Otherwise, we sometimes get: `Uncaught TypeError: a._dropdown.blur is not a function`.

### DIFF
--- a/docs/app/Examples/collections/Menu/Content/MenuExampleDropdownItem.js
+++ b/docs/app/Examples/collections/Menu/Content/MenuExampleDropdownItem.js
@@ -4,7 +4,7 @@ import { Dropdown, Menu } from 'semantic-ui-react'
 const MenuExampleDropdownItem = () => {
   return (
     <Menu vertical>
-      <Dropdown as={Menu.Item} text='Categories'>
+      <Dropdown item text='Categories'>
         <Dropdown.Menu>
           <Dropdown.Item>Electronics</Dropdown.Item>
           <Dropdown.Item>Automotive</Dropdown.Item>
--- a/docs/app/Examples/collections/Menu/Content/MenuExampleSubMenu.js
+++ b/docs/app/Examples/collections/Menu/Content/MenuExampleSubMenu.js
@@ -39,7 +39,7 @@ export default class MenuExampleSubMenu extends Component {
           Messages
         </Menu.Item>
 
-        <Dropdown as={Menu.Item} text='More'>
+        <Dropdown item text='More'>
           <Dropdown.Menu>
             <Dropdown.Item icon='edit' text='Edit Profile' />
             <Dropdown.Item icon='globe' text='Choose Language' />
--- a/docs/app/Examples/collections/Menu/Types/MenuExampleAttached.js
+++ b/docs/app/Examples/collections/Menu/Types/MenuExampleAttached.js
@@ -7,7 +7,7 @@ const MenuExampleAttached = () => {
   return (
     <div>
       <Menu attached='top'>
-        <Dropdown as={Menu.Item} icon='wrench' simple>
+        <Dropdown item icon='wrench' simple>
           <Dropdown.Menu>
             <Dropdown.Item>
               <Icon name='dropdown' />
--- a/docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
+++ b/docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
@@ -13,7 +13,7 @@ export default class MenuExampleVerticalDropdown extends Component {
       <Menu secondary vertical>
         <Menu.Item name='account' active={activeItem === 'account'} onClick={this.handleItemClick} />
         <Menu.Item name='settings' active={activeItem === 'settings'} onClick={this.handleItemClick} />
-        <Dropdown as={Menu.Item} text='Display Options'>
+        <Dropdown item text='Display Options'>
           <Dropdown.Menu>
             <Dropdown.Header>Text Size</Dropdown.Header>
             <Dropdown.Item>Small</Dropdown.Item>
--- a/docs/app/Examples/collections/Menu/Variations/MenuExampleSizeHuge.js
+++ b/docs/app/Examples/collections/Menu/Variations/MenuExampleSizeHuge.js
@@ -15,7 +15,7 @@ export default class MenuExampleSizeHuge extends Component {
         <Menu.Item name='messages' active={activeItem === 'messages'} onClick={this.handleItemClick} />
 
         <Menu.Menu position='right'>
-          <Dropdown as={Menu.Item} text='Language'>
+          <Dropdown item text='Language'>
             <Dropdown.Menu>
               <Dropdown.Item>English</Dropdown.Item>
               <Dropdown.Item>Russian</Dropdown.Item>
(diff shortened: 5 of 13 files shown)


levithomason:
Though this stops the error, we still have the bug where clicking an item closes the menu, but then switching tabs and switching back opens the dropdown again.  This is because it still has focus.

This PR should be updated to actually blur the composite component use case.

eXtreme:
@levithomason 
BTW. I discovered that it is better to do `<Menu.Item as={Dropdown}>` instead of `<Dropdown as={Menu.Item}>` - it does not trigger the error and also generates a correct markup: the first one makes `<div>` for menu.item with dropdown items as `<a>`) while the latter makes `<a>` for menu.item causing nesting of `<a>` (of dropdown.item) inside `<a>` of menu.item/dropdown.

tomitrescak:
@eXtreme would you care to provide a full example? I tried several variants of using menu.item with dropdown but no luck

eXtreme:
@tomitrescak 

``` jsx
<Menu>
    <Menu.Item>
        Foo
    </Menu.Item>
    <Menu.Item as={Dropdown} text="Imma dropdown">
        <Dropdown.Menu>
            <Dropdown.Item>Foo</Dropdown.Item>
            <Dropdown.Item>Bar</Dropdown.Item>
            <Dropdown.Divider />
            <Dropdown.Item>Baz</Dropdown.Item>
        </Dropdown.Menu>
    </Menu.Item>
</Menu>
```

Also works with custom `trigger={<span>bar</span>}`

Chris-R3:
How about impementing a `blur()` method on `Menu.Item`?
Or do you have any suggestions/ideas on how to approach this issue @levithomason?

levithomason:
We'll mostly likely do away with using augmentation for this use case.  The Dropdown can play the role of a MenuItem or Button, but augmentation is not the right tool for the job.  See here: https://github.com/Semantic-Org/Semantic-UI-React/issues/642.

giacomorebonato (approved):

levithomason:
Note https://github.com/Semantic-Org/Semantic-UI-React/pull/659#issuecomment-252990870, this PR is invalid as it circumvents the error by simply removing the feature.  It not longer blurs when augmenting a functional component.  It needs to be updated to blur in all cases.

giacomorebonato:
I am trying to guess what the code should do here...
If I replace  

``` js
_this._dropdown.onBlur();
```

with

``` js
_this._dropdown.props.onBlur()
```

It isn't what you expect right?

levithomason:
No, `blur()` doesn't exist at all.  The reason is that a `ref` returns a DOM Node.  The DOM Node interface has a `blur()` method, but a React `ref` does not work for functional components so there is no DOM Node to call blur on.  Moreover, a `ref` on a composite component does not return a DOM Node either, it returns a reference to the underlying component instance.

So, there are 3 possible returns values from the ref `this._dropdown`, a DOM Node, `null`, and a function.  The current code only handles the DOM Node case.  It blindly fails for the `null` and function cases.

The real hang up here on this is the `null` case, which is what a stateless functional component will return.  We need to blur the Node for the Dropdown to work right, but without a working `ref` there is no way to do that.  So, we're gonna have to do some component wrapping gymnastics to get that to work.

nhducit:
@levithomason It seems this PR didn't fix #961. I can reproduce the #961 issue

Maxhodges:
*bump* any big reason why this can't be merged? I have a production app which is suffering

eXtreme:
@Maxhodges have you tried my "workaround" from [this comment](https://github.com/Semantic-Org/Semantic-UI-React/pull/659#issuecomment-253261831)?

Maxhodges:
ah yes, that worked for me. just nicer if I didn't have to workaround ;) 
cheers! @eXtreme

levithomason:
@Maxhodges refer to the comments I've made above for why this isn't yet merged.

levithomason:
There is a larger issue here that needs solved.  We need a consistent and portable way to deal with refs on stateless functional components.  The same solution for this issue will solve https://github.com/Semantic-Org/Semantic-UI-React/issues/1012.

raptoria:
@eXtreme tried your workaround..but Menu.Item doesn't take a text attribute anymore, it seems. Any work around so that I can see a bit of text associated with the dropdown?

eXtreme:
@raptoria How is that not working for you? I've tried changing an example on http://react.semantic-ui.com/collections/menu and it works fine with text property.

image

From what I understand, it's RSui's augmentation, which makes the "text" property being actually passed to Dropdown component, as you may see from inspecting with react devtools:

image

"t" is actually the "Dropdown" component, which takes "text" property.

The resulting DOM is almost the same as in original SUI (in RSui it has `ui item dropdown` classname, while in SUI it is `ui dropdown item`) but the effect is the same.

Can you provide an example of this workaround not working?

raptoria:
@eXtreme I should clarify. Your workaround works, it just gives me a TypeScript error. I'm a TypeScript n00b so I'm not really sure how to fix it.  I guess the problem has to do with the text attribute not being explicitly defined in MenuItem. 

ERROR in ./src/components/FileMenu/FileMenu.tsx
(18,32): error TS2339: Property 'text' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<MenuItemProps, ComponentState>> & { chil...'.

this is the code my React component is returning:
```
      <Menu inverted className={styles.fileMenu}>
      <Menu.Item as={Dropdown} text='File'>
          <Dropdown.Menu className={styles.dropdownMenu}>
            <Dropdown.Item icon='file outline icon' text='New Order Entry' />
            <Dropdown.Item icon='shopping basket icon' text='New Basket' />
            <Dropdown.Item icon='download icon' text='Load Portfolio' />
            <Dropdown.Item icon='file excel outline icon' text='Export to Excel' />
            <Dropdown.Item icon='sign out icon' text='Quit Application' />
          </Dropdown.Menu>
        </Menu.Item>
        <Menu.Item name='Edit'> Edit &nbsp;<Icon name='caret down icon' /></Menu.Item>
        <Menu.Item name='Tools'> Tools &nbsp;<Icon name='caret down icon' /></Menu.Item>
        <Menu.Item name='Views'> Views &nbsp;<Icon name='caret down icon' /></Menu.Item>
      </Menu>
```
any ideas on how to get rid of the error?

Thanks!!

levithomason:
@raptoria, and others who come here looking for a solution, it is most likely that using augmentation (the `as` prop) **will not be the recommended solution** for this case.  A more appropriate workaround, for now, is to simply add `className='item'` to the Dropdown so it is formatted properly within the Menu.

```jsx
<Menu attached='top'>
  <Dropdown className='item' icon='wrench' />
</Menu>
```

This prevents `Uncaught TypeError: a._dropdown.blur is not a function` since the Dropdown is not attempting to render as a stateless component but remains a class component.

raptoria:
@levithomason thanks! it works perfectly now.

levithomason:

PR #663, opened 2016-10-12
docs(confirm): add trailing . to description

dpkwhan:


levithomason:
Looks like this PR also includes many changes from the Radio PR.

levithomason:
Cool, looks like the extra files are gone.

PR #672, opened 2016-10-13
feat(Dropdown): Refactor dropdown 

jeffcarbs:
This was started as #653 but wound up taking a slightly different approach so I opened this instead.

---

This PR intends to:
- simplify/clean up the `Dropdown` component.
- address some open issues (e.g. #517 and #509).
- separate `Dropdown` and `Select` functionality
  - `Dropdown` is renders a `.ui.dropdown` with a "trigger" and "menu". Meant to be used for static-ish dropdown menus.
  - `Select` will render a dropdown under the hood, but will maintain a value and add search

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Types/Selection.js
+++ b/docs/app/Examples/modules/Dropdown/Types/Selection.js
@@ -1,7 +1,7 @@
 import _ from 'lodash'
 import faker from 'faker'
 import React, { Component } from 'react'
-import { Button, Dropdown, Grid, Header } from 'semantic-ui-react'
+import { Button, Select, Grid, Header } from 'semantic-ui-react'
 
 const getOptions = () => _.times(3, () => {
   const name = faker.name.findName()
@@ -65,7 +65,7 @@ export default class DropdownSelectionExample extends Component {
               <input type='checkbox' checked={multiple} onChange={this.toggleMultiple} /> Multiple
             </label>
           </p>
-          <Dropdown
+          <Select
             fluid
             selection
             multiple={multiple}
--- a/docs/app/Examples/modules/Dropdown/index.js
+++ b/docs/app/Examples/modules/Dropdown/index.js
@@ -4,6 +4,75 @@ import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
 
 import { Message } from 'semantic-ui-react'
 
+// const DropdownExamples = () => (
+//   <div>
+//     <ExampleSection title='Types'>
+//       <ComponentExample
+//         title='Dropdown'
+//         description='A dropdown menu'
+//         examplePath='modules/Dropdown/Types/Dropdown'
+//       />
+//       <ComponentExample
+//         title='Custom Trigger'
+//         description='A dropdown can use a custom trigger element'
+//         examplePath='modules/Dropdown/Types/Trigger'
+//       />
+//       <ComponentExample
+//         examplePath='modules/Dropdown/Types/ImageTrigger'
+//       />
+//       <ComponentExample
+//         title='Uncontrolled'
+//         description='A dropdown works as an uncontrolled component'
+//         examplePath='modules/Dropdown/Types/Uncontrolled'
+//       >
+//         <Message>
+//           See React form docs for an explanation of
+//           <a href='https://facebook.github.io/react/docs/forms.html' target='_blank'>
+//             &nbsp;controlled and uncontrolled&nbsp;
+//           </a>
+//           component patterns.
+//         </Message>
+//       </ComponentExample>
+//       <ComponentExample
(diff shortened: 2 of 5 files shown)


levithomason:
Good deal, LMK when ready for review!

jeffcarbs:
@levithomason - one thing to think about: Is there any way to avoid rendering children with shorthand props here?

Note: Read below for some thoughts on this, but on further consideration I don't think we need to handle this on this PR. I think the goal here should be to split `Dropdown` and `Select` and do some light clean up. Larger refactors should probably be done separately.

---

Currently:

``` js
    const menuChild = Children.only(children)
    const menuClassName = cx(
      useKeyOnly(open, 'visible'),
      menuChild.props.className,
    )

    return (
      <ElementType
        tabIndex={0}
        {...rest}
        className={classes}
        onBlur={this.handleBlur}
        onClick={this.handleClick}
        onFocus={this.handleFocus}
        onMouseDown={this.handleMouseDown}
        ref={this.handleRef}
      >
        {trigger || DropdownText.create(text)}
        {Icon.create(icon)}
        {cloneElement(menuChild, { className: menuClassName })}
      </ElementType>
    )
```

One idea would be to not allow children and just pass the menu as a `menu` prop, although this kinda leads to ugly usage:

``` js
const menu = (
  <Dropdown.Menu>
    <Dropdown.Item text='New' />
    <Dropdown.Item text='Open...' description='ctrl + o' />
    <Dropdown.Item text='Save as...' description='ctrl + s' />
  </Dropdown.Menu>
)

<Dropdown text='File' menu={menu}>
```

Another idea would be to assume the last child is the menu, and only clone that one:

``` js
if (children) {
  // Not sure if this is the actual syntax, but this is the general idea
  const [...trigger, menu] = Children.toArray(children)

  const menuClassName = cx(
    useKeyOnly(open, 'visible'),
    menuChild.props.className,
  )

  return (
    <ElementType
      tabIndex={0}
      {...rest}
      className={classes}
      onBlur={this.handleBlur}
      onClick={this.handleClick}
      onFocus={this.handleFocus}
      onMouseDown={this.handleMouseDown}
      ref={this.handleRef}
    >
      {trigger}
      {cloneElement(menu, { className: menuClassName })}
    </ElementType>
  )
}
```

As much as I dislike dealing with `children`, I kinda think that last one might be the way to go. The goal of disallowing `children` and shorthand (imho) is to give the user absolute control over how the component is ultimately rendered. This would make that possible.

We could then provide shorthand `menu` prop which would let a user define a menu via data.

levithomason:
This component is currently deviating from our pattern of rending only children, or only shorthand.  I'm torn here.  The children are so complicated, I can't ever imagine someone wiring them together correctly.  Consider the ordering of the children is significant and changes between usages, which is currently [causing bugs](https://github.com/Semantic-Org/Semantic-UI-React/issues/642#issuecomment-252327828).

I haven't thought all of these things through, but I think we should take as small of steps as possible.  Would be in favor of 1 PR for the additional props, another PR for any cleanup, and another PR for separating the functionalities (which will be a breaking change).

Also, this is less on my radar since what is there is technically working.  I'm mostly thinking about things like form validation, doc site speed optimizations, and new components.

jeffcarbs:
@levithomason - that's valid. I think I may close this for now and move some of the improvements into the PR I have open for adding missing dropdown docs (https://github.com/Semantic-Org/Semantic-UI-React/pull/682).

So in that PR I'll
- Fix propType definitions/validations. I've seen a number of issues come in related to errant propType validations.
- Add missing examples. This will also help us get a better sense of where we fall short currently.
- Possibly add the `openOnFocus/closeOnBlur` functionality that was added here, but that might be a little out of scope so I may not.

I'll re-open this at some point post-v1

levithomason:
> Fix propType definitions/validations. I've seen a number of issues come in related to errant propType validations.

❤️ We should get those out right away.  My fixes for this are burried in #645. It is blocked as it asserts examples render without errors, but we need to solve how we're gonna generate childKeys before it can be merged.  Otherwise, master will be failing.

Again, a separate PR here could be shipped today really, even in the next hour.

> Add missing examples. This will also help us get a better sense of where we fall short currently.

👍 I can see this PR being a little longer running while offshoot PRs address the shortcomings it reveals.

> Possibly add the openOnFocus/closeOnBlur functionality that was added here, but that might be a little out of scope so I may not.

Agreed, separate PR to close #517 would be preferred.  With tests, it will likely be a large enough PR by itself.

levithomason:
Actually, in light of what I just said regarding Dropdown warnings, I'm just gonna ship that fix now.

levithomason:
@jcarbo warnings fixed in https://github.com/Semantic-Org/Semantic-UI-React/pull/694, releasing now.

PR #674, opened 2016-10-13
Typings for Typescript and Intellisense

tomitrescak:


config/typings.d.ts
```diff
@@ -0,0 +1,2072 @@
...
+  }
+
+  interface SuiContentElement extends SuiBlockElement {
+    content?: any;
+  }
+
+  type ButtonAttached = 'left' | 'right' | 'top' | 'bottom';
+  type LabelPosition = 'left' | 'right';
+
+  export interface IButtonProps extends SuiElement {
```
levithomason: I believe this should be `extends SuiContentElement` since a Button has a `content` prop.

config/typings.d.ts
```diff
@@ -0,0 +1,2072 @@
...
+    'us virgin islands' | 'vn' | 'vietnam' | 'vu' | 'vanuatu' | 'gb wls' | 'wales' | 'wf' | 'wallis and futuna' | 'ws' | 'samoa' |
+    'ye' | 'yemen' | 'yt' | 'mayotte' | 'za' | 'south africa' | 'zm' | 'zambia' | 'zw' | 'zimbabwe';
+
+
+  // BUTTON
+
+  interface SuiElement {
+    /**
+    * An element type to render as (string or function).
+    */
```
levithomason: To cut down on the number of lines here, could we refactor the comments from this:

``` js
/**
 * An element type to render as (string or function).
 */
```

To this:

``` js
```

I would also be fine with no comments in this file.  That way we only have to manage them in one location.

levithomason:
@tomitrescak bump

tomitrescak:
@levithomason I am working on the project with these typings in play, while working on it I am slowly discovering some bugs .. give me about 1-2 more weeks to work on it.

levithomason:
You got it, thanks much.

garyham:
Hi Tomas,

I too have generated a .d.ts file for semantic-ui (wasn't aware of yours).  I cut your version in to my build and only had minor issues on some attributes which were easy to fix.   Great job. 

However, as I made changes to add the missing elements I noticed that you mostly use one set of common props for the sub-component.  For example below the following all extend from SuiContent element, rather than properties unique to them (same for forms).  Is this a policy or a temporary expedient?  

interface IItemStatic {
    Content: (props: IItemProps) => JSX.Element;
    Description: (props: **SuiContentElement**) => JSX.Element;
    Extra: (props: **SuiContentElement**) => JSX.Element;
    Group: (props: IItemGroup) => JSX.Element;
    Header: (props: **SuiContentElement**) => JSX.Element;
    Image: (props: **SuiContentElement**) => JSX.Element;
    Meta: (props: **SuiContentElement**) => JSX.Element;
    (props: IItemProps): JSX.Element;
}
export const Item: IItemStatic;

Rightly or wrongly, in my version I had Item.Image extend it's props from the general Image component then add any missing extras.  I also inherited from _React.Component_  (copying the react-bootstrap guys) rather than _JSX.Element_ which I hadn't seen before.  I also multiply extended from _React.MouseEventHandler, React.MouseEventHandler_ and _React.MouseFocusHandler_ to bring in the event handlers as required which might be overkill but was expedient.

Any thoughts or do you need any help to complete?

Gary

tomitrescak:
Garyham, your changes make a lot of sense, do you want to share your version and maybe we can merge it?

I use JSX.Element as it is the most generic element that can be returned and you still control the list of props it needs. It is not necessary to use React.Component as this represents a Stateful component, while in reality you are using stateless components. But we can discuss this further.

garyham:
Hi Tomas,

I have attached my version (renamed to .txt to allow attachment).  I hadn't appreciated that they were stateless components in which case I should really extend from React.StatelessComponent<T> rather than React.ComponentClass<T>.  I tried that this morning at it works.

Gary

[semantic-ui-react.d.txt](https://github.com/Semantic-Org/Semantic-UI-React/files/552532/semantic-ui-react.d.txt)

tomitrescak:
@garyham are you generating this file from React-Doc? It looks impressive. Couple things to do (in case you are generating):
1. Add javadoc for all props.
2. We need to deal somehow with "as" parameter. When you do "as" that component suddenly accepts parameters from the "as" component. To be hones, I do not know how to do it.

garyham:
Hi Tomas,

Sadly not auto-generated, all hand crafted.  As for 'as', I too struggled with it and in the end used 'any' as a workaround.  I was thinking 'as' should be 'string | React.ReactNode' but I don't know enough about its underlying implementation.

Gary

levithomason:
The as prop is either a string or a function. The value of the as prop is passed to React.createElement().

tomitrescak:
The problem is different.
When you use "as", the element can ultimately use props from the "as" elementent. For example Header as Link, should accept the "to" property. But typescript will scream that there is no "to" property. Same goes for Dropdown as Menuitem and others. There exists a possibility of merging the props via & operator, I'm just not sure how to plug it in. They use it in react-router typings so but I need to study it a bit more.

@levithomason as per auto generation, what is the input to generate the documentation for the website? Maybe we should look into that. In the long run it could be just a simple transform if you would be open to add some meta data for the transform. It's definitely more sustainable in the long run than this masochistic manual approach.

tomitrescak:
@garyham I tried using your typings but I 'm getting "JSX element does not have construct or call signature" on all SUI elements. Did you try using it?

asvetliakov:
@tomitrescak 
I'm doing something like it:

``` tsx
export const DimmableForm = (props: DimmableProps & FormProps) => <Dimmable {...props} component={Form}/>;
export const DimmableSegment = (props: DimmableProps & SegmentProps) => <Dimmable {...props} component={Segment}/>;
```

in semantic-react to get props from both components in client code. I think there is no way to do automatic interface merging in ts definition.

tomitrescak:
@asvetliakov I thought so, this means we'll have to export individual Props from the typing as well, so that users can construct their own composed components. Thanks for letting know!

[EDIT]

Here is an example for React-Router enabled link button.

``` js
import { Button, IButtonProps } from 'semantic-ui-react';
import { Link, ILinkProps } from 'react-router';

export const LinkButton = (props: IButtonProps & ILinkProps) => <Button {...props} as={Link}/>;
```

garyham:
Hi Tomas, I'm using them everyday but I'll have a look at it. 

I install them using typings install --global file:./path/to/semantic-ui-react.d.ts. I won't have chance to look at it until  later today.

tomitrescak:
I needed to get rid of the `import React = __React` line for me to make it work. Typings are so much mess in typescript world ...

garyham:
I think if you bring it in using the _typings_ command it should would.

garyham:
I think i had to add that to make it work using typings command. Can't
remember. Glad it works now.

levithomason:
Is it worth adding `as` support later and shipping the well known props first?  I ask only because even if we cover the basic augmentation cases, it still will not work when a user passes their own component for which we do not have an interface definition.

``` jsx
<Button as={MyUnkonwnButton} />
```

Also, what about all the HTML component cases?

``` jsx
<Button as='div' />
<Button as='button' />
<Segment as='section' />
// ...etc
```

> @levithomason as per auto generation, what is the input to generate the documentation for the website? Maybe we should look into that.

Our prop definition docs are generated using `react-docgen`.  We then take that and parse the docblock strings with `doctrine` to turn them into JSON objects.  This gives us `docs/app/docgenInfo.json`.  You can run `npm run build:docs` and then open that file to see what is available.  Example:

``` json
{
  "src/addons/Confirm/Confirm.js": {
    "methods": [],
    "props": {
      "open": {
        "type": {
          "name": "bool"
        },
        "required": false,
        "docBlock": {
          "description": "Whether or not the modal is visible",
          "tags": []
        }
      },
...
```

garyham:
As an alternative to the currently implemented 'as' would it be possible to change the underlying implementation to accept a string or ReactNode.

**From**

```

<Menu>
  <MenuItem as={Link} to='/home'>
    Home
  </MenuItem>
</Menu>
```

**To**

```

<Menu>
  <MenuItem as={<Link to='/home'/>}>
    Home
  </MenuItem>
</Menu>
```

I appreciate that this is a non-trivial change but it cleanly separates the react-router domain from the semantic-ui domain.  

Apologies if its been suggested and rejected before.

levithomason:
Theoretically, we could do this by updating `getElementType` to first check `React.isValidElement(this.props.as)` and if so `React.cloneElement(this.props.as, ...rest)`.  We'd also need to merge `classNames` so we don't loose any in the clone.  The final result would be that the `as` prop allows a string, function, or element to be passed.

I'm sure there are nuances I'm overlooking here, but I believe it would be possible.  However, it would not negate the `<MenuItem as={Link} />` use case.

levithomason:
Issues coming to mind with this approach, this opens up the possibility to pass children to both the `as` element and the `MenuItem`.  That would not work, only one should win.

Also, we can't simply clone and pass the MenuItem props to the Link since it isn't going to know what to do with them. We'd have to first create an element out of the MenuItem, then merge the elements.  Off the top of my head, this sounds like a bad idea.  Again, this could be explored more and perhaps solved, but I am leaning away from the idea currently.

garyham:
I don't think the MenuItem props should be passed to the Link element as the Link element should already be decorated with whatever props it needs.  I do however think the classNames need to be merged.

hlehmann:
what about something like

```

<Menu>
  <MenuItemAs>
    <Link to='/home'>
      Home
    </Link>
  </MenuItemAs>
</Menu>
```

and forcing MenuItemAs to have only one child

levithomason:
The issue here is that `as` is available on all our components and can accept any other component, not just our other components.  So the possible combinations are infinite.

That said, perhaps `as` is something we just don't support for now.  There are several other prop validations that that cannot be supported in ts as well, like conditionally valid props.

hlehmann:
I personally added `[s:string]:any`
```typescript
  interface SuiElement {
    /**
     * An element type to render as (string or function).
     */
      as?: string | any;
    /**
     * Additional classes.
     */
    className?: string;
    /**
     * Additional props.
     */
    [s:string]:any
  }
```
It allows Intellisense to propose available props but let us add other props like `to`

garyham:
Tomas, Levi,

In case you are still working on this, attached is my most recent version (with .txt rather than .ts extension) which has some fixes (my appalling spelling of accordion, added the addon components, more props extending others to avoid duplication and properly reflect 'is-a' relationships).

The bit I am most unhappy with is the events.  I have components extending from interfaces which implement most of the React mouse, form and key events.  I now think this is incorrect and I should reflect only those implemented.  It works because of most of the components implement ...rest destructuring but I don't like it.

I would have changed it but this is my last update as I'm moving on now that our evaluation is complete.

I think you need to publish a .d.ts file sooner rather than later (and ignore the 'as' issue for the time being) as you have a great library.

Thanks,

Gary
 
[semantic-ui-react.d.txt](https://github.com/Semantic-Org/Semantic-UI-React/files/603540/semantic-ui-react.d.txt)

levithomason:
I agree entirely on publishing sooner than later.  I am not a typescript user so I'm trying to push this along while making assessments based on feedback. Looking to the typescript users for input on what is useful here...

halfmatthalfcat:
@tomitrescak I've been testing out your type definitions in my project. First, thanks a million, this is great. I'm having an issue however. It looks like using components that are defined with optional props such as

`export const Input: (props?: IInputProps) => JSX.Element`

in the context of `<Input action='Login' />` is throwing a compiler error for me: "Property 'action' does not exist on type 'IntrensicAttributes & (IInputProps | undefined)'"

Removing the optional indicator resolves this error though and I haven't seen any adverse consequences having both props and prop-less components with this change..

config/typings.d.ts
```diff
@@ -0,0 +1,2072 @@
...
+    size?: Sizes;
+    sub?: boolean;
+    subheader?: any;
+    textAlign?: Alignments;
+  }
+
+
+  interface IHeaderStatic {
+    Content?: (props: SuiBlockElement) => JSX.Element;
+    SubHeader?: (props: SuiContentElement) => JSX.Element;
```
halfmatthalfcat: This should be `Subheader` instead of `SubHeader`

halfmatthalfcat (changes requested):

psudarsa:
Would someone continue @tomitrescak's work and get at least some part of it committed ? I would like to but i am still new to the typings world

halfmatthalfcat:
I second @psudarsa. @tomitrescak are you still on track to fix/merge? If not, I'd take it over if/when we figure that out.

levithomason:
I'm going to update this with the latest from @garyham and release it. Better to iterate on it in the wild than theorize more here.  Hang tight!

levithomason:
Well folks, this might take a bit longer than anticipated but I'll continue to focus on this.

tomitrescak:
@levithomason This is updated version of @garyham with generics and working with React definition from `types` repository.

https://gist.github.com/tomitrescak/6261ad27dbe43aad03b2497721d751f5

asvetliakov:
@tomitrescak Ambient module definitions are forbidden if included in npm package
You need to name typings file as index.d.ts, remove ```declare module/namespace``` and just do ```export interface {} / export class / export const ```

levithomason:
Thanks guys, we're working on this now.  Once we get local testing working we'll merge and release.

levithomason:
This work has been copied over to #995.  We've tested this out and it all seems to work well.  Will merge and release that PR now.  Thanks to everyone here who helped contribute to typings 🎉

levithomason:

psudarsa:
Thanks for pushing this through!

PR #680, opened 2016-10-14
docs(ComponentExample): fix the warning of using string instead of bo…

dpkwhan:
The type of icon prop in MenuItem is bool, but here a string is used.


levithomason (changes requested):
In this case the `icon` propType in the MenuItem is actually wrong.  The MenuItem does support shorthand icons, see [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Menu/MenuItem.js#L121) in its render method.

We should instead update the propType to allow item shorthand _and_ boolean:

```
  icon: PropTypes.oneOfType([
    PropTypes.bool,
    customPropTypes.itemShorthand,
  ]),
```

levithomason:
I've opened #681 to fix this, will merge soon

PR #685, opened 2016-10-15
docs(Icon): refactor naming and descriptions, use function components

dpkwhan:
1. rename class name and filename
2. add trailing . in description
3. use stateless functional expression


levithomason:
Thanks much!

PR #686, opened 2016-10-16
fix(dropdown): render text when option.value = 0

Chris-R3:
!!0 equals false so the text for a option object with value 0 never rendered after it was selected.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -758,7 +758,9 @@ export default class Dropdown extends Component {
   renderText = () => {
     const { multiple, placeholder, search, text } = this.props
     const { searchQuery, value, open } = this.state
-    const hasValue = multiple ? !_.isEmpty(value) : !!value
+    const hasValue = multiple
+      ? !_.isEmpty(value)
+      : !_.isNil(value) && value !== ''
 
     const classes = cx(
       placeholder && !hasValue && 'default',
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -603,6 +603,60 @@ describe('Dropdown Component', () => {
         .find('.text')
         .should.contain.text(item.text())
     })
+    it('displays if value is 0', () => {
+      const text = faker.hacker.noun()
+
+      wrapperMount(<Dropdown options={[{ value: 0, text }]} selection />)
+
+      // open
+      wrapper.simulate('click')
+
+      // click item
+      const item = wrapper
+        .find('DropdownItem')
+        .simulate('click')
+
+      // text updated
+      wrapper
+        .find('.text')
+        .should.contain.text(item.text())
+    })
+    it('does not display if value is \'\'', () => {
+      const text = faker.hacker.noun()
+
+      wrapperMount(<Dropdown options={[{ value: '', text }]} selection />)
+        .simulate('click')
+        .find('DropdownItem')
+        .simulate('click')
+
+      wrapper
+        .find('.text')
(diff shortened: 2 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -751,7 +751,7 @@ export default class Dropdown extends Component {
   renderText = () => {
     const { multiple, placeholder, search, text } = this.props
     const { searchQuery, value, open } = this.state
-    const hasValue = multiple ? !_.isEmpty(value) : !!value
+    const hasValue = multiple ? !_.isEmpty(value) : value !== ''
```
levithomason: When `value` is `null` or `undefined` then `hasValue` will be true.  We should add a nil check in here to account for those:

``` jsx
const hasValue = multiple 
  ? !_.isEmpty(value) 
  : !_.isNil(value) || value !== ''
```

levithomason (changes requested):
The logic needs updated to account for null/undefined values.

Let's also add two tests.  One that asserts that the text is rendered when the value is `0`, and another that asserts it is not rendered for `null`, `undefined`, and `''`.

src/modules/Dropdown/Dropdown.js
Chris-R3: Actually I had `_.isNil(value)` my first version but then two related tests failed so i changed it to only check for empty strings. Maybe it should be:

``` js
const hasValue = multiple 
  ? !_.isEmpty(value) 
  : !_.isNil(value) && value !== ''
```

I'll  update this PR as soon as possible 😉

levithomason: It could also be that the tests need to be changed to match the new code.  Keep in mind that the tests are just there to assert that the code is doing what it should do.  So, if the code needs updating, it is likely that the tests need updating as well.

levithomason:
@Chris-R3 bump

Chris-R3:
Finally got around to updating this, sorry for the delay.

levithomason:
Looks like a lint issue holding this guy up:

image

Chris-R3:
fixed

levithomason:
Thanks much @Chris-R3, I'll release this soon!

tomitrescak:
Any plan when this will be released? Thanks!

levithomason:
I did not report the version back here, but here have been several releases since this PR.  Checking the log...

levithomason:
Released in `semantic-ui-react@0.58.0` on 2016-10-28:

[CHANGELOG.md#v0580-2016-10-28](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/CHANGELOG.md#v0580-2016-10-28)

tomitrescak:
Hmm, yet even with latest, if value on dropdown is set to {0} that item cannot be selected and it is not pre-filled. I thought that this PR was tackling this.

levithomason:
This PR was dealing with the Dropdown `text` not rendering when a DropdownItem's `value` prop was set to `0`.  It was a bad falsy check.

Would you mind opening a new issue with the steps, expected behavior, actual behavior, and a codepen for debugging?

tomitrescak:
The text is still not rendered when value is set to 0. At least on my end. Do you have an exemplary codepen from which I can start from? I'll report it.

levithomason:
Alrighty, sounds like a bug.  The codepen is linked in the issue template.

littlehome-eugene:
Still having this bug, with 0.64.3

Any workaround would be helpful if merging is not imminent yet.

levithomason:
This PR was merged and released several months ago, the current version is `0.64.8`.  Dropdown items with a value of `0` correctly render their text when made active:

PR #687, opened 2016-10-16
feat(ButtonGroup): Add missing props

jeffcarbs:
Adds missing props for `Button.Group` found in SUI core: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/button.less


src/elements/Button/ButtonGroup.js
```diff
@@ -17,21 +17,33 @@ import {
...
+    secondary, size, toggle, vertical, widths,
   } = props
 
   const classes = cx('ui',
     size,
     color,
     useValueAndKey(attached, 'attached'),
     useKeyOnly(basic, 'basic'),
+    useKeyOnly(compact, 'compact'),
+    useKeyOnly(disabled, 'disabled'),
```
levithomason: Looks like they're all good, except for `disabled`.  I can find [no css for disabled `.ui.buttons`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/button.less#L196) and the markup does not seem to work either:

image

I'll pull this one and merge.

levithomason (approved):

src/elements/Button/ButtonGroup.js
jeffcarbs: Ah you're right! Nice find 👍

levithomason:
Released in `semantic-ui-react@0.56.2`

PR #694, opened 2016-10-17
fix(Dropdown): over zealous propType warnings

levithomason:
This PR fixes over zealous propType warnings from the Dropdown.  Both of these are valid usages:

A Dropdown menu only that currently throws warnings demanding `selection`.

``` jsx
<Dropdown options={[...]} />
```

A Dropdown with a menu child that currently throws warnings demaing `text`.  This is also a doc site example.

``` jsx
<Dropdown trigger={<span>Click Me</span>}>
  <Dropdown.Menu>
    ...
  </Dropdown.Menu>
</Dropdown>
```


levithomason:
Released in `semantic-ui-react@0.56.4`

PR #697, opened 2016-10-18
fix(dropdown): search text displayed in sizer element.

june1123:


levithomason:
This is a hidden element so the text is not visible.  Is it showing up in your case?

june1123:
Sorry,I didn't catch update of css file.
I use version 2.1.8  and it was updated to 2.2.4.

Could you please update usage document? ^^

levithomason:
Updated, thanks

PR #701, opened 2016-10-19
feat(Search|Dropdown): Improve arrow navigation

jeffcarbs:
This improves arrow navigation in both the dropdown and search menus.

**Previously**
When pressing "arrow-down" to move through results once an item was past the bottom the menu's scrollTop would move to that item's scrollTop, which was kinda jarring. It was almost as if you were moving to the next "page" of items, which is pretty non-standard:
before

**After**
When an item is past the bottom, the menu adjusts so that the bottom of the menu is at the item's bottom:
after

This also fixes the specs that were testing this functionality - they weren't actually testing anything 😐

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -713,8 +713,10 @@ export default class Dropdown extends Component {
     const isOutOfUpperView = item.offsetTop < menu.scrollTop
     const isOutOfLowerView = (item.offsetTop + item.clientHeight) > menu.scrollTop + menu.clientHeight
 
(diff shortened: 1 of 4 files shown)


levithomason:
Awesome, any chance we can get this update in the Dropdown in this PR as well (being it's so small)?

jeffcarbs:
This includes the change for both search and dropdown. The title doesn't reflect that because I wasn't sure how to do that within the PR naming convention.

levithomason:
That's what I get for reviewing tired, sorry about that! Released in `semantic-ui-react@0.56.8`.

PR #721, opened 2016-10-21
fix(Modal): Only set position when modal is open

### LINKED ISSUE #712: Modal relies on document as default renderNode, breaks SSR
Line 117 of Modal.js  uses `document` as the mount node if `this.props.mountNode` is absent. Thus if you don't specify `mountNode`, server-side rendering fails because there is no `document`.

[Modal.js:117](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Modal/Modal.js#L117)

The only workaround I can figure out is to track if the parent component has been mounted with a manually set `this.state.hasMounted` and then only render the `Modal` after `componentDidMount` has run.  This makes the markup awkward and adds complexity.

Is there another option? Could Modal.js guard against trying to use `document` when it's not defined? Couldn't it just default to rendering in its parent like other components?

jeffcarbs:
Fixes #712 

The modal has code that continuously calculates its position to:
- ensure it's centered vertically.
- add the scrolling class to the body if need be.

This was being called on component mount, which doesn't make sense since the modal itself may not actually be visible. This updates the the `setPosition` to start when the `Portal` gets mounted (the modal becomes visible) and stop when the `Portal` unmounts (the modal is hidden).

This should:
- reduce the workload being done by modals rendered onto a page. 
- make it so the component doesn't touch the `mountNode` until the page has been fully rendered. This (I think) should fix #712 since the `mountNode` defaults to `document.body` which isn't present server-side.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -77,18 +77,13 @@ class Modal extends Component {
 
   state = {}
 
-  componentDidMount() {
-    debug('componentDidMount()')
-    this.setPosition()
-  }
-
   componentWillUnmount() {
     debug('componentWillUnmount()')
-    this.handleUnmount()
+    this.handlePortalUnmount()
   }
 
-  handleMount = () => {
-    debug('handleOpen()')
+  handlePortalMount = () => {
+    debug('handlePortalMount()')
     const { dimmer } = this.props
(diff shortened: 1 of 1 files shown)


levithomason (approved):
Awesome, nice catch!

PR #731, opened 2016-10-24
perf(props): Remove propTypes from production build

layershifter:
For details - #524

This PR might to do:
- [x] updates contributing guide;
- [x] updates `getUnhandledProps` function;
- [x] adds new sections for common tests.
- [x] updates `ComponentProps` component in docs (we can't rely on `meta` anymore).

---

@levithomason I've updated `Rail` component, waiting for review there.


jeffcarbs:
I think this could work, although I'm not too excited about `if (process.env.NODE_ENV !== 'production') {` everywhere. Especially when `propTypes` is defined as a static prop within the class.

What about this:
- define `props` within the meta as you're doing now. The newly added tests will ensure we're not missing any props.
- Instead of wrapping manually with `NODE_ENV`, use this as part of the build process: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types

I'm still not 100% sure what the `Rails.props` are going to be used for, but I think if we made that a `const` it would get removed during dead code removal since it won't be referenced after the `propTypes` are removed:

``` js
const propValues = {
  close: ['very'],
  position: SUI.FLOATS,
  size: _.without(SUI.SIZES, 'medium'),
}

Rail.propTypes = {
  // snip ...

  /** A rail can be presented on the left or right side of a container. */
  position: PropTypes.oneOf(propValues.position).isRequired,
}
```

Or we could just inline them:

``` js
Rail.propTypes = {
  // snip ...

  /** A rail can be presented on the left or right side of a container. */
  position: PropTypes.oneOf(SUI.FLOATS).isRequired,
}
```

levithomason:
### babel-plugin-transform-react-remove-prop-types

I love this plugin, however, I don't think we can remove the propTypes in any of our builds since the user needs to have them in dev.  If we strip them, then only we'll have them in dev.  We need to only remove them when the user builds for production.

It is worth noting however that this plugin has a `mode: 'wrap'` option which does exactly what was done in this PR.  It wraps propTypes in a NODE_ENV check.  We should probably just use this!

### `NODE_ENV`

I think it is acceptable to use `NODE_ENV === 'production'` since we're a React lib and React has force opted all users into this pattern.  React itself uses these checks and [includes instructions](https://facebook.github.io/react/docs/installation.html) for it in their docs:

> To use React in production mode, set the environment variable `NODE_ENV` to "production".

React router v4 is also [wrapping propTypes](https://github.com/ReactTraining/react-router/blob/v4/modules/Match.js#L105) in a dev check.  So, I think it is here to stay for a while.

---

I've added the plugin with the `mode: 'wrap'` option and tested a build.  Works perfectly 😄 :

**Without Plugin**

``` jsx
Flag.propTypes = {
  as: _lib.customPropTypes.as,

  /** Additional classes. */
  className: _react.PropTypes.string,

  /** Flag name, can use the two digit country code, the full name, or a common alias */
  name: _react.PropTypes.oneOf(Flag._meta.props.name).isRequired
};
```

**With Plugin**

``` jsx
process.env.NODE_ENV !== "production" ? Flag.propTypes = {
  as: _lib.customPropTypes.as,

  /** Additional classes. */
  className: _react.PropTypes.string,

  /** Flag name, can use the two digit country code, the full name, or a common alias */
  name: _react.PropTypes.oneOf(Flag._meta.props.name).isRequired
} : void 0;
```

I've also just found this plugin is included in the preset, [`babel-react-optimize`](https://github.com/thejameskyle/babel-react-optimize).  I'm going to try this out, if it works then I'll update this PR to use the plugin instead of manually adjusting the code.  We'll still need this PR since we'll need the updates to our utils to not rely on propTypes.

levithomason:
## What's done

I've added some perf plugins:

### `babel-plugin-transform-react-constant-elements`

Any const element in a render method, such as `<label>foo</label>`, is instead rendered once at the top of the file and replaced with a variable.  This just prevents needless re-rendering of constant elements.

### `babel-runtime` + `babel-plugin-transform-runtime`

Babel puts helpers in every file to allow future features (spread, class, etc).  The `babel-runtime` is a module with these helpers as exports.  The transform here simply references the module exports instead of re-writing every helper at the top of every file.

### `babel-plugin-transform-react-remove-prop-types`

This plugin wraps our propTypes in a production check so that when users make a production build, the propTypes are removed.

## What's left

### Wrapping prop values

We still need to wrap prop values so they are removed in production. Otherwise we'll have massive lists of prop values left in our files because they are not inlined in propTypes, they are referenced.

We don't inline them because then we wouldn't be able to access the values.  They would be buried in a propType function argument.  We use the values to test component configurations in common tests.  We also plan to use them for a component explorer in the future.  So, I don't think we should inline them.

I'm also considering simplification here.  We _could_ drop the goals of auto testing valid prop value configurations and the component explorer in trade for inline prop values.  That would save a lot of work.  Feedback?

### List of props

We still need a list of props in `_meta` so our utils can use it during runtime.  `getUnhandledProps` must access a list of props so it knows which ones are "unhandled".

Given the amount of work that is there (updating and maintaining every file) I think it is worth considering our own babel-plugin to do this.  It seems straight forward, read the keys of propTypes, and add them to a static array.  We have quite a bit to gamble with on this given the number of hours required to do it by hand.

layershifter:
### List of props

Seems, babel plugin is good idea, I suggested it at the beginning of discussion, it can make code cleaner. I can start on it after `Dimmer` and `Reveal` merge.

#### What it must do?

Collect values of:
- `Component.autoControlledProps` (if its class and extends `autocontrolledComponent`);
-  keys of `Component.defaultProps`;
-  keys of `Component.propTypes`.

And ands it to `Component._meta.props`.

jeffcarbs:
> And adds it to Component._meta.props.

@layershifter - Adds it as `Component._meta.props` **wrapped in `NODE_ENV === 'production'`** so that it can be removed during dead code removal by a user's compiler but still available to them during dev.

layershifter:
I started to work on babel-plugin.

levithomason:
Great, can't wait to see this one come to life!

layershifter:
@levithomason Hey, plugin was completed, take a look [here](https://github.com/layershifter/babel-plugin-sui-props). I've implemented cases for stateless and statefull component.

Seems, we need plugin's code to this PR. Where I can store plugin? I'm thinking about `babel/plugins` directory for code and `test/babel/plugins` for tests.

levithomason:
> Seems, we need plugin's code to this PR. Where I can store plugin? I'm thinking about babel/plugins directory for code and test/babel/plugins for tests.

Exactly what I was thinking 👍 

**EDIT**

It would be nice to also include the README.md for it in the `babel/plugins` directory.  The in/out snippets could be updated as well.

levithomason:
I was just informed by @davezuko that we can use babel-plugin-lodash directly with our project if we structure it in a friendly format on publish.  We just then need to set the `id` option:

https://github.com/lodash/babel-plugin-lodash/issues/85

jeffcarbs:
@layershifter @levithomason - thoughts on making this babel plugin slightly more generic and possibly either shipping it separately or at least making it usable outside this project? I've come across a few places in my app where I've wanted to use the pattern of consuming props and passing the rest through but don't want to rely on `propTypes`.

Instead of (SUI-react specific):

``` js
Component._meta = {
  props: ['children', 'className']
}
```

it could output (more general):

``` js
Component.handledProps = ['children', 'className']
```

Could name it `babel-plugin-handled-props` or something.

layershifter:
@levithomason I've imported code to this PR. 
- I'm having problem with `fs` module, there are no it in webpack's bundle because `target: web`, but I'll need to make assertation on `pluginTests.js` that requires file read of expected. Any ideas there?


- We need solution with wrapping prop values, I'll think that it will be simplier make them inline like in latest `Rail.js` in this PR:

``` js
close: PropTypes.oneOfType([
  PropTypes.bool,
  PropTypes.oneOf(['very']),
]),
```

layershifter:
@levithomason Any response? 😢

levithomason:
Whoops, just got done doing a maintenance sweep but overlooked this one :/ Sorry! 

I try to hit all unread issues/PRs and anything labeled "Ready For Review" each time I do a maintenance sweep. I may not be available the rest of today but will get to this in the next sweep. I'll add the label so as not to miss it again.

jeffcarbs:
> I was just informed by @davezuko that we can use babel-plugin-lodash directly with our project if we structure it in a friendly format on publish. We just then need to set the id option:

I tried this in my project and wanted to report back. I analyzed my webpack-bundled files using https://github.com/th0r/webpack-bundle-analyzer (super cool tool, btw!). Here's the before/after (I think "parsed size" is post-minification, "stat size" is all the file sizes summed):

Before:
screen shot 2016-11-06 at 8 26 30 pm

After:
screen shot 2016-11-06 at 8 27 30 pm

Usage is just:

``` js
webpackConfig.module.loaders.push({
  test: /\.(js|jsx)$/,
  exclude: /node_modules/,
  loader: 'babel',
  query: {
    // ... snip
    plugins: [
      // ... snip
      'lodash', // Run module cherry-picker on lodash
      ['lodash', { 'id': 'semantic-ui-react' }] // Run module cherry-picker on semantic-ui-react
    ],
    // ... snip
```

Not bad for one line of config 😁 . Will be cool to see how much further this number drops once this PR makes it in 👍

levithomason:
Awesome, I also have a branch in progress that generates alias files for all components, so we can:

``` jsx
import Button from 'semantic-ui-react/Button' 
```

@jcarbo I'm actually curios how that worked without that PR... Did you manually add index files for it? 

> babel-plugin-handled-props

I think this is a great idea.

jeffcarbs:
@levithomason - I only added that one config line, no other code. The code in `babel-plugin-lodash` is a little hard to follow, but I'm pretty sure it uses `id` as the root and will try to file modules loaded anywhere underneath. So given:

```
{ 'id': 'semantic-ui-react' }
```

 I think it will do the following conversion:

``` js
// from
import { Button } from 'semantic-ui-react'

// to
import Button from 'semantic-ui-react/dist/commonjs/elements/Button'`
```

layershifter:
Result of lodash plugin looks awesome :+1: 

---

@levithomason So what we do with plugin?

After @jcarbo offer I think we need to do this:
- leave plugin in separate repository and name it `babel-plugin-transform-react-handled-props` (I'll add you to collaborators);
- plugin will move array of props to `Component.handledProps`;
- we choose simple way and make prop values inline: 

``` js
close: PropTypes.oneOfType([
  PropTypes.bool,
  PropTypes.oneOf(['very']),
]),
```

levithomason:
I agree on all counts here.  Let's do it.

layershifter:
I've make clean up of this PR. I'll need to add more tests and make some chore work with plugin :hourglass:

package.json
```diff
@@ -9,8 +9,9 @@
   ],
   "scripts": {
     "docs": "gulp docs",
-    "build": "npm run build:commonjs && npm run build:umd && npm run build:docs",
+    "build": " npm run build:babel-plugins && npm run build:commonjs && npm run build:umd && npm run build:docs",
     "prebuild:commonjs": "rimraf dist/commonjs",
+    "build:babel-plugins": "rimraf lib/* && babel babel/plugins -d lib",
```
levithomason: I believe this goes away now.

package.json
```diff
@@ -29,7 +30,7 @@
     "release:patch": "npm run prerelease && ta-script npm/release.sh patch",
     "start": "npm run docs",
     "start:local-modules": "npm run docs -- --local-modules",
-    "pretest": "npm run build:dll",
+    "pretest": "npm run build:babel-plugins && npm run build:dll",
```
levithomason: I believe this can also be restored.

package.json
```diff
@@ -54,9 +55,15 @@
...
     "babel-plugin-__coverage__": "^11.0.0",
     "babel-plugin-lodash": "^3.1.4",
+    "babel-plugin-react-transform": "^2.0.2",
+    "babel-plugin-transform-react-constant-elements": "^6.9.1",
+    "babel-plugin-transform-react-remove-prop-types": "^0.2.10",
+    "babel-plugin-transform-runtime": "^6.15.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.5.0",
     "babel-preset-stage-1": "^6.5.0",
+    "babel-register": "^6.18.0",
```
levithomason: We don't need `babel-register` any longer, this happened in #821.

levithomason (commented):
Will wait for the plugin and updates before doing a full review, here are a couple cleanup comments for now.

levithomason:
Using the `unhandledProps` pattern in one of our apps right now, would be great to have the babel plugin for it as well :)  It got me thinking whether or not the plugin should add a `this.getUnhandledProps()` method that just reduces `this.props`.

It seems the primary use case of having `this.constructor.handledProps` is to spread unhandled props, so why not provide a method that returns the unhandled props directly?  This could be put behind an option flag if we only want to include the array as the default behavior:

``` json
["transform-react-handled-props", {
  "getUnhandledProps": true
}]
```

layershifter:
Plugin is almost done, I will back there soon 😸

layershifter:
I've added plugin, but initial test-run shows that it's not ready. Working on found bugs.

levithomason:
Thanks for the update 👍

layershifter:
Hey, looks that we are very near to this feature :smiley_cat:

I've made a working build of SUIR with `babel-plugin-transform-react-handled-props@0.2.1`. I'll make necessary changes and get back there :+1:

layershifter:
I'll have issue with `transform-react-constant-elements`, need more time to review for it. However, [facebookincubator/create-react-app #553](https://github.com/facebookincubator/create-react-app/issues/553) it may be problem in plugin.

layershifter:
I've completely updated `Rail` component, all other components can be updated in other small PRs. Tests are running, docs - working. Seems, we're ready :+1: 

@levithomason I've need your review there, when you'll have spare-time :smile_cat:

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -3,11 +3,6 @@ import React, { Component, PropTypes } from 'react'
 
 import { Label, Table } from 'src'
 
-const descriptionExtraStyle = {
-  fontSize: '0.95em',
-  color: '#777',
-}
-
```
levithomason, line 10: It looks like the enums support was mistakenly removed from this file.  Bad merge?

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -25,26 +20,9 @@ export default class ComponentProps extends Component {
     meta: PropTypes.object,
   }
 
-  state = {
-    showEnumsFor: {},
-  }
-
-  toggleEnumsFor = (prop) => () => {
```
levithomason, line 32: Bad merge?

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -25,26 +20,9 @@ export default class ComponentProps extends Component {
     meta: PropTypes.object,
   }
 
-  state = {
```
levithomason, line 28: Bad merge?

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -62,52 +40,6 @@ export default class ComponentProps extends Component {
     )
   }
 
-  renderEnums = (item) => {
```
levithomason, line 65: Bad merge?

docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -145,14 +78,9 @@ export default class ComponentProps extends Component {
...
               <Table.Cell>{this.renderName(item)}</Table.Cell>
               <Table.Cell>{this.requiredRenderer(item)}</Table.Cell>
-              <Table.Cell>
-                {item.type}
-              </Table.Cell>
+              <Table.Cell>{item.type}</Table.Cell>
               <Table.Cell>{this.renderDefaultValue(item.defaultValue)}</Table.Cell>
-              <Table.Cell>
-                {item.description && <p>{item.description}</p>}
-                {this.renderEnums(item)}
```
levithomason, line 154: Bad merge?

src/lib/getUnhandledProps.js
```diff
@@ -1,41 +1,13 @@
...
-  if (!handledProps) {
-    handledProps = []
-
-    if (autoControlledProps) pushUnique(autoControlledProps, handledProps)
-    if (defaultProps) pushUnique(Object.keys(defaultProps), handledProps)
-    if (propTypes) pushUnique(Object.keys(propTypes), handledProps)
-
-    Component.handledProps = handledProps
-  }
+  const { handledProps = [] } = Component
```
levithomason, line 10: How about an inline comment here noting that this is added by the plugin?

test/specs/commonTests.js
```diff
@@ -254,6 +254,27 @@ export const isConformant = (Component, options = {}) => {
...
+    it('Component.handledProps includes all handled props', () => {
+      const computedProps = _.union(
+        Component.autoControlledProps,
+        _.keys(Component.defaultProps),
+        _.keys(Component.propTypes),
+      )
+      const expectedProps = _.uniq(computedProps).sort()
+
+      Component.handledProps.should.to.deep.equal(expectedProps,
+        'It seems that not all props were defined in Component.handledProps, you need to check that they equal to ' +
```
levithomason: `they equal to` -> `they are equal to the`

webpack.umd.config.js
```diff
@@ -24,6 +24,7 @@ const webpackUMDConfig = {
     'react-dom': 'ReactDOM',
   },
   plugins: [
+    new webpack.DefinePlugin(config.compiler_globals),
```
levithomason: This might also be a bad merge.  I don't think we make use any globals in the /src.  Those would be found as double underscored variables (i.e. `__DEV__`).

levithomason (changes requested):
I think I see signs of a bad merge.  I've noted the suspect areas in the inline comments.  Other than this, looks great :)  Can't wait to get this one in!

docs/app/Components/ComponentDoc/ComponentProps.js
layershifter: No, I removed it consciously. Enum's functionality relies on `meta.props`, this PR makes it obsolete, we will remove them all in future PRs. See [this](https://github.com/Semantic-Org/Semantic-UI-React/pull/731#issuecomment-258778858) comment and `Rail`'s update in this PR.

docs/app/Components/ComponentDoc/ComponentProps.js
layershifter: ^

docs/app/Components/ComponentDoc/ComponentProps.js
layershifter: ^

docs/app/Components/ComponentDoc/ComponentProps.js
layershifter: ^

docs/app/Components/ComponentDoc/ComponentProps.js
layershifter: ^

src/lib/getUnhandledProps.js
layershifter: Good idea, I'll add comment immediately.

test/specs/commonTests.js
layershifter: 👍

webpack.umd.config.js
layershifter: Okay, I'll remove it, but it strange - it was introduced in your [commit](https://github.com/Semantic-Org/Semantic-UI-React/pull/731/commits/4a69508f9187d69c0c8d2a260030e17ed63f6897). Maybe it make sense?

layershifter:
I've made update based on your comments 👍

docs/app/Components/ComponentDoc/ComponentProps.js
levithomason: Thanks for the reminder.

levithomason (approved):

levithomason:
Fantastic work!  I'm going to roll this out along with a breaking change for factory keys, #1062.  It will be nice to have a minor bump to group this major library changes together.

tomitrescak:
Just a question; to make this work, do we have to use Babel? I am on pure typescript -> js scenario.

levithomason:
No you don't, but you do need to run dead code elimination (uglify) in order for them to be stripped.  propTypes are wrapped in a conditional by the prop types plugin.  This is done by our build.  So as long as you use the `/dist` (package.json main) and you run uglify with dead code elimination at some point before production, then you should have smaller builds (without propTypes).

levithomason:

tomitrescak:
Would be great if production build would be part of the distro. Uglify tends to take significant time. Thanks for this amazing pr!

tomitrescak:
Guys, I'm not entirely sure how to use this. I upgraded semantic-ui-react to latest 0.64.0, I ran the build process with webpack that includes uglify with dead code elimination and bundle size actually increased 10kb I did not notice any decrease. The only thing that I did in between these build was upgrading react-semantic-ui-to latest.

dvdzkwsk:
@tomitrescak does your build system supply a `process.env.NODE_ENV` global? It looks like this PR wrapped all `propType` definitions in `NODE_ENV` checks, so if those are not able to be resolved statically (and therefore stripped) you may notice an increase. Just a guess.

tomitrescak:
@davezuko what needs to be emitted there? I am emitting 'development' if I'm in development build, otherwise I leave it empty.

[EDIT] I am actually emitting 'production'

```
new webpack.DefinePlugin({
      'process.env': {
        NODE_ENV: 'production'
      }
    })
```

[EDIT2]

The production veriable is specified correctly and code removal is ok, as I am eliminating in the same way mobx-react-devtools. This is the size I am seeing now. It is the same as during development build:

screen shot 2017-01-17 at 11 20 23 am

dvdzkwsk:
@tomitrescak that's exactly what I meant, and it was just a guess at that, so perhaps I'm on the wrong track. I just casually saw this issue and figured I'd inject a thought, but I'll keep thinking on it.

As an aside, I'm curious about that snippet you posted. I'm fairly certain that, at least at some point, that was not valid code for the define plugin due to how it injected globals; you'd actually have to stringify `production` (i.e. `JSON.stringify('production')`) so that the value of `NODE_ENV` was `"production"` (includes the quotes). This was because webpack inserted the value _as is_, so it would ultimately look like:

```js
// without stringifying (NODE_ENV: 'production')
if (process.env.NODE_ENV === 'production') // before compile
if (production === 'production') // after compile, ERROR! production is not defined

// vs (NODE_ENV: JSON.stringify('production') [or just '"production"']).
if (process.env.NODE_ENV === 'production') // before compile
if ("production" === 'production') // after compile, no error
```

This is a whole 'nother tangent, and you'd likely have already encountered this problem in previous builds. I could be totally off base here, but it's all I can think of at the moment.

Aside #2:

Does webpack inject this global into non-webpacked packages? Meaning, you import SUIR, but webpack simply imports it, but does not transform it in any way (i.e. with babel and the lot), so perhaps it is not stripping out that code? Yes, `process.env.NODE_ENV` will be available and defined at runtime, but my gut tells me it is not applying that global statically into the SUIR source. I have honestly not ventured too far down this path (meaning optimizing external libs) so I'm really just spitballing... will do some research. This is most likely the issue.

tomitrescak:
@davezuko that's exactly how I had it with JSON.stringify, I just changed it for this example.

I think it's just BundleAnalyser acting up. When I go the minified source code of semantic-ui-react the PropTypes are missing (in dev build they are there), yet BundleAnalyser reports exactly the same size.

[EDIT] An it's most probably true:

> If you minify/compress your code using the approach recommended in the Webpack documentation, which is to use the UglifyJS plugin, be aware that this tool will report the sizes of modules before they are minified. This is because the statistics generated by webpack --json do not take account of plugins that operate on the bundle as a whole.

layershifter:
> Does this PR wraps `propTypes` and removes them from production build?

Yep, it makes this.

> Why bundle is still have nearly same size?

We used `_meta` on components for tests and docs. Let me show it on `Flag`:

```jsx
const names = [] // long array

function Flag (props) {}
Flag._meta = {
  // something there
  props: {
    name: names
  }
}
```

Yes, `propTypes` will be removed in production build, but `_meta` still require _this long array of names_, so it still be in bundle :cry:

> Guys, make something with it!

Yep, I'm working on it, for example, `Flag` was modified in #1155. But there ~30 components that need to be updated (#524).

Before `Flag` update: 
```
 semantic-ui-react: 965.91 KB (59.5%)
   lodash: 291.54 KB (30.2%)
   <self>: 674.37 KB (69.8%)
```

After: 
```
semantic-ui-react: 961 KB (58.9%)
  lodash: 291.54 KB (30.3%)
  <self>: 669.46 KB (69.7%)
```

> Any ETA?

As soon as I can. Lowering bundle size is my priority in this time.
After all components will be updated, bundle will lose [`SUI.js`](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/SUI.js) and some other definitions.

levithomason:
>bundle will lose SUI.js

We'll likely lose _some_ definitions in this file and it will become drastically smaller since things like `ICONS_AND_ALIASES` will move into propTypes and doc utils (for the icon search in the docs).  However, the file will likely remain since we make heavy use of the first ~30 lines in almost every component.

PR #747, opened 2016-10-25
feat(Item.Image): Add size prop

### LINKED ISSUE #746: Item.Image doesn't honor size prop
The wrapped Image produced by Item.Image doesn't seem to honor the size prop.

Item.Image renders as a wrapped image with ui=false.  Looking at http://semantic-ui.com/views/item.html#image seems to indicate that their wrapped image uses
<div class="ui tiny image"> </div>.

Indeed adding ui=true to the Image produced by Item.Image seems to resolve the size issue.
_Update:_ there is an explicit test ensuring ui=false, so perhaps there's something I don't understand about how semantic is supposed to work.

clemensw:
- Resolves #746
- Reverses an explicit test for ui=false, so needs review.
- Renders similar to the example at http://semantic-ui.com/views/item.html#image
- Images now render left aligned and sized.

### DIFF
--- a/src/views/Item/ItemImage.js
+++ b/src/views/Item/ItemImage.js
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { PropTypes } from 'react'
 import {
   META,
+  getUnhandledProps,
 } from '../../lib'
 import Image from '../../elements/Image'
 
@@ -8,13 +9,21 @@ import Image from '../../elements/Image'
  * An item can contain an image
  **/
 function ItemImage(props) {
-  return <Image {...props} ui={false} wrapped />
+  const { size } = props
+  const rest = getUnhandledProps(ItemImage, props)
+
+  return <Image {...rest} size={size} ui={!!size} wrapped />
 }
 
 ItemImage._meta = {
   name: 'ItemImage',
   parent: 'Item',
   type: META.TYPES.VIEW,
+}
+
+ItemImage.propTypes = {
+  /** An image may appear at different sizes */
+  size: PropTypes.oneOf(Image._meta.props.size),
 }
 
 export default ItemImage
--- a/test/specs/views/Item/ItemImage-test.js
+++ b/test/specs/views/Item/ItemImage-test.js
@@ -13,4 +13,9 @@ describe('ItemImage', () => {
     wrapper.should.have.prop('wrapped', true)
     wrapper.should.have.prop('ui', false)
   })
+
+  it('has ui with size prop', () => {
+    shallow(<ItemImage size='small' />)
+      .should.have.prop('ui', true)
+  })
 })


layershifter:
This PR will break all other except case with `size`, it cannot be merged as is.

---



For a common understanding, we are talking about the example. This example uses `size` prop that will pass to `Image` component, but CSS for `item.image` doesn't handle size (_what is right_).

So, we need add `ui` when there is `size` prop and skip it in other cases.

``` jsx
<Item.Image /> // <div class="ui image">
<Item.Image size="small" /> // <div class="ui small image">
```

I think it's not responsibility of `Image`, so we need fix in `Item.Image`.

``` diff
- import React from 'react'
+ import React, { PropTypes } from 'react'
import {
  META,
+ SUI, 
} from '../../lib'
import Image from '../../elements/Image'

function ItemImage(props) {
-  return <Image {...props} ui={false} wrapped />
+ const { size } = props
+ const rest = getUnhandledProps(ItemImage, props)
+
+ return <Image {...props} size={size} ui={!!size} wrapped />
}

ItemImage._meta = {
  name: 'ItemImage',
  parent: 'Item',
  type: META.TYPES.VIEW,
}

+ ItemImage.propTypes = {
+  /** An image may appear at different sizes */
+  size: PropTypes.oneOf(Image._meta.props.size),
+ },

export default ItemImage
```

``` diff
import React from 'react'
import ItemImage from 'src/views/Item/ItemImage'

describe('ItemImage', () => {
  it('renders Image component', () => {
    shallow(<ItemImage />)
      .should.have.descendants('Image')
  })

  it('is wrapped without ui', () => {
    const wrapper = shallow(<ItemImage />)

    wrapper.should.have.prop('wrapped', true)
    wrapper.should.have.prop('ui', false)
  })

+  it('has ui with size prop', () => {
+    shallow(<ItemImage size='small' />)
+      .should.have.prop('ui', true)
+  })
})
```

levithomason:
Let's also use `ui={!!size}` so the value is coerced to a boolean value.  Otherwise, when the passes a size such as `small` the prop would be `ui='small'`.

layershifter:
@clemensw do you have plans for PR update?

levithomason:
Would like to merge @clemensw's work, however, whoever has time to get to this first let's ship it! :)

clemensw:
Thank you for the review.  I'm seriously confused by the logic behind Semantic UI 's use of the ui class in css.  I've updated the PR based on your suggestions (substituting getUnhandledProps for SUI on the import and rest for props on the Image attributes).  I booleanized size by checking against undefined, but will re-push the PR with your solution.

layershifter:
@clemensw Thanks for updates :smiley_cat: 
@levithomason LGTM, ready to go?

levithomason (approved):

levithomason:
Thanks for the update and help!

PR #748, opened 2016-10-25
feat(Reveal): Add component

### LINKED ISSUE #182: <Reveal /> Component
http://semantic-ui.com/elements/reveal.html

layershifter:
Fixes #182

### DIFF
--- a/src/elements/Reveal/Reveal.js
+++ b/src/elements/Reveal/Reveal.js
@@ -6,20 +6,29 @@ import {
   getElementType,
   getUnhandledProps,
   META,
-  SUI,
+  useKeyOnly,
 } from '../../lib'
+import RevealContent from './RevealContent'
 
 /**
  * A reveal displays additional content in place of previous content when activated.
  */
 function Reveal(props) {
   const {
+    active,
     children,
     className,
+    disabled,
+    effect,
+    instant,
   } = props
 
   const classes = cx(
     'ui',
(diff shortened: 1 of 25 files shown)


Jlassos:
Thanks

layershifter:
We have problem with augmentation on Rotate example, we cannot it use there because `Image` doesn't render children. Or I missed something?

---

Tests and docs are ready, so we need review there.

src/elements/Reveal/Reveal.js
```diff
@@ -0,0 +1,78 @@
...
+  children: PropTypes.node,
+
+  /** Additional classes. */
+  className: PropTypes.string,
+
+  /** A disabled reveal will not animate when hovered. */
+  disabled: PropTypes.bool,
+
+  /** An effect's name that will be applied to Reveal. */
+  effect: PropTypes.oneOf(Reveal._meta.props.effect),
```
levithomason, line 70: Suggest we go with `animation` here to keep in line with the [Button `animated`](http://react.semantic-ui.com/elements/button#animated) prop.

levithomason (approved):
This looks great, thanks!

src/elements/Reveal/Reveal.js
layershifter: I'll make PR for `animated` tomorrow.

Issue #623, opened 2016-10-04
RFC: all event handlers should callback with (event, data)


craig1123:
Hi, I have a simple enhancement to propose.

It would be nice to have access to the options of a Dropdown using the `onMouseEnter` handler. There is obviously already an `onChange` handler which allows me to execute some function when I select an option, but I want to hover over the items as well, triggering some function. 

Thanks, I really enjoy working with stardust.

levithomason:
> /cc @jcarbo @layershifter, speaking of [_always_ passing props](https://github.com/TechnologyAdvice/stardust/pull/619#issuecomment-251504914) with all event handlers ^ 😸 

@craig1123 Your timing was impeccable here, I was just suggesting to the team here that we should always pass `(e, props)` to every handler.  Granted, we don't have an `onMouseEnter` handler in this component.  We'd need to add any event handlers that should receive props.

I think this is useful and we should consider making it a library feature, and not just add it to this one component for this one handler.  I'll leave it at that for now and suggest we start looking at how to achieve this consistently.  Ideas welcomed.

layershifter:
LGTM, I think we need there list of components with such events and update them all.

levithomason:
_This is a living spec open to changes, feel free to comment for changes._ 💬

***

## New Callback Parameters

### Use `event|null`

React's SyntheticEvent should always be passed back, if available.  If there is no event available, such as in Portal `onMount`, pass `null`.

### Use `data` instead of `props`

There are several components which pass back information that is not included in `props`, such as an index or a `state` value.  There may be props that we do not want to pass back at certain times as well.

I propose we then change the name of the `props` callback param to `data` and pass back as much `props` and `state` as we can that makes sense.

### ~~Reverse Parameter Order? (feedback please!)~~

NOTE: As per further discussion below, we decided to keep the order as-is `(e, data)`. Keeping this here for posterity.

~~Given the above, I think it makes sense to reverse the callback parameter order from `(e, data)` to `(data, e)`.  My reasoning is as follows:~~

1. ~~`data` covers the majority of use cases.~~
1. ~~`data` should be preferred over `event` when they both contain the same information (ie `value`).~~
1. ~~`event` is only provided as an escape hatch for edge cases, such as focus.~~
1. ~~`event` is sometimes `null`, making it applicable in fewer cases.~~
1. ~~This will already be a breaking change, which creates an opportunity for this kind of update.~~

## New Documentation

### Current

[...]

PR #749, opened 2016-10-25
feat(Popup): Enabled controlled usage

### LINKED ISSUE #727: Does the <Popup/> Component have 'manual' behavior as the normal popup does?
in the normal semantic ui popup element we are able to control is state as when to show and hide by setting the tiger to 'manual'.
But the react version does'nt seem to have that feature.
So is there anyway we can achieve that result?
any help is much appreciated...

jeffcarbs:
Fixes #727 

Previously the popup was swallowing the `onOpen` event so you couldn't control the `open` prop from outside. This enables that and adds an example in the "Usage" section.

This also:
- Cleans up the order of props passed to the portal to enable more user control. Previously any props derived from the "on" prop would win, now you can override those.
- Rename the popup examples to the new naming scheme.
- Move "Triggers" section of examples to "Usage"

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -355,7 +355,7 @@ class Portal extends Component {
     document.addEventListener('keydown', this.handleEscape)
 
     const { onMount } = this.props
-    if (onMount) onMount(undefined, this.props)
+    if (onMount) onMount(null, this.props)
   }
 
   unmountPortal = () => {
@@ -376,7 +376,7 @@ class Portal extends Component {
     document.removeEventListener('keydown', this.handleEscape)
 
     const { onUnmount } = this.props
-    if (onUnmount) onUnmount(undefined, this.props)
+    if (onUnmount) onUnmount(null, this.props)
   }
 
   render() {
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -287,7 +287,7 @@ class Portal extends Component {
     debug('open()')
 
     const { onOpen } = this.props
-    if (onOpen) onOpen(e)
+    if (onOpen) onOpen(e, this.props)
 
     this.trySetState({ open: true })
   }
@@ -304,7 +304,7 @@ class Portal extends Component {
     debug('close()')
 
     const { onClose } = this.props
-    if (onClose) onClose(e)
+    if (onClose) onClose(e, this.props)
 
     this.trySetState({ open: false })
   }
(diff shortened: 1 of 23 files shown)


src/modules/Popup/Popup.js
```diff
@@ -233,19 +240,32 @@ export default class Popup extends Component {
...
-  onOpen = (event) => {
-    this.coords = event.currentTarget.getBoundingClientRect()
+  handlePortalMount = (e) => {
     if (this.props.hideOnScroll) {
       window.addEventListener('scroll', this.hideOnScroll)
     }
   }
 
+  // Note: We can just pass this through but since we need a custom onOpen
+  // handler it's more clear to define the onClose handler in the same way
```
levithomason: Agreed entirely.

src/modules/Popup/Popup.js
```diff
@@ -233,19 +240,32 @@ export default class Popup extends Component {
...
+  handleClose = (e) => {
+    const { onClose } = this.props
+    if (onClose) onClose(e)
+  }
+
+  handleOpen = (e) => {
```
levithomason, line 260: I know we haven't standardized it yet (#623), but it seems that should just start passing an object as the second param with at least props.  I say "at least" since sometimes it will have values from state, such as `value` or certain auto-controlled props.

Thoughts?

levithomason (approved):
Looks good, one feedback question.

src/addons/Portal/Portal.js
```diff
@@ -355,7 +355,7 @@ class Portal extends Component {
     document.addEventListener('keydown', this.handleEscape)
 
     const { onMount } = this.props
-    if (onMount) onMount()
+    if (onMount) onMount(undefined, this.props)
```
jeffcarbs: @levithomason - this doesn't happen via a user event, so there's no event to pass through. Should we:
- pass undefined, since there's technically no event
- pass an empty object so if people try to access properties like `e.currentTarget` or something it won't die.

jeffcarbs:
@levithomason - I added the props to the callbacks as the second arg. I wound up updating the modal callbacks as well - nothing breaking just passing the additional argument.

One question about the onMount/onUnmount not having an event (see https://github.com/Semantic-Org/Semantic-UI-React/pull/749#discussion_r85033486) otherwise should be g2g.

src/addons/Portal/Portal.js
levithomason: My 0.02 says in that case we pass `null`.  Even with an empty object, the common `e.target.value` would fail.  I think `null` as it says "this is intentionally empty" vs `undefined` which looks like the argument is simply missing.

levithomason (changes requested):

jeffcarbs:
`null` sounds good for those callbacks. Just rebased and pushed the latest, should be g2g 👍

levithomason (approved):

levithomason:
Awesome, thanks again @jcarbo!

PR #756, opened 2016-10-28
style(Button): update docs to the latest standards, add missing prop …

dpkwhan:
1. update docs to the latest standards, 
2. add missing prop 'vertical' and 
3. delete one unused example


levithomason:
Awesome, great catch on the missing `vertical` prop!

PR #757, opened 2016-10-28
feat(Dropdown): support additionLabel components

jayphelps:
Dropdown now wraps additions in `<b>` tags and uses `additionPosition="top"` by default, just like the jQuery semantic-ui plugin.

image

There is also some more subtle changes to align with the jQuery version:
- default additionLabel is now "Add" without the semicolon, and includes the space between the label and the addition. This allows users to use labels which do not wish to have space between the two. 
  - this overall API differs from the jQuery version which uses unsafe templates: `addResult     : 'Add <b>{term}</b>'`. We could instead use callbacks, if even more alignment was preferred. `addResult={term => <span>Add <b>{term}</b></span>`
  - or of course, we could use templates too and use `dangerouslySetInnerHTML` 😨 
- The addition item now has the `addition` className added to it--this can be important for themes

You will notice that, because of a React limitation/design I am passing `text` as an array of elements. Because of the current way the React diffing algorithm works, even though these elements will always be in the same order we must define a `key` for them. This isn't pretty--but at least for me, having the same styling behavior of the wrapped `<b>` is critical to UX.

---

This was my first experience with enzyme/chai-enzyme so lmk if there is a better way to write the prop tests that contain the array of elements. It's not ideal 😞

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleAdditionLabelComponent.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleAdditionLabelComponent.js
@@ -9,7 +9,7 @@ const options = [
   { text: 'Chinese', value: 'Chinese' },
 ]
 
-class DropdownExampleAllowAdditions extends Component {
+class DropdownExampleAdditionLabelComponent extends Component {
   state = { options }
 
   handleAddition = (e, { value }) => {
@@ -18,21 +18,21 @@ class DropdownExampleAllowAdditions extends Component {
     })
   }
 
-  handleChange = (e, { value }) => this.setState({ currentValues: value })
+  handleChange = (e, { value }) => this.setState({ currentValue: value })
 
   render() {
-    const { currentValues } = this.state
+    const { currentValue } = this.state
 
     return (
       <Dropdown
         options={this.state.options}
-        placeholder='Choose Languages'
+        placeholder='Choose Language'
         search
         selection
         fluid
-        multiple
         allowAdditions
-        value={currentValues}
+        additionLabel={<i style={{ color: 'red' }}>Custom Language: </i>}
+        value={currentValue}
         onAddItem={this.handleAddition}
         onChange={this.handleChange}
       />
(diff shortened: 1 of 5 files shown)


levithomason:
This is awesome, thanks!  Would you consider adding an example showing use of a custom `additionLabel`?

jayphelps:
@levithomason to [docs/app/Examples/modules/Dropdown/Usage](https://github.com/Semantic-Org/Semantic-UI-React/tree/master/docs/app/Examples/modules/Dropdown/Usage) ?

something like this?

``` js
render() {
  // etc

  <Dropdown
    additionLabel="Custom: "
    {...etc}
  />

  // etc

  <Dropdown
    additionLabel={<i>Custom: </i>}
    {...etc}
  />
}
```

levithomason:
Exactly.  Two separate example files would be superb.  One showing a text label, and a separate one showing a component example.

jayphelps:
Added requested examples. 🎉

jayphelps:
Note: I forgot to mention that this is technically a breaking change and I have added such in the commit message for commitizen 

> BREAKING CHANGE: Dropdown now now wraps additions in <b> and uses
> additionPosition="top" by default

levithomason:
This looks great!  Just one lint issue to fix that is failing CI:

image

jayphelps:
doh! fixed. I incorrectly assumed it would lint as part of the test runner. Might be helpful to create a pre-commit hook that lints the code.

levithomason:
We used a few pre-commit ~~hooks~~ _solutions_ but found them to be problematic.  We just rely on the PR status.  Though, this obviously needs be a little more clear to developers!  Thanks for the feedback.

levithomason:
Could you clarify something regarding the breaking change? I'd like to release this as a feature rather than an API change.  My understanding:
- The exposed API hasn't change
- The previous additionLabel accepted plain text only
- The default additionPosition changed
- The additionLabel is now bold

My conclusion based on ^ these points is that old additionLabel code will still work, it will just be bold and default to the top of the Dropdown.  I see this as a non-breaking feature.  What are your thoughts?

levithomason:
I'm actually going to release this as a feature and we can adjust if there are issues 👍

jayphelps:
@levithomason okiedokie :shipit:

levithomason:
Released in `semantic-ui-react@0.58.1`

PR #759, opened 2016-10-28
docs(usage): Add usage section to docs

lewisblackwood:
This PR includes a 'Usage' section in the documentation, detailing the installation steps for Semantic UI React and the necessary stylesheet to use the project.

For those unfamiliar with component frameworks such as Semantic UI, it would be handy to have this information upfront in the docs. It would allow new users to get started with the React integration as quickly as possible.

I recently joined the Gitter chat to ask a question around this, and saw that a similar question had already been asked a few hours earlier. @levithomason provided some really useful info, which has formed the basis of this page. He also mentioned PRs for new docs were welcome.

Apologies if I've missed anything around contributing - let me know and I'll fix up.

### DIFF
--- a/docs/app/Views/Usage.js
+++ b/docs/app/Views/Usage.js
@@ -1,0 +1,118 @@
+import React from 'react'
+import pkg from 'package.json'
+import {
+  Container,
+  Header,
+  Segment,
+} from 'src'
+import Logo from '../Components/Logo/Logo'
+import { semanticUIDocsURL, semanticUIRepoURL, semanticUICSSRepoURL } from 'docs/app/utils'
+
+const suiCSSVersion = pkg.devDependencies['semantic-ui-css'].replace(/[~^]/, '')
+
+const Usage = () => (
+  <Container id='usage-page'>
+    <Segment basic textAlign='center'>
+      <Logo centered size='small' />
+      <Header as='h1' textAlign='center'>
+        Semantic-UI-React
+        <Header.Subheader>
+          {pkg.description}
+        </Header.Subheader>
+      </Header>
+    </Segment>
+
+    <Segment basic padded>
+      <Header as='h2' dividing>JavaScript</Header>
+      <p>
+        The Semantic UI React package can be installed via NPM:
+      </p>
+      <Segment>
+        <pre>$ npm install {pkg.name} --save</pre>
+      </Segment>
+      <p>
+        Installing Semantic UI React provides the JavaScript for your components.
+        You'll also need to include a stylesheet to provide the styling for your components.
+        This is the typical pattern for component frameworks, such as Semantic UI or Bootstrap.
+      </p>
+      <p>
+        The method you choose to include the stylesheet in your project will depend on the level
+        of customisation you require.
+      </p>
+    </Segment>
+
+    <Segment basic padded>
+      <Header as='h2' dividing>CSS</Header>
+
+      {/* ----------------------------------------
+       *  Content Delivery Network (CDN)
+       * -------------------------------------- */}
+      <Header as='h3'>Content Delivery Network (CDN)</Header>
+      <p>
+        You can use the default Semantic UI stylesheet by including a Semantic UI CDN link in your
+        <em>index.html</em> file.
+      </p>
+      <p>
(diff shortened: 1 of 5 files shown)


docs/app/Views/Usage.js
```diff
@@ -0,0 +1,116 @@
...
+      </Header>
+    </Segment>
+
+    <Segment basic padded>
+      <Header as='h2' dividing>Usage</Header>
+      <p>
+        The Semantic UI React package can be installed via NPM:
+      </p>
+      <Segment>
+        <pre>$ npm install {pkg.name}</pre>
```
levithomason: Can we add the `--save` flag here to remind users to persist the install to their deps?

Since this new page exists, we can also remove the installation instructions from the Introduction page.  Do you think we should replace that one with a `Setup` section that just links to this `Usage` section?

docs/app/Views/Usage.js
```diff
@@ -0,0 +1,116 @@
...
+        You can use the default Semantic UI stylesheet by including a Semantic UI CDN link in your
+        <em>index.html</em> file.
+      </p>
+      <p>
+        This is the quickest way to get started with Semantic UI React. You won't be able to use
+        custom themes with this method.
+      </p>
+      <Segment>
+        <pre>
+          {'<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.2.4/semantic.min.css"></link>'}
```
levithomason: The doc site uses `semantic-ui-css`, so we can inject the version number from the `package.json` here:

``` jsx
// top of the file
const suiCSSVersion = pkg.devDependencies['semantic-ui-css'].replace(/[~^]/, '')

// update template string
{`<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/${suiCSSVersion}/semantic.min.css"></link>`}
```

This way the instructions always show the same version the doc site itself is running.

docs/app/Views/Usage.js
```diff
@@ -0,0 +1,116 @@
...
+        </pre>
+      </Segment>
+    </Segment>
+
+    {/* ----------------------------------------
+     *  Semantic-UI-CSS package
+     * -------------------------------------- */}
+    <Segment basic padded>
+      <Header as='h2' dividing>Semantic UI CSS package</Header>
```
levithomason: This is great.  Can we update the headers to have some nesting?  Here's an example where I've made `JavaScript` and `CSS` as `h2`s and the three CSS options as `h3` and non-dividing:

image

docs/app/Views/Usage.js
```diff
@@ -0,0 +1,116 @@
...
+        The <a href={semanticUICSSRepoURL}>Semantic UI CSS package</a> is automatically synced with the
+        main Semantic UI repository to provide a lightweight CSS only version of Semantic UI.
+      </p>
+      <p>
+        Semantic UI CSS can be installed as a package in your project using NPM. You won't be able to
+        use custom themes with this method.
+      </p>
+      <Segment>
+        <pre>
+          $ npm install semantic-ui-css
```
levithomason: Let's also add `--save` here.

docs/app/Views/Usage.js
```diff
@@ -0,0 +1,116 @@
...
+        Semantic UI includes Gulp build tools so your project can preserve its own theme changes,
+        allowing you to customise the style variables.
+      </p>
+      <p>
+        Detailed documentation on theming in Semantic UI is provided
+        <a href={`${semanticUIDocsURL}usage/theming.html`}>here</a>.
+      </p>
+      <Segment>
+        <pre>
+          $ npm install semantic-ui
```
levithomason: This one should probably be `--save-dev` given it installs a build system and the final stylesheet is a product of that build.

levithomason (changes requested):

levithomason:
Awesome stuff, thanks!  Left a few comments and questions.

lewisblackwood:
Thanks for the comments 👍

I've moved the introduction section around slightly. I wanted to maintain the flow of the feature list and corresponding sections. Let me know if you'd prefer the link to Usage under a separate sub-heading.

levithomason:
Perfect.  😸

PR #760, opened 2016-10-28
breaking(Reveal): change prop name

layershifter:
This PR renames `effect` prop to `animated` for consistency with `Button` component.


levithomason:
Thanks for the update!

PR #767, opened 2016-10-29
docs(Form): update to latest standards and make minor fixes 

dpkwhan:
1. refactor
2. fix placeholders in read-only example
3. add missing .
4. add one example for inline group fields


levithomason:
Thanks again!

PR #781, opened 2016-11-02
fix(Dimmer): allow children to be clicked

### LINKED ISSUE #780: Dimmer - Inactive dimmer still overlays segment contents, making them unclickable
### Steps
1. Create a Segment with a Dimmer, Loader, and Button
```html
      <Segment>
        <Dimmer active={false} inverted>
          <Loader />
        </Dimmer>
        <Button>Test</Button>
      </Segment>
```

2. Click the button

### Expected Result

Button should be clickable. Cursor should be pointer.

### Actual Result

Button is unclickable. Inspecting the DOM shows that the dimmer node is still present, preventing clicks from items behind it.

### Version

0.58.1

### Testcase

http://codepen.io/ben174/pen/pEMYVG

ben174:
Fixes #780 by adding a `disabled` class when the Dimmer is not active. Also adds a test to ensure the fix isn't regressed.

### DIFF
--- a/src/modules/Dimmer/Dimmer.js
+++ b/src/modules/Dimmer/Dimmer.js
@@ -83,6 +83,7 @@ export default class Dimmer extends Component {
     const classes = cx(
       'ui',
       useKeyOnly(active, 'active'),
+      useKeyOnly(!active, 'disabled'),
       useKeyOnly(inverted, 'inverted'),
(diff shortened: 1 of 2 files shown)


ben174:
Sorry, committed this from my work account. Going to close and fix my branch and re-open this PR.

PR #782, opened 2016-11-02
fix(Dimmer): allow children to be clicked

### LINKED ISSUE #780: Dimmer - Inactive dimmer still overlays segment contents, making them unclickable

ben174:
Fixes #780 by adding a disabled class when the Dimmer is not active. Also adds a test to ensure the fix isn't regressed.

### DIFF
--- a/test/specs/modules/Dimmer/Dimmer-test.js
+++ b/test/specs/modules/Dimmer/Dimmer-test.js
@@ -27,11 +27,16 @@ describe('Dimmer', () => {
         .should.contain.text(text)
     })
 
-    it('renders clickable buttons', () => {
-      const spy = sandbox.spy()
-      const wrapper = mount(<Dimmer><a onClick={spy} href='#'>{faker.hacker.phrase()}</a></Dimmer>)
-      wrapper.find('a').simulate('click')
-      spy.should.have.been.callCount(1)
+  })
+
+  describe('active', () => {
+    it('removes the `disabled` className when true', () => {
+      shallow(<Dimmer active />)
+        .should.not.have.className('disabled')
+    })
+    it('adds the `disabled` className when false', () => {
+      shallow(<Dimmer active={false} />)
+        .should.have.className('disabled')
     })
   })
 
--- a/test/specs/modules/Dimmer/Dimmer-test.js
+++ b/test/specs/modules/Dimmer/Dimmer-test.js
@@ -26,7 +26,6 @@ describe('Dimmer', () => {
       shallow(<Dimmer content={text} />)
         .should.contain.text(text)
     })
-
   })
 
   describe('active', () => {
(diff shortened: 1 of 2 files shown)


layershifter:
We need asserts for `disabled` there, too.

ben174:
Good point, I just tried creating such assertions while disabled and am getting false positives. I've tested it in the wild and am sure it's working correctly, but the test seems to be simulating a click directly against the anchor and bypassing the overlying element. 

Do you know of a way to simulate an actual click on the anchor as if a user was clicking on the DOM, rather than clicking on the element itself?

test/specs/modules/Dimmer/Dimmer-test.js
```diff
@@ -26,6 +26,13 @@ describe('Dimmer', () => {
       shallow(<Dimmer content={text} />)
         .should.contain.text(text)
     })
+
+    it('renders clickable buttons', () => {
```
levithomason: This test is already covered by `common.rendersChildren()` [above](https://github.com/Semantic-Org/Semantic-UI-React/pull/782/files#diff-65fcf9237dea10f9c70c3c54c9a8428eR14).  We can remove this one.

levithomason:
The click simulation is not subject to z-index blocking effects like a user is.  You can simulate an event on any node that can be accessed programmatically.  The event is dispatched via a reference to the node, unlike a user event which is the result of the cursor position and top most element.

Instead, add a test that asserts the `disabled` className is added when `active` is false, and another that asserts `disabled` is _not_ present when `active` is true.  I did not test this, but it should be something like this:

``` jsx
describe('active', () => {
  it('removes the `disabled` className when true', () => {
    shallow(<Dimmer active />)
      .should.not.have.className('disabled')
  })
  it('adds the `disabled` className when false', () => {
    shallow(<Dimmer active={false} />)
      .should.have.className('disabled')
  })
})
```

ben174:
Makes sense, I was trying to write something that actually validates that the item is clickable, rather than trusting that the `disabled` class fixes it. But I suppose tests covering the functionality of `disabled` elsewhere in the suite would be a more appropriate place for that.

levithomason:
Correct, really the sizing would be a CSS test since it is testing what the result of the CSS rule is for that HTML class.  We only want to assert that we are generating the correct markup.

levithomason:
Released in `semantic-ui-react@0.58.2`, thanks much!

PR #783, opened 2016-11-02
feat(Embed): Add component

### LINKED ISSUE #191: <Embed /> Component
http://semantic-ui.com/modules/embed.html

layershifter:
Fixes #191

### DIFF
--- a/src/modules/Embed/Embed.js
+++ b/src/modules/Embed/Embed.js
@@ -41,8 +41,20 @@ export default class Embed extends Component {
     /** An embed can be active. */
     active: PropTypes.bool,
 
+    /** Setting to true or false will force autoplay. */
+    autoplay: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.bool,
+    ]),
+
     /** An embed can specify an alternative aspect ratio. */
     aspectRatio: PropTypes.oneOf(_meta.props.aspectRatio),
+
+    /** Whether to show networks branded UI like title cards, or after video calls to action. */
+    brandedUI: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.bool,
+    ]),
 
     /** Primary content. */
     children: PropTypes.node,
@@ -50,8 +62,26 @@ export default class Embed extends Component {
     /** Additional classes. */
     className: PropTypes.string,
 
+    /** Specifies a default chrome color with Vimeo or YouTube. */
+    color: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.string,
+    ]),
+
     /** Initial value of active. */
     defaultActive: PropTypes.bool,
+
+    /** Whether to show networks branded UI like title cards, or after video calls to action. */
+    hd: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.bool,
+    ]),
+
+    /** Specifies an id for source. */
+    id: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.string,
+    ]),
 
     /** Specifies an icon to use with placeholder content. */
     icon: customPropTypes.itemShorthand,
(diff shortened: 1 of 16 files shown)


layershifter:
Rebased to latest master, we're ready for review :+1:

src/modules/Embed/Embed.js
```diff
@@ -0,0 +1,183 @@
...
+    /** Specifies an id for source. */
+    sourceId: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.string,
+    ]),
+
+    /** Settings to configure video behavior. */
+    sourceSettings: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.shape({
```
levithomason: I think we should flatten these settings as top level props, instead of a nested object. They are applicable only to embedded videos, but I don't think that warrants nesting these props.  Let me know if you disagree and why.

src/modules/Embed/Embed.js
```diff
@@ -0,0 +1,183 @@
...
+    )
+    const rest = getUnhandledProps(Embed, this.props)
+    const ElementType = getElementType(Embed, this.props)
+
+    return (
+      <ElementType {...rest} className={classes} onClick={this.handleClick}>
+        {Icon.create(icon)}
+        {placeholder && <img className='placeholder' src={placeholder} />}
+
+        <EmbedEmbed active={active} src={this.getSource()}>{children}</EmbedEmbed>
```
levithomason: Given that `EmbedEmbed`:
1. used in only one place, in `Embed`
2. only uses props from `Embed`
3. is not an exposed sub component

I would suggest using a method instead of a new component:

``` jsx
renderEmbed = () => {
  const { active, children } = this.props
  if (!active) return null
  if (children) return <div className='embed'>{children}</div>

  return (
    <div className='embed'>
      <iframe
        allowFullScreen=''
        frameBorder='0'
        height='100%'
        scrolling='no'
        src={this.getSource()}
        width='100%'
      />
    </div>
  )
}

render() {
  return (
    // ...snip
    {this.renderEmbed}
  )
}
```

levithomason (changes requested):

layershifter:
Removed `EmbedEmbed` and updated tests, looks better :+1:

src/modules/Embed/Embed.js
```diff
@@ -0,0 +1,198 @@
...
+    if (!active) this.trySetState({ active: true })
+  }
+
+  getSource() {
+    const {
+      source,
+      sourceAutoPlay = true,
+      sourceBrandedUI = false,
+      sourceColor = '#444444',
+      sourceHd = true,
```
layershifter: Just a note, I'm have there same problems with `demand` like in #827, so I left values there instead of `defaultProps`.

src/modules/Embed/Embed.js
```diff
@@ -0,0 +1,198 @@
...
+    ]),
+
+    /** Specifies an id for source. */
+    sourceId: customPropTypes.every([
+      customPropTypes.demand(['source']),
+      PropTypes.string,
+    ]),
+
+    /** Specifies a url to use for embed. */
+    sourceUrl: customPropTypes.every([
```
levithomason: One last update, can we remove the `source` prefixing from these props?  I'm going for parity with the Semantic-UI core settings:

image

levithomason (changes requested):

levithomason:
I'm preparing a release, so I'll go ahead and update the props :)

layershifter:
I'll updating it now :smile:

levithomason:
Whoops, thanks for the rapid reply!

levithomason (approved):

levithomason:
Perfect, will merge and cut a release when tests pass.

levithomason:
Released in `semantic-ui-react@0.60.3`

Issue #765, opened 2016-10-28
Improve doc build time and load time performance


levithomason:
### Build times

Webpack build and load time performance was improved some time ago reducing build times from ~16s down to ~1s.  Something recently has increased the doc build up to ~27s.  This is a massive decrease in performance, 27x worse than it once was.
### Load times

I've also noticed the doc site load times are becoming unacceptably long.

We should investigate the bundle sizes and, at a minimum, skip sourcemaps on the live docs.

```
 27M    ./docs/build
660K    ./docs/build/assets
5.8M    ./docs/build/app.b7becd54b1798794f19a.js
6.1M    ./docs/build/app.b7becd54b1798794f19a.js.map
6.6M    ./docs/build/vendor.b7becd54b1798794f19a.js
7.7M    ./docs/build/vendor.b7becd54b1798794f19a.js.map
```

We should also consider splitting the bundle by page on the doc site to improve performance.
### Development & HMR

The current hot reload transforms do not work with many of our components due to the way the dependencies are resolved with Webpack's `require.context()`.  We should remove the extra transforms and just use vanilla HMR.  This will reduce the development footprint and also restore browser reloads on change.

PR #784, opened 2016-11-02
feat(Image): Render children appropriately

jeffcarbs:
Previously you could render `<Image>` with children but they  #would be passed to the ``, which is not technically valid markup.

Here's debug output from the `rendersChildren` test:
```js
<img className="ui image">
  Children are rendered here :|
</img>
```

This updates the `Image` component to render the `.ui.image` wrapper if children are present, which brings this more inline with the `Input` component. Both are wrapping elements around an HTML element which technically can't take children. However, since there could be other elements inside the wrapper this gives the user more control over how it renders.


layershifter:
@jcarbo May be we can add example in usage section that explains markup?

jeffcarbs:
Added an example for rendering children.
screen shot 2016-11-02 at 1 05 34 pm

I also wound up:
- pulling in those wireframe images into this repo so we can serve them locally instead of hotlinking to SUI-core
- updating the title/descriptions of the Image examples to match SUI-core.

docs/app/Examples/elements/Image/Variations/index.js
```diff
@@ -1,43 +1,42 @@
...
-      title='Shapes'
-      description='An image can have different shapes.'
+      title='Rounded'
+      description='An image may appear rounded.'
       examplePath='elements/Image/Variations/ImageExampleRounded'
     />
-    <ComponentExample examplePath='elements/Image/Variations/ImageExampleCircular'>
-      <Message warning>
-        Perfectly circular images require a perfectly square image file.
-      </Message>
```
levithomason, line 31: I think we should leave this warning in here.  I lost quite a bit time trying to debug why our images were oval and SUI core was circular.  Awkwardly enough, the `circular` class does _not_ make circular images, it makes "fully rounded" images if you will.  There is a similar warning in the SUI core docs.

levithomason (changes requested):
> pulling in those wireframe images into this repo so we can serve them locally instead of hotlinking to SUI-core

I have a suspicion that images are partly what is bloating the bundle / rebuild times, #765.  This PR brings the `docs` task build time up to ~56s for me with bundle/rebuild times to ~16s/3s :(  Compare this to previous bundle/rebuild times of ~7s/900ms.

This may be solvable with some build tweaking, but I'd rather not burden the most common use case (bundle, rebuild on change) for the less common use case (offline dev).

Lastly, we're also exposing unknown globals in the docs which I'd rather avoid.  I think it could confuse some users, but more so, I'd just rather keep the doc examples limited to "vanilla" React/SUIR code and not include any of our own proprietary code or internal conventions (ie `__BASE__`).

We can keep these for now, but I'm not sure there is a bright future for local images unless we solve these other things another way.  One idea is to match our file structure to SUI core docs structure, then just use `/` and rely on the `basename` to handle the rest.  Then, we can turn off any other image handling/processing that might be happening to help rebuild times.

levithomason:
Just one comment regarding the circular warning.

jeffcarbs:
@levithomason - I added the message about circular images back as requested. This should be g2g I think 👍

layershifter:
@jcarbo I aggre with @levithomason, `__BASE__` looks horrible. Do we really need it there?

I looked on Dropdown examples:

``` js
image: { avatar: true, src: `${__BASE__}assets/images/avatar/small/jenny.jpg` },
```

It will be converted `/assets/images/avatar/small/jenny.jpg` in docs site and in local, too. I'm think can simply replace `${__BASE__}` with `/` there.

---

Also, looks there is missing main, test for children rendering.

jeffcarbs:
@layershifter - There was already a test for rendering children: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/elements/Image/Image-test.js#L13. It wasn't failing because it was just rendering children within the `` tag:

```
<img className="ui image">
  Children are rendered here :|
</img>
```

I would've thought React would complain since that's not valid markup but maybe that's an issue withe `enzyme`?

I'll give a shot at removing `__BASE__`. FWIW, it was something introduced [before my time](https://github.com/Semantic-Org/Semantic-UI-React/pull/395/files#diff-1264673dc9aa9600177ee108da530f4eR71) so blame @levithomason 😆

jeffcarbs:
Just removed usage of `__BASE__`.

I realized what it was for - the docs for stardust were served from the `/stardust` path at `technologyadvice.github.io` while the local docs were served from the `/` path. `__BASE__` was used to account for the difference. After switching to SUI-react, both the prod and local docs are served from `/` so we can just use the root absolute path for images (e.g. `/assets/images/...`).

Also, I don't think the images are the culprit with slow build times:
- on first build: they may a little but it's just a raw copy/paste: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/gulp/tasks/docs.js#L56-L59. I can't see how that would drastically decrease performance.
- for rebuilds: they will only be rebuilt if an image file was touched: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/gulp/tasks/docs.js#L139-L148 so the images should have no effect on rebuild time.

I think this is g2g. @levithomason - when you're ready 👍

docs/app/Components/TAAttribution/TAAttribution.js
```diff
@@ -68,7 +68,7 @@ class TAAttribution extends Component {
           inline
           verticalAlign='middle'
           style={imageStyle}
-          src={`${__BASE__}assets/technologyadvice-logo-dark.png`}
+          src={'/assets/technologyadvice-logo-dark.png'}
```
layershifter: `src={'/assets/technologyadvice-logo-dark.png'}` can be replaced with `src='/assets/technologyadvice-logo-dark.png'`

docs/app/Examples/elements/Image/States/ImageExampleHidden.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
 
-const src = `${__BASE__}assets/images/wireframe/image.png`
+const src = '/assets/images/wireframe/image.png'
```
layershifter: Seems, we can move `src` to component:

``` jsx
<Image hidden size='small' src='/assets/images/wireframe/image.png' />
```

docs/app/Examples/elements/Image/Types/ImageExampleImage.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
 
-const src = `${__BASE__}assets/images/wireframe/image.png`
+const src = '/assets/images/wireframe/image.png'
```
layershifter: ^

docs/app/Examples/elements/Image/Types/ImageExampleLink.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
 
-const src = `${__BASE__}assets/images/wireframe/image-text.png`
+const src = '/assets/images/wireframe/image-text.png'
```
layershifter:
``` jsx
<Image
  as='a'
  href='http://google.com'
  target='_blank'
  size='medium'
  src='/assets/images/wireframe/image-text.png'
/>
```

Will look more clean.

docs/app/Examples/elements/Image/Types/ImageExampleWrapped.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
 
-const src = `${__BASE__}assets/images/wireframe/image.png`
+const src = '/assets/images/wireframe/image.png'
```
layershifter: ^

docs/app/Examples/elements/Image/Variations/ImageExampleAvatar.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
 
-const src = `${__BASE__}assets/images/wireframe/square-image.png`
+const src = '/assets/images/wireframe/square-image.png'
```
layershifter: ^

jeffcarbs:
@layershifter - just updated the `src` to put them inline. There were a few instances where the same `src` were used many times in the same example, so in those instances I kept it as a variable.

levithomason:
To be clear, we can/should use the globals (`__BASE__`, `__DEV__`, etc) in _private_ code such as doc site components themselves, I just don't want them to show up in example code snippets (docs/app/Examples/*).

levithomason (approved):

PR #792, opened 2016-11-04
feat(Dropdown): Prevent open when disabled

### LINKED ISSUE #791: Prevent dropdown from opening when disabled
Currently when the dropdown gets the `disabled` prop it sets the `disabled` class. The default `disabled` styling in SUI-css has the rule `pointer-events: none` which prevents the dropdown from opening by preventing mouse events. However this has two downsides:
- You can't use a `Popup` on a disabled dropdown because the `Popup` depends on mouse events to trigger.
- This only prevents opening on mouse events. The dropdown still opens when tab-focused (screenshot of disabled but open via focus):
screen shot 2016-11-03 at 10 33 18 pm

Here's the example on codepen: http://codepen.io/jeffcarbs/pen/yVLgJv?editors=0010

We should prevent the dropdown from opening when in the disabled state in the component itself. FWIW, SUI-core handles this correctly: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.js#L3144-L3146

Some thoughts in terms of how to handle this:
- Preventing the `open` state from being set to true will be tough, if not impossible, given it's an auto controlled component.
- We could just prevent the menu from opening by changing `const menuClasses = open ? 'visible' : ''` to `const menuClasses = (open && !disabled) ? 'visible' : ''`.
  - There's a lot of logic happening that's purely dependent on the `open` state.
  - It would be weird for state to be `open` and the menu to not be open.
  - Not sure why someone would want to, but you wouldn't be able to manually set open AND disabled at the same time. I guess I could see displaying a disabled notifcation in-menu or something.

I'm kinda leaning 1, with the idea that SUI-react (in the uncontrolled usage) won't set state to open if disabled. However, we won't prevent the user from setting it to open in the controlled state if they choose to.

Thoughts?

cc @levithomason  @layershifter

jeffcarbs:
Fixes #791

This PR fixes a few issues with the Dropdown:
- Prevent the dropdown from opening/focused when disabled.
- Respect the user's `tabIndex` if passed. Previously there was no way to prevent the dropdown from being focusable
- Add onOpen/onClose props to fire when the dropdown would've been opened/closed
- Fix issue with selectedIndex. If your dropdown has options that load after the component is mounted (e.g. from a remote API) the selectedIndex would always be -1 instead of the selected value or the first item. This updates the Dropdown to recompute the selected index when the options change.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -295,6 +295,10 @@ export default class Dropdown extends Component {
       debug('value changed, setting', nextProps.value)
       this.setValue(nextProps.value)
     }
+
+    if (!_.isEqual(nextProps.options, this.props.options)) {
+      this.setSelectedIndex(undefined, nextProps.options)
+    }
   }
 
   componentDidUpdate(prevProps, prevState) { // eslint-disable-line complexity
@@ -584,8 +588,8 @@ export default class Dropdown extends Component {
 
   // There are times when we need to calculate the options based on a value
   // that hasn't yet been persisted to state.
-  getMenuOptions = (value = this.state.value) => {
-    const { multiple, search, allowAdditions, additionPosition, additionLabel, options } = this.props
+  getMenuOptions = (value = this.state.value, options = this.props.options) => {
+    const { multiple, search, allowAdditions, additionPosition, additionLabel } = this.props
     const { searchQuery } = this.state
 
     let filteredOptions = options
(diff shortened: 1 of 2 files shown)


jeffcarbs:
@levithomason - this also gets us back to 100% coverage 🎉 . I'd appreciate if you could take a look as it fixes a few things that I'm currently running into using it in my app.

levithomason (approved):

levithomason:
Thanks much 👍

PR #795, opened 2016-11-04
fix(Dimmer): Fix "simple" usage, add disabled prop

jeffcarbs:
I realized that #782 breaks the "simple" example: http://react.semantic-ui.com/modules/dimmer#simple-dimmer. That should be enabled by virtue of having a parent with `ui dimmable dimmed`. See the SUI-core docs: http://semantic-ui.com/modules/dimmer.html#simple-dimmer

Digging into it further, two things stuck out to me:
- The reason the children were unclickable is because the dimmer always had the `visible` class, regardless of whether it was active. `active` controls the opacity, `visible` sets the `visibility`.
- `disabled` is a prop that overrides any active/visible state so we shouldn't be using it for that.

This fixes both issues.
- `visible` is only added when the dimmer is active, which removes the need to force `disabled`, which enables clicking on children and fixes the "simple" usage.
- We now expose `disabled` as a prop which conforms more to our API. A user can set that to override any active/visible state.

### DIFF
--- a/src/modules/Dimmer/Dimmer.js
+++ b/src/modules/Dimmer/Dimmer.js
@@ -37,6 +37,9 @@ export default class Dimmer extends Component {
     /** Shorthand for primary content. */
     content: customPropTypes.contentShorthand,
 
+    /** A disabled dimmer cannot be activated */
+    disabled: PropTypes.bool,
+
     /** Called with (event, props) after user's click. */
     onClick: PropTypes.func,
 
@@ -75,6 +78,7 @@ export default class Dimmer extends Component {
       children,
       className,
       content,
+      disabled,
       inverted,
       page,
       simple,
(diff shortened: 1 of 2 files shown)


test/specs/modules/Dimmer/Dimmer-test.js
```diff
@@ -29,13 +29,13 @@ describe('Dimmer', () => {
   })
 
   describe('active', () => {
-    it('removes the `disabled` className when true', () => {
+    it('adds the `transition visible` className when true', () => {
```
levithomason: The common test can be configured to assert a className that is different than the prop name.  Should be able to replace this describe block with:

``` jsx
common.propKeyOnlyToClassName(Dimmer, 'active', {
  className: 'active transition visible',
})
```

src/modules/Dimmer/Dimmer.js
```diff
@@ -75,31 +78,32 @@ export default class Dimmer extends Component {
...
       page,
       simple,
     } = this.props
 
     const classes = cx(
       'ui',
-      useKeyOnly(active, 'active'),
-      useKeyOnly(!active, 'disabled'),
+      useKeyOnly(active, 'active transition visible'),
+      useKeyOnly(disabled, 'disabled'),
```
levithomason, line 90: How about adding a common test for this?

``` jsx
common.propKeyOnlyToClassName(Dimmer, 'disabled')
```

levithomason (changes requested):
This is great, thanks much.  Couple updates to tests then we can merge this guy.

src/modules/Dimmer/Dimmer.js
jeffcarbs: Added 👍

jeffcarbs:
Added those tests and removed the ones for "active" that were testing duplicate functionality already covered by the common tests. Should be g2g 👍

levithomason (approved):

PR #800, opened 2016-11-06
docs(Modal): fix icon prop in Button, fix labeled prop, add index.js and trailing . in description

dpkwhan:


levithomason:
Thanks much!

PR #809, opened 2016-11-08
breaking(Dropdown|Label): Multi-select customization

jeffcarbs:
# Breaking Change

Removes the `removable` prop from `Label`. This isn't in SUI-core and it didn't really make sense since it would show the icon without handling the click in some way.

### Upgrade Guide

- `removable ` is no longer a valid prop
- You must pass an `onRemove` for the remove icon to show up

**BEFORE**
```jsx
<Label removable />
```

**AFTER**
```jsx
<Label onRemove={handleRemove} />
```

----

Multiple improvements to `<Dropdown multiple />`
- Set label to "active" on click
- Allow custom rendering of labels
- Add handler for label click

**Active**
screen shot 2016-11-08 at 12 01 20 pm

**Custom Label**
screen shot 2016-11-08 at 12 00 35 pm

Also adds an `active` prop to the `Label` element. This is actually a valid class name in SUI-core for `Label` outside of the context of a `Dropdown`:
screen shot 2016-11-08 at 11 58 16 am
The one on the right has the `active` class applied.

Also adds `removeIcon` prop to be able to customize the icon that triggers `onRemove`

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMultipleCustomLabel.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMultipleCustomLabel.js
@@ -1,0 +1,27 @@
+import React from 'react'
+import { Dropdown, Icon, Label } from 'semantic-ui-react'
+
+const options = [
+  { text: 'One', value: 1 },
+  { text: 'Two', value: 2 },
+  { text: 'Three', value: 3 },
+]
+
+const renderLabel = (label, index, props) => ({
+  color: 'blue',
+  content: `Customized label - ${label.text}`,
+  icon: 'check',
+})
+
+const DropdownExampleMultipleCustomLabel = () => (
+  <Dropdown
+    multiple
+    selection
+    fluid
+    options={options}
+    placeholder='Choose an option'
+    renderLabel={renderLabel}
+  />
+)
+
+export default DropdownExampleMultipleCustomLabel
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMultipleCustomLabel.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMultipleCustomLabel.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Dropdown, Icon, Label } from 'semantic-ui-react'
+import { Dropdown } from 'semantic-ui-react'
 
 const options = [
   { text: 'One', value: 1 },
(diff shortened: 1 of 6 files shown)


jeffcarbs:
@levithomason - The tests were failing are related to [`assertValidShorthand`](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/commonTests.js#L652-L657). This generates an element via `createShorthand` using some default props (`shorthandDefaultProps`) and ensures that it's in the shallow rendering of the element.

This works when the values of `shorthandDefaultProps` are literals, however there's no way to make this work when the component passes an event handler to the shorthand like this PR is trying to do: https://github.com/Semantic-Org/Semantic-UI-React/pull/809/files#diff-fde3c67bba82fd0efc2088a5fe1a4686R207

```
Icon.create(removeIcon, { onClick: this.handleRemove })
```

One option would be to lighten the expectation:

``` diff
-const renderedShorthand = createShorthand(ShorthandComponent, mapValueToProps, value, shorthandDefaultProps)
const element = createElement(Component, { ...requiredProps, [propKey]: value })

-shallow(element).should.contain(renderedShorthand)
+ shallow(element).should.have.descendants(name)
```

although you lose some granularity in the test with that approach. Also may get false positives since the selection isn't specific (e.g. if a component renders two `Icon`s a test that should fail may pass).

For now I've tests for the `removeIcon` manually. If we run into this again we may want to address the issue here.

levithomason:
Ugh, my review has been pending here for a day.  I apparently didn't hit submit, so I'll submit yesterday's now and take another look at your latest.  Apologies!

levithomason:
> For now I've tests for the removeIcon manually. If we run into this again we may want to address the issue here.

Agreed, some of the common tests (like this one) are getting pretty hairy.  We _could_ assert each of their prop names/values match, except for props with function values where we'd only assert the prop name matches and the value is "a function".  Will let this one cook for now.

jeffcarbs:
@levithomason - Addressed the feedback, should be g2g 👍 

For the `removeIcon` prop I added a more clear description explaining that the icon would trigger `onRemove` and also added `.demand(['onRemove'])` since the `removeIcon` is only rendered if `onRemove` is provided.

levithomason:
@jcarbo would mind adding a simplified before / after migration guide to the PR description?  You can reference https://github.com/Semantic-Org/Semantic-UI-React/pull/805.

levithomason:
Released in `semantic-ui-react@0.60.0`

meze:
Now `<Label>test</Label>` triggers `Failed prop type: 'removeIcon' prop in 'Label' requires props: 'onRemove.'` ;(

layershifter:
Thanks @meze. I've opened #827 for this problem.

PR #810, opened 2016-11-08
Dropdown.js bug fix

sarkistlt:
Fix for issue if you use array from state, as options for dropdown. Now if you pass options dynamically component don't update at all, only if you use 'ref'. [this.refs[DropdownComponent].forceUpdate]

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -255,6 +255,7 @@ export default class Dropdown extends Component {
   componentWillMount() {
     if (super.componentWillMount) super.componentWillMount()
     debug('componentWillMount()')
+    this.state.options = this.props.options
     const { open, value } = this.state
 
     this.setValue(value)
@@ -266,6 +267,8 @@ export default class Dropdown extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    this.setState({options: this.props.options});
+    
     super.componentWillReceiveProps(nextProps)
     debug('componentWillReceiveProps()')
     // TODO objectDiff still runs in prod, stop it
@@ -296,7 +299,7 @@ export default class Dropdown extends Component {
       this.setValue(nextProps.value)
     }
 
-    if (!_.isEqual(nextProps.options, this.props.options)) {
+    if (!_.isEqual(nextProps.options, this.state.options)) {
       this.setSelectedIndex(undefined, nextProps.options)
     }
   }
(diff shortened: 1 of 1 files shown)


levithomason:
Can you provide steps to reproduce the issue this is solving?  I'm not sure there is a bug here.

sarkistlt:
sure. 

```
...
let {options, defaultOptions} = this.state;

return (
                    <div className='file-form' key={idx}>
                        <label>{field[propName].label}</label>
                        {options ?
                            <Dropdown
                                ref={propName}
                                placeholder={field[propName].label}
                                id={propName}
                                fluid
                                multiple
                                selection
                                search
                                defaultValue={defaultOptions}
                                options={options}
                            /> : null}
                    </div>
                );
...
```

I'm sorry there are no issue with 'options', issue with 'defaultValue'.
I set 'defaultValue' in 'componentDidMount', so when component rendered 'defaultOptions' is empty, but when I set 'defaultOptions', it doesn't rerender component at all. 
this.forceUpdate(), doesn't work in this case too, it works only if I manually call from ref, like:
`this.refs[propName].forceUpdate()`

levithomason:
Great, thanks for the code snippet.  Can you also describe the expected behavior and the actual behavior?

sarkistlt:
**expected:**
Rerender dropdown component to visually see all selected by default options, as soon as I set  'defaultOptions' state.

**actual:**
After I set 'defaultOptions' state, I still see only empty select component, but when I click on it, it immediately shows all selected by default options.

levithomason:
It might help to note there is no `defaultOptions` prop.  You can set / update options at any time by using the `options` prop.

I apologies I still don't quite follow the issue here.  I've put together this Codepen that shows how to use `defaultValue` which asynchronously populated options.  Is this what you're trying to achieve?

http://codepen.io/levithomason/pen/rWOKRe

sarkistlt:
Yes correct, and 'defaultOptions' is not prop, its my state, which I use to pass default value. `defaultValue={defaultOptions}`

levithomason:
Is`defaultOptions` pre-populated on state before the initial render?  Default props in React are only applied on the first render.

sarkistlt:
No it's pre-populated after, in `componentDidMount`, so on the first render `defaultOptions` is empty.

levithomason:
Ah, that is the issue.  This is not a bug.  It is a little unintuitive on the part of React, but the `defaultValue` prop is only applied to `value` on the first render.  After that, all changes to `defaultValue` are ignored.

If you'd like to set the value asynchronously, use `value` instead.  This is the "controlled" component pattern.  The value you pass will _always_ be the rendered value.  So, you'll need to capture new values in `onChange` and set those on state as well.

### React Docs

This explains how to use controlled/uncontrolled components in React.  You'd be looking to use the "controlled" pattern.

https://facebook.github.io/react/docs/forms.html

### Semantic-UI-React Dropdown Examples

We have examples showing both usages as well.

Uncontrolled: http://react.semantic-ui.com/modules/dropdown#uncontrolled
Controlled: http://react.semantic-ui.com/modules/dropdown#remote

levithomason:
I should have also included:

### React deafultValue

https://facebook.github.io/react/docs/uncontrolled-components.html#default-values

PR #814, opened 2016-11-09
feat(Table.Cell): make Table.Cell selectable

### LINKED ISSUE #812: [Table] Add Possibility to Make Table.Cell Selectable
In semantic ui table not only row, but cell can be selectable, using `<td className="selectable">`, but current react bindings don't have this ability

saitonakamura:
WIP

I suggest that 3 steps should be done

- [x] make Table.Cell selectable
- [x] test it
- [x] document it

Am I right?

fixes #812


jeffcarbs:
@saitonakamura - looks good!

For tests, see `test/specs/collections/Table/TableCell-test.js`. You should be able to just follow the pattern of using `common.propKeyOnlyToClassName(TableCell, 'selectable')`

For documentation, we try to follow the SUI core docs exactly. So you should add an example like this http://semantic-ui.com/collections/table#selectable-cell.
- build an example component [like this](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/collections/Table/Variations/TableExampleSelectable.js)
- add it to the appropriate index file [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/collections/Table/Variations/index.js#L45-L51)

levithomason:
@saitonakamura Let's also rename `TableExampleSelectable.js` => `TableExampleSelectableRow.js`.  This way we can have `TableExampleSelectableCell.js`.  The index file linked by @jcarbo should also be updated to point to the renamed example.

saitonakamura:
I've added tests and implemented examples

But I didn't implement the inverted example because its broken on the official site, it's the same code as `inverted selectable row` and hasn't any `selectable cells` in it. I belive that we shouldn't worry about it very much and should just wait for official docs fix (of course I'll add an issue if there isn't one)

I also noticed using `"` brackets instead of `<code>` such as `Fixed "single line" tables`. Current official documentation using `Fixed <code>single line</code> tables`. Is there any reason for that or I just can fix it right away?

jeffcarbs:
Looks good 👍  The second inverted example in the SUI-core docs I think is there by mistake since, as you mentioned, it's just the same markup as the inverted row.

I see you're referring to the "single line" in the message above the second example here: http://semantic-ui.com/collections/table.html#fixed. I don't think there's any reason it's quoted in our docs vs being code. Feel free to update it to use `<code>` to match the SUI-core docs.

saitonakamura:
> Feel free to update it to use `<code>` to match the SUI-core docs.

done

I'm guessing I should rebase from master?

jeffcarbs:
LGTM 👍 You can rebase from the latest master if you'd like.

@levithomason - I think this is g2g but you may want to give a quick look.

saitonakamura:
Oh, I'm guessing I've done something wrong
I'm new to git (came from hg), so what should I do
Plan B: remove the PR and fork, do everything from scratch (it not too much)

saitonakamura:
I created another PR #829 to clean up this mess

PR #821, opened 2016-11-10
chore(build): simplify and fix hmr

levithomason:
I'm beginning to address some of the build issues.  This PR:

- **Fixes hot module reloading**, the doc site actually reloads properly in dev now 🎉 
- trims the install footprint by removing unnecessary react and babel plugins
- rewrites the gulpfile and tasks to plain node, no babel-register precompile

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,27 +8,6 @@
     "lodash"
   ],
   "env": {
-    "development": {
-      "plugins": [
-        ["react-transform", {
-          "transforms": [{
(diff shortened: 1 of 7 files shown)


levithomason:
@Semantic-Org/react-collaborators Aside from this PR, I also pushed some updates that affect how dependencies work.  I suggest you `rm -rf node_modules && npm i` to get a fresh install.  Not only are packages different, but the tree needs to be flattened for some of them.

PR #823, opened 2016-11-10
fix(Portal): ignore document click if no node or portal

### LINKED ISSUE #750: IE Modal error
I'm receiving the following error when I close the modal window through a button click event in IE.

"Unable to get property 'contains' of undefined or null reference"

I do not receive any error if i close the modal by clicking outside of it though.
I'm using the latest version of semantic-ui-react and my version of IE is Version 11.0.9600.18500

levithomason:
This PR makes a few fixes to the Portal, including one that should fix the IE Modal issue:

1. ignore document click if not mounted and open, fixes #750 
1. consolidates `if (this.state.open)` checks to the `renderPortal()` method, instead of checking every time `renderPortal()` is called.
1. remove `portal` event handlers before adding duplicates in `renderPortal()`, this should fix a memory leak

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -157,8 +157,8 @@ class Portal extends Component {
   handleDocumentClick = (e) => {
     const { closeOnDocumentClick, closeOnRootNodeClick } = this.props
 
-    // If event happened in the portal, ignore it
-    if (this.portal.contains(e.target)) return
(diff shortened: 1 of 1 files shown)


levithomason:
@jcarbo sanity check?

jeffcarbs:
LGTM 👍

levithomason:
Released in `semantic-ui-react@0.60.3`

PR #828, opened 2016-11-10
fix(Dropdown|Label): Remove overzealous prop validations

### LINKED ISSUE #827: bug(customPropTypes): demand validation is broken with defaultProps
Thanks to @meze for report in #809.

----

#809 introduced new propType validation for `Label`:

```js
removeIcon: customPropTypes.every([
  customPropTypes.demand(['onRemove']),
  customPropTypes.itemShorthand,
]),
```

And seems there is bug with `defaultProps` and `customPropTypes.demand`. Looks that validation for  `removeIcon` is called **always** even it's not passed to component.

```jsx
<Label>
  <Icon name='mail' /> 23
</Label>
```

```
Warning: Failed prop type: `removeIcon` prop in `Label` requires props: `onRemove`.
    in Label (created by LabelExampleBasic)
    in LabelExampleBasic
```

_This_ issue in Label can be solved with:

```diff
- static defaultProps = {
-  removeIcon: 'delete',
-}

+ {onRemove && Icon.create(removeIcon, { onClick: this.handleRemove })}
- {onRemove && Icon.create(removeIcon || 'delete', { onClick: this.handleRemove })}
``` 

However, I don't like this because it will not solve real problem. Any better proposals?

jeffcarbs:
Fixes #827

### DIFF
--- a/src/elements/Label/Label.js
+++ b/src/elements/Label/Label.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { Component, PropTypes } from 'react'
 
@@ -106,17 +107,8 @@ export default class Label extends Component {
     /** Adds an "x" icon, called with (event, props) when "x" is clicked. */
     onRemove: PropTypes.func,
 
-    /**
-     * Shorthand for Icon to appear as the last child and trigger onRemove.
-     *
-     * TODO: This should use customPropTypes.itemShorthand. However, it has a
-     * default value so it's always set, which causes the `disallow([children])`
-     * prop validation to fail.
-     */
-    removeIcon: PropTypes.oneOfType([
-      PropTypes.node,
-      PropTypes.object,
-    ]),
+    /** Shorthand for Icon to appear as the last child and trigger onRemove. */
+    removeIcon: customPropTypes.itemShorthand,
 
     /** A label can appear as a ribbon attaching itself to an element. */
     ribbon: PropTypes.oneOfType([
@@ -129,10 +121,6 @@ export default class Label extends Component {
 
     /** A label can appear as a tag. */
     tag: PropTypes.bool,
-  }
-
-  static defaultProps = {
-    removeIcon: 'delete',
   }
 
   static _meta = _meta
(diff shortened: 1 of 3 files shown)


jeffcarbs:
@levithomason - this fixes an annoying/incorrect prop validation warning in `Dropdown` and `Label` (see discussion in #827). A quick turnaround on this would be appreciated 😄 

Aside from the `demand(['onRemove'])` warning, having a default for `removeIcon` also caused a warning in `customPropTypes.itemShorthand` via `disallow([children])`. I added a TODO comment so we can possible address in the future and for now just manually specified `oneOf(object|node)` which we're doing in `Dropdown` and `Search`.

levithomason:
Hm, I read up on this but I think this fix should be to the default and not the validation.  My reasoning is that the `removeIcon` prop is not valid with children, so defaulting a value means we're forcing a conflict when children are used.

I would think then that the issue is that there should not be a default, since having one would break certain use cases.  Let me know your thoughts on this.

test/specs/elements/Label/Label-test.js
```diff
@@ -45,17 +45,6 @@ describe('Label', () => {
   })
 
   describe('removeIcon', () => {
-    it('has no icon without onRemove', () => {
-      shallow(<Label />)
-        .should.not.have.descendants('Icon')
-    })
-
-    it('has delete icon by default', () => {
```
levithomason, line 53: I believe this test would still apply since there is a default icon, it is just not applied in `defaultProps`, true?

jeffcarbs: > I would think then that the issue is that there should not be a default, since having one would break certain use cases. Let me know your thoughts on this.

As per this comment, I removed the default entirely so you'd have to pass the `removeIcon` in addition to `onRemove`. That does seem a little redundant though since the 99% use-case would be to have the `removeIcon` be `delete`. I'll add it back and add the test. I'll just use:

``` js
const removeIconShorthand = _.isUndefined(removeIcon) ? 'delete' : removeIcon
```

jeffcarbs:
Ok this should be g2g :)

levithomason:
Released in `semantic-ui-react@0.60.4`.

PR #832, opened 2016-11-10
fix(Accordion) fix wrong indexes when using Accordion.Title and Accordion.Content

### LINKED ISSUE #773: Accordion indexes
### Steps

```
<Accordion>
  <Accordion.Title>First title</Accordion.Title>
  <Accordion.Content>First content</Accordion.Content>
  <Accordion.Title>Second title</Accordion.Title>
  <Accordion.Content>Second content</Accordion.Content>
  <Accordion.Title>Third title</Accordion.Title>
  <Accordion.Content>Third content</Accordion.Content>
</Accordion>
```
### Expected Result

First title -> activeIndex = 0
Second title -> activeIndex = 1
Third title -> activeIndex = 2
### Actual Result

First title -> activeIndex = 0
Second title -> activeIndex = 2
Third title -> activeIndex = 4
### Version

0.58.0

saitonakamura:
RFC

fixes #773 

I fix wrong indexes but I find the code kinda ugly. It would be better if we don't need to copy index into local variable, but without it its buggy, i'm not sure why.

Another way: still use `child.index` but divide it by two like this
```
return Children.map(children, (child, i) => {
  const currentIndex = i / 2 
  // 0 / 2 = 0
  // 1 / 2 = 1
  // 2 / 2 = 1
  // 3 / 2 = 1
  // etc.
  ...
})
```
But its hard to read

### DIFF
--- a/src/modules/Accordion/Accordion.js
+++ b/src/modules/Accordion/Accordion.js
@@ -101,24 +101,28 @@ export default class Accordion extends Component {
   renderChildren = () => {
     const { children } = this.props
     const { activeIndex } = this.state
+    let titleIndex = 0
+    let contentIndex = 0
 
-    return Children.map(children, (child, i) => {
+    return Children.map(children, (child) => {
       const isTitle = child.type === AccordionTitle
       const isContent = child.type === AccordionContent
 
       if (isTitle) {
-        const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === i
+        const currentIndex = titleIndex
+        const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === currentIndex
         const onClick = (e) => {
-          this.handleTitleClick(e, i)
-          if (child.props.onClick) child.props.onClick(e, i)
+          this.handleTitleClick(e, currentIndex)
+          if (child.props.onClick) child.props.onClick(e, currentIndex)
         }
+        titleIndex++
         return cloneElement(child, { ...child.props, active: isActive, onClick })
       }
 
       if (isContent) {
-        // content must be the a sibling too title
-        // it is active if the active title index that precedes it is active
-        const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === i - 1
+        const currentIndex = contentIndex
+        const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === currentIndex
+        contentIndex++
         return cloneElement(child, { ...child.props, active: isActive })
       }
 
--- a/test/specs/modules/Accordion/Accordion-test.js
+++ b/test/specs/modules/Accordion/Accordion-test.js
@@ -77,6 +77,29 @@ describe('Accordion', () => {
       wrapper
         .should.have.state('activeIndex', -1)
     })
+    it('is called with (event, index) on AccordionTitle click', () => {
+      const spy = sandbox.spy()
+      const titles = mount(
+        <Accordion>
+          <Accordion.Title onClick={spy} />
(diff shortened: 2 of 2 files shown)


src/modules/Accordion/Accordion.js
```diff
@@ -101,24 +101,28 @@ export default class Accordion extends Component {
...
+    let contentIndex = 0
 
-    return Children.map(children, (child, i) => {
+    return Children.map(children, (child) => {
       const isTitle = child.type === AccordionTitle
       const isContent = child.type === AccordionContent
 
       if (isTitle) {
-        const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === i
+        const currentIndex = titleIndex
```
levithomason, line 112: Could we remove `currentIndex` from this block and the isContent block and just use `titleIndex` and `contentIndex` respectively?  I think it is a little more clear.

test/specs/modules/Accordion/Accordion-test.js
```diff
@@ -77,6 +77,29 @@ describe('Accordion', () => {
       wrapper
         .should.have.state('activeIndex', -1)
     })
+    it('is called with (event, index) on AccordionTitle click', () => {
```
levithomason: Thanks for this test. Could we also add one more test similar to this that asserts that the correct title/content are "active" given an `activeIndex` of 0, 1, and 2?  Something like this added to `describe('activeIndex')`, note I did not test this so it could have issues:

``` jsx
it('sets the correct pair of title/content active', () => {
  const wrapper = shallow(
    <Accordion>
      <Accordion.Title />
      <Accordion.Content />
      <Accordion.Title />
      <Accordion.Content />
      <Accordion.Title />
      <Accordion.Content />
    </Accordion>
  )
  wrapper.setProps({ activeIndex: 0 })
  wrapper.childAt(0).should.have.prop('active', true)
  wrapper.childAt(1).should.have.prop('active', true)

  wrapper.setProps({ activeIndex: 1 })
  wrapper.childAt(2).should.have.prop('active', true)
  wrapper.childAt(3).should.have.prop('active', true)

  wrapper.setProps({ activeIndex: 2 })
  wrapper.childAt(4).should.have.prop('active', true)
  wrapper.childAt(5).should.have.prop('active', true)
})
```

levithomason (changes requested):

src/modules/Accordion/Accordion.js
saitonakamura: I did that at first

```
if (isTitle) {
  const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === titleIndex
  const onClick = (e) => {
    this.handleTitleClick(e, titleIndex)
    if (child.props.onClick) child.props.onClick(e, titleIndex)
  }
  titleIndex++
  return cloneElement(child, { ...child.props, active: isActive, onClick })
}

if (isContent) {
  const isActive = _.has(child, 'props.active') ? child.props.active : activeIndex === contentIndex
  contentIndex++
  return cloneElement(child, { ...child.props, active: isActive })
}
```

But if you do this, than all titles will have index 3 in the callback. Like the lambda gets not only lazy evaluation, but lazy creation, or value is passed by reference.

test/specs/modules/Accordion/Accordion-test.js
saitonakamura: Of course, sure I can add this

levithomason (approved):

levithomason:
Released in `semantic-ui-react@0.60.7`.

PR #844, opened 2016-11-13
fix(Docs): fix find() usage for IE

layershifter:
This PR is feedback for #750 - modal is works correctly now 👍 

But, there is little problem with [`find()`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/find) usage, it's not supported in IE at all :weary So, I replaced it with lodash's `_find()`.


levithomason:
Thanks for the confirmation and fix 👍

PR #845, opened 2016-11-13
fix(debug): add bug work around

levithomason:
Debug localStorage flags are borked.  I've opened a PR to fix them: https://github.com/visionmedia/debug/issues/330

This PR adds a workaround for now.


levithomason:
Released in `semantic-ui-react@0.60.5`

PR #854, opened 2016-11-14
feat(Dropdown): Enable custom DropdownItem render

### LINKED ISSUE #852: Dropdown: Customize selected label for single-select
### Background
For the `Dropdown` component in SUI-core, dropdown options can have both `text` and `name` properties to distinguish b/w how the option should look within the dropdown menu and how it should look as the selected item (see a more in depth description here: https://github.com/Semantic-Org/Semantic-UI-React/pull/682#discussion_r84160476)

In my app I'm using a node for the `text` prop for each dropdown option to show more complex markup. Here's a rough example: http://codepen.io/jeffcarbs/pen/VmjmYo?editors=0010:
screen shot 2016-11-14 at 1 05 36 pm

For multi-select, you can now specify the `renderLabel` to customize how the selected items should be rendered.

There's no such equivalent to be able to customize how/what the single-select should render the selected item. In that same example, when you select something it renders the same complex node: 
screen shot 2016-11-14 at 1 32 51 pm

I'd like it to show something else (e.g. just the simple text)

### Question

Should/could we:
- Enable specifying different text/name options (names of those props could change)?
- Use the `renderLabel` for both single/multiple? The third argument to that function is the dropdown props so one could dynamically do something different for single/multiple.
- Add a different `render*` function for that?

Thoughts?

jeffcarbs:
Fixes #852

See the linked issue and also this comment for relevant discussion: https://github.com/Semantic-Org/Semantic-UI-React/pull/682#discussion_r84160476

In the SUI-core dropdown, you can pass both the `text` and `name` props for each option. 
- `text` - the text that gets rendered within the closed `Dropdown` when that option is selected
- `name` - the text that gets rendered for a given option within the dropdown menu.

If no `name` is provided, `text` is used for both. See the second example ("Trending Repos") for an example of the distinction: http://semantic-ui.com/modules/dropdown.html#inline. Notice how within the dropdown the options are title-cased whereas when you select an option it is rendered lowercased:
screen shot 2016-11-14 at 2 21 20 pm

This PR implements what essentially is the `name` prop, although I named it `content` to fit into the pattern established throughout the rest of the repo. It's shorthand that takes the place of `children` and is rendered alongside the other shorthand props (icon, flag, etc) if given.

This isn't a breaking change since the `text` prop is used if the `content` prop is missing.

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleItemContent.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleItemContent.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import { Dropdown, Header } from 'semantic-ui-react'
+
+const options = [
+  { text: 'Mobile',
+    value: 1,
+    content: <Header icon='mobile' content='Mobile' subheader='The smallest size' /> },
+  { text: 'Tablet',
+    value: 2,
+    content: <Header icon='tablet' content='Tablet' subheader='The size in the middle' /> },
+  { text: 'Desktop',
+    value: 3,
+    content: <Header icon='desktop' content='Desktop' subheader='The largest size' /> },
+]
+
+const DropdownExampleItemContent = () => (
+  <Dropdown
+    selection
+    fluid
+    options={options}
+    placeholder='Choose an option'
+  />
+)
+
+export default DropdownExampleItemContent
--- a/docs/app/Examples/modules/Dropdown/Usage/index.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/index.js
@@ -44,6 +44,11 @@ const DropdownUsageExamples = () => (
       description='A "multiple" dropdown can render customized label for selected items.'
       examplePath='modules/Dropdown/Usage/DropdownExampleMultipleCustomLabel'
     />
+    <ComponentExample
+      title='Item Content'
+      description='A dropdown item can be rendered differently inside the menu.'
+      examplePath='modules/Dropdown/Usage/DropdownExampleItemContent'
+    />
 
   </ExampleSection>
 )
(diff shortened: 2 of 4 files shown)


levithomason:
Cool, this looks great.  The only thing I think we should change is the prop name `content`.  The DropdownItem `text` prop will become `content` before we hit v1 for consistency with all other components, it is just left over from before we standardized.

Looking at SUI core, they use `data-text=""` for the value rendered in the Dropdown when active.  I think this makes perfect sense given that the Dropdown prop for the same value space is `text`.  It then stands to reason that the Dropdown `text` is set by the DropdownItem `text`.  Whereas, `content` is always the primary content of the component, in this case, the content of the DropdownItem (which only appears in the Menu).

I believe this means we ought to reverse the two props from what they are now in this PR, correct?

levithomason:
One alternative would be to add `renderText` to the Dropdown and pass it the currently active DropdownItem props.  This deviates from SUI core a bit, but extends the `renderLabel` idea.  Depending on the majority use case, one of these (`renderText` vs DropdownItem `text` prop) will be better for users I would assume.

jeffcarbs:
> It then stands to reason that the Dropdown text is set by the DropdownItem text. Whereas, content is always the primary content of the component, in this case, the content of the DropdownItem (which only appears in the Menu).

I think that's right and aligns with naming that's currently in the PR. Here's an example of SUI core dropdown: http://codepen.io/jeffcarbs/pen/dOXvmQ

```
<div class="item" data-value="0" data-text="Option 1 text">Option 1 content</div>
```
- `Option 1 text` === `text prop` === value rendered in the Dropdown when selected
- `Option 1 content` === `content prop` === content of the DropdownItem.

I'd advocate for the `content` prop over `renderText` since it aligns with the pattern we've established throughout the app of `content` being shorthand for primary content of a component and it more closely maps to SUI-core. If there's interest in a `renderText` prop in the future I think we can consider adding it.

levithomason:
Thanks for the clarification, I was reading the `DropdownExampleItemContent` backward, thinking it was showing the `Header` as the rendered text (which is in `content`).

levithomason:
Released in `semantic-ui-react@0.60.9`.

PR #860, opened 2016-11-15
perf(getUnhandledProps): replace lodash with vanilla js

### LINKED ISSUE #859: Very low performance caused by `getUnhandledProps`
I have Form and Table inside of Container. And Table would be updated asynchronously.

I found that the UI is laggy when updating state, (updating table data), and eventually the browser may crash if I update table twice and more.

According to the CPU profiles, it seems that `getUnhandledProps` consumes too much time. (it calls baseClone which is very time-consuming)

selection_458

How I can avoid of `getUnhandledProps`?

### Version
0.59.0

```javascript
import React from 'react'
import { Container, Form, Table, Icon, Button, Modal, Segment, Dimmer, Loader } from 'semantic-ui-react'

export default class FaildedMessage extends React.Component {
  constructor(props) {
    super(props)
    this.state = {
      messages: [],
      searching: false,
    }
  }

  search() {
    this.setState({ searching: true})
    setTimeout(() => {
      this.setState({
        searching: false,
        messages: [1,2,3,4,5,6,7,8,9,10]
      })
    }, 1000)
  }

  render() {
    const { messages, searching } = this.state

    return (
      <Container>
        <Form size="small">
          <Form.Group>
            <Form.Button
              basic
              primary
              loading={searching}
              onClick={e => {
                e.preventDefault()
                this.search()
                // sorry, the statement below is pasted here by mistake. please remove it.
                //this.setState({ searching: !searching })
              }}
              style={{
                position: 'absolute',
                right: 0,
                bottom: 0,
              }}
            >查询</Form.Button>
          </Form.Group>
        </Form>

        <Segment basic size="small" style={{ marginTop: '2rem', padding: 0 }}>
          <Dimmer inverted active={searching}>
          <Loader inverted />
          </Dimmer>
          <Table basic>
          <Table.Header>
            <Table.Row>
              <Table.HeaderCell>ID</Table.HeaderCell>
              <Table.HeaderCell></Table.HeaderCell>
            </Table.Row>
          </Table.Header>
          <Table.Body>
          {
            messages.map(m => (
              <Table.Row key={m}>
                <Table.Cell>{m}</Table.Cell>
                <Table.Cell textAlign="right">
                  <Modal dimmer="blurring" size="small" trigger={<Button basic primary icon="info" circular size="mini" />}>
                    <Modal.Header>alert</Modal.Header>
                    <Modal.Content>hehe</Modal.Content>
                    <Modal.Actions>
                      <Button.Group>
                        <Button basic color="red">delete</Button>
                        <Button basic color="red">cancel</Button>
                      </Button.Group>
                    </Modal.Actions>
                  </Modal>
                </Table.Cell>
              </Table.Row>
            ))
          }
          </Table.Body>
          <Table.Footer>
          <Table.Row>
          <Table.HeaderCell colSpan="8">
            </Table.HeaderCell>
            </Table.Row>
          </Table.Footer>
          </Table>
        </Segment>
      </Container>
    )
  }
}

```

levithomason:
Fixes #859 

**TL;DR** Conclusion at the bottom.

As with the linked issue, I've also noticed some severe performance issues, mostly in Grid and Grid column.  It appears there is a memory leak somewhere in getUnhandledProps due to lodash.

Calculating unhandled props can be expensive due to the several nested object references it has to traverse to first calculate the _handled_ props: `autoControlledProps`, `defaultProps` keys, `props` keys.  It then needs to loop the props again to omit all those keys.  This can be done several times per render, per element.  This alone should not be responsible for the results we're seeing below, however.

This PR replaces lodash with vanilla JS, **resulting in >12,000 times performance improvement** .

It also caches _handledProps_ so they are only calculated once per component (handled props never change).  This will be fixed permanently once #731 is merged as we'll have a static array of handled props baked into every component.

# Rough Perf Analysis

## No cache, lodash (baseline)

After lots of testing, I notice that the Grid and Grid column render times seem increase every render for the particular view I am testing against.  Note the instance count and render count remains constant, but the time required is growing rapidly.

### First Render - `1s`

image

### Second Render - `3s`

image

### Third Render - `27s`

image

### First Render - `1.5s`

image

### Second Render - `21s`

image

### Third Render - `1.3s`

image

### Additional tests...

I found the `21s` odd here, so I ran renders 4 and 5 with results of `3.6s` and `2.8s`.  This seems odd.

## No cache, vanilla JS

**All three runs completed in <=2ms (approx.)** 😮.  This certainly means, lodash is at fault here.

image

## Cache and vanilla JS

Using vanilla JS is obviously sufficient, however, I tested vanilla with cache for completeness.  The results were not significantly different from vanilla JS without cache.

# Conclusion

Something in lodash is _extremely_ non-performant given the usage in our `getUnhandledProps()` lib method.  Using vanilla JS increases performance by up to >12,000 times.  Caching makes no difference and is not necessary.

### DIFF
--- a/src/lib/getUnhandledProps.js
+++ b/src/lib/getUnhandledProps.js
@@ -1,4 +1,16 @@
-import _ from 'lodash'
+const handledPropsCache = {}
+
+/**
+ * Push all `source` array elements to the `target` array if they don't already exist in `target`.
+ *
+ * @param {Array} source - An array of elements to add to the `target`
+ * @param {Array} target - An array to receive unique elements from the `source`
+ * @returns {Array} Mutated `target` array
+ */
+const pushUnique = (source, target) => source.forEach(x => {
+  if (target.indexOf(x) === -1) target.push(x)
+})
+
 /**
  * Returns an object consisting of props beyond the scope of the Component.
  * Useful for getting and spreading unknown props from the user.
@@ -7,13 +19,27 @@ import _ from 'lodash'
  * @returns {{}} A shallow copy of the prop object
  */
 const getUnhandledProps = (Component, props) => {
-  const handledProps = _.union(
-    Component.autoControlledProps,
-    _.keys(Component.defaultProps),
-    _.keys(Component.propTypes),
-  )
+  const { _meta: { name }, autoControlledProps, defaultProps, propTypes } = Component
 
-  return _.omit(props, handledProps)
+  let handledProps = handledPropsCache[name]
+
+  // ----------------------------------------
+  // Calculate handledProps once and cache
+  // ----------------------------------------
+  if (!handledProps) {
+    handledProps = []
+
+    if (autoControlledProps) pushUnique(autoControlledProps, handledProps)
+    if (defaultProps) pushUnique(Object.keys(defaultProps), handledProps)
+    if (propTypes) pushUnique(Object.keys(propTypes), handledProps)
+
+    handledPropsCache[name] = handledProps
+  }
+
+  return Object.keys(props).reduce((acc, prop) => {
+    if (handledProps.indexOf(prop) === -1) acc[prop] = props[prop]
+    return acc
+  }, {})
 }
 
 export default getUnhandledProps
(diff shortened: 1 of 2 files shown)


levithomason:
Released in `semantic-ui-react@0.60.10`.

markerikson:
Hah.  I'm playing around with Semantic-UI-React on a sample project for the first time, and _just_ ran into this behavior myself (in my case, with a tabbed menu acting as parent to some panels with a couple simple forms).  After adding a table and a small form,the time to switch tabs became extremely noticeable.  A perf profile showed the issue you found here, and fortunately I ran a quick search before filing another issue :)

PR #877, opened 2016-11-16
add meteor url for usage

Maxhodges:
merge this 877 before https://github.com/Semantic-Org/Semantic-UI-React/pull/878

PR #878, opened 2016-11-16
add meteor segment

Maxhodges:
merge this one first 
https://github.com/Semantic-Org/Semantic-UI-React/pull/877

### DIFF
--- a/docs/app/Views/Usage.js
+++ b/docs/app/Views/Usage.js
@@ -111,6 +111,19 @@ const Usage = () => (
           import '.../semantic/dist/semantic.min.css';
         </pre>
       </Segment>
+
+      {/* ----------------------------------------
+       *  Semantic-UI-Meteor package
+       * -------------------------------------- */}
+      <Header as='h3'>Semantic UI Meteor package</Header>
+      <p>
+        The <a href={semanticUIMeteorRepoURL}>Semantic UI Meteor package</a> is the official way to install Semantic-UI for Meteor applications.
+      </p>
+      <p>
+        Semantic UI Meteor can be installed as a package in your project using NPM. See the repo for detailed instructions.
+      </p>
+     
+
     </Segment>
   </Container>
 )


levithomason:
The variable `semanticUIMeteorRepoURL` is `undefined` in that scope, let's pull these two PRs into one.

levithomason:
Looking at this note, it looks like it is not related to CSS, but to installing SUI for Meteor in general.  I'm not sure this note makes sense in this repo since the Meteor install depends on jQuery.  This is not compatible with React.  I don't think it should be added to the React docs.

I'm not a Meteor user, so please correct me if I'm mistaken here.

Maxhodges:
​The meteor install does allow you to use Semantic-UI without React (using
Meteor's Blaze templating system)​, OR you can use Semantic-UI-React
instead. The benefit of this package is that it allows Meteor to build the
LESS and gives you control over which components to load, and from what
theme to load them (in our case we've created our own theme and override
many of the SUI default theme's component styles with our own. This Meteor
packages makes that easy to do, whereas simply linking to a minified CSS
file wouldn't provide the same latitude at all.

I'll discuss with my CTO to be sure and get back to you.

levithomason:
OK, I believe this is best left to the Meteor repo.  IMHO the React repo should only include information regarding integrating with React. Meteor users would then use the Meteor repo for information on Meteor usage.

For information on creating custom builds and themes for non-meteor users, I wouldn't want to propose Meteor as the path to doing this since there is already a SUI theming build.

Per the last half of [my comment here](https://github.com/Semantic-Org/Semantic-UI-React/issues/908#issuecomment-261815386), I'd like to add something similar to this repo for customizing CSS builds with your React components.  I'm hesitant to further fracture and confuse the path to theming, so I'd like to get with Jack on this as well before embarking.  It seems there is a broader need here, a simple way to customize and ship themes with _any_ integration that does not dictate the JS implementation.

I have several ideas for this and will reach out to Jack regarding them once I get a proposal together.

PR #886, opened 2016-11-17
fix(Input): pass all handlers to the html input

### LINKED ISSUE #884: Support onKeyDown in Input
To be able to allow/disallow different keys in input fields it would be nice if the `Input` component would have an `onKeyDown` prop where I could provide my custom event handler. If I call `e.preventDefault()` in the `onKeyDown` event handler the `onChange` handler would not be called.

This might be as easy as to add the `onKeyDown` into `htmlInputPropNames` array?

### Version
0.60.9

levithomason:
Fixes #884

We currently pass `onChange` to the HTML input element but others are spread on the root node as unhandled props.  This caused inconsistent behavior in the event object since the `onChange` receives the event before the root node.

This PR passes all event handlers to the HTML input, resolving any issues with event manipulation, such as `e.preventDefault()`.

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -17,29 +17,37 @@ import Icon from '../../elements/Icon'
 import Label from '../../elements/Label'
 
 export const htmlInputPropNames = [
-  // React
-  'selected',
-  'defaultValue',
-  'defaultChecked',
+  // REACT
+  'selected', 'defaultValue', 'defaultChecked',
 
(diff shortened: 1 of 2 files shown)


levithomason:
Released in `semantic-ui-react@0.61.1`.

lextoc:
Doesn't seem to be present in `"version": "0.86.0"`, is it removed? Is there a way to force a regex on user input via event listeners?

Issue #885, opened 2016-11-17
Search Module can't be made a11y compliant


fracmak:
Search autocompletes modules are made accessible using aria role="combobox" which requires a number of aria attributes and ids that are not possible to apply to the search component because the underlying input/searchResults can't be given any extra props.

According to https://www.w3.org/TR/wai-aria/roles#combobox an autocomplete needs the following aria attributes on an input

* role="combobox"
* aria-expanded="true" (true if results are showing, false otherwise)
* aria-autocomplete="inline"
* aria-owns="owned_listbox" (id to the search results)
* aria-activedescendant="selected_option" (id for the selected result, either this needs to be updated, or the id of the results needs to dynamically switch)

### Version
0.60.4

levithomason:
This is important, and also a massive undertaking.  We should probably consider using something like [react-a11y](https://github.com/reactjs/react-a11y) or [eslint-plugin-jsx-a11y](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) to assist in identifying issues with components.  Maybe start with warnings, and try to address them as we can moving forward.

Is this something you're interested in heading up?

fracmak:
I'm happy to lend a hand with it. I've been using both of those plugins, though neither of them are very thorough and seems like react-a11y has gone dormant. I've been running my components through [HTML_CodeSniffer](https://github.com/squizlabs/HTML_CodeSniffer), but I'm still hammering out the best way to integrate it into automated testing (their npm package isn't compatible with webpack). 

I see the main deficiencies with a11y support are two fold, we're missing aria tags across all components, and in some places we're using simple div elements when more specific elements are necessary, but there's also issues there too, it seems semantic-ui's css file doesn't reset ul's properly (leaves padding on them) which makes it hard to use them properly for things like menus. 

Enough rambling, how would you like me to approach what needs to be done? Is a good starting place hooking up the eslint and react libraries and see what it reports then start compiling a list of the deficiencies of the other components that we can create issues for like we do the missing components from Semantic main?

levithomason:
I think we should first try to get this automated as part of the current test workflow.  I don't have the bandwidth at present, however, scanning their README it seems we should be able to cook something up using their [PhantomJS runner](https://github.com/squizlabs/HTML_CodeSniffer#phantomjs) or [node & jsdom](https://github.com/squizlabs/HTML_CodeSniffer#node--jsdom) solution.

[...]

PR #890, opened 2016-11-17
test(lint): Added eslint-plugin-jsx-a11y plugin in warning mode

fracmak:
Based on feedback to #885 this adds the eslint-plugin-jsx-a11y in warning mode to showcase where it thinks we're failing. 

Currently have 93 warnings.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,9 @@
 {
+  "plugins": [
+    "jsx-a11y"
+  ],
   "extends": [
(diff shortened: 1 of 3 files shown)


levithomason:
There are a number of warnings being reported in the `/test` directory.  I think we should exclude these (they are non-user facing).  Could you please update `/test/.eslintrc` to turn off a11y rules?

fracmak:
Done!

levithomason:
Thanks much :D

PR #892, opened 2016-11-18
feat(Rating): Rating now keyboard navigable and visually impaired accessible

fracmak:
Related to #885 

Accessibility issues addressed:
* Module is now keyboard accessible by adding tabIndex and keyPress handlers for space and enter activation
* added role=radio/radiogroup, aria-checked, aria-posinset, aria-setsize to properly indicate to visually impaired users the kind of module and what values they're selecting

W3C Documentation:
https://www.w3.org/TR/wai-aria-1.1/#radiogroup
https://www.w3.org/TR/wai-aria-1.1/#radio

### DIFF
--- a/src/modules/Rating/RatingIcon.js
+++ b/src/modules/Rating/RatingIcon.js
@@ -41,7 +41,7 @@ export default class RatingIcon extends Component {
     if (onClick) onClick(e, index)
   }
 
-  handleKeyPress = (e) => {
+  handleKeyUp = (e) => {
     const { onClick, index } = this.props
 
     if (onClick) {
@@ -76,7 +76,7 @@ export default class RatingIcon extends Component {
       <i role='radio' tabIndex={0}
         {...rest}
         className={classes}
-        onKeyPress={this.handleKeyPress}
+        onKeyUp={this.handleKeyUp}
         onClick={this.handleClick}
         onMouseEnter={this.handleMouseEnter}
       />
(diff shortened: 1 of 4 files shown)


src/modules/Rating/RatingIcon.js
```diff
@@ -52,7 +70,16 @@ export default class RatingIcon extends Component {
...
       'icon'
     )
+    const rest = _.omit(this.props, _.keys(RatingIcon.propTypes))
 
-    return <i className={classes} onClick={this.handleClick} onMouseEnter={this.handleMouseEnter} />
+    return (
+      <i role='radio' tabIndex={0}
+        {...rest}
+        className={classes}
+        onKeyPress={this.handleKeyPress}
```
levithomason: Should this be `onKeyUp`?  Key press is fired on key down and repeatedly fires when held.  Whereas, the `keyup` event only fires once on release.  I believe `onKeyUp` is the correct event handler for adding button like attributes to non-button elements, since it mirrors the mouse click behavior, only invoked once and invoked on release.

levithomason (commented):
This looks great, just one question for clarification 👍

src/modules/Rating/RatingIcon.js
fracmak: Ah, yes, I debated this a bunch originally. I was under the impression key press only fired after a key down followed by a key up which is why i chose it. But I could be wrong, which would make key up a good alternative. Unfortunately because of the holiday I'm not certain when I'll be able to make the update. Feel free to change it if you have time, otherwise I'll pick it back up next week at the latest

fracmak:
keypress changed to keyup, should be good to go now

levithomason (approved):

levithomason:
Thanks much, I'll release this today.

PR #893, opened 2016-11-18
feat(Portal|Modal|Confirm): Better uncontrolled usage support

### LINKED ISSUE #891: Modal: Better uncontrolled usage support
The SUI-core Modal supports various options for closing. See the "DOM Settings" section here: http://semantic-ui.com/modules/modal.html#settings

The following actions are called onClick of something matching the associated selector:
```js
  close    : '.close, .actions .button',
  approve  : '.actions .positive, .actions .approve, .actions .ok',
  deny     : '.actions .negative, .actions .deny, .actions .cancel'
```

- `close` just closes the modal
- `approve` calls `onApprove` and closes the modal unless the return value from `onApprove` is `false`
- `deny` calls `onDeny` and closes the modal unless the return value from `onDeny` is `false`.

This could be possible by attaching a click listener to the modal content and checking the `currentTarget` for one of these classes or perhaps a `data-(close|approve|deny)` attribute. This doesn't really feel very React-y, so open to other suggestions but I really think we should support something like this. Currently there's no way to:
- use the `Confirm` component without managing `open` state yourself.
- implement an `X` to close a `Modal` without wiring it to some state you're managing outside.

IMHO the most common use-cases for all of our components should be handled out of the box without the need to wrap things yourself.

One issue I see is that the modal technically doesn't manage `open` state, the `Portal` does. So we'd either have to:
- implement this approve/deny functionality within `Portal` which feels kinda wrong
- turn `Modal` into an `AutoControlled` component and have it manage it's own `open` state and pass it to `Portal`. From a user perspective, I don't think anything would change, we'd just have some duplicative functionality.

We could also forego the approve/deny functionality and add a `closeOnCloseClick` prop to `Portal` which would cause it to close if something with the `data-close` attribute was clicked within the `Portal`. I actually think that would be pretty reasonable.

Thoughts?

jeffcarbs:
Fixes #891

A few updates that I think will have a big impact in terms of usability of these components:
- Adds a `closeOnCloseClick` to `Portal` which makes it so that if an element with the `data-close` attribute within the portal is clicked, it will close the portal.
- Update the `Confirm` component so clicking confirm/cancel will close the modal (makes use of `closeOnCloseClick`). This enables uncontrolled usage and I've updated all of the `Confirm` examples to make use of that where possible.
- Adds a `closeIcon` prop to `Modal` which renders an icon in the modal which, onClick, will trigger the modal to close (makes use of `closeOnCloseClick`). NOTE: With the default SUI styling, the icon appears outside of the modal. This is not an error on our side - the placement of the icon matches the [SUI-core example](http://semantic-ui.com/modules/modal.html#/usage) and there are SUI CSS rules for `.ui.modal > .close` which intentionally put it out there. This seemed weird to me so I just wanted to call attention to it not being a bug:
screen shot 2016-11-18 at 12 32 14 am

### DIFF
--- a/src/addons/Confirm/Confirm.js
+++ b/src/addons/Confirm/Confirm.js
@@ -86,7 +86,7 @@ class Confirm extends Component {
     const rest = getUnhandledProps(Confirm, this.props)
 
     return (
-      <Modal open={open} size='small' onOpen={this.handleOpen} onClose={this.handleClose} closeOnCloseClick {...rest}>
+      <Modal open={open} size='small' onOpen={this.handleOpen} onClose={this.handleClose} {...rest}>
         {header && <Modal.Header>{header}</Modal.Header>}
         {content && <Modal.Content>{content}</Modal.Content>}
         <Modal.Actions>
--- a/src/addons/Confirm/Confirm.js
+++ b/src/addons/Confirm/Confirm.js
@@ -1,68 +1,101 @@
-import _ from 'lodash'
-import React, { PropTypes } from 'react'
+import React, { Component, PropTypes } from 'react'
 
 import { getUnhandledProps, META } from '../../lib'
 import Button from '../../elements/Button'
 import Modal from '../../modules/Modal'
 
-/**
- * A Confirm modal gives the user a choice to confirm or cancel an action
- * @see Modal
- */
-function Confirm(props) {
-  const { open, cancelButton, confirmButton, header, content, onConfirm, onCancel } = props
-  const rest = getUnhandledProps(Confirm, props)
-
-  // `open` is auto controlled by the Modal
-  // It cannot be present (even undefined) with `defaultOpen`
-  // only apply it if the user provided an open prop
-  const openProp = {}
-  if (_.has(props, 'open')) openProp.open = open
-
-  return (
-    <Modal {...openProp} size='small' onClose={onCancel} {...rest}>
-      {header && <Modal.Header>{header}</Modal.Header>}
-      {content && <Modal.Content>{content}</Modal.Content>}
-      <Modal.Actions>
-        <Button onClick={onCancel}>{cancelButton}</Button>
-        <Button primary onClick={onConfirm}>{confirmButton}</Button>
-      </Modal.Actions>
-    </Modal>
-  )
-}
-
-Confirm._meta = {
+const _meta = {
   name: 'Confirm',
   type: META.TYPES.ADDON,
 }
 
-Confirm.propTypes = {
-  /** Whether or not the modal is visible */
-  open: PropTypes.bool,
+/**
+ * A Confirm modal gives the user a choice to confirm or cancel an action
+ * @see Modal
+ */
+class Confirm extends Component {
+  static propTypes = {
+    /** Whether or not the modal is visible */
+    open: PropTypes.bool,
 
-  /** The cancel button text */
-  cancelButton: PropTypes.string,
+    /** The cancel button text */
+    cancelButton: PropTypes.string,
 
-  /** The OK button text */
-  confirmButton: PropTypes.string,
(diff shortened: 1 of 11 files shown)


src/addons/Confirm/Confirm.js
```diff
@@ -1,68 +1,101 @@
...
+
+    const { onOpen } = this.props
+    if (onOpen) onOpen(e, this.props)
+  }
+
+  handleClose = (e) => {
+    const { onCancel, onClose } = this.props
+    if (onClose) onClose(e, this.props)
+
+    // Only call onCancel if this was closed was some other way besides an
```
levithomason, line 92: typo `was closed was some`

src/addons/Portal/Portal.js
```diff
@@ -182,6 +200,15 @@ class Portal extends Component {
   // Component Event Handlers
   // ----------------------------------------
 
+  handlePortalClick = (e) => {
+    if (!this.props.closeOnCloseClick) return
+    if (!domUtils.closest(e.target, '[data-close]')) return
```
levithomason, line 205: I'm not sure of an alternative proposal yet, but I don't think we should be leaving the virtual DOM to accomplish this.  I see that this allows the user to define any element to be the "close" element and that this is important.

I'll give this some more thought and see what other proposals I can come up with.  I think the goal should be that we do not read from the real DOM though.

jeffcarbs: I think the main alternative to address the issues raised in the related issue are to make Modal an auto controlled component that handles "open" and have it implement
- closeIcon, and handle closing on click
- actions, which would be an array of buttons, with an "action" prop or something to determine if they are approve/deny/close buttons and handle them appropriately

What I like about the approach in this PR is:
- it's lightweight
- it stays consistent with to SUI-core (although they use classes, which are more brittle than data attributes imo)
- works regardless of how the user defines the underlying markup (children/shorthand)
- is usable outside of Modal, in any of our components that use portal (e.g. Adding a close button to a pop up) or in any user-defined components that use Portal directly

jeffcarbs:
@levithomason - ping 😉

levithomason:
I'm hesitant on this one, I really don't think we should be reaching into the DOM for this.  I've considered a few solutions, but I believe it is best to leave this to a controlled pattern until we solve it using the virtual DOM API.

levithomason:
@davezuko do you have any wise ideas on parent-child communication in this regard?  The only other thing I can think of is providing some Modal method that would return a wrapped "close" component.  It would intercept the click and close the Modal, but this requires an obtuse API like:

```jsx
const { MyModal, Close } = Modal.createWithClose()

<MyModal>
  <Close />
</MyModal>
```

Where the parent-child relationship can be baked in when the create factory is invoked.

We have a very similar problem for Sidebars which have a related Pusher component.  The pattern is this need to have two components that can be placed anywhere in the tree but have a communication relationship established before they are rendered.

levithomason:
Given we've merged the Modal closeIcon in #932 it would be good so see what this PR looks like after having those changes implemented here.  This PR still has really great improvements outside of that.

jeffcarbs:
Just rebased this. The updates to `Confirm` are making use of the `data-close` attribute. I think to keep the same functionality without we'll have to introduce shorthand for the Modal. I can take a stab at that, probably on a separate PR though. @levithomason - thoughts?  If so, we can mark this PR as `blocked`

levithomason:
Blocked makes sense to me.  Modal shorthand would be awesome :D

levithomason:
We now support a close icon and there is a shorthand PR underway #1508.  Should be unblocked here soon, finally!

levithomason:
Modal shorthand is also now completed.  Unblocking this one.

levithomason:
Closing for #1599

PR #774, opened 2016-10-30
feat(Sidebar): add component

UnbrandedTech:
Submitted passing test and structure for Sidebar

PR #905, opened 2016-11-20
feat(Sidebar): add component

### LINKED ISSUE #197: <Sidebar /> Component
http://semantic-ui.com/modules/sidebar.html

fracmak:
Fixes #197, Replaces PR #774.

Adds component with examples

### DIFF
--- a/src/collections/Sidebar/Sidebar.js
+++ b/src/collections/Sidebar/Sidebar.js
@@ -9,6 +9,8 @@ import {
   getElementType,
   useKeyOnly,
 } from '../../lib'
+import SidebarPushable from './SidebarPushable'
+import SidebarPusher from './SidebarPusher'
 
 const _meta = {
   name: 'Sidebar',
@@ -51,6 +53,10 @@ class Sidebar extends Component {
     /** Sidebar width */
     width: PropTypes.oneOf(Sidebar._meta.props.width),
   }
+
+  static Pushable = SidebarPushable
+
+  static Pusher = SidebarPusher
 
   static defaultProps = {
     direction: 'left',
--- a/src/collections/Sidebar/Sidebar.js
+++ b/src/collections/Sidebar/Sidebar.js
@@ -66,13 +66,30 @@ class Sidebar extends Component {
     'visible',
   ]
 
+  state = {}
+
+  startAnimating = (duration = 500) => {
+    clearTimeout(this.stopAnimatingTimer)
+
+    this.setState({ animating: true })
+
+    this.stopAnimatingTimer = setTimeout(() => this.setState({ animating: false }), duration)
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.visible !== this.props.visible) {
+      this.startAnimating()
+    }
+  }
+
   render() {
+    const { animating } = this.state
     const { animation, className, children, direction, visible, width } = this.props
 
     const classes = cx(
       'ui',
       'sidebar',
-      'animating', // should be dynamically set only when animating
+      useKeyOnly(animating, 'animating'),
       direction,
       width,
       animation,
--- a/src/collections/Sidebar/Sidebar.js
+++ b/src/collections/Sidebar/Sidebar.js
@@ -1,0 +1,84 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import {
+  AutoControlledComponent as Component,
+  META,
+  customPropTypes,
+  getUnhandledProps,
+  getElementType,
+  useKeyOnly,
+} from '../../lib'
+
+const _meta = {
+  name: 'Sidebar',
+  type: META.TYPES.MODULE,
+  props: {
+    animation: ['overlay', 'push', 'scale down', 'uncover', 'slide out', 'slide along'],
+    direction: ['top', 'right', 'bottom', 'left'],
+    width: ['very thin', 'thin', 'wide', 'very wide'],
+  },
+}
+
+/**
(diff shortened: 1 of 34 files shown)


levithomason:
Wow, thanks!  I'll get a review on this as soon as I can 😸

kevflynn:
So excited for this one 🐶

kevflynn:
@levithomason - I know you're super-busy - any chance you can review this week? 😁

levithomason (changes requested):
We need to figure out an API for the pushable/pusher.  Given that these can appear anywhere in the tree, my first thought is a component/subcomponent.  I believe it is only applicable to the Sidebar, so perhaps Sidebar subcomponents would make sense?

```jsx
<Sidebar.Pushable>
  ...
  <Sidebar />
  <Sidebar.Pusher />
<Sidebar.Pushable />
```

This also follows our [`<Dimmer.Dimmable />`](react.semantic-ui.com/modules/dimmer#dimmer) pattern and allows us to add props and features to each individually.

### Pushable `body`

There is also the need to make the `body` element `pushable` for full page Sidebars.  We could again borrow from the `Dimmer.Dimmable` component and provide a `page` prop which, like a `page` Dimmer affects the `body`, it would make the entire page pushable:

```jsx
<Sidebar.Pushable page />
```
This might even make more sense as a Sidebar prop, since there is no need to place the "pushable" element in the tree and no need to support "pushable" children since it is only adding/removing the className to the `body`:
```jsx
<Sidebar page />
```

This would also need to render the Sidebar to the body using the Portal.  It also would need to render a `pusher` to the body.

Inspecting SUI core docs for full page Sidebars, this might be problematic.  The pusher needs to wrap the React app mount node in order to work.  I'm not sure on how to go about this since modifying React's render tree will break it.  I'll leave it at this for now.  Here is the markup required by SUI, per the core docs:

```html
<body class="pushable">
  <div class="sidebar"></div>
  <div class="pusher">
      <div class="root">my react app</div>
  </div>
</body>
```

Given the tricky issues here, I'm OK adding page level pushers on a separate PR as I suspect it is going to be fairly involved.

### Pusher

I think the `<Sidebar.Pusher />` API makes sense in and applies to all use cases here.  It seems it is always required. LMK if you disagree or see a better API.

#### Dimmed Pusher

The pusher can also be dimmed, http://semantic-ui.com/modules/sidebar#dimmed.  My first thought on this is `<Sidebar.Pusher dimmed />`.  Let's add this for now.

I think in an ideal workflow something like a `<Pusher dimmable />` would dim itself when the Sidebar was active.  I have a couple ideas for this, but again, let's consider auto dimming on a separate PR and this one will just add the simple prop `dimmed`.

levithomason:
I've left an initial review, this looks really great so far, thanks!

fracmak:
Ah, I like the suggestions, I'll get on making them. Though with the holidays it might have to wait until next week. Hopefully I'll get to it sooner. I'll comment when it's ready for another review

fracmak:
Added Sidebar.Pushable and Sidebar.Pusher with docs.

levithomason:
### Dynamic `animating` className

I've updated the Sidebar to use a dynamic animating className.  This was the reason all the inline `overflow` styles were required.  I didn't want users thinking that they had to add inline styles to get it to work, so I removed them all.

### Using Segments

I also updated Pushables in all  examples to augment the Segment and added a `basic` Segment child to the Pusher.  This was the reason all the inline `background` styles were required.  I remove those inline styles as well.  This will be handled automatically once #813 is merged.

I've based this markup here off of the examples here: http://semantic-ui.com/modules/sidebar.html#/examples.  The Pushable is a Segment while the Pusher first child is a `basic` Segment.  This gives the proper styling and alignment for non page level Sidebars.

levithomason:
Releasing this now, massive thanks @UnbrandedTech!

comewithmeifyouwanttolive:
Don't thank me, thank @fracmak! He did every thing, literally.

UnbrandedTech:
Don't thank me, thank @fracmak. He Literally did all the work!

levithomason:
Oh snap!  Getting my friends confused there :D Thanks @fracmak!

fracmak:
woo! excited to see this land!

vmptk:
Any good practice how to hide the sidebar when click on any place outside its area and on a menu item?

fracmak:
I would say using a click handler on the outer element that's controlling the visibility of the sidebar would be enough since click events bubble. You can then add a click handler to the sidebar itself to preventDefault() if the click wasn't on a link.

src/collections/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,107 @@
...
+  ]
+
+  state = {}
+
+  startAnimating = (duration = 500) => {
+    clearTimeout(this.stopAnimatingTimer)
+
+    this.setState({ animating: true })
+
+    this.stopAnimatingTimer = setTimeout(() => this.setState({ animating: false }), duration)
```
fracmak, line 76: I was thinking about it, any reason we aren't using the [transitionend event](https://developer.mozilla.org/en-US/docs/Web/Events/transitionend) to trap the animation finishing instead of hard coding 500ms?

levithomason: I actually did not know about this event, thanks for pointing it out.  The primary reason I added this hack was because we have a Transition component that will soon be merged and take over this task, #813.

Support for this event seems solid, I'd merge a PR updating this bit of code if you want to open one.

PR #906, opened 2016-11-20
feat(docs): show enum values

fracmak:
Simple feature to show the enum values of proptypes in a popup

screenshot 2016-11-20 01 19 07


levithomason:
This PR / doc approach will be outdated once #731 is merged since we will then inline all enums.  `Component._meta.props` will no longer contain prop enums.

However, this is a great stopgap for now.  Can we change the icon to something like:

```
{enum} (hover to view)
```

fracmak:
I updated it removing the icon in favor of the original text, also fixed an issue where polymorphic types weren't showing the tooltip. 

Fresh screenshot:

screenshot 2016-11-21 09 27 27

levithomason:
I wanted to show all enums when there were only a few options, which spiraled into some a number of other fixes and updates to the props table.

I added an enums display that will be in line with the function signature display once we update docs for those.  They are shown in the **description** area:

image

### Show all toggle (closed)

image

### Show all toggle (open)

image

PR #910, opened 2016-11-21
Added TextArea Example

iamcaleberic:
- Added TextArea Example

### DIFF
--- a/docs/app/Examples/addons/TextArea/Types/TextAreaExample.js
+++ b/docs/app/Examples/addons/TextArea/Types/TextAreaExample.js
@@ -6,10 +6,10 @@ class TextAreaExample extends Component {
     return (
       <Form>
         <Form.Field>
-          <Form.TextArea name='TextArea' label='TextArea' placeholder='TextArea placeholder text' rows='3'/>
+          <Form.TextArea name='TextArea' label='TextArea' placeholder='TextArea placeholder text' rows='3' />
         </Form.Field>
       </Form>
-    ) 
+    )
   }
 }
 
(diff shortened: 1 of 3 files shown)


docs/app/Examples/addons/TextArea/Types/TextAreaExample.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react'
+import { Form, TextArea } from 'semantic-ui-react'
+
+class TextAreaExample extends Component {
```
levithomason: We prefer to use a stateless functional component when possible, could you update this to use a stateless functional component?  I think we should also show the simplest default `<Textarea />` usage here.  The current usage is really a feature of the `<Form />`.

On a side note, `<Form.TextArea />` is a `<TextArea />` wrapped in a `<Form.Field />` so the wrapping form field here is creating a duplicate nested form field.

```jsx
const TextAreaExample = () => (
  <TextArea placeholder='TextArea placeholder text' rows='3' />
)
```

docs/app/Examples/addons/TextArea/Types/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+const TextAreaTypesExamples = () => (
+  <ExampleSection title='Types'>
```
levithomason: We mirror Semantic-UI core doc headings and examples.  Then, all extra examples we're placing under a "Usage" heading, for now.  Let's update `Types` here to `Usage`, since Semantic-UI has no CSS definition page that shows "Types" for a textarea.

levithomason (changes requested):
Thanks much contributing!  I've left a couple requests for updates.

levithomason (approved):

levithomason:
Thanks much!  CI will ship this in a few minutes.

PR #916, opened 2016-11-22
fix(Input): spread input props on the input child

### LINKED ISSUE #909: fix(Input): onChange callback does not pass the value of the input when using Input with children
Similar to this open issue: https://github.com/Semantic-Org/Semantic-UI-React/issues/907

When using the Input component with children, you have to use the onChange callback with the child input tag, and no second data parameter is passed to the callback.

**Steps to Reproduce**
1. Create a Input with children, you must use your own <input /> tag as per the Docs. The semantics <Input> tag will not run the onChange callback, you must pass as an option in your <input /> tag.
2. pass the onChange option with a callback option, have the callback console.log the arguments.

**Expected**
The onChange callback function should receive the input value as a second parameter in the callback.

**Result**
The callback option does not receive the value of the input as the second parameter.

**Testcase**
http://codepen.io/tintin301/pen/rWyBvw

levithomason:
Fixes #909 

This PR spreads props on the `input` child when using children in advanced input variations.  It also updates the docs, hoisting all `<input />` props to the `<Input />` parent.

I've spread the props so that the `input` child props win, for backward compatibility.  Though, the suggested use here is to use the `input` child as a placeholder, without props.  I've updated the docs to reflect this as well.

### DIFF
--- a/docs/app/Examples/elements/Input/Variations/InputExampleActions.js
+++ b/docs/app/Examples/elements/Input/Variations/InputExampleActions.js
@@ -8,8 +8,8 @@ const options = [
 ]
 
 const InputExampleActions = () => (
-  <Input action>
-    <input type='text' placeholder='Search...' />
+  <Input type='text' placeholder='Search...' action>
+    <input />
     <Select compact options={options} defaultValue='articles' />
     <Button type='submit'>Search</Button>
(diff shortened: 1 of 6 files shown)


levithomason:

PR #917, opened 2016-11-22
feat(FormField): pass `required` to the control

### LINKED ISSUE #911: FormField: `required` prop not passed to control
### Steps
1. Pass `required` prop to form field, e.g. `<Form.Field control={Input} required />`

### Expected Result
`required` prop would be passed to `Input`

### Actual Result
`required` prop not passed to `Input`

### Version
0.61.2

### Testcase
http://codepen.io/jeffcarbs/pen/JbWLYa

----

The issue is that `Form.Field` has a `required` prop (which applies the `required` class to the wrapping `div.field` element) so it "consumes" the prop. This means it never gets passed to the `Input`. As you can see in the codepen where I manually used `<Input required />`, modern browsers will automatically prevent the form submission if a required field is blank (among other validations). I think in addition to using the `required` prop for the className the `Form.Field` should pass `required` down to the control.

levithomason:
Fixes #911 

This PR updates the FormField to pass the `required` prop to the `control` component.

```jsx
<Form.Field control='input' required />
//=> has child: <input required>

<Form.Input required />
//=> has child: <Input required />
//=> has child: <input required />
```

### DIFF
--- a/src/collections/Form/FormField.js
+++ b/src/collections/Form/FormField.js
@@ -56,7 +56,7 @@ function FormField(props) {
   // ----------------------------------------
   // Checkbox/Radio Control
   // ----------------------------------------
-  const controlProps = { ...rest, children, type }
+  const controlProps = { ...rest, children, required, type }
 
   // wrap HTML checkboxes/radios in the label
   if (control === 'input' && (type === 'checkbox' || type === 'radio')) {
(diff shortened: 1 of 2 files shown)


levithomason:

su6838354:
when the component is select/dropdown, it may not effective.

levithomason:
Correct, there is no `required` prop for a Dropdown (nor Select).  There is no DOM node for the browser to "require" a value from.  The Dropdown is constructed from HTML and CSS.

You should instead use form validation, #407.

PR #920, opened 2016-11-22
feat(TextArea): add onChange(e, data) callback signature

### LINKED ISSUE #907: TextArea onChange callback does not pass the Input object
TextArea - onChange callback does not pass the input object. Most other Form fields I have tried does this.

### Steps
See also Testcase

1. Create a TextArea and give it a onChange callback function.
2. In the callback function print the arguments.

### Expected Result

The callback function should pass the input object as the second parameter. 
This would be consistent with other form fields.

### Actual Result

The onChange callback for TextArea has only one argument.

### Version
0.61.1

### Testcase
http://codepen.io/anon/pen/LbxqKM

levithomason:
Fixes #907 

This PR adds the `onChange(e, data)` signature to the TextArea.  This component is now a class to house the `handleClick` state.  Old tests were also removed (covered by common) and `onChange` tests were added.

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Component, PropTypes } from 'react'
 import {
   customPropTypes,
   getElementType,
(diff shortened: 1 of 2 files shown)


levithomason:

PR #925, opened 2016-11-23
Added Select example

iamcaleberic:
- Added Select Example


levithomason:
I added a comment showing the shape of `countryOptions`, will merge on pass.  Thanks much!

PR #928, opened 2016-11-23
fix(Accordion): Fixed incorrect defaultActiveIndex check

### LINKED ISSUE #772: Accordion does not work properly when defaultActiveIndex is 0
set defaultIndex to 0, the panel will not unfold. but rest of the index, it works properly.

My code here (v0.58.0)

```
import { Accordion } from 'semantic-ui-react'
import faker from 'faker'
import _ from 'lodash'

const panels = _.times(3, () => ({
  title: faker.lorem.sentence(),
  content: faker.lorem.paragraphs(),
}))

export default class AccordionPanelsPropExample extends Component {
  render() {
    return (
      // at index 0 not working, but rest index working fine. 1 or 2
      <Accordion defaultActiveIndex={0} panels={panels} />
    )
  }
}
```

Arthelon:
Fixes #772

This PR fixes the incorrect check of the props defaultActiveIndex in the componentWillMount hook. Instead of using the ! operator, which evaluates falsy values like 0 to true, the new condition checks for an undeclared variable or the number -1.

### DIFF
--- a/src/modules/Accordion/Accordion.js
+++ b/src/modules/Accordion/Accordion.js
@@ -83,7 +83,7 @@ export default class Accordion extends Component {
     // Otherwise, on first render we're allowed to set state for a prop that might have a default.
     // The default prop should always win on first render.
     // This default check should then be removed.
-    if (!this.props.defaultActiveIndex) {
+    if (typeof this.props.defaultActiveIndex === 'undefined' || this.props.defaultActiveIndex === -1) {
       this.trySetState({ activeIndex: -1 })
     }
   }
--- a/src/modules/Accordion/Accordion.js
+++ b/src/modules/Accordion/Accordion.js
@@ -83,9 +83,10 @@ export default class Accordion extends Component {
     // Otherwise, on first render we're allowed to set state for a prop that might have a default.
     // The default prop should always win on first render.
     // This default check should then be removed.
-    if (typeof this.props.defaultActiveIndex === 'undefined' || this.props.defaultActiveIndex === -1) {
+    if (typeof this.props.defaultActiveIndex === 'undefined') {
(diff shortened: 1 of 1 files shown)


Arthelon:
@levithomason Sorry for the trouble. Fixed.

src/modules/Accordion/Accordion.js
```diff
@@ -83,7 +83,7 @@ export default class Accordion extends Component {
     // Otherwise, on first render we're allowed to set state for a prop that might have a default.
     // The default prop should always win on first render.
     // This default check should then be removed.
-    if (!this.props.defaultActiveIndex) {
+    if (typeof this.props.defaultActiveIndex === 'undefined' || this.props.defaultActiveIndex === -1) {
```
levithomason: I believe this should only have the `undefined` check, true?  We only want to set the activeIndex if it was not already set.

If the prop were already set to `-1` then the auto controlled state will already be `-1` and we don't want to try to set it again.

LMK your thoughts on this.

Arthelon: @levithomason the typeof check for any undeclared or undefined variable will be "undefined"

levithomason: So, I am proposing that I _believe_ this is the correct logic:

```diff
-if (typeof this.props.defaultActiveIndex === 'undefined' || this.props.defaultActiveIndex === -1) {
+if (typeof this.props.defaultActiveIndex === 'undefined') {
  // ...
}
```

Arthelon: Yes, that would make more sense. Thanks for the catch!

levithomason:
Released in `semantic-ui-react@0.61.5`.

PR #930, opened 2016-11-23
feat(Button): Prevent click on disabled button

### LINKED ISSUE #929: Better handle `<Button disabled />`
Similar issue to #791

By default, setting a button to `disabled` only disables pointer events. You can still tab-focus and press enter to "click" it. We should prevent clicks (both mouse clicks and enter) on disabled `Button` within the component itself.

Also, like the Dropdown issue, there's no way to show a popup on a disabled `Button` (which I'd imagine is a very common use-case, letting the user know why the button is disabled). I don't think we should manually set the `pointer-events`, but we should make it possible for the user to do so while still keeping the "disabled" behavior of preventing the default action.

See the following codepen for examples: http://codepen.io/jeffcarbs/pen/gLWBRj

jeffcarbs:
Fixes #929 

- Prevent clicks on a disabled button (both mouse clicks and "clicks" via focus + pressing enter). Also called `preventDefault` on the click event when the button is disabled to prevent form submission (if within a form) or link following (if an `a`).
- Pass back props to the `onClick` and documented the `onClick` prop as spec'd in https://github.com/Semantic-Org/Semantic-UI-React/issues/623#issuecomment-261018287

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -1,5 +1,5 @@
 import cx from 'classnames'
-import React, { PropTypes } from 'react'
+import React, { Component, PropTypes } from 'react'
 
 import {
   customPropTypes,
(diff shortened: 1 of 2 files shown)


levithomason:
Looks good, thanks!

PR #932, opened 2016-11-23
feat(Modal): Add closeIcon

### LINKED ISSUE #436: Modal: Support showing Close (X) Icon
It would be nice if we can specify a prop to show close (x) icon on the top right corner, which would close the model and fires onClose or onHide.

jeffcarbs:
Fixes #436

Given #893 is a little controversial due to the reliance on the DOM and data props, this solves a more acute need of adding a close icon to the `Modal`, which is standard in SUI core (see http://semantic-ui.com/modules/modal.html#/usage): 
screen shot 2016-11-23 at 10 57 12 am

While this mixes shorthand with children, I think it's reasonable given:
- We already do this elsewhere (see `Message`: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Message/Message.js#L74-L81)
- Pending a solution the question posed in https://github.com/Semantic-Org/Semantic-UI-React/pull/893#issuecomment-262409730, there'd be no way to implement the "close" behavior with a close icon with children without controlling the `open` state yourself, which I think is overkill for something as simple/common as the close icon.
- It's not there by default, so you have to "opt-in" by passing `closeIcon='whatever'`. And even then, by default it's absolutely positioned so it doesn't affect the rest of the markup at all.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -51,6 +51,7 @@ class Modal extends Component {
     closeIcon: PropTypes.oneOfType([
       PropTypes.node,
       PropTypes.object,
+      PropTypes.bool,
     ]),
 
     /** A modal can reduce its complexity */
@@ -217,9 +218,11 @@ class Modal extends Component {
     const portalProps = _.pick(unhandled, portalPropNames)
     const ElementType = getElementType(Modal, this.props)
 
+    const closeIconName = closeIcon === true ? 'close' : closeIcon
+
     const modalJSX = (
       <ElementType {...rest} className={classes} style={{ marginTop }} ref={c => (this._modalNode = c)}>
-        {Icon.create(closeIcon, { onClick: this.handleClose })}
+        {Icon.create(closeIconName, { onClick: this.handleClose })}
         {children}
       </ElementType>
     )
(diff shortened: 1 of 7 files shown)


src/modules/Modal/Modal.js
```diff
@@ -45,9 +47,18 @@ class Modal extends Component {
     /** Additional classes. */
     className: PropTypes.string,
 
+    /** Icon */
+    closeIcon: PropTypes.oneOfType([
+      PropTypes.node,
```
levithomason, line 52: How about allowing boolean as well?  I would think `closeIcon` alone would be the primary use case for this prop.

levithomason:
I believe this also fixes #436, if so, want to update the description to close it on merge?

jeffcarbs:
@levithomason - you're right, added that issue to the description 👍

levithomason:
Cool, what do you think about the boolean prop for the default `close` icon?  https://github.com/Semantic-Org/Semantic-UI-React/pull/932#discussion_r89366895

jeffcarbs:
Added in da00365 👍

levithomason:
👍 will merge on pass

levithomason (approved):

levithomason:
Released in `semantic-ui-react@0.61.4`.

PR #950, opened 2016-11-28
feat(Card): pass props to onClick

### LINKED ISSUE #949: Card  onClick() callback function signature.
Hello
I am using the Card component inside map to render multiple cards.
What I want is to be able to click the card and go to a link using  onClick (I use react-router with browser history)

in the function on handleItemClick how can I know what card is the function called from?
I have used Menu.Items before and I can get that info with the name prop by passing event and {name} together 
  handleItemClick = (e, { name }) => this.setState({ activeItem: name })

Is there any prop like name that i can pass here, I could not find it in the docs.
and I am sorry if this is not the right way to ask for help , I am very new to github. 
### Version
0.61.6

jeffcarbs:
Fixes #949

### DIFF
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -66,7 +66,13 @@ export default class Card extends Component {
     /** Shorthand for CardMeta. */
     meta: customPropTypes.itemShorthand,
 
-    /** Render as an `a` tag instead of a `div` and called with event on Card click. */
+    /**
+     * Called after user's click. When passed, the component render as an `a`
+     * tag by default instead of a `div`.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - Card props.
+     */
     onClick: PropTypes.func,
 
     /** A Card can be formatted to raise above the page. */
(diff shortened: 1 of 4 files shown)


src/views/Card/Card.js
```diff
@@ -66,7 +66,13 @@ export default class Card extends Component {
     /** Shorthand for CardMeta. */
     meta: customPropTypes.itemShorthand,
 
-    /** Render as an `a` tag instead of a `div` and called with event on Card click. */
+    /**
+     * Called after user's click. When passed, the component render as an `a`
+     * tag by default instead of a `div`.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - Relevant data.
```
levithomason: I used the phrase "Relevant data" as an example to convey the intent in the spec, however, I think we should be as descriptive as possible in the actual docs.  In this case, we can say `data` is props since that is all that is being passed.

levithomason (changes requested):

jeffcarbs:
@levithomason - updated 👍

levithomason (approved):

PR #951, opened 2016-11-28
breaking(Form|RatingIcon|Search|SearchResult): Pass data to event handlers

### LINKED ISSUE #623: RFC: all event handlers should callback with (event, data)

jeffcarbs:
Breaking Changes!
-------------------

The following components have breaking changes in their event handler props.

### Form

onClick | (e, formData) | (e, { ...props, formData })

### RatingIcon

onClick | (e, index) | (e, { ...props, index })
onKeyUp | (e, index) | (e, { ...props, index })
onMouseEnter | (e, index) | (e, { ...props, index })

### Search

NOTE: The handler name changed from `onChange` to `onResultSelect`. The change was made because:
- onChange doesn't really make sense in this context since nothing is really changing.
- given the result is being returned, onChange made even less sense.
- it maps closer to the SUI-core event (`onSelect`). The reason to not use `onSelect` is because that's a built-in html event type for when a user selects text.

Handler | New Name | Previous Signature | Current Signature
------- | -------- | ------------------ | -----------------
onChange | onResultSelect | (e, result) | (e, result) _no change_

### SearchResult

onClick | (e, id) | (e, { ...props, id })

----

Fixes #623

This updates the common tests for event handling to ensure any handlers that we wrap are called with `(event, data)` as arguments.

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -110,19 +110,44 @@ export default class Search extends Component {
     // Callbacks
     // ------------------------------------
 
-    /** Called with the React Synthetic Event on Search blur. */
+    /**
+     * Called on blur.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
     onBlur: PropTypes.func,
 
-    /** Called with the React Synthetic Event, the selected result. */
-    onChange: PropTypes.func,
+    /**
+     * Called when a result is selected.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
+    onResultSelect: PropTypes.func,
 
-    /** Called with the React Synthetic Event and current value on search input change. */
+    /**
+     * Called with the React Synthetic Event and current value on search input change.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
     onSearchChange: PropTypes.func,
 
-    /** Called with the React Synthetic Event on Search focus. */
+    /**
+     * Called on focus.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
     onFocus: PropTypes.func,
 
-    /** Called with the React Synthetic Event on Dropdown mouse down. */
+    /**
+     * Called on mousedown.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
     onMouseDown: PropTypes.func,
 
     // ------------------------------------
@@ -251,13 +276,11 @@ export default class Search extends Component {
   // Document Event Handlers
   // ----------------------------------------
 
-  // onChange needs to receive a value
-  // can't rely on props.value if we are controlled
-  handleChange = (e, result) => {
-    debug('handleChange()')
+  handleResultSelect = (e, result) => {
+    debug('handleResultSelect()')
     debug(result)
-    const { onChange } = this.props
-    if (onChange) onChange(e, result)
+    const { onResultSelect } = this.props
+    if (onResultSelect) onResultSelect(e, result)
   }
 
   closeOnEscape = (e) => {
@@ -294,9 +317,9 @@ export default class Search extends Component {
     // prevent selecting null if there was no selected item value
     if (!result) return
 
-    // notify the onChange prop that the user is trying to change value
+    // notify the onResultSelect prop that the user is trying to change value
     this.setValue(result.title)
-    this.handleChange(e, result)
+    this.handleResultSelect(e, result)
     this.close()
   }
 
@@ -313,7 +336,7 @@ export default class Search extends Component {
   handleMouseDown = (e) => {
     debug('handleMouseDown()')
     const { onMouseDown } = this.props
-    if (onMouseDown) onMouseDown(e)
+    if (onMouseDown) onMouseDown(e, this.props)
     this.isMouseDown = true
     // Do not access document when server side rendering
     if (!isBrowser) return
@@ -337,7 +360,7 @@ export default class Search extends Component {
     this.tryOpen()
   }
 
-  handleItemClick = (e, id) => {
+  handleItemClick = (e, { id }) => {
     debug('handleItemClick()')
     debug(id)
     const result = this.getSelectedResult(id)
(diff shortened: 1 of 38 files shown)


jeffcarbs:
@levithomason - thoughts on this approach? 

The alternative would be to have a separate common test like:
```js
common.passesPropsToEventHandler(Component, 'onClick')
```

The issue with the current approach is that non-built-in events (e.g. `onItemClick`) aren't tested. That being said, I don't think the suggestion here would work either since there's no way to reliably trigger the event anyway, so we're not really gaining anything.

Also I know we try to avoid big refactors, but assuming we go with this approach (updating the 'handles events' spec) I think it makes sense to just fix all of the handlers at once. These files currently have failing specs:
- AccordionTitle
- BreadcrumbSection
- Checkbox
- Dropdown Component
- Form
- MenuItem
- Search Component
- SearchResult
- Select
- Step
- TableFooter
- TableHeaderCell

levithomason:
I like the idea of doing these all at once as it makes for less breaking changes and a clear upgrade path for users.

My idea for solving the handled/unhandled callbacks is precisely what you've done.  Though, I we can't assert the `data` param is `props` because sometimes it has other data like `state.value` in `TextArea` and `Input`.  There may be others that have arbitrary data as well.  Instead, we should match against an empty object `{}` for now.  Any further data specific logic I think should be the responsibility of individual component test.

jeffcarbs:
`calledWithMatch` is an alias for `calledWith(sinon.match(arg1), sinon.match(arg2), ...)` and `sinon.match(object)` does an incomplete match on objects. So `calledWithMatch(event, props)` will pass as long as the second arg is an object that includes each key/value in `props`, even if there are additional key/value.

jeffcarbs:
@levithomason - ready for review 👍 

One question I had about `Dropdown`. Currently we pass `{ ...this.props, value: proposedValue }` as the second argument to `onChange` but only pass the props to the other handlers. Should we also pass the value to the rest of the event handlers? I was thinking something like:

```js
getHandlerProps = () => {
  return { ...this.props, value: this.state.value }
}

onChange = (e, value /* <- proposed value */) => {
  // ...
  props.onChange(e, { ...this.getHandlerProps(), value })
}

onClick = (e) => {
  // ...
  props.onClick(e, this.getHandlerProps())
}
```

levithomason:
>So calledWithMatch(event, props) will pass as long as the second arg is an object that includes each key/value in props, even if there are additional key/value.

This is correct, however, the test would have failed because the value in the `data` object is overridden as the event value and  `props.value != e.target.value`.  So, we can not be certain that all the props key/value pairs will appear in `data`.

>Should we also pass the value to the rest of the event handlers?

Initially, I don't think this is necessary in this case.  When the props change the ACC copies those values to state so the `props.value` will always equal `state.value`, except in the case of the `onChange` where we need to communicate the proposed new value.

That aside, it is a good idea IMO to have a SSOT for the handler props so I still like this idea.  It makes testing easier and the component stronger.

src/modules/Search/Search.js
```diff
@@ -110,19 +110,44 @@ export default class Search extends Component {
...
+     * Called when a result is selected.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
+    onResultSelect: PropTypes.func,
 
-    /** Called with the React Synthetic Event and current value on search input change. */
+    /**
+     * Called with the React Synthetic Event and current value on search input change.
```
levithomason: Let's drop the signature info from the description here as well.

layershifter:
For clarity, `RatingIcon` is internal component and it can't be used outside `Rating` component. I think it's bad idea to export it.

test/specs/modules/Accordion/AccordionTitle-test.js
```diff
@@ -1,43 +1,9 @@
...
 
-import Accordion from 'src/modules/Accordion/Accordion'
 import AccordionTitle from 'src/modules/Accordion/AccordionTitle'
 
 describe('AccordionTitle', () => {
   common.isConformant(AccordionTitle)
   common.rendersChildren(AccordionTitle)
   common.propKeyOnlyToClassName(AccordionTitle, 'active')
-
-  describe('onClick', () => {
```
levithomason, line 13: Did we intend to drop these tests or update them?

levithomason (changes requested):

levithomason:
### Doc Examples

I'm guessing there are several doc site examples that need to be updated to reflect the new changes as well.  Would finding handler prop names in the docs, `/on[A-Z]\w+=\{/`, and updating where necessary.

### RatingIcon

>For clarity, RatingIcon is internal component and it can't be used outside Rating component. I think it's bad idea to export it.

@layershifter Good call. We should make this a private component in this case.  Simply prefixing its name with `_` will make it private.  The common tests will then automatically assert that it _is not_ exposed anywhere in the API.

test/specs/commonTests.js
```diff
@@ -304,10 +304,18 @@ export const isConformant = (Component, options = {}) => {
           'You may need to hoist your event handlers up to the root element.\n'
         )
 
+        let expectedArgs = [eventShape]
+        let errorMessage = 'was not called with (event)'
+
+        if (_.has(Component.propTypes, listenerName)) {
+          expectedArgs = [eventShape, props]
```
levithomason, line 311: I see all tests are passing, but how does this work for things like the Input onChange where the `data` arg's `value` key != to the `props` value key?  Same for Checkbox's `checked` prop.

test/specs/modules/Accordion/AccordionTitle-test.js
jeffcarbs: Intentional given that all the functionality is already being tested by the event-related tests within `isConformant`

test/specs/commonTests.js
jeffcarbs: `props` here isn't the Component's full props object, it's just:
```
const props = {
  ...requiredProps,
  [listenerName]: handlerSpy,
  'data-simulate-event-here': true,
}
```

So we're really just testing that arbitrary props get passed through (e.g. that `data-simulate-event-here` prop). Tests for additional data, like `checked` of a `Checkbox`, should be handled in that Component's test file.

jeffcarbs:
I started looking into making `RatingIcon` a private component and I don't really think we should:
- There are currently no other private components, the common tests for them don't actually work out of the box.
- While `RatingIcon` it's not usable outside of `Rating`, neither is `Dropdown.Divider` usable outside of `Dropdown`, nor `ItemImage` outside of `Item`.
- Our goal is to provide a Component API to give someone full control in rendering all SUI components, so we need to export it to allow full control over a `Rating`.

levithomason (approved):

levithomason:
That logic makes sense to me, thanks @jcarbo.

levithomason:

PR #967, opened 2016-12-01
feat(Checkbox|FormField) support `label` shorthand

### LINKED ISSUE #939: FormField support `label` shorthand
I would like to pass some html to my label in my radio button. But the html is not rendered.

```
<Form.Field>
    <label>Quel est le motif de votre inscription ?</label>
    <Field name="test" component={DefaultRadio} type="radio" label="<b>Value : 0</b>" value="0"/>
    <Field name="test" component={DefaultRadio} type="radio" label="Value : 1" value="1"/>
</Form.Field>
```
`Field` is a specific component from redux-form

```
class DefaultRadio extends Component {
    render() {
        const {input, meta, ...rest} = this.props;
        return (
            <Form.Field control={Radio} {...input} {...rest} />
        );
    }
}
```
### Expected Result
.... **Value : 0**

### Actual Result
.... Value : 0

### Version
0.58.2

layershifter:
Addresses part of #939.

This PR:
* cleanups components and tests code;
* adds docs about `label` shorthand;
* improves `label` shorthand for `Checkbox` and `FormField` components.


test/specs/modules/Checkbox/Checkbox-test.js
```diff
@@ -41,6 +37,15 @@ describe('Checkbox', () => {
     })
   })
 
+  describe('defaultChecked', () => {
+    it('sets the initial checked state', () => {
+      // consoleUtil.disableOnce()
```
levithomason: Is this meant to be uncommented or removed?

docs/app/Examples/modules/Checkbox/Types/CheckboxExampleShorthand.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Checkbox, Divider } from 'semantic-ui-react'
+
+const CheckboxExampleCheckbox = () => (
+  <div>
+    <Checkbox label={{ children: 'Prop label', className: 'foo' }} />
+    <Divider />
+    <Checkbox label={<label><b>Element label</b></label>} />
```
levithomason: We're listing each shorthand as separate examples.  Ideally, shorthand examples immediately follow the example showing the regular usage.  I think we should split this into two examples and show them immediately after the first Checkbox example:

image

levithomason (changes requested):

levithomason:
Though this solves the Checkbox usage, I don't see the FormField updates as the title states.

levithomason:
I'm thinking if a user does this:

```jsx
<Form.Field label={<MyLabel />} control={<MyControl />} />
<Form.Field label={{ children: 'Name' }} control={<input name='name' />} />
```

That we should pass that ^ value through the factory as well.  Currently, we assume it is a string and render it nested in a `<label />`.

We can do this on another PR if you like and just leave this one to the Checkbox label updates.  LMK.

test/specs/modules/Checkbox/Checkbox-test.js
layershifter: Looks like cruft.

layershifter:
I've added updates for `FormField` and splited shorthand examples 👍

layershifter:
> We can do this on another PR if you like and just leave this one to the Checkbox label updates. LMK.

I think we don't need touch `control` in this PR.

levithomason (approved):

levithomason:
Awesome work, thanks :D 

I updated the PR description to not close #939 since the PR adding shorthand to the FormField label should close that one.

layershifter:
Hm... It seems there happened misunderstanding.

----

This PR improved `label` shorthand on `Checkbox` and `FormField`, but didn't touch `control` shorthand on `FormField`. Now we can to do this:

```jsx
<FormField label='my label' />
<FormField label={123} />
<FormField label={{ children: 'foo' }} />
<FormField label={<div className='label' />} />
<FormField label={<MyComponent />} />
```

But, we can't do this:
```jsx
<FormField control={<input name='name' />} />
```

In this case `control` shorthand will go to [this branch](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Form/FormField.js#L96) and will fail due `createElement` :)

So, I think #939 can be closed, but we need to open new issue about `control` shorthand improvement.

levithomason:
Indeed, thanks for clarifying.  I've closed the other issue.

PR #972, opened 2016-12-01
feat(StatisticGroup): Add missing props

### LINKED ISSUE #970: Statistic size Prop doesn't work inside Statistic.Group
```jsx
<Statistic.Group widths="4">
  <Statistic size="mini" color="black" value={value} label="label" />
  <Statistic size="mini" color="red" value={value} label="label" />
  <Statistic size="mini" color="orange" value={value} label="label" />
  <Statistic size="mini" color="green" value={value} label="label" />
</Statistic.Group>

This should display 4 mini-sized statistics "in a row" but they are always regular size
Is there other way to specify the size?

jeffcarbs:
Fixes #970

This adds props for `color`, `size`, and `inverted` to the `StatisticGroup`.

### DIFF
--- a/src/views/Statistic/StatisticGroup.js
+++ b/src/views/Statistic/StatisticGroup.js
@@ -14,11 +14,14 @@ import {
 import Statistic from './Statistic'
 
 function StatisticGroup(props) {
-  const { children, className, horizontal, items, widths } = props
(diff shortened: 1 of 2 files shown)


levithomason:
Thanks Jeff! Released in `semantic-ui-react@0.61.9`.

PR #973, opened 2016-12-01
fix(Modal): prevent unnecessary renders

### LINKED ISSUE #969: Modal continuously updating
Whenever a modal is open and you check the updates being done by React, the modal is being updated constantly even when doing nothing. I don't know if this is intended or not, but I imagine it puts strain on the rendering.

### Steps
1. Go to: http://react.semantic-ui.com/modules/modal
2. Click on a modal button
3. Enable `Trace React Updates` in the React tab of the Web Developer Console.
4. ...

### Expected Result
No continuous updating of the modal

### Actual Result
Continuously updating the modal as shown here:
Continuously updating modal

### Version
Checked versions 0.61.7 and 0.61.8.

jeffcarbs:
Fixes #969 

We continuously check the height of the `Modal` to determine if it should be scrolling or not. When we made `Modal` autocontrolled, it broke the check we were using to determine if we should call `setState`.

This fixes the check and makes it faster and more future-proof so if more keys are added to state it won't unexpectedly break again.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -173,23 +173,27 @@ class Modal extends Component {
     if (this._modalNode) {
       const { mountNode } = this.props
       const { height } = this._modalNode.getBoundingClientRect()
+
+      const marginTop = -Math.round(height / 2)
       const scrolling = height >= window.innerHeight
 
-      const newState = {
-        marginTop: -Math.round(height / 2),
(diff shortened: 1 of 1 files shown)


levithomason:
Thanks much!

levithomason:
Released in `semantic-ui-react@0.61.9`.

PR #977, opened 2016-12-02
docs(Icon): add icon search

levithomason:
Finding icons can be a pain.  The current icon set examples are far too large to be of any use with the code editor provided.  You need to scroll up and down for pages to make changes and see your results.  Users probably also have no idea that there are other examples at the bottom of that massive page.

This PR removes the code editor icon set examples and replaces them with an icon search.  It moves the other examples to an "Examples" tab for clarity.

# Search Features
- auto focused search input on mount / load, so you can search instantly
- show names and aliases that "contain" the query
- optionally show names/alias that are "similar to" query (default on)
- click to copy

# Demo


src/collections/Menu/Menu.js
```diff
@@ -153,7 +153,7 @@ class Menu extends Component {
     const { activeIndex } = this.state
 
     return _.map(items, (item, index) => {
-      return createShorthand(MenuItem, val => ({ content: val }), item, {
+      return createShorthand(MenuItem, val => ({ content: val, name: val }), item, {
```
levithomason, line 156: When using shorthand items as an array of strings, the strings were not passed to the `name` prop of the menu item, so the callback name was `undefined`.  I've added the literal shorthand values to the name prop so the callbacks contain the name.

PR #987, opened 2016-12-05
fix(Sidebar) sidebar is a module, not a collection

fracmak:
woops, the only place that correctly shows sidebar as a module is in the docs, everywhere else it was incorrectly listed as a collection.


levithomason:
Whoops, thanks for the fix!

levithomason:

Issue #978, opened 2016-12-03
Portal `mouseOver` should be `mouseEnter`


levithomason:
The `mouseOver` event is called on [enter _and_ on leave](http://jsfiddle.net/ZCWvJ/7/).  This is creating duplicate timers for opening the portal.  If you add a `mouseOverDelay` this bug is very obvious.  The portal will open after the mouse leaves the element even though the timer is cleared.

**This is a simple fix** as we simply need to replace all `mouseOver` references with `mouseEnter` in the Portal, Modal, Popup including docs and tests.

Unfortunately, this is a breaking change since all `[mM]ouseOver` props will become `[mM]ouseEnter`.

## Bug demo with mouseOverDelay={1000}

![http://g.recordit.co/Z7Aei4wnUw.gif](http://g.recordit.co/Z7Aei4wnUw.gif)

## Fix demo with mouseEnterDelay={1000}

![http://g.recordit.co/tyM2LqQPVT.gif](http://g.recordit.co/tyM2LqQPVT.gif)

## Why is this happening?

The mouse over event is creating two "open" timers, one on mouse enter and another on mouse leave.  The mouse leave event is only clearing one of those, the latest one.  The first open timer is left running in memory.  When it expires, it opens the portal, even though the mouse has left and the _latest_ open timer has been canceled.

dvdzkwsk:
WIP PR: https://github.com/Semantic-Org/Semantic-UI-React/pull/989.

I'd be willing to go through and maintain backwards compatibility by using `mouseOver` if it exists and `mouseEnter` does not, while also logging a warning, but seeing as Stardust is pre-1.0.0 I'd prefer this to be discussed before just implementing it, since I'm not convinced it's worth adding a deprecation warning to such a version.

levithomason:
Awesome, thanks for starting this.  Agreed on skipping deprecation warnings.  It is inline with our development phase warnings at the top of the README:

>Hey, we're in development. Prior to reaching [v1.0.0][6]:
>
>1. **MINOR** versions represent **breaking changes**
>1. **PATCH** versions represent **fixes _and_ features**
>1. There are **no deprecation warnings** between releases
>1. You should consult the [**CHANGELOG**][18] and related issues/PRs for more information

[...]

PR #989, opened 2016-12-06
breaking(Portal|Popup): replace mouseOver event listener with mouseEnter

dvdzkwsk:
# Breaking Change

Fixes https://github.com/Semantic-Org/Semantic-UI-React/issues/978.

This PR replaces all `mouseOver`-related `<Portal />` events with `mouseEnter` to be consistent with `mouseLeave`. This affects the `<Portal />` component directly, as well as any which implement it, which at this time includes `<Popup />`.

### Upgrade Guide

All related `mouseOver` props should now be `mouseEnter`. This will also impact the logic of these handlers since they are now only called when the element is entered, not (incorrectly) on enter and exit.

**BEFORE**
```jsx
const onMouseOver = () => {
  console.log('Entered (or exited!!) the element in the portal')
}

<Portal onMouseOver={onMouseOver} trigger={<h2>hello</h2>} />
```

**AFTER**
```jsx
const onMouseEnter = () => {
  console.log('Entered the element in the portal')
}

<Portal onMouseEnter={onMouseEnter} trigger={<h2>hello</h2>} />
```

### DIFF
--- a/docs/app/Components/IconSearch/IconSearch.js
+++ b/docs/app/Components/IconSearch/IconSearch.js
@@ -75,7 +75,7 @@ export default class IconSearch extends Component {
   renderIconColumn = (name) => (
     <Popup
       key={name}
-      mouseOverDelay={500}
+      mouseEnterDelay={500}
       inverted
       hoverable={false}
       size='mini'
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -76,7 +76,7 @@ class Portal extends Component {
     mouseLeaveDelay: PropTypes.number,
 
     /** Milliseconds to wait before opening on mouse over */
-    mouseOverDelay: PropTypes.number,
+    mouseEnterDelay: PropTypes.number,
 
     /**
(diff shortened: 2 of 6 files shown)


levithomason:
I know you're a  busy fellow, any progress here @davezuko?  Icon docs are borked without this guy.

dvdzkwsk:
No, perhaps I will take a stab again tonight. There's a failing test or two that I'm absolutely stumped on, but maybe a fresh look will solve it.

levithomason:
Cool, LMK.

levithomason:
Nice, this looks good.  Want to add some breaking change notes as laid out in https://github.com/Semantic-Org/Semantic-UI-React/issues/978#issuecomment-265064306?  Also, the title will then be `breaking(Portal): ...`.

dvdzkwsk:
@levithomason how is this?

levithomason:
Looks great, thanks!

levithomason:
Released in `semantic-ui-react@0.63.0`.

PR #990, opened 2016-12-06
docs(Popup): make allowed content types statically analyzable

dvdzkwsk:
This is presumably an issue with all `propType` definitions that reuse propTypes defined in another variable. I traced it down to `react-docgen` being unable to parse the union type because it viewed it as dynamic, which is technically true because that type definition is referenced from a variable and as such the propType is resolved at runtime.

## Before
screen shot 2016-12-05 at 9 42 50 pm

## After
screen shot 2016-12-05 at 9 40 50 pm

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -21,7 +21,6 @@ const _meta = {
   name: 'Popup',
   type: META.TYPES.MODULE,
   props: {
-    content: [PropTypes.string, PropTypes.node],
     on: ['hover', 'click', 'focus'],
     positioning: [
       'top left',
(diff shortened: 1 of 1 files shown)


src/modules/Popup/Popup.js
```diff
@@ -21,7 +21,6 @@ const _meta = {
   name: 'Popup',
   type: META.TYPES.MODULE,
   props: {
-    content: [PropTypes.string, PropTypes.node],
```
dvdzkwsk, line 24: Also this looked to be in error, since it is not an enum but a propType.

levithomason:
Yikes, thanks.  Since `node` is "anything that can be rendered" it includes `string`, so we can actually simplify this to `PropTypes.node`.  React PropTypes [docs](https://facebook.github.io/react/docs/typechecking-with-proptypes.html#react.proptypes):

>  // Anything that can be rendered: numbers, strings, elements or an array
>  // (or fragment) containing these types.
>  optionalNode: React.PropTypes.node,

dvdzkwsk:
Good call, was not even paying attention to the types to be honest. Patch incoming.

dvdzkwsk:
Pushed and renamed commit accordingly

levithomason:
Thanks will merge on pass.

levithomason:

PR #991, opened 2016-12-06
fix(Accordion): allow contentShorthand for panel prop

wassname:
Allow accordian panel contents to be elements instead of just strings


levithomason:
Looks good, thanks much!

levithomason:

PR #995, opened 2016-12-07
feat(typings): add first draft typings

levithomason:
Updated version of #674


levithomason:

PR #1001, opened 2016-12-07
fix(List): Added aria roles for list/list-item

fracmak:
List elements are missing the appropriate roles to indicate what is being displayed.

Documentation: https://www.w3.org/TR/wai-aria/roles#list

### DIFF
--- a/test/specs/elements/List/List-test.js
+++ b/test/specs/elements/List/List-test.js
@@ -30,6 +30,22 @@ describe('List', () => {
   common.propValueOnlyToClassName(List, 'size')
   common.implementsVerticalAlignProp(List)
 
+  describe('role', () => {
+    const items = ['Name', 'Status', 'Notes']
+
+    it('is accessibile with no items', () => {
+      const wrapper = shallow(<List />)
+
+      wrapper.should.have.prop('role', 'list')
+    })
+
+    it('is accessibile with items', () => {
+      const wrapper = shallow(<List items={items} />)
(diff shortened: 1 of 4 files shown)


test/specs/elements/List/List-test.js
```diff
@@ -37,12 +37,14 @@ describe('List', () => {
       const wrapper = shallow(<List />)
 
       wrapper.find('ListItem').should.have.lengthOf(0)
+      wrapper.should.have.prop('role', 'list')
```
levithomason: Would prefer we add separate `it()` blocks for the accessibility assertions:

```jsx
it('is accessible', () => {...})
// or 
it('has proper aria props', () => {...})
```

Something to this effect.  The idea is to keep the `it()` strings descriptive to exactly what is being asserted inside of them.

levithomason (changes requested):
Just one comment regarding organizing tests.  Looks great otherwise 👍

levithomason:
The test organization used in https://github.com/Semantic-Org/Semantic-UI-React/pull/1006 is exactly what I was thinking.  Let's do that here as well.

fracmak:
tests reorganized

levithomason:
Made a trivial fix here, mostly to debug why CI is suddenly not running at all... Hm.

levithomason:
So odd, gonna have to double back and test these manually.

levithomason:

PR #1002, opened 2016-12-07
fix(Icon): added aria-hidden attribute to icon

fracmak:
Icons are primary for decoration or visual style and should be hidden from screen readers. Additional markup can be added to allow screen readers to read the icons if necessary, but by default all icons should be hidden.

Documentation: http://fontawesome.io/accessibility/

### DIFF
--- a/test/specs/elements/Icon/Icon-test.js
+++ b/test/specs/elements/Icon/Icon-test.js
@@ -28,4 +28,12 @@ describe('Icon', () => {
     shallow(<Icon />)
       .should.have.tagName('i')
   })
+
+  describe('aria', () => {
+    it('should add aria-hidden to icon', () => {
+      const wrapper = shallow(<Icon />)
+
+      wrapper.should.have.prop('aria-hidden', 'true')
+    })
+  })
 })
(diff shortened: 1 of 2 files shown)


test/specs/elements/Icon/Icon-test.js
```diff
@@ -25,7 +25,9 @@ describe('Icon', () => {
   common.rendersChildren(Icon)
 
   it('renders as an <i> by default', () => {
-    shallow(<Icon />)
-      .should.have.tagName('i')
+    const wrapper = shallow(<Icon />)
+
+    wrapper.should.have.tagName('i')
+    wrapper.should.have.prop('aria-hidden', 'true')
```
levithomason: Same comment here, let's pull these assertions into their own `it()` block.

levithomason (changes requested):

test/specs/elements/Icon/Icon-test.js
fracmak: updated

levithomason (approved):

levithomason:
Again, will manually test and release this later.

levithomason:

PR #1003, opened 2016-12-07
docs(MenuExampleVerticalDropdown): doesn't work twice

tomjnsn:
Setting the Dropdown `as={Menu.Item}` makes the dropdown only work the first time when clicked on. 

In addition, it makes the outer element an anchor tag which creates invalid HTML when you nest links in the dropdown which is probably a fairly common use case I would think. Although browsers do the right thing and work on the nested anchor tags it just doesn't feel right.

I borrowed from the DropdownExamplePointingTwo.js code that solves the problem by not setting the dropdown to a Menu.Item and adds in the necessary classes to show that it is a dropdown. That allows it to work multiple times on expanding and gives correct underlying HTML structure for putting anchor tags in the dropdown.

Just as a side question, since I'm fairly new to the Semantic-UI-React world: is that considered good practice to add in the classes like that or should it be something that is done under the covers by the component?

### DIFF
--- a/docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
+++ b/docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
@@ -13,7 +13,7 @@ export default class MenuExampleVerticalDropdown extends Component {
       <Menu secondary vertical>
         <Menu.Item name='account' active={activeItem === 'account'} onClick={this.handleItemClick} />
         <Menu.Item name='settings' active={activeItem === 'settings'} onClick={this.handleItemClick} />
-        <Dropdown as={Menu.Item} text='Display Options'>
+        <Dropdown text='Display Options' pointing='left' className='link item'>
           <Dropdown.Menu>
             <Dropdown.Header>Text Size</Dropdown.Header>
             <Dropdown.Item>Small</Dropdown.Item>


levithomason:
Thanks much for tackling this!  We want to solve this without use of `className` by adding props to the Dropdown itself.  See https://github.com/Semantic-Org/Semantic-UI-React/issues/642.

tomjnsn:
Ah yeah that's sort of what I was wondering. Thanks @levithomason !

levithomason:
I've re-commited your changes and forced push to trigger a new build.  We had an outage with CircleCI that was just fixed.

levithomason (changes requested):
See https://github.com/Semantic-Org/Semantic-UI-React/pull/1003#issuecomment-265563564

levithomason:
@tomjnsn any update here?

tomjnsn:
Sorry I think I'm lost here. What were you looking for from me?

docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
```diff
@@ -13,7 +13,7 @@ export default class MenuExampleVerticalDropdown extends Component {
       <Menu secondary vertical>
         <Menu.Item name='account' active={activeItem === 'account'} onClick={this.handleItemClick} />
         <Menu.Item name='settings' active={activeItem === 'settings'} onClick={this.handleItemClick} />
-        <Dropdown as={Menu.Item} text='Display Options'>
+        <Dropdown text='Display Options' pointing='left' className='link item'>
```
levithomason, line 16: Here, we would want to use a prop `linkItem` rather than a `className='link item'`.  The Dropdown actually already has this prop in its source, it is just commented out.  We need to enable that prop and add a description to it.

Then, there is a test stubbed for this prop in `Dropdown-test.js`.  That also need to be uncommented and updated to pass.

The final result will be:

```jsx
<Dropdown text='Display Options' pointing='left' linkItem>
```

levithomason (changes requested):
Here is a little more explicit instruction on what needs to happen here.  Let me know if it is still not clear.

docs/app/Examples/collections/Menu/Types/MenuExampleVerticalDropdown.js
tomjnsn: Ah ok, sorry for being dense 😄  I was just thinking about the doc change not the longer term, more complete, fix for the code. I'll see what I can do.

levithomason:
This work is being continued in #659 and will be merged soon.  Thanks for the initiative.

PR #1005, opened 2016-12-08
fix(Dropdown): fix crash when setting defaultOpen on a search dropdown

fracmak:
Because the this.open() call happens inside componentWillMount(), the search input hasn't been built yet and the this._search.focus() crashes. This simply moves the open call to a componentDidMount() function

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -884,7 +884,7 @@ export default class Dropdown extends Component {
 
     const { disabled, onOpen, search } = this.props
     if (disabled) return
-    if (search) this._search.focus()
+    if (search && this._search) this._search.focus()
     if (onOpen) onOpen(e, this.props)
 
     this.trySetState({ open: true })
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -335,9 +335,13 @@ export default class Dropdown extends Component {
...
     if (super.componentWillMount) super.componentWillMount()
     debug('componentWillMount()')
-    const { open, value } = this.state
+    const { value } = this.state
 
     this.setValue(value)
+  }
+
+  componentDidMount() {
+    const { open } = this.state
```
levithomason: Because open() sets state, this results in setting state on mount which is an antipattern.  We have a lint rule for this enabled, although, because it is indirect and abstracted away, it is not picking up on it.

I think we should instead add guarding logic where the focus() is called and only call it if there is a node present.

levithomason (changes requested):

src/modules/Dropdown/Dropdown.js
fracmak: updated

levithomason (approved):

levithomason:
Thanks much!

levithomason:

PR #1006, opened 2016-12-08
fix(Dropdown): added role/aria attributes to Dropdown and DropdownItem

fracmak:
This will fail tests until #1005 lands, but I wanted to get feedback on what I've done and the problems I ran into making the Dropdown truly accessible. 

**Things accomplished in this pull request**
* adds a role="listbox" to the normal Dropdown component
* adds a role="combobox" to the Dropdown component in search mode, with DropdownMenu getting role="listbox"
* adds a role="option" to the DropdownItem component
* adds aria flags for disabled, busy, expanded on the Dropdown component
* adds aria flags for disabled, checked, selected on DropdownItem component
* adds aria flags for multiselectable on Dropdown normal component, and DropdownMenu search components
* sets an aria-autocomplete="list" on the input field for search Dropdowns

So this mostly works in bringing it up to some basic accessibility standards. 

So let me dump some thoughts here for discussion:

**listbox/option vs menu/menuitem**
So one thing that will probably jump out to you is how similar listbox/option is to menu/menuitem. I've been reading up a lot on them and menu's are truly used for desktop style menus, where you drop down an entry and trigger an action and it goes away. This is backed up by this article: https://www.marcozehe.de/2014/03/11/easy-aria-tip-7-use-listbox-and-option-roles-when-constructing-autocomplete-lists/ Dropdowns are really a fancy select vs a real menu. EXCEPT when Dropdowns are added to Menu components, then they really are menu/menuitems. So I added an if statement to check if the as prop === Menu.Item. Not 100% certain the best way to switch the dropdown item from role=option to role=menuitem (though menuitemcheckbox might be better, need to ponder that one), thoughts?

**combobox and search dropdowns**
So this is where things get a little weird. Dropdowns can act like combobox's in search mode. So the roles and markup changes under those circumstances. I'm taking a lot of this from here: https://www.w3.org/TR/wai-aria-1.1/#combobox which has a nice pretty example to follow that follows a similar stucture to this Dropdown. I had to add the type="text" to the input to get the role="textbox", and the listbox role and aria-multiselectable had to move to the DropdownMenu since aria-multiselectable isn't a valid option for combobox. Now this leads into the next problem. 

**problems with ids**
Most accessibility attributes use page ids to link separate dom elements together. So for the combobox, it wants to have 2 ids:
* aria-owns and aria-controls both need an id to identify the listbox the combobox is controlling
* aria-activedescendant needs an id of the selected dropdown menu item

The first solution that comes to mind for me is to allow an id prop, with a default prop of some ComponentName + incrementingCounter. Thoughts?

**sub dropdowns**
Whenever support gets added for sub menus, the flag aria-haspopup=true needs to be set, thoughts on where to put that so it isn't forgotten?

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -18,6 +18,7 @@ import {
 } from '../../lib'
 import Icon from '../../elements/Icon'
 import Label from '../../elements/Label'
+import Menu from '../../collections/Menu'
 
 import DropdownDivider from './DropdownDivider'
 import DropdownItem from './DropdownItem'
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
     return _.findIndex(options, ['value', value])
   }
 
+  getDropdownAriaOptions = (ElementType) => {
+    const { loading, disabled, search, multiple } = this.props
+    const { open } = this.state
+    const ariaOptions = {
+      role: 'listbox',
+      'aria-busy': loading,
+      'aria-disabled': disabled,
+      'aria-expanded': !!open,
+    }
+    if (search) {
+      ariaOptions.role = 'combobox'
+    } else if (ElementType === Menu.Item) {
+      ariaOptions.role = 'menuitem'
+    } else {
+      ariaOptions['aria-multiselectable'] = multiple
+    }
+    return ariaOptions
+  }
+
+  getDropdownMenuAriaOptions() {
+    const { search, multiple } = this.props
+    const ariaOptions = {}
+
+    if (search) {
+      ariaOptions['aria-multiselectable'] = multiple
+      ariaOptions.role = 'listbox'
+    } else if (getElementType(Dropdown, this.props) === Menu.Item) {
+      ariaOptions.role = 'menu'
+    }
+    return ariaOptions
+  }
+
   // ----------------------------------------
   // Setters
   // ----------------------------------------
@@ -1058,15 +1091,10 @@ export default class Dropdown extends Component {
   }
 
   renderMenu = () => {
-    const { children, header, search, multiple } = this.props
+    const { children, header } = this.props
     const { open } = this.state
     const menuClasses = open ? 'visible' : ''
-    const ariaOptions = {}
-
-    if (search) {
-      ariaOptions['aria-multiselectable'] = multiple
-      ariaOptions.role = 'listbox'
-    }
+    const ariaOptions = this.getDropdownMenuAriaOptions()
 
     // single menu child
     if (children) {
@@ -1147,17 +1175,7 @@ export default class Dropdown extends Component {
     )
     const rest = getUnhandledProps(Dropdown, this.props)
     const ElementType = getElementType(Dropdown, this.props)
-    const ariaOptions = {
-      role: 'listbox',
-      'aria-busy': loading,
-      'aria-disabled': disabled,
-      'aria-expanded': !!open,
-    }
-    if (search) {
-      ariaOptions.role = 'combobox'
-    } else {
-      ariaOptions['aria-multiselectable'] = multiple
-    }
+    const ariaOptions = this.getDropdownAriaOptions(ElementType, this.props)
 
     return (
       <ElementType
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -981,6 +981,8 @@ export default class Dropdown extends Component {
     return (
       <input
         value={searchQuery}
+        type='text'
+        aria-autocomplete='list'
         onChange={this.handleSearchChange}
         className='search'
         name={[name, 'search'].join('-')}
@@ -1056,20 +1058,26 @@ export default class Dropdown extends Component {
   }
 
   renderMenu = () => {
-    const { children, header } = this.props
+    const { children, header, search, multiple } = this.props
     const { open } = this.state
     const menuClasses = open ? 'visible' : ''
+    const ariaOptions = {}
+
+    if (search) {
+      ariaOptions['aria-multiselectable'] = multiple
+      ariaOptions.role = 'listbox'
+    }
 
     // single menu child
     if (children) {
       const menuChild = Children.only(children)
       const className = cx(menuClasses, menuChild.props.className)
 
-      return cloneElement(menuChild, { className })
+      return cloneElement(menuChild, { className, ...ariaOptions })
     }
 
     return (
-      <DropdownMenu className={menuClasses}>
+      <DropdownMenu {...ariaOptions} className={menuClasses}>
         {createShorthand(DropdownHeader, val => ({ content: val }), header)}
         {this.renderOptions()}
       </DropdownMenu>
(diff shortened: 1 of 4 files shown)


levithomason:
This one will require a little more focus than I have ATM, marked it for review for later.

fracmak:
no rush, the dropdown is a really complicated module. I've gone through like 20 iterations of my implementation. I think I'm at a stopping point, I've been updating the description of this ticket as I refine my approach.

levithomason:
I've fixed the outage with CircleCI, rebased this branch, and force pushed to trigger a new build.

src/modules/Dropdown/Dropdown.js
```diff
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
...
+    const ariaOptions = {
+      role: 'listbox',
+      'aria-busy': loading,
+      'aria-disabled': disabled,
+      'aria-expanded': !!open,
+    }
+    if (search) {
+      ariaOptions.role = 'combobox'
+    } else if (ElementType === Menu.Item) {
+      ariaOptions.role = 'menuitem'
```
levithomason: This role will be added automatically by the Menu.Item's aria props.  When augmenting components with the `as` prop, you essentially merging components.

I don't think the Dropdown should be aware of other components in this way.  Let's pull this one.

src/modules/Dropdown/Dropdown.js
```diff
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
...
+    const { loading, disabled, search, multiple } = this.props
+    const { open } = this.state
+    const ariaOptions = {
+      role: 'listbox',
+      'aria-busy': loading,
+      'aria-disabled': disabled,
+      'aria-expanded': !!open,
+    }
+    if (search) {
+      ariaOptions.role = 'combobox'
```
levithomason: Can we add this up where the object is being created on line 763?

```jsx
role: search ? 'combobox' : 'listbox',
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
...
+    } else if (ElementType === Menu.Item) {
+      ariaOptions.role = 'menuitem'
+    } else {
+      ariaOptions['aria-multiselectable'] = multiple
```
levithomason, line 768: Also would like to have this up top with the object, just so all the aria props are in one location.  Note that a multi-select dropdown may or may not be searchable.  I suspect the aria prop for multiple should be added in both cases.

It could be that this should only be applied when the Dropdown is a `select` Dropdown?  The difference being that the normal Dropdown is the "menu only style" and the `select` prop creates behavior like a `<select />` element.

LMK

src/modules/Dropdown/Dropdown.js
```diff
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
...
+    }
+    return ariaOptions
+  }
+
+  getDropdownMenuAriaOptions() {
+    const { search, multiple } = this.props
+    const ariaOptions = {}
+
+    if (search) {
```
levithomason, line 777: Same note here, this might need to be `select` so that we're differentiating between a "menu only" style dropdown and a `<select />` element style Dropdown.

http://react.semantic-ui.com/modules/dropdown#selection
^ Not searchable, but is a "select".

http://react.semantic-ui.com/modules/dropdown#search-selection
^ Exactly the same as a "select" Dropdown above, but you can search for items.

http://react.semantic-ui.com/modules/dropdown#dropdown
^ Regular dropdown like a menu at the top of an desktop application.

src/modules/Dropdown/Dropdown.js
```diff
@@ -755,6 +756,38 @@ export default class Dropdown extends Component {
...
+      ariaOptions['aria-multiselectable'] = multiple
+      ariaOptions.role = 'listbox'
+    } else if (getElementType(Dropdown, this.props) === Menu.Item) {
+      ariaOptions.role = 'menu'
```
levithomason: Hm, I see the necessity for checking if the Dropdown is a MenuItem now since this is a child and there is no other way to get this role added with the current components.

I'll have to think about this one.  This component is doing so much.  We've considered ways to break it up and also make the trigger/menu separate components.  We also have a need to add an explicit prop for the "as menu item" and "as button" use cases.  Either of these paths would make this aria prop addition cleaner as well.

levithomason (changes requested):
I've left some initial feedback.  This one might be longer running.  I'm seeing more indications here that we should probably stop adding code to this component and separate its concerns sooner than later.

I'm thinking we need to completely separate the `<Select />` and `<Dropdown />` components and create some abstract component that powers these two, handling the commonalities (keyboard nav, etc) in one place.

Not sure where this will go from here just yet.

src/modules/Dropdown/Dropdown.js
fracmak: Ya, unfortunately this aria option is only valid for listbox (and grid, tablist, tree)

https://www.w3.org/TR/wai-aria/states_and_properties#aria-multiselectable

It's not technically valid to put it on a combobox, even if we set it to false.

src/modules/Dropdown/Dropdown.js
fracmak: Ok, I see what you're saying about using 'selection' instead of 'search'.

src/modules/Dropdown/Dropdown.js
fracmak: I do feel like splitting this up at least internally between the different types of dropdowns would help in keeping all these aria roles straight.

src/modules/Dropdown/Dropdown.js
fracmak: So I remember why I did search instead of selection, this needs to match the "search" if statement on the Dropdown component, because the search prop switches the Dropdown to be a combobox role, which means DropdownMenu needs to be a listbox role. With selection, we're covered because Dropdown becomes a listbox and DropdownMenu isn't needing any additional markup.

fracmak:
So I was looking into how to do the Menu component and I feel that normally it's a `listbox/option`, unless it has a Dropdown as a Menu.Item AND is fixed top/bottom or attached top/bottom, at which point the entire menu becomes a `menu/menuitem`. So is it kosher to peer into your children to check their types and see if there's a Dropdown in there? Would love some ideas on how to detect that.

levithomason:
Unfortunately, I know of no good way to check your children.  This is problematic in React.  Perhaps we just wrap this up where it is for now and come back later if/when we split things up and refactor at that time?

If you agree on this, let's just pull the `.skip()` from the last commit.  If that test will not be applicable, let's comment it out with a TODO.  There is a lint rule that prevents only/skip in tests that should have actually failed this build.

fracmak:
I've removed all references of Menu from the Dropdown component, I also added in an aria-hidden to the hidden select menu to squash some errors I'm seeing in my project. This should be good to go for now

fracmak:
While I say this is good to go, there's still an outstanding issue regarding aria-activedescendant and aria-owns/controls for the search/combobox but this is a good leap in the right direction

levithomason (approved):

levithomason:
Great, thanks for the work.  The attention to detail and correctness is much appreciated.  We'll get there with small steps :)

levithomason:
Released in `semantic-ui-react@0.63.2`.

PR #1007, opened 2016-12-08
docs(StepExampleGroups): remove object destructuring

tomjnsn:
Although this isn't a bug and isn't necessary, I'd like to propose reworking the Step examples that use object destructuring. All of the other examples that I've seen in the docs that have subcomponents don't use destructuring. Because of that, it is easy to get mixed up on the destructuring when looking at tags without the base component name in there. It looks like maybe there's a Group tag, etc. Once it is understood then developers can do the destructuring and be aware, but this creates a sort of lowest common denominator for all styles of programming and keeps it consistent with what I've observed in the rest of the examples.

If this is agreeable I can go through and change the other Step examples that have the same destructuring in them.

### DIFF
--- a/docs/app/Examples/elements/Step/Groups/StepExampleGroups.js
+++ b/docs/app/Examples/elements/Step/Groups/StepExampleGroups.js
@@ -1,7 +1,6 @@
 import React from 'react'
 import { Icon, Step } from 'semantic-ui-react'
 
-const { Content, Description, Group, Title } = Step
 const steps = [
   { icon: 'truck', title: 'Shipping', description: 'Choose your shipping options' },
   { active: true, icon: 'payment', title: 'Billing', description: 'Enter billing information' },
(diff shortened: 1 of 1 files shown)


levithomason:
Indeed, this would be fantastic.  I meant to double back and standardize all examples to use sub-components without destructuring.  I'd gladly merge as many of these as you are able to put up!

tomjnsn:
Cool, I'll do the rest of the Step examples that need it. Anywhere else you can think of that should have the destructuring pulled out? I can do those as well.

tomjnsn:
Oh, and do you prefer that each are an individual pull request so your naming stays consistent or can I just do something like docs(StepExamples*): remove object destructuring as one pull request?

levithomason:
Not sure if there are others off the top of my head, would have to do a search.

Separate PRs is great.  Reviews are easier to manage that way.  Also, if there are changes to one, it doesn't hold up the others.

tomjnsn:
I'll remember to do separate PRs next time.

PR #1008, opened 2016-12-08
docs(Step): remove object destructuring for Step subcomponents

tomjnsn:
This is as discussed previously on another pull request. Removing the ES6 destructuring of the Step subcomponents (Group, Title, etc.) to make the examples consistent with other examples in the docs.

Searched for any other subcomponent destructuring in the examples but didn't find any so looks like these were the only ones.


levithomason:
Awesome, thanks much!

PR #1031, opened 2016-12-15
feat(docs): Active className on Sidebar item

pradel:
Added an active style when the menu is active on left Sidebar.

kapture 2016-12-15 at 16 03 47


pradel:
Travis failed due to npm install error.

levithomason:
Thanks, rebuilding.

PR #1035, opened 2016-12-15
fix(Button): Disabled Button with Label leaves Label enabled

### LINKED ISSUE #1026: Disabled Button with Label leaves Label enabled
### Steps
1. Create a disabled Button with label.

### Expected Result
Both the button and the label should be disabled.

### Actual Result
Only the button is disabled, the label is not.

This is because the `disabled` prop is passed to the `button` child.  It should instead be used to create a `disabled` className for the wrapping `div`.

### Version
v0.62.1

### Testcase
http://codepen.io/levithomason/pen/pNxwMb

luski:
fixes #1026

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -260,7 +260,7 @@ class Button extends Component {
       return (
         <ElementType {...rest} className={containerClasses} onClick={this.handleClick}>
           {labelPosition === 'left' && labelElement}
-          <button className={classes}>
+          <button className={classes} tabIndex={tabIndex}>
             {Icon.create(icon)} {content}
           </button>
           {(labelPosition === 'right' || !labelPosition) && labelElement}
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -256,7 +256,7 @@ class Button extends Component {
     if (labelElement) {
       let classes = cx('ui', baseClasses, 'button', className)
       if (className && className.indexOf('disabled') !== -1) {
-        classes = cxd(classes, {disabled: false})
+        classes = cxd(classes, { disabled: false })
       }
       const containerClasses = cx('ui', labeledClasses, 'button', className, wrapperClasses)
       debug('render label:', { classes, containerClasses }, this.props)
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -1,4 +1,5 @@
 import cx from 'classnames'
+import cxd from 'classnames/dedupe'
 import _ from 'lodash'
 import React, { Component, PropTypes } from 'react'
 
@@ -218,7 +219,6 @@ class Button extends Component {
       useKeyOnly(basic, 'basic'),
       useKeyOnly(circular, 'circular'),
       useKeyOnly(compact, 'compact'),
-      useKeyOnly(disabled, 'disabled'),
       useKeyOnly(fluid, 'fluid'),
       useKeyOnly(icon === true || icon && (labelPosition || !children && !content), 'icon'),
       useKeyOnly(inverted, 'inverted'),
@@ -232,6 +232,7 @@ class Button extends Component {
       useKeyOrValueAndKey(attached, 'attached'),
       useValueAndKey(floated, 'floated'),
     )
+    const wrapperClasses = cx(useKeyOnly(disabled, 'disabled'))
     const rest = getUnhandledProps(Button, this.props)
     const ElementType = getElementType(Button, this.props, () => {
       if (!_.isNil(label) || !_.isNil(attached)) return 'div'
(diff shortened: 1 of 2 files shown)


levithomason:
Thanks for the PR!  The `disabled` class makes an element partially transparent (faded).  We need to only apply the disabled class to the container and not the `button` also when there is a `label`.  Otherwise, the button gets duplicate opacity rules making it extra transparent.

Note the button opacity in this animation as the disabled class is added and removed.  The container is also already disabled:

luski:
It means that in the example http://codepen.io/levithomason/pen/pNxwMb given in https://github.com/Semantic-Org/Semantic-UI-React/issues/1026 both cases are wrong, but in a different way. The first case is wrong because a label is not disabled and the second because a button is "double disabled". I fixed both cases in the commit below

levithomason:
Thanks for the update.  Would appreciate some tests for the logic surrounding the `disabled` handling.  Here's the lint issue failing CI:

```sh
/home/ubuntu/Semantic-UI-React/src/elements/Button/Button.js
  256:32  error    A space is required after '{'             object-curly-spacing
  256:48  error    A space is required before '}'            object-curly-spacing
```

test/specs/elements/Button/Button-test.js
```diff
@@ -116,6 +116,16 @@ describe('Button', () => {
       shallow(<Button label='hi' />)
         .should.have.className('labeled')
     })
+    it('adds the disabled className to the root element only when the disabled attribute is set', () => {
+      const wrapper = shallow(<Button label='foo' disabled />)
+      wrapper.should.have.className('disabled')
+      wrapper.children().forEach(child => child.should.not.have.className('disabled'))
+    })
+    it('adds the disabled className to the root element only when the disabled className is set', () => {
```
levithomason: This test is covered by `common.isConformant()` which asserts that all props and classNames are passed to the root element.  We can remove this one.

test/specs/elements/Button/Button-test.js
```diff
@@ -116,6 +116,16 @@ describe('Button', () => {
       shallow(<Button label='hi' />)
         .should.have.className('labeled')
     })
+    it('adds the disabled className to the root element only when the disabled attribute is set', () => {
```
levithomason: This test is covered by `common.propKeyOnlyToClassName(Button, 'disabled')` which asserts the prop is converted into a className.  We can remove the test here.

levithomason (commented):
I did not get time to do a full review here, but I left some comments regarding the tests.  The current tests should already be covered by `common` tests.

I believe this means we'll need to find some more specific test cases to cover the added functionality here.  I'll come back later and make some suggestions.

luski:
Something like that 1ed599b? The added functionality is just making sure the container has disabled class and children not, so I limited the test case to check if label and button inside the wrapper don't have the 'disabled' class, becouse common tests check if the wrapper has 'disabled' class if I understood you

src/elements/Button/Button.js
```diff
@@ -250,8 +251,11 @@ class Button extends Component {
     }
 
     if (label) {
-      const classes = cx('ui', baseClasses, 'button', className)
-      const containerClasses = cx('ui', labeledClasses, 'button', className)
+      let classes = cx('ui', baseClasses, 'button', className)
+      if (className && className.indexOf('disabled') !== -1) {
+        classes = cxd(classes, { disabled: false })
```
levithomason: This line here is failing Codecov as it is not invoked during tests.  That said, we don't support component features through className but only props.  I believe this should be removed.  Let me know if you believe otherwise and why.

>Here is the coverage hint from the Codecov browser extension:
>
>image

levithomason (changes requested):
Updated tests look good, thanks.  I've left one comment as to why Codecov is failing this PR.

src/elements/Button/Button.js
luski: Yep, you're right. I thought the className checking must be added because there was className used in the example provided in #1026.

luski:
@levithomason Does it still need author feedback? I made changes weeks ago

levithomason:
My apologies for the slow response.  I just had a new baby :P  I'll try to get this as soon as I am able.

luski:
@levithomason Congratulations!!! Especially that my wife is pregnant as well :) No worries, no rush, I was just curious if you remember about my PR

levithomason:
I've rebased this to master and fixed conflicts.  There is one issue with the current solution.  Since it circumvents passing the `disabled` prop down, the HTML button still has a tab stop when the disabled prop is set.

luski:
Done. Sorry it took so long...

levithomason (approved):

levithomason:
No worries at all.  As a maintainer, I completely understand delays.  Thanks for seeing it through 😄 

I've retriggered the build and will merge on pass.

levithomason:

PR #1038, opened 2016-12-16
breaking(Dropdown): add options shorthand

### LINKED ISSUE #470: Dropdown: support an array of string|element options
Currently, the Dropdown takes an array of plain objects via the `options` prop.  These objects are used to create the `<Dropdown.Item />` children.  Each object is spread as props on each item:

``` js
// Dropdown.js renderItems()

return _.map(options, (opt, i) => (
  <DropdownItem
    key={`${opt.value}-${i}`}
    active={isActive(opt.value)}
    onMouseDown={e => e.preventDefault()}
    {...opt}
  />
))
```

We should also accept an array of strings or elements.  An array of strings would be used as the item text and value.  Whereas, an array of elements would be cloned.

We created `src/factories` for this exact purpose.  They handle mapping a primitive value to props, spreading props objects, and cloning elements while merging className and spreading props.  We should then use `createFactory` to create a `createDropdownItem` factory, see `src/factories/index.js`.  We would then use the `createDropdownItem` factory in place of the `_.map()` callback used above.

levithomason:
# TODO

- [ ] Fix bug with allow additions examples (key warning, no addition added)
- [x] Change SSOT for getting items to return props
- [ ] fix bug with duplicate keys (i.e. fix tests)

# Breaking Changes!

This PR updates the Dropdown to allowing defining options with any shorthand value: number, string, props object, or another element.

## `options` prop renamed to `items`

As a design decision, shorthand props are provided for each subcomponent.  Example, a `Modal` with `<Modal.Header>Foo</Modal.Header>` can be configured as `<Modal header='Foo' />.  The Dropdown `options` prop is used to create `<Dropdown.Item />`s, hence, the correct shorthand prop name is `items`.

**Before**
```jsx
const options = [{ key: 1, text: 'One', value: 1 }, ...]
<Dropdown options={options} />
```

**After**
```jsx
const items = [{ key: 1, text: 'One', value: 1 }, ...]
<Dropdown items={items} />
```

## New `search` callback signature

The `items` are used to create `<DropdownItem />`s internally.  Custom `search` functions, which used to receive the original `items` objects, now receive `<DropdownItem />` instances.  Use the instance's `props` to access the `text`, `value`, and other `props` as necessary.

**Before**

```jsx
const options = [{ key: 1, text: 'One', value: 1 }, ...]

const search = (options, query) => {
  // options = [{ key: 1, text: 'One', value: 1 }, ...]
  return options.filter(opt => opt.text === query)
}

<Dropdown search={search} options={options} />
```
**After**
```jsx
const items = [{ key: 1, text: 'One', value: 1 }, ...{}]

const search = (options, query) => {
  // items = [<Dropdown.Item key={1} text='One' value=1 />, ...]
  return items.filter(item => item.props.text === query)
}

<Dropdown search={search} items={items} />
```

## New `renderLabel` callback signature

Multiple select Dropdowns can render a custom `Label` for each active `DropdownItem`.  This callback, which used to receive a single `option` object, now receives a `<DropdownItem />` instance.  Use the instance's `props` to access the `text`, `value`, and other `props` as necessary.

**Before**

```jsx
const options = [{ key: 1, text: 'One', value: 1 }, ...]

const renderLabel = (item, index, defaultLabelProps) => {
  // item = { key: 1, text: 'One', value: 1 }
  return defaultLabelProps
}

<Dropdown renderLabel={renderLabel} options={options} />
```
**After**
```jsx
const items = [{ key: 1, text: 'One', value: 1 }, ...]

const renderLabel = (item, index, defaultLabelProps) => {
  // item = <Dropdown.Item key={1} text='One' value={1} />
  return defaultLabelProps
}

<Dropdown renderLabel={renderLabel} items={items} />
```

***

Fixes #470 

This adds shorthand factory support to the Dropdown `options`.


src/modules/Dropdown/Dropdown.js
```diff
@@ -230,11 +230,8 @@ export default class Dropdown extends Component {
     /** Controls whether or not the dropdown menu is displayed. */
     open: PropTypes.bool,
 
-    /** Array of Dropdown.Item props e.g. `{ text: '', value: '' }` */
-    options: customPropTypes.every([
-      customPropTypes.disallow(['children']),
-      PropTypes.arrayOf(PropTypes.shape(DropdownItem.propTypes)),
-    ]),
+    /** Array of Dropdown.Item shorthand */
+    options: customPropTypes.itemShorthand,
```
jeffcarbs: I think this should be `collectionShorthand`

jeffcarbs:
This is going to affect methods like: 

```js
getItemByValue = (value) => {
  const { options } = this.props

  return _.find(options, { value })
}
```

I kinda think we're gonna need to:
- update anywhere that's using `this.props.options` to use `this.getMenuOptions()`
- update `getMenuOptions` to return an array of `DropdownItem`
- update anything using `getMenuOptions` to use `option.props` instead of just `option`

Passing it through the factory first would allow us to predictably access a prop (text, value, whatever) of an item.

Also if we do it this way, maybe update `getMenuOptions` to `getDropdownItems` to be more clear as to what it's doing.

levithomason:
Thanks for the extra eyes. Indeed, I'll be sure to make these changes prior to merging.  This PR is half baked atm.

UnbrandedTech:
I'm not sure this is the place, but react-widgets has a nice ish API allowing options to be an array of objects and two optional props. One to specify the display value and another for the value to be selected.

```
Const people = [{id:1, name:"James"},{id:2, name:"Bob"}]
<Dropdown valueField="id" textField="name" data={people} \>
```

levithomason:
@UnbrandedTech use `content` for the display value in the menu and `text` for the active value. 

http://react.semantic-ui.com/modules/dropdown#item-content

levithomason:
Closing this PR as it has gotten too stale.  Also, I'm not sure this work is worth it.  The gains are minor and the repercussions are major.

PR #1039, opened 2016-12-16
docs(layouts): add layout examples

### LINKED ISSUE #836: Basic semantic-ui examples
It would be very nice to have the basic sui examples ( http://semantic-ui.com/usage/layout.html ) using these components. It would be a really big help (even without any business logic).

iamcaleberic:
## TODO

- [x] Theming
- [x] Responsive
- [x] Grid
- [x] Attached Content
- [x] Bootstrap Migration
- [x] Homepage
- [x] Sticky
- [x] Fixed Menu
- [x] Login Form

Fixes #836


iamcaleberic:
@levithomason Would like your thoughts on current implementation before i continue with the porting 
#### issues currently having
 - Routing i.e in currently main routes file ? Do I need to change this for the layout examples
 - Showing source for layout examples

levithomason:
Here's what I'm thinking.

### Structure

We put layout examples in `/docs/app/layouts` to keep them separate from the component examples.  Then, we make the routes dynamic by creating a Webpack context of this directory.

I'm thinking the layouts need to render as a full page view without the sidebar in order to be effective.  I'll need to rework some components to make this happen.

### Showing source

Again, I think this needs to be done differently that in the current examples.  What if we showed a circular floating button in one of the corners of the every layout, when clicked, we could show the source.  Perhaps we would show source in a modal over the layout itself, or, we could toggle the entire page from the rendered view to the source editor and back.

I think the primary concern here is that the source will be very large so we'll likely need to take up the entire page, or at least most of it.

***

If you can move the layout to be in the directory structure proposed above, I'll come back to this PR and update it to handle the routes and showing source for layouts as soon as I can.

iamcaleberic:
- I agree on the layouts being full page view
- I think we can toggle between the view and the source editor for showing source

levithomason:
Great!  If you want to take a stab at any of that, please do.  Otherwise, I'll get to this as soon as I am able.

docs/app/Views/Layouts.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react'
+import { Header, Segment, Container, Divider, Card } from 'semantic-ui-react'
+
+const LayoutsExample = () => (
+  <Container as='div'>
```
levithomason: Heads up, the default component class is `div` so we can remove all `as='div'` props since it is redundant.

docs/app/Views/Layouts.js
```diff
@@ -0,0 +1,88 @@
...
+    <Segment masthead vertical>
```
levithomason: There is no `masthead` prop, let's remove that.

docs/app/Views/Layouts.js
```diff
@@ -0,0 +1,88 @@
...
+      <Container as='div'>
+        <Header as='h1'> Layouts
+          <div className='sub header'>Examples of using Semantic UI React</div>
+        </Header>
+        <Divider hidden />
+      </Container>
+    </Segment>
+    <Container as='div' className='main intro'>
```
levithomason: Let's also remove `className='main intro'`, we shouldn't have any extra classNames on the layouts as we have no additional styles to apply.  If the layouts require additional custom styling on top of the default theme, then we'll have to reconsider how we go about that.

levithomason (changes requested):

docs/app/Views/Layouts.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react'
+import { Header, Segment, Container, Divider, Card } from 'semantic-ui-react'
+
+const LayoutsExample = () => (
+  <Container as='div'>
+    <Segment masthead vertical>
+      <Container as='div'>
+        <Header as='h1'> Layouts
+          <div className='sub header'>Examples of using Semantic UI React</div>
```
levithomason: Here we can use `<Header.Subheader />` instead of `className`.

levithomason:
I'm adding route support for the view ATM, I'll address some of the other issues as well.

levithomason:
OK, I've stubbed out the routes and layout files themselves.  I also included a back button automatically for the layout views.  You should be good to simply update the layout files.  Let me know if you run into any issues 👍

levithomason (approved):

iamcaleberic:
@levithomason thanks alot

levithomason:
No problem, ping me when the other layouts are in and I'll make sure to get the full page source and editor going.

iamcaleberic:
@levithomason I was updating the stream from master but alot has changed quite significantly and I have alot of conflicts.
Tried fixing some but am a bit stuck now. Please assist.

levithomason:
Yep, I'll label this issue for assistance and @layershifter or myself will get to it soon.

layershifter:
I've updated to latest `master`, updated routes and performed some fixes. LGTM.
I will review existing layout examples.

layershifter:
I've updated the start comment with TODO, working on other examples.

iamcaleberic:
@layershifter thank you for fixing the issues. I updated the checklist I had four layout examples left minus the theming layout you working on and some styling for the existing ones.

layershifter:
@iamcaleberic Thanks for the follow up. I updated checklist again, I saw that we have some existing examples, however I mark only that I checked.

iamcaleberic:
@layershifter managed to get all the layouts up but without styling

layershifter (commented):
Updated all examples, fixed styling and merge conflicts. @levithomason we're ready for initial review there

BTW, @iamcaleberic thanks for your work there!

iamcaleberic:
@layershifter thanks for the assist in completing this....

levithomason:
Thank you @iamcaleberic and @layershifter for pushing this one over the finish line!

levithomason:

jjwilliams42:
This is awesome! Thank you so much for this contribution, I've been looking for full blown layout examples. I was wondering, I noticed the "Home Page" example on regular SUI has a responsive hamburger menu, does SUI React not support this?

snekse:
Yeah, I think @jackjwilliams found a bug. 

https://semantic-ui.com/examples/homepage.html
vs.
https://react.semantic-ui.com/layouts/homepage

PR #1040, opened 2016-12-16
fix(Header): Remove duplicate class

jeffcarbs:
Using the `subheader` shorthand would result in the `sub header` classes being applied twice:
```
<Header subheader='hey' />

<div class="sub header sub header">hey</div>
```

This fixes that.


levithomason:
Oh, of course!  Looks like some related tests need updated.

jeffcarbs:
Fixed those specs, should be g2g 👍

levithomason:
Released in `semantic-ui-react@0.62.2`

PR #1043, opened 2016-12-16
feat(Checkbox): Indeterminate state

jeffcarbs:
Adds support for `indeterminate` state for checkbox. See usage in SUI-core here: http://semantic-ui.com/modules/checkbox.html#grouped-checkboxes

screen shot 2016-12-16 at 3 52 22 pm

Initially I thought to just implement this as a plain `indeterminate` prop. However, that would prevent using `indeterminate` with uncontrolled usage since you'd have to maintain the `indeterminate` state and optionally pass the prop.

Instead, this makes `indeterminate` an auto-controlled prop, so to default a toggle to `indeterminate` you pass `defaultIndeterminate` or you could control the usage by passing the `indeterminate` prop.

This follows the standard usage pattern of `indeterminate` in HTML checkboxes (as well as SUI core): you can't get into the indeterminate state by user interaction, only via javascript. If a checkbox starts as indeterminate, once clicked the indeterminate state will be set to false for the rest of its lifecycle and it will function as a normal checkbox.

### DIFF
--- a/docs/app/Examples/modules/Checkbox/States/CheckboxExampleIndeterminate.js
+++ b/docs/app/Examples/modules/Checkbox/States/CheckboxExampleIndeterminate.js
@@ -1,0 +1,8 @@
+import React from 'react'
+import { Checkbox } from 'semantic-ui-react'
+
+const CheckboxExampleIndeterminate = () => (
+  <Checkbox label='This checkbox is indeterminate' defaultIndeterminate />
+)
+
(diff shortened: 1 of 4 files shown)


levithomason:
Made a minor fix to lifecycle method signatures and order.  Crazy there is no `indeterminate` HTML attr for this!  Will merge on pass.

levithomason:
Released in `semantic-ui-react@0.62.3`.

PR #1044, opened 2016-12-16
fix(typings): fix typings spelling on `Accordion`

ericl85:
Typescript typings were spelled `Accordian` instead of `Accordion`.


levithomason:
Thank you!

levithomason:
Released in `semantic-ui-react@0.62.3`.

PR #1053, opened 2016-12-19
feat(typings): Added sidebar, Update to Search and Textarea

koenvg:
Added Sidebar typings, Fixed callback onChange of Textarea and added onResultSelect prop to Search.

Sidenote: Why does Textarea have such a different callback than everything else?


levithomason:
Thanks much!

PR #1062, opened 2016-12-20
breaking(shorthand): only generate keys from strings and numbers

### LINKED ISSUE #1057: fix(Table): Cannot convert a Symbol value to a string
### Steps
1. Create table with shorthand and use a custom element

### Expected Result
Should render Table

### Actual Result
Errors with `Cannot convert a Symbol value to a string`

### Version
0.62.3

### Testcase
http://codepen.io/UnbrandedTech/pen/mOvGxj?editors=0010

levithomason:
## Breaking Changes!

When passing an array of props objects or an array of elements, such as `items` to a Menu, you must now define a [React `key`](https://facebook.github.io/react/docs/lists-and-keys.html#keys).

### Before

SUIR attempted to autogenerate keys for objects and elements.

```jsx
const items = [
  { name: 'Home' },
  <Menu.Item name='Search' />,
]

<Menu items={items} />
```

### After

Note, any unique unchanging `key` will work.

```jsx
const items = [
  { key: 'home', name: 'Home' },
  <Menu.Item key='search' name='Search' />,
]

<Menu items={items} />
```

### Unchanged

Shorthand strings and numbers continue to be used as the element's `key`.  This means shorthand string/number values must still be unique in a list.  If they are not, use an object or element and define a unique `key`.

***

Fixes #1057

### New Key Generation

This PR simplifies auto child key generation.  We now only generate keys for shorthand string and number primitive values.  If the user passes a props object or element, they must include a `key`.

This is faster, simpler, and was the original intent of auto key generation (handling elements from strings/numbers).

### Styles are now merged

Just as className is merged when using shorthand object or elements, style objects are now merged.

### DIFF
--- a/docs/app/Examples/collections/Breadcrumb/Types/BreadcrumbExampleProps.js
+++ b/docs/app/Examples/collections/Breadcrumb/Types/BreadcrumbExampleProps.js
@@ -2,9 +2,9 @@ import React from 'react'
 import { Breadcrumb } from 'semantic-ui-react'
 
 const sections = [
-  { content: 'Home', link: true },
-  { content: 'Store', link: true },
-  { content: 'T-Shirt', active: true },
+  { key: 'Home', content: 'Home', link: true },
+  { key: 'Store', content: 'Store', link: true },
+  { key: 'Shirt', content: 'T-Shirt', active: true },
 ]
 
 const BreadcrumbExampleProps = () => (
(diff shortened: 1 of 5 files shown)


jeffcarbs:
@levithomason - what's the status of this PR?

This fixes a bug introduced in `0.62.2` and there are things merged into later versions that we'd like to use but can't without refactoring much of our shorthand-related code.

levithomason:
@jcarbo the updated key generation and style object merging is done. The docs and view tests can be done in separate PRs. I'll be busy through the holiday weekend. If you'd like to cleanup and get tests passing then we can merge this. Otherwise, I hope put time towards this next week.

levithomason:
Oh, this constitutes a breaking change as shorthand objects and elements no longer generate keys. We'll need to update the description and title to reflect this.

levithomason:
@jcarbo Heads up, this is going out on the next minor bump (tonight).  I know you were waiting for it.

levithomason:

PR #1066, opened 2016-12-20
feat(AutoControlledComponent): Default values

### LINKED ISSUE #764: Support defaultProps in AutoControlledComponent
This was fixed for the Accordion but likely exists in other auto controlled components. 

From #763:

The `defaultActiveIndex` is overridden in the Accordion on `componentWillMount`.  This is an edge case, but we need to update the `AutoControlledComponent` to defer to default props when calling `trySetState` in `componentWillMount`.

The real solve here is likely adding `static defaultAutoControlledProps`.  Currently, the `AutoControlledComponent` tells devs to use `trySetState` in `componentWillMount` instead of setting defaults for auto controlled props.  This is to prevent the actual defaults from always winning, however, if we had our own auto controlled defaults then we could apply them correctly.

jeffcarbs:
Fixes #764

I noticed a bug using `AutoControlledComponent` where initial state for non-AutoControlled props was getting clobbered in the `componentWillMount` of the ACC. This is because ACC uses `this.state = ...` instead of `this.setState` to set the state.

Example of the bug:
```js
class MyComponent extends AutoControlledComponent {
  state = { someField: 'default' }

  render () => <div />
}

// Actual
shallow(<MyComponent />).state() === {}

// Expected
shallow(<MyComponent />).state() === { someField: 'default' }
```

In fixing that bug, I came across the #764 issue which I realized could be solved simultaneously. The initial state is set within the constructor which means it's available in the ACC's `componentWillMount`.  I updated the `componentWillMount` to use the initial state if the value of the prop was not defined via props.

For example:
```js
class MyComponent extends AutoControlledComponent {
  state = { someField: 'default' }

  static autoControlledProps = ['someField']

  render () => <div />
}

shallow(<MyComponent />).state() === { someField: 'default' }
shallow(<MyComponent defaultSomeField='propDefault' />).state() === { someField: 'propDefault' }
shallow(<MyComponent someField='prop' />).state() === { someField: 'prop' }
```

This seems pretty intuitive to me. The props are being passed directly to state, so to specify a "global default" outside of props, you set the state yourself in the constructor (or use the es7 property initializer, which is basically just sugar on the same thing).

----

_Edit_ As an aside, I've been using `AutoControlledComponent` as a base for some of my own components via:
```js
import AutoControlledComponent from 'semantic-ui-react/dist/commonjs/lib/AutoControlledComponent'
```

I think it's a really cool abstraction and lets you write super flexible/reusable components that either do all the work (control the state) or let the user do it from higher up.

@levithomason - Thoughts on open-sourcing it as a separate package? Its only dependency is `lodash` (which you could probably even drop) so would probably be pretty simple.

### DIFF
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -52,10 +52,11 @@ export const getAutoControlledStateValue = (propName, props, state, includeDefau
     const defaultProp = props[getDefaultPropName(propName)]
     if (defaultProp !== undefined) return defaultProp
 
-    // initial state
-    // check for an object as React initializes state as `null` if it is `undefined`
-    const initialState = ({}).toString.call(state) === '[object Object]' && state[propName]
-    if (initialState) return initialState
+    // initial state - state may be null or undefined
+    if (state) {
+      const initialState = state[propName]
+      if (initialState !== undefined) return initialState
+    }
   }
 
   // React doesn't allow changing from uncontrolled to controlled components,
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -91,11 +91,16 @@ export default class AutoControlledComponent extends Component {
       // It is not reasonable to decipher the difference between props from the parent and defaultProps.
       // Allowing defaultProps results in trySetState always deferring to the defaultProp value.
       // Auto controlled props also listed in defaultProps can never be updated.
+      //
+      // To set defaults for an AutoControlled prop, you can set the initial state in the
+      // constructor or by using an ES7 property initializer:
+      // https://babeljs.io/blog/2015/06/07/react-on-es6-plus#property-initializers
       const illegalDefaults = _.intersection(autoControlledProps, _.keys(defaultProps))
       if (!_.isEmpty(illegalDefaults)) {
         console.error([
-          'Do not set defaultProps for autoControlledProps,',
-          'use trySetState() in constructor() or componentWillMount() instead.',
+          'Do not set defaultProps for autoControlledProps. You can set defaults by',
+          'setting state in the constructor or using an ES7 property initializer',
+          '(https://babeljs.io/blog/2015/06/07/react-on-es6-plus#property-initializers)',
           `See ${name} props: "${illegalDefaults}".`,
         ].join(' '))
       }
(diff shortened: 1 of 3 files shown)


levithomason:
I think this sounds sane, using the initial state as the "default" AC prop values.  Another way of thinking about this is that it is the "default state", since, the props are copied to state immediately.

I'm gonna update this PR to undo the Accordion hacks as well just to ensure it is working correctly.

src/lib/AutoControlledComponent.js
```diff
@@ -30,27 +30,33 @@ const getDefaultPropName = (prop) => `default${prop[0].toUpperCase() + prop.slic
...
-  const prop = props[propName]
-  const defaultProp = props[defaultPropName]
-
-  const hasProp = prop !== undefined
-  const hasDefaultProp = defaultProp !== undefined
-
-  // defaultProps & props
-  if (includeDefaultProps && !hasProp && hasDefaultProp) return defaultProp
-  if (hasProp) return prop
+export const getAutoControlledStateValue = (propName, props, state, includeDefaults = false) => {
```
levithomason, line 45: Updated this function to handle initial state.  Since this is the SSOT when it comes to how we derive AC state I think it makes sense here.  It also prevents an extra loop of state in CWM.

test/specs/lib/AutoControlledComponent-test.js
```diff
@@ -32,6 +32,11 @@ describe('extending AutoControlledComponent', () => {
     TestClass = createTestClass({ autoControlledProps: [], state: {} })
   })
 
+  it('does not throw with a `null` state', () => {
```
levithomason, line 35: Extra test to catch the edge case I ran into.  If you have `state = undefined` React will initialize state as `null`.  This then throws as we look for keys on `null`.  I added a comment about this and a check for it in `getAutoControlledStateValue`.

levithomason (commented):
I updated the Accordion to use the new initial state pattern in the constructor.  I consolidated the "default state" logic into the `getAutoControlledStateValue` function as well.  Finally, I added another test to cover an edge case.

LMK if anything looks out of place.

src/lib/AutoControlledComponent.js
```diff
@@ -30,27 +30,33 @@ const getDefaultPropName = (prop) => `default${prop[0].toUpperCase() + prop.slic
...
+  if (propValue !== undefined) return propValue
+
+  if (includeDefaults) {
+    // defaultProps
+    const defaultProp = props[getDefaultPropName(propName)]
+    if (defaultProp !== undefined) return defaultProp
+
+    // initial state
+    // check for an object as React initializes state as `null` if it is `undefined`
+    const initialState = ({}).toString.call(state) === '[object Object]' && state[propName]
```
jeffcarbs: I'd maybe do something like:
```js
if (state) {
  const initialState = state[propName]
  if (initialState !== undefined) return initialState
}
```

- `state` will only every be: an object, null, or undefined. I think this is a more straightforward check than the `toString` check.
- `if (initialState) return initialState` wouldn't let `null` through, which should be a valid default value.
- It makes it the same style as the `propValue` and `defaultProp` returns.

jeffcarbs:
Left a comment but otherwise I think this is g2g 👍

levithomason:
I try to avoid extra nesting / branch logic when possible.  Perhaps went too far in that case, updated :)

levithomason:
Released in `semantic-ui-react@0.63.1`.

PR #1071, opened 2016-12-21
fix(Dropdown): prevent overzealous onAddItem calls

### LINKED ISSUE #1064: Dropdown onAddItem gets called on value change when text and value aren't equal
### Steps
1. Create a Select Dropdown with AllowAdditions enabled (and search so you can add them)
2. Set the options so that the text and the values are not equal (for me I had unique id's for values).
3. Add an onAddItem function
4. Whenever the values of the dropdown are changed it calls onAddItem whether or not a new item has changed.

### Expected Result
The onAddItem should only be called when a new item is added not every time the value changes.

### Actual Result
onAddItem is called every time the selected item changes.

### Version
0.63.2

### Testcase
http://codepen.io/anon/pen/XNGNpa

Thanks :)

keeslinp:
Fixes #1064 

The values were being compared against the text to determine whether a new item had been added and so whenever the two weren't the same it gave a false positive and called the function at the wrong time. I switched it to how it should be.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -522,7 +522,7 @@ export default class Dropdown extends Component {
     if (!value || !open) return
 
     // notify the onAddItem prop if this is a new value
-    if (onAddItem && !_.some(options, { text: value })) {
+    if (onAddItem && !_.some(options, { value })) {
       onAddItem(e, { ...this.props, value })
     }
 
@@ -624,7 +624,7 @@ export default class Dropdown extends Component {
     if (item.disabled) return
 
     // notify the onAddItem prop if this is a new value
-    if (onAddItem && !_.some(options, { text: value })) {
+    if (onAddItem && !_.some(options, { value })) {
       onAddItem(e, { ...this.props, value })
     }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -514,18 +514,18 @@ export default class Dropdown extends Component {
 
   selectHighlightedItem = (e) => {
     const { open } = this.state
-    const { multiple, onAddItem, options } = this.props
-    const value = _.get(this.getSelectedItem(), 'value')
+    const { multiple, onAddItem } = this.props
+    const item = this.getSelectedItem()
+    const value = _.get(item, 'value')
 
     // prevent selecting null if there was no selected item value
     // prevent selecting duplicate items when the dropdown is closed
     if (!value || !open) return
 
     // notify the onAddItem prop if this is a new value
-    if (onAddItem && !_.some(options, { value })) {
+    if (onAddItem && item.additional) {
       onAddItem(e, { ...this.props, value })
     }
-
     // notify the onChange prop that the user is trying to change value
     if (multiple) {
       // state value may be undefined
@@ -611,7 +611,7 @@ export default class Dropdown extends Component {
   handleItemClick = (e, item) => {
     debug('handleItemClick()')
     debug(item)
-    const { multiple, onAddItem, options } = this.props
+    const { multiple, onAddItem } = this.props
     const { value } = item
 
     // prevent toggle() in handleClick()
@@ -624,7 +624,7 @@ export default class Dropdown extends Component {
     if (item.disabled) return
 
     // notify the onAddItem prop if this is a new value
-    if (onAddItem && !_.some(options, { value })) {
+    if (onAddItem && item.additional) {
       onAddItem(e, { ...this.props, value })
     }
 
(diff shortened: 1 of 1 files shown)


jeffcarbs:
This line would need to be updated as well: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L708. It determines whether the "new item" should show up as an option.

I'm not sure this really addresses the issue. It wouldn't work in any case where the text/value were different. Consider the following scenario:
```js
const options = [
  { text: 'John', value: 'user_1' },
  { text: 'Mary', value: 'user_2' },
  { text: 'Bob',  value: 'user_3' }
]

<Dropdown options={options} search />
```

User types `John`

Comparing against `text` would be correct here. If you compared against `value`, almost anything would be considered a new item.

The only case we can reliably handle is where `text` and `value` are the same. I think the way to handle more complex usage is to add a `isNewItem` prop which would let the user pass a function that determines if something is a new item or not. It could default to the naive `text` check to handle the most basic case if the user doesn't pass one in.

This would also allow a user to use the "add new" functionality with more complex options. E.g.

```js
const someRemoteData = [
  { id: 'user_1', name: John },
  { id: 'user_2', name: Mary },
  { id: 'user_3', name: Bob  }
]

const options = someRemoteData.map((user) => ({
  { text: <div>{user.name}</div>, value: user.id },
})

<Dropdown options={options} search isNewItem={(value, props) => !_.some(someRemoteData, (user) => user.name === value) } />
```

keeslinp:
If you typed in 'John' wouldn't the value that is received in be 'user_1' and that would be compared against the values in the options list? That's what happens to me when I try it out in codepen at least. I haven't had a chance to verify that against my PR though cause I'm not at that machine. I'll do some more testing to make sure it all does work.

Thanks for pointing out that other spot, I didn't catch that :).

I do like your idea of the isNewItem function. There are a lot of edge cases that'd be intriguing there. I'll see if I can write that up.

edit: One thing I did just think of is that is someone typed in 'user_1' it would probably think that was the same as 'John'. I can't really think of a solution to that. If you're using unique id's the odds of someone typing that on accident are slim. Considering that problem though, I think you're right. We probably would need to have a user-defined function to cover those strange cases.

jeffcarbs:
> If you typed in 'John' wouldn't the value that is received in be 'user_1' and that would be compared against the values in the options list?

No, the "new" item that gets added is an option with `text` and `value` equal to the search query. So in the case where you type `John` the "new" item will be `{ text: 'John', value: 'John' }`. So `John` would be getting compared to `user_1`, `user_2`, etc. This is why this only works when the text/value are the same for all options.

@levithomason - I realized there's a separate convo happening on https://github.com/Semantic-Org/Semantic-UI-React/issues/1064. Want to move it here?

You mentioned not wanting to add more props to `Dropdown` and while I agree that it's getting a bit much I don't see much of a way around adding something like `isNewItem` if we want to support this functionality completely.

levithomason:
👍 to keeping convo here.


>I don't see much of a way around adding something like `isNewItem`

The primary issue is in step 4, "Whenever the values of the dropdown are changed it calls onAddItem whether or not a new item has changed."

The reason for this is in how we call `onAddItem`:
```js
// in selectHighlightedItem() and handleItemClick()

if (onAddItem && !_.some(options, { text: value })) {
  onAddItem(e, { name, value })
}
```
We simply need to call `onAddItem` _only_ when the user is actually adding an item, but, we have no declarative way of identifying the "additional item" from a "regular item".  We are naively attempting to do so with text/value.  My argument is that we shouldn't be delegating this identification to the user because it can be prevented entirely with a little better design.

Simply adding a `data-additional-item` prop to the `addItem` when inserting it would give us a declarative way of identifying the "additional item":

```js
// in selectHighlightedItem() and handleItemClick()

if (item['data-additional-item']] && onAddItem) {
  onAddItem(e, item)
}
```

Now, the issue is solved and no one ever has to attempt to differentiate the "additional item" from "regular items".  We always know exactly which item is the `addItem`, regardless of text/value.

jeffcarbs:
@levithomason - if we're going to go with a prop, we should avoid a `data` prop. Just add an `additional` (or something) to `DropdownItem`.

This still doesn't solve the issue of _when_ to show the additional item. We're relying on the same naive `text` check for that as well: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L708

We could:
- just always show the "additional item" if there's any search input, but I kinda think that's poor UX.
- not add the item manually but instead require the user to pass it as one of the `options` with the `additional` prop. This would require the user to store more state on their side to add the option to the `options` prop they pass in.

levithomason:
I don't think there is an issue with when to show the additional item. The user can only see the text / content, so any search that does not yield an item containing an exact match of the query should result in showing the additional item. 

We should just update the logic to do this. So, if you search for "foo" we show the item "food" but also show "add foo". 

I really think this is something that should just work and serve the user's intuition. I'm not seeing a good reason to make this more complex and configurable.

keeslinp:
I updated it to add an 'additional' property to the Add new item. It all seems to work as far as I can tell and passes all of the tests.

levithomason:
Released in `semantic-ui-react@0.63.2`.

PR #1073, opened 2016-12-21
fix(typings): use boolean `vertical` Segment attribute

dennari:


levithomason:
Thanks much!

levithomason:
Released in `semantic-ui-react@0.63.1`.

PR #1077, opened 2016-12-22
fix(Components): render with children or shorthand of `0`

### LINKED ISSUE #1069: Table.Cell with child {0} doesn't render
### Version
0.63.0

<Table.Cell>{0}<Table.Cell> renders a blank cell. I guess there is some false check on the child. If you render a list of numbers in a table and one happens to be 0, it will not display that value. 

Workaround is to convert the value to a string, but this shouldn't really be necessary.

lusarz:
Fixes #1069

### DIFF
--- a/src/collections/Breadcrumb/BreadcrumbDivider.js
+++ b/src/collections/Breadcrumb/BreadcrumbDivider.js
@@ -20,7 +20,8 @@ function BreadcrumbDivider(props) {
   const rest = getUnhandledProps(BreadcrumbDivider, props)
   const ElementType = getElementType(BreadcrumbDivider, props)
 
-  if (!_.isNil(icon)) return Icon.create(icon, { ...rest, className: classes })
+  const iconElement = Icon.create(icon, { ...rest, className: classes })
+  if (iconElement) return iconElement
 
   let breadcrumbContent
   if (_.isNil(content)) {
--- a/src/collections/Breadcrumb/BreadcrumbDivider.js
+++ b/src/collections/Breadcrumb/BreadcrumbDivider.js
@@ -22,7 +22,13 @@ function BreadcrumbDivider(props) {
 
   if (!_.isNil(icon)) return Icon.create(icon, { ...rest, className: classes })
 
-  return <ElementType {...rest} className={classes}>{content || (_.isNil(children) ? '/' : children)}</ElementType>
+  let breadcrumbContent
+  if (_.isNil(content)) {
+    breadcrumbContent = _.isNil(children) ? '/' : children
+  } else {
+    breadcrumbContent = content
+  }
+  return <ElementType {...rest} className={classes}>{breadcrumbContent}</ElementType>
 }
 
 BreadcrumbDivider._meta = {
--- a/src/collections/Breadcrumb/BreadcrumbDivider.js
+++ b/src/collections/Breadcrumb/BreadcrumbDivider.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
@@ -19,9 +20,9 @@ function BreadcrumbDivider(props) {
   const rest = getUnhandledProps(BreadcrumbDivider, props)
   const ElementType = getElementType(BreadcrumbDivider, props)
 
-  if (icon) return Icon.create(icon, { ...rest, className: classes })
+  if (!_.isNil(icon)) return Icon.create(icon, { ...rest, className: classes })
 
-  return <ElementType {...rest} className={classes}>{content || children || '/'}</ElementType>
+  return <ElementType {...rest} className={classes}>{content || (_.isNil(children) ? '/' : children)}</ElementType>
 }
 
 BreadcrumbDivider._meta = {
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -468,6 +468,11 @@ export const rendersChildren = (Component, options = {}) => {
     shallow(createElement(Component, requiredProps, child))
       .should.contain(child)
   })
+
+  it('renders child number with 0 value', () => {
+    shallow(createElement(Component, requiredProps, 0))
+      .should.contain.text('0')
+  })
 }
 
 // ----------------------------------------
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -556,6 +556,10 @@ export const implementsCreateMethod = (Component) => {
       isValidElement(Component.create(123))
         .should.equal(true)
     })
+    it(`creates a ${name} from a number 0`, () => {
+      isValidElement(Component.create(0))
+        .should.equal(true)
+    })
     it(`creates a ${name} from a props object`, () => {
       isValidElement(Component.create({ 'data-foo': 'bar' }))
         .should.equal(true)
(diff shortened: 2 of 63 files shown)


test/specs/collections/Table/TableCell-test.js
```diff
@@ -42,5 +42,11 @@ describe('TableCell', () => {
 
       wrapper.text().should.equal(content)
     })
+
+    it('render 0 value', () => {
```
layershifter: It makes sense to move it to [commonTests.rendersChildren()](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/commonTests.js#L454). It will test all components that renders children.

layershifter (changes requested):

layershifter:
Hello, @lusarz 

Thanks for PR, I've added review comment, after it's realization we can see more components that need to be updated.

lusarz:
Hello @layershifter,
Thanks for review, I'll update my PR soon

test/specs/collections/Table/TableCell-test.js
lusarz: I moved this to commonTests.rendersChildren(). I am not sure if this check:
`if (Component._meta.name !== 'Embed') {...}`

is ok, but I need it because `<Embed>{0}</Embed>` is rendered as `<Icon/>`

test/specs/commonTests.js
```diff
@@ -468,6 +468,13 @@ export const rendersChildren = (Component, options = {}) => {
     shallow(createElement(Component, requiredProps, child))
       .should.contain(child)
   })
+
+  if (Component._meta.name !== 'Embed') {
```
layershifter: Let's remove this check. You've have failure because [`Embed`](https://github.com/Semantic-Org/Semantic-UI-React/blob/2bfaa3563d3dde230ad54b0e1bdaedba04bf43af/test/specs/modules/Embed/Embed-test.js#L23) asserts use `requiredProps` option:

```js
  common.rendersChildren(Embed, {
    requiredProps: { active: true },
  })
```

Let's use factory there:

```diff
- it('renders child number with 0 value', ()=> {
-   const wrapper = shallow(<Component>{0}</Component>)
-   wrapper.text().should.equal('0')
- })
+ it('renders child number with 0 value', () => {
+ shallow(createElement(Component, requiredProps, 0))
+   .should.contain.text('0')
+ })
```

layershifter (changes requested):

levithomason:
Awesome, this is a great start.  We'll also need to add this check to conditionals for other props that should be rendered.  Here are some examples:

**Button.js**

[`if (label) { ... }`](https://github.com/Semantic-Org/Semantic-UI-React/blob/50b324165fd2908d0b4aab66d3b02e3a61886a9e/src/elements/Button/Button.js#L79) ln 79
[`if (icon && !label) { ... }`](https://github.com/Semantic-Org/Semantic-UI-React/blob/50b324165fd2908d0b4aab66d3b02e3a61886a9e/src/elements/Button/Button.js#L98) ln 98

These should be:
`if (!_.isNil(label)) { ... }`
`if (!_.isNil(icon) && _.isNil(label)) { ... }`

This way, `label={0}` will still render a label with a `0`.  Though there is no icon name `0`, I think we should still add the check for safety.

You can find capture most of these by searching `/src` with the regex `if \((?!children)\w+\) (?:return|\{)`.  This will exclude matches like `if (onClick) onClick(e, this.props)`, of which there are many.

jeffcarbs:
I think for shorthand props we should let the `create` shorthand factory handle instantiating the element and use that for conditionals.

For example (from https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Header/Header.js#L54-L58):

```js
  const iconElement = Icon.create(icon)
  const imageElement = Image.create(image)

  if (iconElement || imageElement) { /* ... */ }
```

In some cases, where a prop can be shorthand or boolean you'd have to add an additional check like:
```js
  const iconElement = Icon.create(icon)

  if (icon === true || iconElement) { /* ... */ }
```

Technically you wouldn't need the `=== true` but I think it would make it more clear.

test/specs/commonTests.js
```diff
@@ -468,6 +468,11 @@ export const rendersChildren = (Component, options = {}) => {
     shallow(createElement(Component, requiredProps, child))
       .should.contain(child)
   })
+
```
lusarz, line 471: I removed check if Component is `Embed` from tests and replaced it with `createElement` method
Thanks @layershifter for help

lusarz:
I added a few additional checks like:
`if (!_.isNil(label)) { ... }`
`if (!_.isNil(icon) && _.isNil(label)) { ... }`

layershifter (approved):

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -19,9 +20,9 @@ function BreadcrumbDivider(props) {
   const rest = getUnhandledProps(BreadcrumbDivider, props)
   const ElementType = getElementType(BreadcrumbDivider, props)
 
-  if (icon) return Icon.create(icon, { ...rest, className: classes })
+  if (!_.isNil(icon)) return Icon.create(icon, { ...rest, className: classes })
 
-  return <ElementType {...rest} className={classes}>{content || children || '/'}</ElementType>
+  return <ElementType {...rest} className={classes}>{content || (_.isNil(children) ? '/' : children)}</ElementType>
```
layershifter: Looks, that `content` might have check for `nil` too. If content will `0`, we will show children there.

lusarz: Updated

levithomason:
I think we have 2 items left here:

1. @jcarbo makes a really great point in https://github.com/Semantic-Org/Semantic-UI-React/pull/1077#issuecomment-268896252.

1. We also have the common test `implementsShorthandProp()` which powers all `implementsFooProp()` tests.  The base shorthand prop test should assert that `0` is also rendered.  Similar to what has been added for shorthand children.

@lusarz Let us know if you are up for these last two items and/or if you need some assistance.


>P.S.
>
>@jcarbo @layershifter Not that long ago I noticed GitHub now allows maintainers to push to a fork's PR branch.  So, if folks need assistance and you have time, you can add their remote and update the PR's branch 👍

levithomason (changes requested):
See my last PR level comment ^

lusarz:
According to [@levithomason comment](https://github.com/Semantic-Org/Semantic-UI-React/pull/1077#issuecomment-269502919), I added two commits:

1. [Check if falsy after instantiate](https://github.com/Semantic-Org/Semantic-UI-React/pull/1077/commits/76f2795f51e40584d62c763065e5901f87abebfe) - I found two places in code to change

2. [Test for shorthand properties](https://github.com/Semantic-Org/Semantic-UI-React/pull/1077/commits/313eb897be8f542790da09eb5322e86fab11127f) - I had problem with pass tests for `implementsHTMLInputProp` method, and I filtered checks with `if (propKey !== 'input') ` - please check if it should be make better

Please check if these changes fix problems, otherwise I think I need some assistance to finish this PR.

levithomason (approved):

levithomason:
I pulled this down and took a look at the input issue.  The solution proposed seems necessary.  I've only added a comment as to why it is there.

This looks great.  Will merge on pass, thanks much!

levithomason:
Released in `semantic-ui-react@0.63.3`.

PR #1078, opened 2016-12-22
fix(docs): importing `../common` on doc site can be edited

lusarz:


levithomason:
I made a couple stylistic updates and fixed some other issues we had in this component.  Will merge on pass, thanks much!

PR #1085, opened 2016-12-23
feat(Grid): add `container` prop

tarang9211:
Is this what we were going for?

### DIFF
--- a/docs/app/Examples/collections/Grid/ResponsiveVariations/GridExampleContainer.js
+++ b/docs/app/Examples/collections/Grid/ResponsiveVariations/GridExampleContainer.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Grid, Image } from 'semantic-ui-react'
+
+const GridExampleContainer = () => (
+  <Grid container columns={3}>
+    <Grid.Column>
+      <Image src='http://semantic-ui.com/images/wireframe/image.png' />
+    </Grid.Column>
+    <Grid.Column>
+      <Image src='http://semantic-ui.com/images/wireframe/image.png' />
+    </Grid.Column>
+    <Grid.Column>
+      <Image src='http://semantic-ui.com/images/wireframe/image.png' />
+    </Grid.Column>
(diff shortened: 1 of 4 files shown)


tarang9211:
@levithomason does that look alright to you? do we need docs for this feature too?

levithomason:
Yep!  Now, we just need an example.  Our examples mirror SUI core examples.  Here is the core grid container example: http://semantic-ui.com/collections/grid#containers

You can reference our Grid examples in `docs/app/Examples/collections/Grid`.  The `container` example would go into the [Responsive Variations](http://react.semantic-ui.com/collections/grid#responsive-variations) section.

tarang9211:
I'm on this @levithomason

tarang9211:
@levithomason added the example to the docs site. let me know if this is good.

levithomason:
Great. I made one update in the last commit, 8af2a62.  Prop strings cannot be multi-line.  Looks good otherwise, I'll merge this on pass.

tarang9211:
oh ok, sorry about that. the reason I did that was because eslint threw an error. Didn't know we could methods on react props.

levithomason:
No worries!  Figured I show the concatenation trick while at it.  FWIW:

Backticks (`` ` ``) preserve whitespace, including newlines:

```js
const multiline = `Line 1
Line2
Line2`
```

But regular quotes (`'` and `"`) do not:

```jsx
const error = 'line 1
line 2'
//=> Uncaught SyntaxError: Invalid or unexpected token
```

A nice trick is to use `[].join()` to join multiple lines.  This is not only far faster than string concatenation, but it also allows better indenting without adding lots of whitespace in the string:

```jsx
const nice = [
  'line 1',
  'line 2',
  'line 3',
].join('\n')
```

To create a single line string, just `join('')` instead 👍

tarang9211:
Awesome, thanks for that tip!

levithomason:
Released in `semantic-ui-react@0.63.3`.

PR #1086, opened 2016-12-25
docs(Dropdown): added comments for shape of options

fobbyal:
took the liberty to add the shape of the options on examples. Just thought it would be helpful. I had to find the `../common.js` file myself.

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Types/DropdownExampleSearchSelectionTwo.js
+++ b/docs/app/Examples/modules/Dropdown/Types/DropdownExampleSearchSelectionTwo.js
@@ -3,6 +3,8 @@ import { Dropdown } from 'semantic-ui-react'
 
 import { stateOptions } from '../common'
 
+// stateOptions = [ { value: 'AL', text: 'Alabama' }, ...  ]
+
 const DropdownExampleSearchSelectionTwo = () => (
   <Dropdown placeholder='State' search selection options={stateOptions} />
 )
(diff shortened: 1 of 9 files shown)


docs/app/Examples/modules/Dropdown/Types/DropdownExampleSearchSelectionTwo.js
```diff
@@ -3,6 +3,16 @@ import { Dropdown } from 'semantic-ui-react'
 
 import { stateOptions } from '../common'
 
+/** stateOptions shape **/
+/**
+[
+  { value: 'AL', text: 'Alabama' },
```
levithomason: Let's shorten all examples that have single-line option objects:

```js
/** stateOptions shape **/
/**
[
  { value: 'AL', text: 'Alabama' },
  { value: 'AK', text: 'Alaska' },
  { value: 'AZ', text: 'Arizona' },
  ...
]
**/
```
Into this style comment:
```js
// stateOptions = [{ value: 'AL', text: 'Alabama' }, ...{}]
```

levithomason (changes requested):
Awesome!  This was on my bucket list.  I've left a couple comments on formatting.  We reserve `/**` doc blocks for external documentation. I'd like to keep the comments as small as possible.  I've left two examples of how I'd like to refactor these.  LMK if there are any questions.  Thanks again!

fobbyal:
amended the commit. please take a look... i had to do multi-line on some of them because they didn't really fit if they were one liners. let me know what you think

levithomason:
Thanks much!

levithomason (approved):

PR #1088, opened 2016-12-25
Dist in repository

andriytyurnikov:
added `babel-cli` to `devDependencies`,
added `dist` to repository, for those who wish to install from GitHub,
I guess we could use pre-release hook that would commit `dist` into repository,
another thing that I've noted - looks like some build scripts are not in repo, like `npm/release.sh`:
`"release:major": "... && ta-script npm/release.sh major && npm run prerelease",` or we should have ta-scripts as devDependency maybe?


layershifter:
> added dist to repository, for those who wish to install from GitHub,

It's very shady idea, IMHO.

levithomason:
>added babel-cli to devDependencies

We don't use the `babel-cli`.

>added dist to repository, for those who wish to install from GitHub

Please install from NPM.  It includes `src` and `dist`.  We do not support any other installation methods.

>we should have ta-scripts as devDependency maybe?

Have another look in [package.json](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/package.json#L119) ;)

andriytyurnikov:
> We don't use the babel-cli.

I believe you are mistaken here, as `babel` command was moved from `babel` package to `babel-cli` some time ago.
I don't have global installation of babel on my machine - and I had real problem with this script:
`package.json` line 16 ` "build:commonjs": "babel src -d dist/commonjs" while trying to make a custom build

> Please install from NPM. It includes src and dist. We do not support any other installation methods.

Well, I've ended up with fork. Just highlighting that some people install from repository, to empower  active contribution. See bootstrap guys https://github.com/twbs/bootstrap/tree/v4-dev/dist/css

As for `ta-scripts` -- those are in place indeed, my bad

levithomason:
>I believe you are mistaken here...

You are correct, I overlooked this one.  I appreciate the correction and have pushed a [fix commit](https://github.com/Semantic-Org/Semantic-UI-React/commit/9b661ac0ebd6760f52ee821a645e8e6423c69893).

PR #1100, opened 2016-12-31
fix(tabIndex): use -1 when disabled, allow override

### LINKED ISSUE #966: Input/Dropdown/Button should not have tab stops when disabled
### Steps
1. Create a disabled Input, disabled search Dropdown, disabled Button
2. Use the tab key to navigate to the Input, Dropdown, or Button

### Expected Result

The tab key should skip focus for disabled Inputs, Dropdowns, and Buttons.  The text should not be selected, neither should there be a tab stop here.

### Actual Result

Even though the user can not enter text into the controls, there is a tab stop for them.  The text is highlighted and the active blue border is applied.  This is against the native behavior of HTML elements.

### Version
60.1.8

### Testcase
~~http://codepen.io/levithomason/pen/QGOPPG~~
**Updated to add button:** http://codepen.io/jeffcarbs/pen/QGQgPJ

image

levithomason:
Fixes #966 

Disabled components were focusable as their `tabIndex` was not properly set.  Some components did not allow setting `tabIndex`.

Updates:

- Disabled components default to `tabIndex={-1}`
- Required `tabIndex` props default to `0`
- All components now allow `tabIndex` to be set explicitly
- Disabled component default `tabIndex` values (-1) can be overridden
- PropTypes accept string _or_ number `tabIndex` values

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -154,11 +154,17 @@ class Button extends Component {
     /** A button can be formatted to show different levels of emphasis */
     secondary: PropTypes.bool,
 
-    /** A button can be formatted to toggle on and off */
-    toggle: PropTypes.bool,
-
     /** A button can have different sizes */
     size: PropTypes.oneOf(_meta.props.size),
+
+    /** A button can receive focus. */
+    tabIndex: PropTypes.oneOfType([
+      PropTypes.string,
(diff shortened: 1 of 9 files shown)


levithomason:
Released in `semantic-ui-react@0.63.4`.

PR #1101, opened 2016-12-31
feat(Dropdown): add `openOnFocus` and `closeOnBlur`

### LINKED ISSUE #517: feat(Dropdown): add `openOnFocus` and `closeOnBlur`

tarang9211:
Fixes #517

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
@@ -4,9 +4,9 @@ import { friendOptions } from '../common'
 
 const DropdownExampleCloseOnBlur = () => (
   <div>
-    <Dropdown placeholder='I close on blur' closeOnBlur fluid selection options={friendOptions} />
-    <br />
-    <Dropdown placeholder='I stay open on blur' closeOnBlur={false} fluid selection options={friendOptions} />
+    <Dropdown placeholder='I close on blur' closeOnBlur selection options={friendOptions} />
+    {' '}
+    <Dropdown placeholder='I stay open on blur' closeOnBlur={false} selection options={friendOptions} />
   </div>
 )
 
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
@@ -1,0 +1,9 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+import { friendOptions } from '../common'
+
+const DropdownExampleCloseOnBlur = () => (
+  <Dropdown placeholder='Select Friend' closeOnBlur fluid selection options={friendOptions} />
+)
+
+export default DropdownExampleCloseOnBlur
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
@@ -3,7 +3,11 @@ import { Dropdown } from 'semantic-ui-react'
 import { friendOptions } from '../common'
 
 const DropdownExampleCloseOnBlur = () => (
-  <Dropdown placeholder='Select Friend' closeOnBlur fluid selection options={friendOptions} />
+  <div>
+    <Dropdown placeholder='I close on blur' closeOnBlur fluid selection options={friendOptions} />
+    <br />
+    <Dropdown placeholder='I stay open on blur' closeOnBlur={false} fluid selection options={friendOptions} />
+  </div>
 )
 
 export default DropdownExampleCloseOnBlur
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
@@ -1,0 +1,9 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+import { friendOptions } from '../common'
+
+const DropdownExampleOpenOnFocus = () => (
+  <Dropdown placeholder='Select Friend' openOnFocus fluid selection options={friendOptions} />
+)
+
+export default DropdownExampleOpenOnFocus
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
@@ -3,7 +3,11 @@ import { Dropdown } from 'semantic-ui-react'
 import { friendOptions } from '../common'
 
 const DropdownExampleOpenOnFocus = () => (
-  <Dropdown placeholder='Select Friend' openOnFocus fluid selection options={friendOptions} />
+  <div>
+    <Dropdown placeholder='I stay open on focus' openOnFocus fluid selection options={friendOptions} />
+    <br />
+    <Dropdown placeholder='I close when not focussed' openOnFocus={false} fluid selection options={friendOptions} />
+  </div>
 )
 
 export default DropdownExampleOpenOnFocus
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
@@ -4,9 +4,9 @@ import { friendOptions } from '../common'
 
 const DropdownExampleOpenOnFocus = () => (
   <div>
-    <Dropdown placeholder='I stay open on focus' openOnFocus fluid selection options={friendOptions} />
-    <br />
-    <Dropdown placeholder='I close when not focussed' openOnFocus={false} fluid selection options={friendOptions} />
+    <Dropdown placeholder='I open on focus' openOnFocus selection options={friendOptions} />
+    {' '}
+    <Dropdown placeholder='I do not open on focussed' openOnFocus={false} selection options={friendOptions} />
   </div>
 )
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -326,6 +326,8 @@ export default class Dropdown extends Component {
     noResultsMessage: 'No results found.',
     renderLabel: ({ text }) => text,
     selectOnBlur: true,
+    openOnFocus: true,
+    closeOnBlur: true
   }
 
   static autoControlledProps = [
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -405,8 +405,7 @@ export default class Dropdown extends Component {
       if (!this.isMouseDown) {
         const { openOnFocus } = this.props
         debug('mouse is not down, opening')
-        if (!openOnFocus) return
-        this.open()
+        if (openOnFocus) this.open()
       }
       if (!this.state.open) {
         document.addEventListener('keydown', this.openOnArrow)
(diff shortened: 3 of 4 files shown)


tarang9211:
@levithomason This PR seems a little vague as of now since I don't have a 100% understanding of the exact issue. But what I did was added the two props to the `Dropdown.js` file.  Let me know what additional changes to make and I'll be more than happy to implement and learn!

src/modules/Dropdown/Dropdown.js
```diff
@@ -296,6 +296,10 @@ export default class Dropdown extends Component {
     /** The text displayed in the dropdown, usually for the active item. */
     text: PropTypes.string,
 
+    openOnFocus: PropTypes.bool,
```
levithomason, line 300: Let's add some descriptions to these:

"Whether or not the menu should open when the dropdown is focused."

And:

"Whether or not the menu should close when the dropdown is blurred."

levithomason:
Now, the `open` method needs to return early if `openOnFocus` is false. Similar to how it currently exits early if `disabled`. 

The same is true for the `close` method and `closeOnBlur`. 

Both of these new props should be added to `defaultProps` and set to `true`. This way, the current behavior is preserved. 

We should also have two new doc examples, one for each prop. These would go in the "Usage" section. 

Finally, we'll add some tests after all this is done.

tarang9211:
So, what's the difference between the menu and the dropdown? The menu is in the dropdown?

So basically open the dropdown when the user focusses it by tabbing to it?

And close the dropdown when you tab away?

tarang9211:
@levithomason let me know how this PR looks as of now. Time permitting, could you clarify my questions above?

levithomason:
>So, what's the difference between the menu and the dropdown? The menu is in the dropdown?

Correct.

>So basically open the dropdown when the user focusses it by tabbing to it?
>And close the dropdown when you tab away?

Exactly.  These features already exist, however, they are always "on" and there is no way to turn them off.  This PR is adding the ability to turn them off.

On this note, the logic should actually be added to the `componentDidUpdate` where it is calling open/close on focus/blur:

https://github.com/tarang9211/Semantic-UI-React/blob/35bd319e09132096c6fa3f2844c4f29bf36e9005/src/modules/Dropdown/Dropdown.js#L405

https://github.com/tarang9211/Semantic-UI-React/blob/35bd319e09132096c6fa3f2844c4f29bf36e9005/src/modules/Dropdown/Dropdown.js#L419

tarang9211:
So the logic added to the open & close methods currently is incorrect?

levithomason:
Indeed, my mistake on the initial instructions.  The goal is to allow the user to enable/disable this behavior.  My suggestion would be to get the doc examples written and wired up, then play with them until you feel it is working correctly.  I can then review the progress.

I thought I'd have more time to dive into this and give guidance, however, it would help me if you could push some further attempts at it on your own and I'll try to help guide you.

tarang9211:
No worries. Hope my questions weren't too silly! I'll have some example for you ready in a few hours.

tarang9211:
@levithomason Hi! I've made the changes and added example in the last three commits. Please could you review these for me? 💯

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+import { friendOptions } from '../common'
+
+const DropdownExampleOpenOnFocus = () => (
+  <Dropdown placeholder='Select Friend' openOnFocus fluid selection options={friendOptions} />
```
levithomason: Let's show both cases here as well.

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+import { friendOptions } from '../common'
+
+const DropdownExampleCloseOnBlur = () => (
+  <Dropdown placeholder='Select Friend' closeOnBlur fluid selection options={friendOptions} />
```
levithomason: How about we show both true/false cases for comparison?

```jsx
<div>
    <Dropdown
      closeOnBlur
      options={friendOptions}
      placeholder='I close on blur' 
      selection
    />
    {' '}
    <Dropdown
      closeOnBlur={false}
      options={friendOptions}
      placeholder='I stay open on blur' 
      selection
    />
  </div>
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -403,7 +403,9 @@ export default class Dropdown extends Component {
     if (!prevState.focus && this.state.focus) {
       debug('dropdown focused')
       if (!this.isMouseDown) {
+        const { openOnFocus } = this.props
         debug('mouse is not down, opening')
+        if (!openOnFocus) return
```
levithomason: Small syntactical change request: `if (openOnFocus) this.open()`, since there is an explicit 1:1 correlation between the condition and the action in this case.

src/modules/Dropdown/Dropdown.js
```diff
@@ -417,7 +419,9 @@ export default class Dropdown extends Component {
     } else if (prevState.focus && !this.state.focus) {
       debug('dropdown blurred')
       if (!this.isMouseDown) {
+        const { closeOnBlur } = this.props
         debug('mouse is not down, closing')
+        if (!closeOnBlur) return
```
levithomason: Same here: `if (closeOnBlur) this.close()`.

levithomason (changes requested):
Just had a new 👶 so might be a little slow on follow up, but I managed to get some reviews in on this.  We're getting real close, thanks much!

tarang9211:
@levithomason Congratulations on the 👶! Hope he/she is doing well!

i've made the changes in the last commit. I also have a .mov file that I created to show how it's working currently. Don't think Github supports that, where would be the best place to send you that file?

I also see this error when I use the arrow keys to move through the menu items. Here's a screenshot of that specific error. 
screen shot 2017-01-10 at 12 27 06 pm

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
```diff
@@ -3,7 +3,11 @@ import { Dropdown } from 'semantic-ui-react'
 import { friendOptions } from '../common'
 
 const DropdownExampleOpenOnFocus = () => (
-  <Dropdown placeholder='Select Friend' openOnFocus fluid selection options={friendOptions} />
+  <div>
+    <Dropdown placeholder='I stay open on focus' openOnFocus fluid selection options={friendOptions} />
```
levithomason:
```diff
-I stay open on focus
+I open on focus
```

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleOpenOnFocus.js
```diff
@@ -3,7 +3,11 @@ import { Dropdown } from 'semantic-ui-react'
...
+    <br />
+    <Dropdown placeholder='I close when not focussed' openOnFocus={false} fluid selection options={friendOptions} />
```
levithomason:
```diff
-I close when not focussed
+I do not open on focus
```

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCloseOnBlur.js
```diff
@@ -3,7 +3,11 @@ import { Dropdown } from 'semantic-ui-react'
 import { friendOptions } from '../common'
 
 const DropdownExampleCloseOnBlur = () => (
-  <Dropdown placeholder='Select Friend' closeOnBlur fluid selection options={friendOptions} />
+  <div>
+    <Dropdown placeholder='I close on blur' closeOnBlur fluid selection options={friendOptions} />
+    <br />
```
levithomason: Let's drop the `fluid` prop and go with `{' '}` here so these are inline elements rather than full page vertically stacked elements.  Just helps keep the docs more concise.

levithomason (changes requested):

levithomason:
Thanks for the updates, I'll pull this branch down and give it a go as soon as I am able.  I left a couple more comments for now 👍 Thanks again.

tarang9211:
@levithomason Done! Just made the changes you suggested.

src/modules/Dropdown/Dropdown.js
```diff
@@ -409,8 +418,9 @@ export default class Dropdown extends Component {
     } else if (prevState.focus && !this.state.focus) {
       debug('dropdown blurred')
       if (!this.isMouseDown) {
+        const { closeOnBlur } = this.props
         debug('mouse is not down, closing')
-        this.close()
+        if (closeOnBlur) this.close()
```
tarang9211, line 423: Does this line look fine to you, logically?

src/modules/Dropdown/Dropdown.js
```diff
@@ -395,8 +403,9 @@ export default class Dropdown extends Component {
     if (!prevState.focus && this.state.focus) {
       debug('dropdown focused')
       if (!this.isMouseDown) {
+        const { openOnFocus } = this.props
         debug('mouse is not down, opening')
-        this.open()
+        if (openOnFocus) this.open()
```
tarang9211, line 408: Does this line look fine to you, logically?

levithomason: 👍

src/modules/Dropdown/Dropdown.js
levithomason: 👍

levithomason:
Perfect thank you!

levithomason (approved):

tarang9211:
@levithomason what about tests? 😸

levithomason:

Yes, tests would be great.  Didn't want to hold this one up too much longer for you, since there are tests covering the basic use already.  If you want to start a new PR for that, we just need to copy some of the existing tests and update then to work for this case 👍

tarang9211:
Cool, any example tests I could look at for this? The other test case was kind of easy since I just had to check if the component had the class name appended to it. Not the case for this one, right?

levithomason:
We add a `describe()` block for each prop, so `describe('openOnFocus', () => { ... })` for example.  Then you can find other focus/blur tests and use those for inspiration.  Example:

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/modules/Dropdown/Dropdown-test.js#L150

tarang9211:
Sounds good. Did you get a chance to review this comment? 

**Alright, hmm. I'm seeing something weird. When I go to the dropdown with closeOnBlur being true, and tab out of it to the adjacent one, the first one closes. It shouldn't right?**

levithomason:
Just beat me to it!  Was about to comment the same.  I checked the source as well, not sure right off hand why that is.  Let's investigate and fix it on the next PR.

tarang9211:
wait a sec, shouldn't it be an if else statement. And in the else, just do `this.open()`?

src/modules/Dropdown/Dropdown.js
```diff
@@ -409,8 +418,9 @@ export default class Dropdown extends Component {
     } else if (prevState.focus && !this.state.focus) {
       debug('dropdown blurred')
       if (!this.isMouseDown) {
```
mull, line 420: Hey guys I know I'm late to the party. I wanted to discuss this line with you. So I'm doing a **search** Dropdown and I don't want it to open on focus. This works fine while tabbing through the form, but clicking on a form field is also focusing on it, yet it'll still open. 

It makes sense that the menu opens if you click on the arrow, but I think for search it's a bit unintuitive if it opens when one clicks on the input.

Thoughts?

PR #1103, opened 2017-01-01
fix(typings): make dimmer in ImageProps optional

### LINKED ISSUE #1102: Typings: dimmer property in ImageProps is not optional.
## Steps
We cannot use ``<Image />`` tag without ``dimmer`` property.
```
[ts] Property 'dimmer' is missing in type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ImageProps, ComponentState>> & ...
```

### Expected Result
It should be OK to use without ``dimmer``.

### Version
0.63.4

### Testcase
http://codepen.io/anon/pen/dNbzEb

rokoroku:
It fixes #1102.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -508,7 +508,7 @@ interface ImageProps extends ReactMouseEvents<any> {
   disabled?: boolean;
 
   /** Shorthand for Dimmer. */
-  dimmer:any;
+  dimmer?: any;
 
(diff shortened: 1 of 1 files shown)


layershifter:
@rokoroku Thanks for PR. LGTM, we should merge this 👍

layershifter (approved):

PR #1098, opened 2016-12-30
feat(Typings: Support Modular Imports): #1080

cschmidli:
Support for modular imports for typescript users, see #1080 for more details.

What's been changed:
* added index.d.ts file for each component
* added index.d.ts in src folder for shared types
* added index.d.ts on root (which is automatically added by typescript IDEs)
* removed typings folder
* added a npm script step to copy the ```*.d.ts``` to the ```dist/commonjs``` folder

please don't wait too long with merging this one as it will block changes to new typings (or i'll have to manually fix the merge conflicts :) )

PR #1121, opened 2017-01-04
feat(typings): support modular imports

cschmidli:
see #1098, just the rebased version.


levithomason:
Thanks much!

levithomason:

Issue #1104, opened 2017-01-02
No way to make search input fluid


wallslide:
http://react.semantic-ui.com/modules/search

There is a `fluid` property on the Search component, however that only affects the search results that appear below the input. I'd like to make the input itself take up the width of its container. This can be accomplished by making the underlying input a `ui fluid input` as mentioned [here](https://github.com/Semantic-Org/Semantic-UI/issues/3084). 

I think the easiest fix is to add another boolean property called `fluidInput` to the Search component to control this behavior.

levithomason:
I see.  Perhaps we should accept a shorthand prop to configure the Input child component:

```jsx
<Search input={{ fluid: true }} />
```

Rohanhacker:
#1123 should fix this (my first PR)

![screenshot from 2017-01-05 14-01-04](https://cloud.githubusercontent.com/assets/9512969/21673551/80069ab0-d34f-11e6-80c9-931686e280d3.png)

![screenshot from 2017-01-05 14-01-35](https://cloud.githubusercontent.com/assets/9512969/21673558/865aee2a-d34f-11e6-99ec-8cab7bb74fed.png)

levithomason:
Fixed in #1123, thanks!

kunokdev:
How do we find possible values for this prop (input @ Search) other than searching within issues?

brianespinosa:
@kunokdev the available props for **all** components are located in the documentation: https://react.semantic-ui.com/modules/search

kunokdev:
@brianespinosa I think you misunderstood me.

![image](https://user-images.githubusercontent.com/13220754/32230187-ddd579ce-be52-11e7-9fb2-5f4aedeb56c0.png)

Default value is `text` string, while solution to this was object: `{ fluid: true }`. 
What else is possible? Here we simply can't know what is possible, object? string? which string values? probably 'text', 'number'...oh what? Actual solution is object? There are infinite possibilities for object values...which ones do we use? How do we get to know that other than searching issues? Is there some list of possible values?

brianespinosa:
@kulakowka I see what you are saying. With shorthand props like `input` they are referencing another SUIR component. So in this case, the `Input` component (https://react.semantic-ui.com/elements/input) would be the place to look for the available props. Looking at this in the docs, I agree that it could be helpful to show this.

@levithomason what if we added links next to these props in the docs to the component that they are referencing to make it easier to follow? Or would that break the way the docs are built currently?

[...]

PR #1123, opened 2017-01-05
fix(Search): Search input can be made fluid

Rohanhacker:
Fixed issue #1104


jeffcarbs:
Looks like this PR updates the permissions for all files. 
screen shot 2017-01-05 at 10 01 16 am

@Rohanhacker - can you revert that and only include the code changes?

Rohanhacker:
@jcarbo sorry my bad reverted and fixed it

src/modules/Search/Search.js
```diff
@@ -508,11 +516,12 @@ export default class Search extends Component {
   // ----------------------------------------
 
   renderSearchInput = () => {
-    const { icon, placeholder } = this.props
+    const { icon, placeholder, input } = this.props
     const { value } = this.state
 
     return (
       <Input
+        fluid={input.fluid}
```
levithomason: This should use our shorthand `Input.create` method.  Our shorthand methods allow creating elements from strings, numbers, props objects, and other elements.  This way, the user has more flexible control for defining the search input.  We can also pass some default props to use with the search input:

```jsx
return Input.create(input, {
  value,
  placeholder,
  onBlur: this.handleBlur,
  onChange: this.handleSearchChange,
  onFocus: this.handleFocus,
  onClick: this.handleInputClick,
  input: { className: 'prompt', tabIndex: '0', autoComplete: 'off' },
  icon,
})
```

src/modules/Search/Search.js
```diff
@@ -166,6 +166,11 @@ export default class Search extends Component {
     /** A search can have its results take up the width of its container. */
     fluid: PropTypes.bool,
 
+    /** A search input can take up the width of its container . */
+    input: PropTypes.shape({
```
levithomason: This will be a shorthand value, `customPropTypes.itemShorthand`.  Our shorthand prop type allows strings, numbers, props objects, and other elements.  These shorthand prop values are used with our shorthand `create()` methods, see below.

src/modules/Search/Search.js
```diff
@@ -176,6 +181,9 @@ export default class Search extends Component {
     minCharacters: 1,
     noResultsMessage: 'No results found.',
     showNoResults: true,
+    input: {
```
levithomason: Once the shorthand changes are made, the default value here would just be `text`.  That will ensure the `Input.create()` shorthand creates a `type='text'` input by default.

We don't explicitly need to set the `type`, it is just that we need a non-nil value to work with.  Otherwise, the `Input.create()` shorthand factory would not create an element.  All factories return null if passed null/undefined.

levithomason (changes requested):

src/modules/Search/Search.js
jeffcarbs: @levithomason - Given the code in this snippet, if you pass any `input` prop it will override the default `input` props. For example:

```jsx
<Search>
// yields an input of
<input type="text" value="" class="prompt" tabindex="0" autocomplete="off">

<Search input={{fluid: true}} />
// would yield an input of
<input type="text" value="" class="fluid">
// but we'd desire
<input type="text" value="" class="prompt fluid" tabindex="0" autocomplete="off">
```

levithomason: Hm, was this tested?  The default props should be merged, not replaced.

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/factories.js#L84

src/modules/Search/Search.js
Rohanhacker: I have changed this to customPropTypes.itemShorthand but I getting this error when using Input.create()
screenshot from 2017-01-06 16-39-37

levithomason: I've opened https://github.com/Semantic-Org/Semantic-UI-React/pull/1140 to add this missing method.  Sorry about that, didn't realize it was missing!

levithomason: Alright, just merged the update.  You can now update this PR with the latest master and you'll have the `Input.create()` method 👍

levithomason:
See my latest comments here: https://github.com/Semantic-Org/Semantic-UI-React/pull/1123#discussion_r95270911

src/modules/Search/Search.js
Rohanhacker: thanks @levithomason

Rohanhacker:
implemented requested changes

levithomason:
Almost done here, one more comment left: https://github.com/Semantic-Org/Semantic-UI-React/pull/1123/files#r94852421

And the file permissions seem to be changed still from `100644 → 100755`.

levithomason (approved):

levithomason:
Great, thanks much for all the work here!

levithomason:

PR #1127, opened 2017-01-06
docs(MenuExampleSecondaryPointing): Missing Menu Labels

AnsonT:
Missing content in Menu.Item


levithomason:
This is intentional.  The "Start Cased" `name` prop will be used for the child content if there is not one.  See the example at http://react.semantic-ui.com/collections/menu#menu.

PR #1131, opened 2017-01-06
fix(Input): allow conditional rendering of children

kyrasteen:
Hey there - first of all - big fan of semantic-ui-react!

I have found an issue when trying to render conditional children inside of `<Input>`. When the condition is not met, the child is null, causing the internal check of `child.type` to error. Wondering if you would consider adding my change to support conditional rendering?


levithomason:
This is a necessary improvement, thanks much.  I've updated the syntax a bit to match the existing.  Will merge on pass.

levithomason:

PR #1146, opened 2017-01-11
feat(Popup): add vertical offset

goloveychuk:
In current implementation it's impossible to change vertical position of popup (without ugly hacks).
This pr changes this.


goloveychuk:
it's not correct. let me fix.

goloveychuk:
fixed, force pushed

src/modules/Popup/Popup.js
```diff
@@ -76,6 +76,9 @@ export default class Popup extends Component {
     /** Horizontal offset in pixels to be applied to the popup */
     offset: PropTypes.number,
 
+    /** Vertical offset in pixels to be applied to the popup */
+    verticalOffset: PropTypes.number,
```
levithomason: How about allowing an X and Y offset in the current offset prop instead of adding a new prop? 

```jsx
offset={[5, 10]}

// vs

offset={5}
verticalOffset={10} 
```

We could still allow a single value to apply a horizontal offset only,so it is backwards compatible.

levithomason (changes requested):
This PR is failing due to missing test coverage. There is an existing test for the offset property that can copied and updated to test vertical offset logic. 

Other than this is I've left one API comment. Thanks for the work here!

src/modules/Popup/Popup.js
goloveychuk: yea, your api is better. will change tomorrow

levithomason:
Looks like this PR is falling behind.  Shall we close it, or update and merge?

goloveychuk:
we can close it.

src/collections/Message/Message.d.ts
```diff
@@ -9,7 +9,7 @@ import { default as MessageList } from './MessageList';
 export interface MessageProps {
   [key: string]: any;
 
-	/** An element type to render as (string or function). */
+  /** An element type to render as (string or function). */
```
layershifter, line 12: Lint fix. Quite strange issue.

layershifter (commented):
@levithomason I've updated this PR to latest `master` and implimented functionality.  However, I see there a potential problem, `offset` tests are testing nothing in fact.

levithomason:
@layershifter Could you expound a bit on "testing nothing"?

layershifter:
@levithomason take a look at the assertions in the `offset` tests, they only check what `Popup` is visible. But the offsets and positions aren't tested in any way.

levithomason:
Oh, I see, existing tests were updated.  We need new tests that assert this is working.

levithomason:
This has gotten stale and we don't hear much about this feature from the community.  Closing for housekeeping.  Happy to merge a complete community PR for this feature if submitted.

PR #1148, opened 2017-01-12
fix(Dropdown): respect `closeOnBlur={false}`

tarang9211:


src/modules/Dropdown/Dropdown.js
```diff
@@ -420,7 +420,11 @@ export default class Dropdown extends Component {
       if (!this.isMouseDown) {
         const { closeOnBlur } = this.props
         debug('mouse is not down, closing')
-        if (closeOnBlur) this.close()
+        if (closeOnBlur) {
```
tarang9211: @levithomason this is what I meant. With this piece of code, when I tab away from this dropdown, it stays open until a user clicks somewhere else on the screen. Let me know your thoughts.

tarang9211:
@levithomason any idea what's going with the codecov failing?

src/modules/Dropdown/Dropdown.js
levithomason: Hm, I'll mark this for review to investigate.  I can't think of a reason we'd need to set `open` again when it is already open.  Thanks for the note here.

tarang9211: @levithomason sounds good. let me know how we can fix this.

levithomason:
Codecov is failing because there were new branches of logic added that do not have test cases.

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.github/CONTRIBUTING.md#coverage

tarang9211:
yea it complained about the if/else block not having code coverage.

tarang9211:
@levithomason Do we have any feedback or review on this?

levithomason:
I'll get a review on the functionality as is for now.  However, this PR needs more coverage before it can be merged.  The added branch logic needs a test, this is why Codecov is failing the PR.

levithomason:
OK, so the issue it turns out is that blurring calls `handleBlur()` which causes the selected item to become active in `selectHighlightedItem()`.  This method would then close the Dropdown, if it is not a multiple Dropdown.

This logic flow worked fine before adding `closeOnBlur`.  Instead, I've removed the `close()` logic from `selectHighlightedItem()` and that method only activates the item.  I've explicitly called `close()` in `selectItemOnEnter()` and `handleBlur()` with the appropriate logic.

With this fix, we're able to avoid having to re-open the Dropdown on blur as was originally the case in this PR.

src/modules/Dropdown/Dropdown.js
```diff
@@ -547,7 +547,6 @@ export default class Dropdown extends Component {
     } else {
       this.setValue(value)
       this.handleChange(e, value)
-      this.close()
```
levithomason, line 550: This was the bug, we were closing every time an item was activated.  Even when `closeOnBlur` was false.  Since, this method was called on blur due to blur activating the selected item.

src/modules/Dropdown/Dropdown.js
```diff
@@ -902,9 +906,7 @@ export default class Dropdown extends Component {
 
   scrollSelectedItemIntoView = () => {
     debug('scrollSelectedItemIntoView()')
-    // Do not access document when server side rendering
-    if (!isBrowser) return
-    const menu = document.querySelector('.ui.dropdown.active.visible .menu.visible')
```
levithomason, line 907: This fixes another bug.  If multiple Dropdowns were open, this would select the first one on the page.  This meant the Menu for a Dropdown not belonging to this one was being operated on.

I've fixed that by selecting _this_ Dropdown's Menu.

tarang9211:
@levithomason Ah ok that makes sense. I did notice that the dropdown was closing after making a selection on an item. Sounds good, thank you for the review and guidance!

levithomason:

PR #1150, opened 2017-01-12
style(Search): several cleanups

layershifter:
I missed the moment when #1123 was merged, so I'll make several style cleanups in this PR.

This PR:
- sorts props inside their blocks;
- sorts imports;
- propTypes cleanups (#524);
- other moments.

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -24,14 +24,6 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
-const _meta = {
-  name: 'Search',
-  type: META.TYPES.MODULE,
-  props: {
-    size: _.without(SUI.SIZES, 'medium'),
-  },
-}
-
 /**
  * A search module allows a user to query for results from a selection of data
  */
@@ -44,11 +36,23 @@ export default class Search extends Component {
     // Behavior
     // ------------------------------------
 
+    /** Initial value of open. */
+    defaultOpen: PropTypes.bool,
+
+    /** Initial value. */
+    defaultValue: PropTypes.string,
+
     /** Shorthand for Icon. */
     icon: PropTypes.oneOfType([
       PropTypes.node,
       PropTypes.object,
     ]),
+
+    /** Controls whether or not the results menu is displayed. */
+    open: PropTypes.bool,
+
+    /** Placeholder of the search input. */
+    placeholder: PropTypes.string,
 
     /**
      * One of:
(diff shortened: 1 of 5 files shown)


src/modules/Search/index.d.ts
```diff
@@ -34,6 +34,7 @@ interface SearchProps extends ReactMouseEvents<HTMLInputElement>, ReactFocusEven
     /** Shorthand for Icon. */
     icon?: any;
 
+    /** A search can show a loading indicator. */
```
layershifter, line 37: Added inline comment to prop.

src/modules/Search/Search.js
```diff
@@ -640,15 +639,13 @@ export default class Search extends Component {
...
-      useValueAndKey(aligned, 'aligned'),
       useKeyOnly(category, 'category'),
       useKeyOnly(focus, 'focus'),
       useKeyOnly(fluid, 'fluid'),
-
-      className,
+      useKeyOnly(loading, 'loading'),
+      useValueAndKey(aligned, 'aligned'),
       'search',
+      className,
```
layershifter, line 643: `className` is always last in our pattern.

src/modules/Search/Search.js
```diff
@@ -612,11 +615,7 @@ export default class Search extends Component {
 
     if (!menuContent) return
 
-    return (
-      <SearchResults className={resultsClasses}>
-        {menuContent}
-      </SearchResults>
-    )
+    return <SearchResults className={resultsClasses}>{menuContent}</SearchResults>
```
layershifter, line 613: This expression is too simple to be multiline.

src/modules/Search/Search.js
```diff
@@ -512,25 +517,23 @@ export default class Search extends Component {
...
-        placeholder,
-        onBlur: this.handleBlur,
-        onChange: this.handleSearchChange,
-        onFocus: this.handleFocus,
-        onClick: this.handleInputClick,
-        input: { className: 'prompt', tabIndex: '0', autoComplete: 'off' },
-        icon,
-      })
-    )
+    return Input.create(input, {
```
layershifter, line 518: We don't really need parens there.

src/modules/Search/Search.js
```diff
@@ -169,17 +169,19 @@ export default class Search extends Component {
     /** A search input can take up the width of its container. */
     input: customPropTypes.itemShorthand,
 
-    size: PropTypes.oneOf(_meta.props.size),
-
+    /** A search can show a loading indicator. */
```
layershifter, line 164: Inline comment.

src/modules/Search/Search.js
```diff
@@ -169,17 +169,19 @@ export default class Search extends Component {
...
     loading: PropTypes.bool,
+
+    /** A search can have different sizes. */
```
layershifter, line 167: ^

src/modules/Search/index.d.ts
levithomason: Thanks, we should also update the ts file comment as well.

src/modules/Search/Search.js
levithomason: Let's also update the ts file.

src/modules/Search/Search.js
levithomason: ^ 😄

layershifter:
I've updated ts-definition file and updated propTypes for #524.

levithomason:
Thanks much!

levithomason:

PR #1151, opened 2017-01-13
feat(typings): add closeOnBlur and openOnFocus

levithomason:
Adds newly added props to typings.  Also sorts said props in Dropdown.


levithomason:

PR #1154, opened 2017-01-13
fix(Portal): take focus on  mount and restore on unmount

fracmak:
All portal based elements (modals, popups, etc) are not keyboard accessible because they're usually injected at the end of the dom and not in the flow of the document. This pull request fixes that by putting focus on the popup when open (except when closeOnTriggerBlur is true) and restores the focus to the last known focus point on close.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -341,7 +341,7 @@ class Portal extends Component {
     if (!this.state.open) return
     debug('renderPortal()')
 
-    const { children, className } = this.props
+    const { children, className, closeOnTriggerBlur } = this.props
 
     this.mountPortal()
 
@@ -363,6 +363,14 @@ class Portal extends Component {
     )
 
     this.portal = this.node.firstElementChild
+    // don't take focus away from portals that close on blur
+    if (!closeOnTriggerBlur) {
+      this.previousActiveElement = document.activeElement
+      this.portal.setAttribute('tabindex', '-1')
+      this.portal.style.outline = 'none'
+      // wait a tick for things like popups which need to calculate where the popup shows up
+      setTimeout(() => this.portal && this.portal.focus())
+    }
 
     this.portal.addEventListener('mouseleave', this.handlePortalMouseLeave)
     this.portal.addEventListener('mouseenter', this.handlePortalMouseEnter)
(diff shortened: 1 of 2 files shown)


src/addons/Portal/Portal.js
```diff
@@ -363,6 +363,14 @@ class Portal extends Component {
     )
 
     this.portal = this.node.firstElementChild
+    // don't take focus away from portals that close on blur
+    if (!closeOnTriggerBlur) {
+      this.triggerNode = document.activeElement
```
levithomason: Nit: Could we use another name here?

If a portal is triggered programmatically, the `activeElement` will not necessarily be the `triggerNode`.  These are only the same when the user clicked on a trigger, such as a button.  How about something like `previousActiveElement` or similar?

src/addons/Portal/Portal.js
```diff
@@ -363,6 +363,14 @@ class Portal extends Component {
...
+      this.portal.setAttribute('tabindex', '-1')
+      this.portal.style.outline = 'none'
+      // wait a tick for things like popups which need to calculate where the popup shows up
+      setTimeout(() => this.portal && this.portal.focus())
```
levithomason, line 372: I'd like to pull this side effect out of the render method if at all possible.  I haven't the bandwidth at the moment to find a better place myself, but open to suggestions.

levithomason (changes requested):

src/addons/Portal/Portal.js
fracmak: Technically this code is being called by either componentDidMount or componentDidUpdate depending on when 'open' is set to true which is after the render method. Or are you saying you'd like me to make a named helper function that's called here? There's no later life cycle function I can use to trigger this behavior

src/addons/Portal/Portal.js
fracmak: done

src/addons/Portal/Portal.js
levithomason: Oh, I see, thanks for the clarification.  From my initial pass I assumed this was called in a `render()` method itself.  -1 for assumptions :)

levithomason (approved):

levithomason:

Issue #1072, opened 2016-12-21
fix(typings): unable to set many typical html element props


dennari:
I was running in to problems trying to set the `id` attribute for `Segment`. Currently the typings don't allow this in TSX, even if it would be perfectly ok in JSX. The same is true for a number of other standard html attributes and Semantic-UI-React-Components.

One option would be to change
```typescript
interface SegmentProps {
```
to

```typescript
interface SegmentProps extends React.HTMLProps<HTMLElement> {
```
giving the ability to set a huge amount of additional attributes, including `id`. This however floods the intellisense with attributes that are probably unneeded 95% of the time.

Another option would be to add an index type:
```typescript
interface SegmentProps {
  ...
  [key: string]: any;
}
```
which would then allow any property to be set without flooding intellisense.

Any thoughts?

levithomason:
I would assume the latter would be more useful to users.   That said, I'd like to get some feedback on this from our ts users before merging a fix.

/cc @tomitrescak @hlehmann @JasonRitchie @garyham @asvetliakov @halfmatthalfcat @psudarsa

You all have been active in our ts decisions in the past, let us know your thoughts here.

asvetliakov:
For now you can use module augmentation i think
```ts
declare module "semantic-ui-react" {
interface SegmentProps {
   id?: string;
}
// or
interface SegmentProps extends React.HTMLProps<HTMLElement> {}
}

```


Personally i'm doing 1) approach for elements where passing html attributes does make sense (input, form, etc...)

The 2) approach destroys parameter validation since you can pass anything.

levithomason:
>The 2) approach destroys parameter validation since you can pass anything.

Good point, this seems rather unfavorable.

dennari:
> The 2) approach destroys parameter validation since you can pass anything.

The validation still works as expected for the explicitly specified properties.

asvetliakov:
Yes, seems i was wrong about 2. Latest compiler version works as expected.

levithomason:
If there is an agreement on 2, I'd merge a PR for that.  It would be nice _not_ to have the common HTML props flooding the completions.  I think the intent is for users to see the custom component props.

levithomason:
For clarity, we're ready for a community PR here implementing this pattern for all components:

```ts
interface SegmentProps {
  ...
  [key: string]: any;
}
```

layershifter:
### TODO

- [x] #1216 Button
- [x] #1159 Container
[...]

PR #1159, opened 2017-01-14
style(Container|Divider|Header|Icon): propTypes cleanups and typings update

layershifter:

Affects following components:

- `Container`
- `Divider`
- `Header`
- `Icon`


src/elements/Header/Header.js
```diff
@@ -80,29 +95,43 @@ function Header(props) {
...
 Header.propTypes = {
   /** An element type to render as (string or function). */
   as: customPropTypes.as,
 
-  /** Additional classes. */
-  className: PropTypes.string,
+  /** Attach header  to other content, like a segment. */
+  attached: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.oneOf('top', 'bottom'),
```
levithomason:
```diff
-PropTypes.oneOf('top', 'bottom'),
+PropTypes.oneOf(['top', 'bottom']),
```

src/elements/Header/index.d.ts
```diff
@@ -1,17 +1,22 @@
-import { SemanticCOLORS, SemanticFLOATS, SemanticSIZES, SemanticTEXTALIGNMENTS } from '../..';
 import * as React from 'react';
+import {
+  SemanticCOLORS,
+  SemanticFLOATS,
+  SemanticTEXTALIGNMENTS
+} from '../..';
+
+type HEADER_SIZES = 'mini' | 'tiny' | 'small' | 'medium' | 'large' | 'huge'
```
levithomason: Looks like this is only used in one place.  Suggest inlining it.

src/elements/Icon/index.d.ts
```diff
@@ -71,4 +78,4 @@ interface IconGroupProps {
   size?: SemanticSIZES;
```
levithomason: I believe this should be `ICON_SIZES`, correct?

levithomason (changes requested):

layershifter:
TODO: update `content?: any;` to `React.Node` (`customPropTypes.contentShorthand`).

src/elements/Icon/index.d.ts
layershifter: Yep, I also update Group's `propTypes`.

layershifter:
We need merge #1165 before this, #1165 contains `implementsTextAlignProp` test update.

levithomason (approved):

levithomason:
Merged #1165, fixed conflicts and unblocked.  Will merge on pass.

levithomason:

PR #1160, opened 2017-01-14
chore(package): remove transform-react-constant-elements

layershifter:
We [decided](https://github.com/Semantic-Org/Semantic-UI-React/pull/731#issuecomment-271340793) not to use it at #731, but I missed the included dependency, this PR removes it.


levithomason:

PR #1164, opened 2017-01-15
typings(Popup, Menu): Fixed Popup inverted and MenuItem content types

dstpierre:


src/collections/Menu/index.d.ts
```diff
@@ -135,7 +135,7 @@ interface MenuItemProps extends ReactMouseEvents<HTMLElement> {
   color?: SemanticCOLORS;
 
   /** Shorthand for primary content. */
-  content?: string;
+  content?: any;
```
levithomason: This is actually `PropTypes.node` which is:

>Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types.

https://facebook.github.io/react/docs/typechecking-with-proptypes.html

levithomason (changes requested):

src/collections/Menu/index.d.ts
levithomason: Per @layershifter in https://github.com/Semantic-Org/Semantic-UI-React/pull/1159#issuecomment-272907848, let's use `React.Node`.

dstpierre:
I used `React.ReactNode` instead of the `any`.

levithomason (approved):

levithomason:
Thanks much!

levithomason:

PR #1165, opened 2017-01-15
style(Loader|Reveal|Segment): propTypes cleanups and typings update

layershifter:

Affects following components:

- `Loader`
- `Reveal`
- `Segment`

### DIFF
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -859,7 +859,7 @@ export const implementsImageProp = (Component, options = {}) => {
 /**
  * Assert that a Component correctly implements the "textAlign" prop.
  * @param {React.Component|Function} Component The component to test.
- * @param {array} alignments Array of possible alignment positions.
+ * @param {array} [alignments] Array of possible alignment positions.
  * @param {Object} [options={}]
  * @param {Object} [options.requiredProps={}] Props required to render the component.
  */
--- a/test/specs/commonTests.js
+++ b/test/specs/commonTests.js
@@ -858,7 +858,6 @@ export const implementsImageProp = (Component, options = {}) => {
  * @param {Object} [options.requiredProps={}] Props required to render the component.
  */
 export const implementsTextAlignProp = (Component, options = {}) => {
-  const { requiredProps = {} } = options
   const { assertRequired } = commonTestHelpers('implementsTextAlignProp', Component)
 
   describe('aligned (common)', () => {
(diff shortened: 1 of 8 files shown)


test/specs/commonTests.js
```diff
@@ -858,31 +858,13 @@ export const implementsImageProp = (Component, options = {}) => {
...
-  const { requiredProps = {} } = options
   const { assertRequired } = commonTestHelpers('implementsTextAlignProp', Component)
 
   describe('aligned (common)', () => {
     assertRequired(Component, 'a `Component`')
 
     _noDefaultClassNameFromProp(Component, 'textAlign', options)
     _noClassNameFromBoolProps(Component, 'textAlign', options)
-
-    _.each(Component._meta.props.aligned, (propVal) => {
```
levithomason, line 870: The removed portion of the test is the useful part.  It makes the assertion that the proper `className` is added.  I see it was removed as it looped over `_meta.props.aligned`.  In fact, this test was doing nothing as the `aligned` prop name was changed to `textAlign`!

Instead of removing these assertions, however, let's loop over `SUI.TEXT_ALIGNMENTS` directly:

```jsx
_.each(SUI.TEXT_ALIGNMENTS, (propVal) => {
  ...
})
```

levithomason (changes requested):
Just one note about the common test.

test/specs/commonTests.js
layershifter: I'm think we can do there something better :thinking: We can use this pattern with other common tests that relies on `meta`. Thoughts?

```js
export const implementsTextAlignProp = (Component, alignments = SUI.TEXT_ALIGNMENTS,  options = {}) => {
// ...
  _.each(alignments, (propVal) => {
  // ...
  })
}
```

levithomason: That would also work.  I checked all the usages and all components that implement `TEXT_ALIGNMENTS` implement all of them.  I'm OK with the default pattern as well.

test/specs/commonTests.js
```diff
@@ -854,17 +859,36 @@ export const implementsImageProp = (Component, options = {}) => {
 /**
  * Assert that a Component correctly implements the "textAlign" prop.
  * @param {React.Component|Function} Component The component to test.
+ * @param {array} alignments Array of possible alignment positions.
```
levithomason: This can be optional:
```diff
-@param {array} alignments
+@param {array} [alignments]
```

layershifter: Indeed :smile:

levithomason:
Awesome, thank you!

levithomason (approved):

levithomason:

PR #1167, opened 2017-01-16
style(Input): propTypes cleanups and typings update

layershifter:

Affects `Input` component.


levithomason:
Fantastic, thanks much!

levithomason:

Issue #1147, opened 2017-01-11
Dropdown: Icon isn't displayed in selection


rbideau:
### Steps
1. Create a select Dropdown with options with text and icon
2. Select an option

### Expected Result
The icon and text should be displayed as the selected option

### Actual Result
Only the text is displayed , not the icon

### Version
v0.63.5

### Testcase
[semantic-ui codepen](https://codepen.io/anon/pen/PWZrLb)
[semantic-ui-react codepen](http://codepen.io/anon/pen/qRbzeW)

levithomason:
### Workaround
I'm a little torn on which way this should go.  A Dropdown option (`Dropdown.Item`) can define both `text` and `content`.  The item's `text` value is displayed in the Dropdown when the item is active.  The `content` is used to show one value in the menu and a different value in the Dropdown when the item is active, see http://react.semantic-ui.com/modules/dropdown#item-content. So, to add text and an icon to the Dropdown for an active item, the item's `text` needs to contain both:

```jsx
const options = [
  { value: '0', text: <span><Icon name='man' /> Man</span> },
  { value: '1', text: <span><Icon name='woman' /> Woman</span> },
  { value: '2', text: <span><Icon name='genderless' /> Other</span> },
]
```

This will render what you are looking for.  Forked example http://codepen.io/levithomason/pen/apNOZp.

### Proper fix
That said, I think your use case is intuitive and should probably work.  What we'd need to do is render the `children` of the active item instead of the item's `text` prop only.  This way, how ever the item ended up rendering in the menu is the how it will render in the Dropdown when active.

This approach will need to consider the fact that the `content` should not be rendered in the Dropdown.  Also, if there are multiple children, then the item's `children` will be an array that we'll need to wrap with another element (i.e. `div`).

Likely, the proper implementation here will be to take the active item's props, remove the `content` and `children`, render a new DropdownItem with said props, and take its children.  This way, we get the proper children given the DropdownItem's props configuration.  The [logic for choosing the text](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L961) would need to handle something like:

```jsx
const { children, content, ...rest } = activeItem.props
const _text = DropdownItem.create(rest).props.children
```

rbideau:
Thank you for your quick feedback.

Your workaround does works but generates the following warnings :
```
Warning: Only strings and numbers are supported as <option> children
[...]

PR #1172, opened 2017-01-17
fix(Dropdown): display icon in selected text

rbideau:
First attempt at fixing #1147

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -974,12 +974,21 @@ export default class Dropdown extends Component {
     } else if (text) {
       _text = text
     } else if (open && !multiple) {
-      _text = _.get(this.getSelectedItem(), 'text')
-    } else if (hasValue) {
+      _text = this.renderComplexeText(value)
+    } else if (hasValue && multiple) {
+      // renderLabels will handle image and icon in multiple dropdown
       _text = _.get(this.getItemByValue(value), 'text')
+    } else if (hasValue && !multiple) {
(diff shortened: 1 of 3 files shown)


levithomason:
@rbideau what is the status on this one?

rbideau:
I didn't spent any time on it lately, but I'm still stuck at the same point as I said in #1147.

It's in [line 990 of src/modules/Dropdown/Dropdown.js](https://github.com/Semantic-Org/Semantic-UI-React/pull/1172/files#diff-609214949026e594dbfd98481664c928R990) of this PR.

levithomason:
The DropdownItem's create method does not set children:

https://github.com/Semantic-Org/Semantic-UI-React/pull/1172/files#diff-39c9e693d6f8be797267ef676395ef14R171

```diff
-DropdownItem.create = createShorthandFactory(DropdownItem, value => value)
+DropdownItem.create = createShorthandFactory(DropdownItem, children => ({ children }))
```

The callback above is invoked when the shorthand value is a string, number, or an array.  This change will map those values to the children of the DropdownItem so that:

```jsx
DropdownItem.create('item 1')
// => <DropdownItem>item 1</DropdownItem>
```

rbideau:
This doesn't work because when there is an icon, `rest` is an object and then [mapValueToProps is not called in factories/createShorthand](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/factories.js#L37-L39)

levithomason:
This PR has gotten stale and seems abandoned.  Closing for housekeeping.  Still happy to merge an updated and complete community PR.

PR #1179, opened 2017-01-18
fix(Divider|FormInput): fix the broken typings

### LINKED ISSUE #1180: The typings for FormInput and Divider are broken
See #1179 for a fix

dennari:
Fixes #1180.

### DIFF
--- a/src/collections/Form/index.d.ts
+++ b/src/collections/Form/index.d.ts
@@ -8,7 +8,7 @@ import {
     SemanticWIDTHSSTRING
 } from '../..';
 import * as React from 'react';
-import { InputBaseProps } from '../../elements/Input/index';
+import { InputProps } from '../../elements/Input/index';
 import { DropdownProps } from '../../modules/Dropdown/index';
 
 
@@ -173,7 +173,7 @@ interface FormGroupProps {
 
 export const FormGroup: React.ComponentClass<FormGroupProps>;
 
-interface FormInputProps extends InputBaseProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
+interface FormInputProps extends InputProps {
 }
 
 export const FormInput: React.ComponentClass<FormInputProps>;
--- a/src/collections/Form/index.d.ts
+++ b/src/collections/Form/index.d.ts
@@ -8,7 +8,7 @@ import {
     SemanticWIDTHSSTRING
 } from '../..';
 import * as React from 'react';
-import { InputProps } from '../../elements/Input/index';
+import { InputBaseProps } from '../../elements/Input/index';
 import { DropdownProps } from '../../modules/Dropdown/index';
 
 
@@ -173,7 +173,7 @@ interface FormGroupProps {
 
 export const FormGroup: React.ComponentClass<FormGroupProps>;
 
-interface FormInputProps extends InputProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
+interface FormInputProps extends InputBaseProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
 }
 
 export const FormInput: React.ComponentClass<FormInputProps>;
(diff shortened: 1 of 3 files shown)


src/collections/Form/index.d.ts
```diff
@@ -173,7 +173,7 @@ interface FormGroupProps {
 
 export const FormGroup: React.ComponentClass<FormGroupProps>;
 
-interface FormInputProps extends InputProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
+interface FormInputProps extends InputBaseProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
```
levithomason: Hm, why exclude the `onChange` prop from the FormInputProps?  All Input props are supported in the Form control shorthand components.  I may be missing something here.

dennari: This is the error I got before:

```
> tsc


176 interface FormInputProps extends InputProps, ReactFormEvents<HTMLInputElement>, ReactFocusEvents<HTMLInputElement> {
              ~~~~~~~~~~~~~~

node_modules/semantic-ui-react/dist/commonjs/collections/Form/index.d.ts(176,11): error TS2320: Interface 'FormInputProps' cannot simultaneously extend types 'InputProps' and 'ReactFormEvents<HTMLInputElement>'.
  Named property 'onChange' of types 'InputProps' and 'ReactFormEvents<HTMLInputElement>' are not identical.
```

So `ReactFormEvents<HTMLInputElement>` already includes the `onChange` prop with an incompatible signature. This is probably not the cleanest possible fix, but it works at least.

src/elements/Divider/index.d.ts
```diff
@@ -34,4 +34,4 @@ interface DividerProps {
   vertical?: boolean;
 }
 
-export const Container: React.StatelessComponent<DividerProps>;
+export const Divider: React.StatelessComponent<DividerProps>;
```
layershifter, line 37: Thanks, copy-paste is evil.

src/collections/Form/index.d.ts
layershifter: I think we can just remove `ReactFormEvents<HTMLInputElement>` and `ReactFocusEvents<HTMLInputElement>` there, because we don't [expose](https://github.com/Semantic-Org/Semantic-UI-React/issues/1072#issuecomment-268848820) non SUIR props (#1072).

levithomason: The signature for this:
```jsx
<Form.Input onChange={ ... } />
```
Should be identical to this:
```jsx
<Input onChange={ ...  } />
```
Wihch is:
```ts
onChange(event: React.SyntheticEvent<HTMLInputElement>, data: InputOnChangeData) => void
```

So the `onChange` signature from `ReactFormEvents<HTMLInputElement>` is not correct for this case since it does not include the `data` argument.  I'm not a ts user so I'm not exactly sure of the fix here, however, I wanted to note the actual signatures.

rokoroku: How about adding   
`onChange`, `onInput`, `onSubmit` methods to the `InputProps` and removing  `extend ReactFormEvents`? (If these methods are available)

dennari: @layershifter's suggestion seems to me to be the correct one in this case

layershifter (approved):
I've returned `onChange` on `Input` typings. LGTM, ready to merge 👍

levithomason:
Will this not throw the same error in FormInput about duplicate onChange props with different signatures?

layershifter:
@levithomason nope, problem was there:
```
(176,11): error TS2320: Interface 'FormInputProps' cannot simultaneously extend types 'InputProps' and 'ReactFormEvents<HTMLInputElement>'.
  Named property 'onChange' of types 'InputProps' and 'ReactFormEvents<HTMLInputElement>' are not identical.
```

`FormInputProps` extended `InputProps` and  `ReactFormEvents`, while the last two have the same `onChange` method, but with different signatures. [After changes](https://github.com/Semantic-Org/Semantic-UI-React/pull/1179/files#diff-c606f242c4664fddae43b8c98ee81854L176) in this PR, `FormInputProps` will extend only `InputProps`.

bychwa:
Can someone merge this? Am stuck with this error! :-)

levithomason:

PR #1181, opened 2017-01-18
Test the typings by trying to render a simple component server-side

dennari:
As it stands, the typings seem to break quite easily in new releases. This PR adds a smoke test that tries to render a simple component server-side with typescript. It's run in Circle after the normal test set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist/
 docs/build/
 docs/app/docgenInfo.json
 dll/
+.vscode/
 
 .DS_Store
 .idea/
--- a/circle.yml
+++ b/circle.yml
@@ -12,4 +12,6 @@ test:
     - npm run lint
     - npm run tsd:lint
   post:
+    - npm run build:commonjs
+    - npm run tsd:test
     - bash <(curl -s https://codecov.io/bash)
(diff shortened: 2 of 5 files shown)


levithomason:
Thanks for the PR, we should certainly have some kind of tests around ts.  I cannot see any obvious reason that the Icon export is importing as `undefined` however.

levithomason:
Is there some way to write this as a loop so that all directories are tested for consistency?  We very much favor abstracted automated tests when possible.

[Here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/docs/examples-test.js) we test all doc site examples to ensure they render without errors.  It is an easy win smoke test across the board to ensure the initial render is at least successful.

Another example is the [isConformant](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/commonTests.js#L92) common test for all components.  It ensures the component is properly setup in the library and has base functionality.  Would be great to have something similar for ts.  Potential automated assertions:

- assert every component directory has an `index.d.ts`
- assert that it exports a component of the same name as the folder
- assert that it exports a subcomponent for every other js file in the directory
- assert that it can render the default export

You can see examples of how we use configurable `commonTests` in any test file, like [Button-test.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/elements/Button/Button-test.js).

PR #1206, opened 2017-01-23
Add required prop to Dropdown

mull:
As a Dropdown can easily be used in a form it makes sense to pass the `required` prop down to the `<input>` for HTML5 validation (browser will complain if its value is empty.)

I couldn't get the test working but I believe I wrote it correctly, any pointers? The `test:watch` doesn't seem to be doing what it's supposed to... :open_mouth: 

This PR is still incomplete though as when we select an item the `<input>` value is emptied. Any ideas on working around that?

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -257,6 +257,9 @@ export default class Dropdown extends Component {
      */
     renderLabel: PropTypes.func,
 
+    /** Pass required prop onto input element */
+    required: PropTypes.bool,
+
     /** A dropdown can have its menu scroll. */
     scrolling: PropTypes.bool,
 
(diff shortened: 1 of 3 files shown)


mull:
(I need to fix the commit message as well.)

mull:
Upon further inspection this would require an `<input>` element for a non-search Dropdown as well.

levithomason:
The Dropdown has no formal browser compatible form control (i.e. input element).  Because of this, there is no input for the browser to validate.

The input that this PR touches is the input that takes a search query, not the input that holds the value.  That is because the value is stored in state only.

This is not the case with other controls, such as an `<Input />` which does have a valid browser form control that we can pass props such as `required` to.

The correct solution here is going to be implementing a form validation solution.  See #407 for several options.

PR #1208, opened 2017-01-23
style(Dimmer): update typings

layershifter:
This PR cleanups and updates typings for #1072.

### DIFF
--- a/src/modules/Dimmer/index.d.ts
+++ b/src/modules/Dimmer/index.d.ts
@@ -27,7 +27,7 @@ interface DimmerProps {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onClick?: (event: React.MouseEvent<HTMLInputElement>, data: DimmerProps) => void;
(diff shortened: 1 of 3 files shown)


src/modules/Dimmer/index.d.ts
```diff
@@ -26,15 +27,15 @@ interface DimmerProps {
...
+  onClick?: (event: React.MouseEvent<HTMLInputElement>, data: DimmerProps) => void;
 
   /**
    * Handles click outside Dimmer's content, but inside Dimmer area.
    *
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onClickOutside?: React.MouseEventHandler<HTMLDivElement>;
+  onClickOutside?: (event: React.MouseEvent<HTMLInputElement>, data: DimmerProps) => void;
```
levithomason: Should this and `onClick` not be using `<HTMLDivElement>`?

layershifter: Fixed

PR #1210, opened 2017-01-23
chore(package): support for jsnext:main

### LINKED ISSUE #862: Support jsnext:main and module fields
This is fairly low on the priority list, especially since `v1.0.0` is still looking to be shipped, but I wanted to document it for record keeping.

It would be great for SUIR to provide a bundle that maintains ES6 imports/exports so that bundlers that natively understand ES6 modules (webpack 2, rollup) can [tree shake](https://blog.engineyard.com/2016/tree-shaking) its usage in order to produce smaller bundles. Adding support for `babel-plugin-lodash` is already a great start, especially for webpack 1.0 and similar; this change would take it a step further by allowing bundlers to perform the same thing natively without the overhead of additional tooling.

**Note:** it's important that the code still be preprocessed with babel, because we should not assume anything else about the user's environment.

I believe that this is important piece of functionality, especially as the library grows, because it means that users don't have to be as concerned about the library's weight as a whole and can reap the benefits of targeted bundles without additional work on their part, especially as more bundlers implement tree shaking natively.

See:
* https://github.com/ReactTraining/react-router/pull/3672
* https://github.com/reactjs/redux/issues/1042

### Steps
1. Import SUIR in a bundler that understands ES6 modules.
2. Do not use a plugin to transform imports (such as `babel-plugin-lodash`)

### Expected Result
Unused SUIR code should not be included in the final bundle.

### Actual Result
All SUIR code is put into the bundle regardless of whether or not it was used.

### Version
all

layershifter:
Fixes #862.

Webpack 2 was released, let's make happy its users 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,9 @@
   "name": "semantic-ui-react",
   "version": "0.64.3",
   "description": "The official Semantic-UI-React integration.",
+  "jsnext:main": "dist/es/index.js",
   "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
   "files": [
     "src",
     "dist",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,17 +2,21 @@
   "name": "semantic-ui-react",
   "version": "0.64.3",
   "description": "The official Semantic-UI-React integration.",
+  "jsnext:main": "dist/es/index.js",
   "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
```
layershifter, line 7: `module` field for [Rollup](https://github.com/rollup/rollup/wiki/pkg.module)

layershifter:
Fixed lint issue.

levithomason:
@davezuko would love your eyes here.

dvdzkwsk:
Is there any harm in always using the `lodash` plugin? I'm not sure whether lodash yet offers a package with ES6 modules, but even if they did this plugin shouldn't hurt, right?

Rest looks good.

layershifter:
@davezuko You're right, plugin returned 😄

levithomason (approved):

levithomason:

PR #1215, opened 2017-01-24
fix(Portal): do not take focus after first render

### LINKED ISSUE #1199: Modal TextArea / Input Bug
When typing in a TextArea or Input (in a modal), the TextArea / Input loses focus on keypress.

https://codepen.io/methodx/pen/OWpYaW?editors=0010

levithomason:
Fixes #1199 

This PR addresses a bug introduced in #1154.  The portal takes focus on render for keyboard access, however, it takes focus on _every_ render.  This means if you have a stateful Portal (Modal, Popup, etc) then it will take focus away from form controls on every render.

This PR simply adds a check/test for the initial render and only takes focus on the initial render.  This way, Portal components are still keyboard accessible but also do not take focus from other controls after the initial render.

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentProps.js
+++ b/docs/app/Components/ComponentDoc/ComponentProps.js
@@ -49,7 +49,7 @@ export default class ComponentProps extends Component {
       .join(', ')
 
     const paramDescriptions = params.map(param => (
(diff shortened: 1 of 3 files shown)


levithomason:

PR #1216, opened 2017-01-24
style(Button): update typings and propTypes usage

layershifter:

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -1,5 +1,5 @@
-import _ from 'lodash'
 import cx from 'classnames'
+import _ from 'lodash'
 import React, { Component, PropTypes } from 'react'
 
 import {
@@ -16,37 +16,14 @@ import {
 } from '../../lib'
 import Icon from '../Icon/Icon'
 import Label from '../Label/Label'
-
 import ButtonContent from './ButtonContent'
 import ButtonGroup from './ButtonGroup'
 import ButtonOr from './ButtonOr'
 
 const debug = makeDebugger('button')
 
-const _meta = {
-  name: 'Button',
-  type: META.TYPES.ELEMENT,
-  props: {
-    animated: ['fade', 'vertical'],
-    attached: ['left', 'right', 'top', 'bottom'],
-    color: [
-      ...SUI.COLORS,
-      'facebook',
-      'twitter',
-      'google plus',
-      'vk',
-      'linkedin',
-      'instagram',
-      'youtube',
-    ],
-    floated: SUI.FLOATS,
-    labelPosition: ['right', 'left'],
-    size: SUI.SIZES,
-  },
-}
-
 /**
- * A Button indicates a possible user action
+ * A Button indicates a possible user action.
  * @see Form
  * @see Icon
  * @see Label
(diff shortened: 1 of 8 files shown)


src/elements/Button/Button.js
```diff
@@ -16,37 +16,14 @@ import {
...
-      'linkedin',
-      'instagram',
-      'youtube',
-    ],
-    floated: SUI.FLOATS,
-    labelPosition: ['right', 'left'],
-    size: SUI.SIZES,
-  },
-}
-
```
layershifter, line 47: It's gone 😄

src/elements/Button/Button.js
```diff
@@ -16,37 +16,14 @@ import {
...
 /**
- * A Button indicates a possible user action
+ * A Button indicates a possible user action.
```
layershifter, line 26: Dots.

src/elements/Button/Button.js
```diff
@@ -240,22 +228,21 @@ class Button extends Component {
...
     )
     const rest = getUnhandledProps(Button, this.props)
     const ElementType = getElementType(Button, this.props, () => {
       if (!_.isNil(label) || !_.isNil(attached)) return 'div'
     })
-
-    let computedTabIndex
-    if (!_.isNil(tabIndex)) computedTabIndex = tabIndex
-    else if (disabled) computedTabIndex = -1
-    else if (ElementType === 'div') computedTabIndex = 0
```
layershifter, line 252: Replaced with method `computeTabIndex`, looks much cleaner to me.

src/elements/Button/index.d.ts
```diff
@@ -133,30 +142,59 @@ interface GroupProps {
...
+  /** Groups can be formatted to appear on dark backgrounds. */
+  inverted?: boolean;
 
-  /** Groups can be formatted as labeled icon buttons */
+  /** Groups can be formatted as labeled icon buttons. */
   labeled?: boolean;
 
-  /** Groups can have different sizes */
+  /** Groups can hint towards a negative consequence. */
+  negative?: boolean;
```
layershifter, line 164: `Button.Group` typings had missing props.

src/index.d.ts
```diff
@@ -2,8 +2,6 @@ export type InputType = 'color' | 'date' | 'datetime' | 'datetime-local' | 'emai
 
 export type SemanticCOLORS = 'red' | 'orange' | 'yellow' | 'olive' | 'green' | 'teal' | 'blue' | 'violet' | 'purple' | 'pink' | 'brown' | 'grey' | 'black' | 'twitter' | 'google plus' | 'facebook';
```
levithomason: Looks like a bug here.  Only the true color values are available in every component (red, orange, etc).  The social colors are only available in buttons.

levithomason (changes requested):

src/index.d.ts
layershifter: Indeed, I've updated type.

levithomason (approved):

src/elements/Button/Button.js
levithomason: 🎉

PR #1219, opened 2017-01-24
fix(ComponentExample): use explicit babel presets

levithomason:
The esnext support introduced some babel config that the in-browser editor doesn't understand, #1210.  This means you cannot edit the doc examples.  This PR hard codes the babel presets to overcome the logic surrounding the preset selection.


levithomason:

PR #1224, opened 2017-01-25
fix(Button): fix `tabIndex` in typings

layershifter:
`tabIndex` is optional prop, thanks to @wcatron for [notification](https://github.com/Semantic-Org/Semantic-UI-React/commit/7c2bfee9d9029119e3ccf4fcd84bc430eb97900c#commitcomment-20610490).


levithomason:

PR #1225, opened 2017-01-25
fix(Icon): Incorrect definition in typings

layershifter:
Incorrect definition,  thanks to @wcatron for [notification](https://github.com/Semantic-Org/Semantic-UI-React/commit/727177039daf00c9ffafe14f72d83eb65a6c1a83#commitcomment-20612546).


levithomason:

PR #1226, opened 2017-01-25
Clean up mixed imports in debug.js

marshallswain:
This module was using a mix of es6 and CommonJS imports/exports when it clearly needs CommonJS's dynamic requires.  It has been updated to use CommonJS throughout.

Without this fix, I get the error `require is not defined` in StealJS projects, because it doesn't support mixing syntaxes in a single file the way that Webpack does.

### DIFF
--- a/src/lib/debug.js
+++ b/src/lib/debug.js
@@ -1,4 +1,4 @@
-import isBrowser from './isBrowser'
+const isBrowser = require('./isBrowser')
 let _debug
 const noop = () => undefined
 
@@ -36,7 +36,7 @@ if (isBrowser && process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !
  * debug('Some message')
  * @returns {Function}
  */
-export const makeDebugger = (namespace) => {
+const makeDebugger = exports.makeDebugger = (namespace) => {
   return _debug(`semanticUIReact:${namespace}`)
 }
 
@@ -46,4 +46,4 @@ export const makeDebugger = (namespace) => {
  * import { debug } from 'src/lib'
  * debug('Some message')
  */
-export const debug = makeDebugger('log')
+exports.debug = makeDebugger('log')


marshallswain:
This is the only error preventing me from using this module with [this workaround for using StealJS](https://github.com/Semantic-Org/Semantic-UI-React/pull/1221#issuecomment-275163935) posted by @levithomason

levithomason:
Provided you are importing from `dist`, this change should not be necessary since the built version of this module uses `require()` syntax already:

**semantic-ui-react/dist/commonjs/lib/debug.js**

```jsx
Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.debug = exports.makeDebugger = undefined;

var _isBrowser = require('./isBrowser');

var _isBrowser2 = _interopRequireDefault(_isBrowser);

// ...snip

var makeDebugger = exports.makeDebugger = function makeDebugger(namespace) {
  return _debug('semanticUIReact:' + namespace);
};

// ...snip

var debug = exports.debug = makeDebugger('log');
```

levithomason:
Let me know if I'm over looking anything.

marshallswain:
When I use the dist versions, I get the errors below.  This change lets me use the `src` directly and has no errors when I load the individual modules.

screen shot 2017-01-25 at 11 01 44 am

marshallswain:
I would really appreciate it if we could merge this.  It has a much smaller footprint than the other PR, and I don't think the changes made are anything scary.  If anything, it makes it cleaner to not mix syntax, even though Webpack doesn't have a problem with it.  I did find another way for me to fix the errors that occur when I try to use the `umd` build, but I really prefer to load the individual modules.  I totally understand that we don't want to break stuff with this, but hopefully this is a small enough change that it's acceptable.  If not, I can keep my own copy up to date and use it in my apps.  I'm not intending to be a burden.  😇

Or I suppose I can switch back to Webpack.  :)

levithomason:
The only thing is that it removes ES imports, so there is no tree shaking.  The ~~debug~~ isBrowser module is something that should be able to be dropped from production builds, however, the current change would mean the module will always be included even when not used.

We're currently working toward better ES module support and smaller builds.  This makes me hesitant to merge this.  Also, there is no declarative way to know why this one module is written this way and there are no tests to ensure it is, so it is likely it will break again in the future.

Taken as a whole, I'm thinking the solution here is probably an update in the build system and not necessarily an update in the library.

>I did find another way for me to fix the errors that occur when I try to use the umd build, but I really prefer to load the individual modules

Good news, I'm currently working on individual module support for UMD modules ;)  Also, this will include better individual module support for all modules as we'll likely end up with a single core import so that individual modules are minimally lightweight and do not include boilerplate.

Let's forgo this change here and see if we can solve things on the other front.

marshallswain:
Ah.  That makes sense.  Thanks for your help.

PR #1228, opened 2017-01-25
fix(docs): fix usage of arrow function

layershifter:


docs/app/index.ejs
```diff
@@ -44,7 +44,7 @@
       function removeDimmer() {
         dimmer.style.opacity = '0'
 
-        setTimeout(() => {
+        setTimeout(function() {
```
layershifter, line 47: [Arrow functions](http://caniuse.com/#feat=arrow-functions) are not supported by IE.

levithomason: Whoops! Habit, thanks!

levithomason:
Docs redeployed.

PR #1231, opened 2017-01-26
style(Grid): update typings and propTypes usage

layershifter:


levithomason (approved):
🉑

PR #1233, opened 2017-01-27
perf(package): add stats, optimize, add individual umd components

### LINKED ISSUE #864: perf(lodash): remove use of _.omit (and optimize lodash)
# `_.omit`

Per https://github.com/Semantic-Org/Semantic-UI-React/pull/860, we've **improved the render times of all components by ~12,000 times** with a [4 line](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/getUnhandledProps.js#L39) vanilla `_.omit` replacement.

The issue here was that omit makes copies of its object arguments, the latter argument appears to be deep copied with circular reference checks.  That second argument happened to be `props`, which of course includes children, which of course include children, ...omg.

Every component uses `getUnhandledProps`, on every render, so every update caused every component to make a deep copy of its props and children's props all the way down the render tree util it reached the end of the tree.  Then, repeat this for every node in the tree.  It is obvious why Grid and Grid column were top offenders, they are mostly likely to contain the most deeply nested children.

We have no need to ever make a full copy of objects like this.  We ought to replace all uses of `_.omit` and other methods that copy objects and replace them with our own simple util.

# Lodash must go :(

This would be a massive undertaking but we should, unfortunately, consider completely replacing lodash if we're serious about performance.  This could be done with a shim module that replaces one method at a time (e.g. `lib/lodash`) and uses the lodash method if it has no replacement.

### _What? Why would you?_

It's cherry-picked by `babel-plugin-lodash`, so it the footprint is small, right?

```jsx
import omit from 'lodash/omit'
```

Webpack ^ this **alone** and you get this bundle:

- 96 modules
- 3,172 lines
- `83.6 kB`

_All_ of this code is lodash.  Its internal util dependency tree is _massive_, even when cherry-picking methods. This one method ends up taking what can be done in 4 LOC and turns it into an 84kb module with 96 dependencies.  This is utterly insane.

### _Isn't it really performant, sometimes faster than native?_

Sure, however, as shown above and in #860, we achieved >12,000 times performance gain with a 4 line vanilla JS method.

### _How large is our cherry-picked lodash dependency?_

It's 30% of the uncompressed library size, 278kb.  When Semantic-UI-React is minified and propTypes are striped, it is only `288K` itself.

image

levithomason:
### Doc site bundle stats

[app-stats.md](https://github.com/Semantic-Org/Semantic-UI-React/blob/3aa72bc76aac05c526e2b14a6ed4687995cd11af/stats-app.md)

### `semantic-ui-react` UMD Library Stats

[library-stats.md](https://github.com/Semantic-Org/Semantic-UI-React/blob/3aa72bc76aac05c526e2b14a6ed4687995cd11af/stats-library.md)

Stats for the UMD module bundles also include stats for every component individually.

## Lodash Optimization

Fixes #864

The [`lodash-webpack-plugin`](https://github.com/lodash/lodash-webpack-plugin) was tested during this process, however, very little gain was realized as our usage required most options to be enabled.  The errors it caused in PhantomJS and tests were extremely cryptic and unclear.  I opted to omit this plugin for simplicity and assurance that nothing would break.

We have already implemented the babel plugin for cherry-picking and removed the massive and slow `_.omit` method.

## UMD bundle and individual components

All components are now individually published as UMD modules.  Shared code between components is now included in `semantic-ui-react-core.js`.

All components can be used as a single bundle, `semantic-ui-react.js`.  Include `semantic-ui-react-core.js` in this case as well.

## Original, minified, and gzipped

Every UMD component now comes in full `.js`, compressed `.js.gz`, minified `.min.js`, and minified + compressed `.min.js.gz`.

## Stats in VC

Stats for the doc site _and_ the UMD modules are included in the repo now.  They are generated on every build.

# TODO

### Only strip propTypes in minified UMD modules

Currently, all UMD modules strip propTypes.  It would be nice to retain them in the unminified builds and only remove them in the minified builds.

### Ensure stats are generated for PRs

UMD stats are most helpful since they show the size of every file with every combination of build and compression.  However, **UMD stats are not built in development** when using `npm start`.

Ideally, the stats files would be updated and committed with every PR so we can keep tabs on changes in the library size.  Users must `npm run build` then commit to get a clean build of the stats.


levithomason:
As nice as this was, it will need redone with the new build system.  We should also save this work for when move to a monorepo.  Closing as it is stale.

PR #1234, opened 2017-01-27
Fix typings for Menu Item onClick handler

shanehyde:
The typings are missing a correct definition for onClick on Menu.Item to add the props as the 2nd parameter.

The typings for Button are correct, so this is a copy of the onClick definition from there for Menu.Item


shanehyde:
I will fix this up and resubmit

PR #1235, opened 2017-01-27
fix(Menu): fix typings for Menu.Item

shanehyde:
The typings are missing a correct definition for onClick on Menu.Item to add the props as the 2nd parameter.

The typings for Button are correct, so this is a copy of the onClick definition from there for Menu.Item

### DIFF
--- a/src/collections/Menu/index.d.ts
+++ b/src/collections/Menu/index.d.ts
@@ -159,7 +159,7 @@ interface MenuItemProps extends ReactMouseEvents<HTMLElement> {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onClick?: (event: React.MouseEvent<HTMLElement>, data?:MenuItemProps) => void;
+  onClick?: (event: React.MouseEvent<HTMLElement>, data: MenuItemProps) => void;
 
   /** A menu item can take right position. */
   position?: 'right';
--- a/src/collections/Menu/index.d.ts
+++ b/src/collections/Menu/index.d.ts
@@ -154,6 +154,13 @@ interface MenuItemProps extends ReactMouseEvents<HTMLElement> {
   /** Internal name of the MenuItem. */
   name?: string;
 
+  /**
+   * Called after user's click.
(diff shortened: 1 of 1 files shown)


src/collections/Menu/index.d.ts
```diff
@@ -154,6 +154,13 @@ interface MenuItemProps extends ReactMouseEvents<HTMLElement> {
   /** Internal name of the MenuItem. */
   name?: string;
 
+  /**
+   * Called after user's click.
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onClick?: (event: React.MouseEvent<HTMLElement>, data?:MenuItemProps) => void;
```
squirly: there should be space between `data?:` and `MenuItemProps`

layershifter: `data` will be [always](https://github.com/shanehyde/Semantic-UI-React/blob/647986443152232a9793039b8645779f90832dff/src/collections/Menu/MenuItem.js#L90) passed.

```diff
-onClick?: (event: React.MouseEvent<HTMLElement>, data?:MenuItemProps) => void;
+onClick?: (event: React.MouseEvent<HTMLElement>, data: MenuItemProps) => void;
```

shanehyde:
I have made both those changes and added them to the PR

levithomason:

tilor:
Menu.Item throws a compiler error in 0.64.5 (this error did not happen in 0.64.0):

error TS2430: Interface 'MenuItemProps' incorrectly extends interface 'ReactMouseEvents<HTMLElement>'.
  Types of property 'onClick' are incompatible.
    Type '(event: MouseEvent<HTMLElement>, data: MenuItemProps) => void' is not assignable to type 'EventHandler<MouseEvent<HTMLElement>>'.

dylankiss:
I get the same error when using 0.64.5.

levithomason:
@layershifter ?

layershifter:
I've forget to make a response there, it was fixed in #1264.

PR #1246, opened 2017-01-29
feat(Button): add `floated` prop to Button.Group

mohammedkhadar:
This pull request fixes #1245

PR #1248, opened 2017-01-29
fix(Button): add `floated` to typings of ButtonGroup

layershifter:
#1246 added prop, but typings  were not done there.


levithomason:
Thanks again. Would be great to get a smoke test for this so that we aren't missing props between the two. I keep forgetting to update typings. 

We could assert the interface for every component has props for every propType key.

levithomason:

Issue #1244, opened 2017-01-28
feat(Dropdown): add `closeOnChange` prop


apiv:
### Steps
1. With a single-selection dropdown, select an item. The menu closes on select.
![single](https://d17oy1vhnax1f7.cloudfront.net/items/0M190h0s0J16190d0F2Q/Screen%20Recording%202017-01-28%20at%2002.59%20PM.gif?v=2bcd869a)
2. With a multi-selection dropdown, the menu always stays open.
![multi](https://d17oy1vhnax1f7.cloudfront.net/items/2y092g243u1P0I100L3t/Screen%20Recording%202017-01-28%20at%2003.00%20PM.gif?v=39b29b73)

### Proposed changes
There are use cases for both menu open/close behaviors, but in some cases, I may want to close the dropdown on select. I'm proposing that we add a `closeOnChange` or `closeOnSelect` option to `Dropdown`, defaulting to `true` for single selects and `false` for multi-selects.

### Version
v0.64.4

### Testcase
http://react.semantic-ui.com/modules/dropdown#search-selection
http://react.semantic-ui.com/modules/dropdown#multiple-search-selection

I'm willing to submit a PR for this if we decide this option is worth adding to `Dropdown`.

levithomason:
I'm all for this change. Let's go with `closeOnChange`. 

I'd think this would only close on changes invoked by user interaction, such as, click and enter. This way programmatic updates don't usurp user control. Thoughts?

apiv:
I think that fits the pattern, since changing `value` does not trigger `onChange`, only user interaction does. I'll submit a pull request shortly.

levithomason:
@apiv Heads up, if the PR description contained `fixes #1244` it would auto close the issue.  However, the string must be one of the [supported string values](https://github.com/blog/1506-closing-issues-via-pull-requests).  The actual PR included `fulfills issue #1244` so the issue remained open after merge.

Fixed in #1244, thanks for the work!

andrewleith:
I'm having some trouble with this.  When I use a dropdown without `closeOnChange`, it automatically closes when I select an option, which is exactly what I want.  When I add the `simple` prop to the dropdown (because I would like this menu to open on hover), it no longer closes when I select an option.  Why would the `simple` prop even affect this behavior at all?

Adding `closeOnChange` kind of works: it closes the dropdown, but only after I make a *second* selection.

Any advice?

levithomason:
[...]

PR #1252, opened 2017-01-30
feat(Dropdown): add closeOnChange prop

apiv:
fulfills issue #1244, adding a `closeOnChange` prop to Dropdown, with usage example added to the docs.

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/index.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/index.js
@@ -17,6 +17,12 @@ const DropdownUsageExamples = () => (
     />
 
     <ComponentExample
+      title='Close On Change'
+      description='A multiple selection dropdown can close when the user changes its value.'
+      examplePath='modules/Dropdown/Usage/DropdownExampleCloseOnChange'
+    />
+
+    <ComponentExample
       title='Uncontrolled'
       description='A dropdown can behave like an uncontrolled input.'
       examplePath='modules/Dropdown/Usage/DropdownExampleUncontrolled'
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -85,6 +85,13 @@ export default class Dropdown extends Component {
     /** Whether or not the menu should close when the dropdown is blurred. */
     closeOnBlur: PropTypes.bool,
 
+    /**
+     * Whether or not the menu should close when a value is selected from the dropdown.
+     * By default, multiple selection dropdowns will remain open on change, while single
+     * selection dropdowns will close on change.
+     */
+    closeOnChange: PropTypes.bool,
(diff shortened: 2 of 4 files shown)


apiv:
onBlur/onFocus are firing incorrectly with this option, looking into it now

docs/app/Examples/modules/Dropdown/Usage/index.js
```diff
@@ -17,6 +17,12 @@ const DropdownUsageExamples = () => (
     />
 
     <ComponentExample
+      title='Close On Change'
+      description='A multiselect can close when the user changes its value.'
```
levithomason: Nit: for consistency with the props, can we go with `A multiple selection dropdown...` here?

src/modules/Dropdown/Dropdown.js
```diff
@@ -85,6 +85,9 @@ export default class Dropdown extends Component {
     /** Whether or not the menu should close when the dropdown is blurred. */
     closeOnBlur: PropTypes.bool,
 
+    /** Whether or not the menu should close when a value is selected from the dropdown */
```
levithomason: Since the default is hidden from the user (logic based in the method) perhaps we should note the default here.  Let's add the period to the end of this sentence while we're at it.

src/modules/Dropdown/Dropdown.js
```diff
@@ -480,6 +483,15 @@ export default class Dropdown extends Component {
     if (onChange) onChange(e, { ...this.props, value })
   }
 
+  closeOnChange = (e) => {
+    const { closeOnChange, multiple } = this.props
+    const shouldClose = _.isUndefined(closeOnChange)
+      ? !multiple
+      : closeOnChange
```
levithomason, line 494: The false branch of this logic is untested, this is why Codecov is failing the PR:

image

>**tip**: the coverage data overlayed here in GH is added by the [Codecov extension](https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf?hl=en-US)

Even though the other branches are executed during tests, we don't actually have assertions for this new feature.  We'll need a new `describe('closeOnChange', ... )` block in the `Dropdown-test.js` where we can test this new prop.

levithomason (changes requested):
Looks great so far, couple small nits and some tests will finish it up.

src/modules/Dropdown/Dropdown.js
apiv: Thanks for the heads up, specs added, changes made and re-pushed

levithomason (approved):

levithomason:
Superb, thanks much!  I'll release this within the hour.

levithomason:
As promised, released in `semantic-ui-react@0.64.5`.

PR #1254, opened 2017-01-30
style(Message): update typings and remove propTypes

layershifter:

---

Also, it converts `Message` to statefull components and update `onDismiss` behavior.


levithomason:
Perfect, thanks.  I made one trivial update to the `MessageItem` description (added "list").

levithomason:

PR #1256, opened 2017-01-30
fix(DropdownItem): add className to element shorthands

levithomason:
When passing elements to the `text` or `description` prop for a `DropdownItem`, the appropriate `className` values were not added.  The `className` was only added for primitive value shorthand.

This PR moves the `className` values to the `defaultProps` callback for each shorthand so they are always added to all resulting shorthand elements.


levithomason:

Issue #1243, opened 2017-01-28
bug(Dropdown): `multiple search` dropdown should refocus the `input` element on select of an option


apiv:
### Steps
1. Type search term in multiple select
2. Click one of the results

### Expected Result
The result should be added to the list, and I should be able to continue typing to search for another option.

### Actual Result
The result is added to the list, but focus is taken away from the input box. In order to search for another item, I need to click back into the input. Backspacinc still works, and still removes the label from the list when pressed.

![actual](https://d17oy1vhnax1f7.cloudfront.net/items/1H0G2u2V1b1e1F1A2n1U/Screen%20Recording%202017-01-28%20at%2002.53%20PM.gif?v=4b3428b8)

### Version
v0.64.4

### Testcase
http://react.semantic-ui.com/modules/dropdown#multiple-search-selection

More than willing to submit a PR for this if the maintainers of this repo agree with the expected behavior above.

levithomason:
Agree! This will prove tricky given the side effects of focus and blur (opens/closes the dropdown, fires on focus, etc). These things will somehow need to be circumvented so that we aren't firing unnecessary state changes and callbacks when an item is activated. 

Also, it seems this should only happen on user click of an item.

levithomason:
Fixed in #1247

PR #1257, opened 2017-01-30
feat(Dropdown): multiple search dropdown should refocus the search on select of an option

apiv:
fixes issue #1243 

Before:
before

After:
after

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -422,8 +422,8 @@ export default class Dropdown extends Component {
       } else {
         document.addEventListener('keydown', this.moveSelectionOnKeyDown)
         document.addEventListener('keydown', this.selectItemOnEnter)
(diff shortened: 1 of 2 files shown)


levithomason:
Checked out the code here and tried the component.  Looks great, thanks!  Could we get an assertion for this behavior as well?  There is a `describe('selecting items', ... ')` around line 1094 (nested in `describe('multiple', ... )`) in `Dropdown-test.js` that could house the `it()` for this.

The `wrapperMount` method will mount the dropdown to the document so you can check the `activeElement` in tests for focus assertions.

apiv:
Sure thing, glad to contribute, our app uses `semantic-ui-react` as the base for most of our UI. I've added the assertion and rebased / re-pushed.

levithomason:
Superb, thanks for the rapid update 👍

levithomason:

PR #1258, opened 2017-01-31
fix(Editor): use handledProps instead of propTypes

levithomason:
The editor uses propTypes to generate completions.  However, we now strip propTypes in production.  The docs, therefore, have no prop completions.

This PR gets prop completions from the `handledProps` array, which includes all props.


levithomason:

PR #1263, opened 2017-01-31
fix(docs): fix checkbox inputType docs to use correct props

fracmak:
Also changed the verbiage a bit to clear up some confusion we had over at my office


levithomason:
Yep, that one was old news!  Thanks much :)

PR #1265, opened 2017-01-31
style(Label): update typings and remove propTypes

layershifter:


levithomason:
💥

levithomason:

PR #1266, opened 2017-01-31
feat(docs): restore enum expansion on docs page

fracmak:
This restores the enum display in the docs lost by #731

To get it to work I'm now parsing the docgen to get the correct values which has it's limitations. docgen is very literal when it comes to programatic propTypes, so I've tried expanding the easy ones but there's a few occurrences that are more difficult to make work (see last screenshot for an example)
screenshot 2017-01-31 15 32 16
screenshot 2017-01-31 15 32 20
screenshot 2017-01-31 15 32 34

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentProps.js
+++ b/docs/app/Components/ComponentDoc/ComponentProps.js
@@ -2,6 +2,12 @@ import _ from 'lodash'
 import React, { Component, PropTypes } from 'react'
 
 import { Label, Table } from 'src'
+import { SUI } from 'src/lib'
+
+const descriptionExtraStyle = {
+  fontSize: '0.95em',
+  color: '#777',
+}
 
 /**
  * Displays a table of a Component's PropTypes.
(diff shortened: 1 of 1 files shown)


docs/app/Components/ComponentDoc/ComponentProps.js
```diff
@@ -65,6 +84,64 @@ export default class ComponentProps extends Component {
     )
   }
 
+  expandEnums = (value) => {
+    const parts = value.split('.')
+    if (parts[0] === 'SUI') {
+      return SUI[parts[1]]
+    } else if (parts[1] === '_meta') {
+      // these look like Sidebar._meta.props.animation
+      return _.get(this.props, `meta.props.${parts[3]}`, value)
```
levithomason: `_meta.props` is in the process of going away and the values are being inlined into the `propTypes` checkers, [example](https://github.com/Semantic-Org/Semantic-UI-React/pull/1216/files#diff-234ffdf68307872a89a512e653bea71aL29).  You can see progress [here](https://github.com/Semantic-Org/Semantic-UI-React/issues/524).  Given this, we're removing all traces of `_meta.props` from the codebase.

In this case, we should probably just not parse the `meta.props` here at all, for now.  Most the enums will show up automatically because the values can be parsed by react-docgen.  Those that cannot will be those that have more complicated checkers, which, we'll have to solve another way.

levithomason (changes requested):
Thanks, I'd like to get this support back.  There are some reasons it was removed initially.  I've left a comment here to address that I think will allow us to add the basic functionality back in.

docs/app/Components/ComponentDoc/ComponentProps.js
fracmak: Removed

levithomason (approved):

levithomason:
Thanks!

PR #1273, opened 2017-02-02
fix(Portal): fix several event related behaviors

### LINKED ISSUE #1076: Popup doesn't work with modal
The popup appears, but under the modal. Give it a try. Not really sure if this is React specific or CSS. Apologies if it's the latter.

### LINKED ISSUE #1251: Bug(Popup): Clicking on another Popup's trigger doesn't close open Popups.
### Steps
1. Open an on="click" popup
2. Click on another trigger for a different on="click" popup while the first is still open
### Expected Result
The first Popup should close and the second should open

### Actual Result
Both Popups stay open.

### Version
0.64.4
### Testcase
Here's an example:
http://codepen.io/anon/pen/XpVqKq

It happens on the doc page as well:

screen shot 2017-01-30 at 9 53 14 am

### LINKED ISSUE #1271: fix(Portal): focus is stolen when clicking away to close
Introduced in #1154, the Portal steals focus from other controls on close.

### Steps
1. Create a popup that opens on click
1. Click the trigger to open the popup
1. Click an input element outside of the popup to focus it

### Expected Result
The popup should close and the input should receive focus.

### Actual Result
The popup closes, but the popup's trigger steals focus from the input.

### Version
0.64.5

### Testcase
http://codepen.io/levithomason/pen/MJVoQY

levithomason:
Fixes #1271 Do not steal focus on portal close
Fixes #1251 Do not stop propagation of events
Fixes #1076 Popups in Modals open in the correct position

This PR also:
- fixes the `mountNode=null` bug in for the Modal
- adds Modal prop `closeOnDimmerClick`
- keeps a Portal open when blurring the trigger and focusing the Portal

### DIFF
--- a/docs/app/Examples/modules/Dimmer/States/DimmerExampleActive.js
+++ b/docs/app/Examples/modules/Dimmer/States/DimmerExampleActive.js
@@ -1,15 +1,15 @@
 import React from 'react'
-import { Dimmer, Segment } from 'semantic-ui-react'
+import { Dimmer, Image, Segment } from 'semantic-ui-react'
 
 const DimmerExampleActive = () => (
   <Segment>
     <Dimmer active />
 
     <p>
-      <img src='http://semantic-ui.com/images/wireframe/short-paragraph.png' />
+      <Image src='/assets/images/wireframe/short-paragraph.png' />
     </p>
     <p>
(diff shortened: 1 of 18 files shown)


levithomason:

PR #1274, opened 2017-02-02
fix(Modal): add missing typings

levithomason:
Typings updates weren't pushed before merging #1273


levithomason:

PR #1279, opened 2017-02-03
fix(Dropdown): fix arrow keys after no results

### LINKED ISSUE #1275: bug(Dropdown): Arrow keys are not working after `No results found.`
### Steps
1. Go to [search-selection dropdown](http://react.semantic-ui.com/modules/dropdown#search-selection)
2. Type something (e.g. `af`)
3. If you will use arrow up and down keys, everything is working as expected
3. Add one letter to search field, so there will be no results (`afs`)
4. Remove this letter (`af`)
5. Arrow keys are not working anymore (`Uncaught TypeError: Cannot read property 'disabled' of undefined`)

levithomason:
Fixes #1275

The Dropdown attempts to set the `selectedIndex` (highlighted item) to `undefined` after encountering the "No results." message.

This PR resets the `selectedIndex` to `0` after every search change.  This is inline with SUI core and was our original intention for this component.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -709,7 +709,7 @@ export default class Dropdown extends Component {
     if (search && newQuery && !open) this.open()
 
     this.setState({
-      selectedIndex: this.getEnabledIndices()[0],
+      selectedIndex: 0,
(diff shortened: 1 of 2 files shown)


levithomason:

Issue #1144, opened 2017-01-10
feat(Accordion): support Title 'as' prop in dynamic accordion titles


uc2pac:
There is only one way to create dynamic panels for accordion. 

`const panels = () => {
    title: string,
    content: custom
}`

There are cases when we need to create accordion titles with different components inside. For example some labels on the right. But since dynamic title can be only string there is no way to achieve that.

Can you please allow custom type or a way to use 'as' prop for dynamic titles?

levithomason:
Indeed, the TODO [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Accordion/Accordion.js#L67) would allow this.  The panel should be a component, then, it can be defined using shorthand like all our other components.  This would allow the `title` prop to be a number, string, props object, or another element.

We can also take a short term PR that just updates the type check for this prop.  Note that you can pass an element currently and it will work, you will just see an overzealous prop warning in development mode :/

rkostrzewski:
I'm happy to implement the TODO for Accordion component (`* TODO: AccordionPanel should be a sub-component`) but I'm unsure whether I understand correctly what is meant by this. Should it look like this? (I guess there would be no need for `<Accordion panels={...}` then)

```
<Accordion>
  <Accordion.Panel>
    <Accordion.Title>
       <Icon name='dropdown' />
       What is a dog?
    </Accordion.Title>
    <Accordion.Content>
       <p>
         A dog is a type of domesticated animal. Known for its loyalty and faithfulness,
         {' '}it can be found as a welcome guest in many households across the world.
       </p>
    </Accordion.Content>
  </Accordion.Panel>
  ...
</Accordion>
```

levithomason:
@rkostrzewski The goal is to keep both.  You can reference other components like the Message, Menu, and Header to see how subcomponents and children relate to shorthand props.

levithomason:
Released in `semantic-ui-react@0.66.0`.

PR #1281, opened 2017-02-04
feat(Accordion): support panel's title as custom element #1144

rkostrzewski:
Short term fix for #1144

I've also added custom key to panels object named `key` because React is throwing errors when using own components as title or content. Using `panel.title` as source of component key at ``<AccordionTitle key={`${panel.title}-title`} active={isActive} onClick={onClick}>`` obviously does not work in such scenario as `panel.title` becomes `[Object object]` when cast to string.


src/modules/Accordion/Accordion.js
```diff
@@ -63,14 +63,16 @@ export default class Accordion extends Component {
...
      * Mutually exclusive with children.
      * TODO: AccordionPanel should be a sub-component
      */
     panels: customPropTypes.every([
       customPropTypes.disallow(['children']),
       PropTypes.arrayOf(PropTypes.shape({
+        key: PropTypes.string,
         active: PropTypes.bool,
-        title: PropTypes.string,
+        title: customPropTypes.contentShorthand,
```
levithomason, line 66: If this prop is changed to shorthand, which we'd love, then its value should be passed to:

```jsx
AccordionTitle.create(title)
```

The `create()` method needs to be added to the title and content components to support this.  You can reference other simple components with shorthand prop for subcomponents and children, such us, the Header.  The Card is a good example of a component that has many levels of children.

levithomason (changes requested):
The

src/modules/Accordion/Accordion.js
rkostrzewski: Thanks - I didn't quite get that from contributing guidelines.

Corrected (hopefully correctly :smile:)

test/specs/modules/Accordion/Accordion-test.js
```diff
@@ -254,10 +254,28 @@ describe('Accordion', () => {
...
+
+    it('adds custom element title and custom element content sibling children', () => {
+      const panels = [{
+        key: 'panel-1',
+        title: (<h1>{faker.lorem.sentence()}</h1>),
+        content: (<h2>{faker.lorem.paragraph()}</h2>),
+      }]
+      const wrapper = mount(<Accordion panels={panels} />)
+
+      expect(wrapper.childAt(0).containsMatchingElement(panels[0].title)).to.equal(true)
```
rkostrzewski: Oh, I've made a mistake - this should still have 'title' class added

rkostrzewski: Fixed. Also added example to docs showcasing the feature.

src/modules/Accordion/Accordion.js
levithomason: No worries, our shorthand API and tools were developed after the guide and the guide was, unfortunately, not adequately updated.  Sorry about that!

src/modules/Accordion/AccordionTitle.js
```diff
@@ -43,17 +50,29 @@ export default class AccordionTitle extends Component {
     parent: 'Accordion',
   }
 
+  static create = createShorthandFactory(AccordionTitle, content => ({ content }))
```
levithomason: This method will need to be placed after the class is created, otherwise, AccordionTitle will be undefined in this context:


```jsx
class AccordionTitle extends Component {
  // ...
}

AccordionTitle.create = createShorthandFactory(AccordionTitle, content => ({ content }))
```

There is also a common test for this:

**AccordionTitle-test.js**
```js
common.implementsCreateMethod(AccordionTitle)
```

src/modules/Accordion/AccordionTitle.js
```diff
@@ -43,17 +50,29 @@ export default class AccordionTitle extends Component {
...
+
   handleClick = (e) => {
     const { onClick } = this.props
 
     if (onClick) onClick(e, this.props)
   }
 
+  renderContent(content) {
+    return isValidElement ? content : (
```
levithomason: The `isValidElement` is missing its argument `isValidElement(content)`, however, this can be removed entirely and the `content` always rendered in place.

Also, we try to avoid all styling and leave this to the user's CSS theme of choice.  Let's remove the extra wrapping `div` and inline style by going with two explicit returns in the [render()](url) method:

```jsx
if (!_.isNil(content)) {
  return (
    <ElementType {...rest} className={classes} onClick={this.handleClick}>
      <Icon name='dropdown' />
      {content}
    </ElementType>
  )
}

return (
  <ElementType {...rest} className={classes} onClick={this.handleClick}>
    {children}
  </ElementType>
)
```

docs/app/Examples/modules/Accordion/Types/AccordionExamplePanelsProp.js
```diff
@@ -1,10 +1,22 @@
-import React from 'react'
-import { Accordion } from 'semantic-ui-react'
+import React, { PropTypes } from 'react'
+import { Accordion, Icon } from 'semantic-ui-react'
 import faker from 'faker'
 import _ from 'lodash'
 
-const panels = _.times(3, () => ({
-  title: faker.lorem.sentence(),
+const CustomTitle = ({ text }) => (
```
levithomason: Would prefer we create a separate example that shows shorthand element usage and leave this one for shorthand string usage.

levithomason (changes requested):
I've left a few comments, LMK if you have any issues or questions.  Also, much of our latest practices for docs are not well documented (my apologies) but I can help out if you like once these first round of changes are made.

src/modules/Accordion/AccordionTitle.js
rkostrzewski: Actually, AccordionTitle is defined in such scenario and has right value ([as shown in babel example](https://babeljs.io/docs/plugins/transform-class-properties/)). Nevertheless I've placed it after the class is created because there was some issue with code coverage not seeing the line as covered.

src/modules/Accordion/AccordionTitle.js
rkostrzewski: Oh my. I've fixed isValidElement issue and left the logic intact (apart from inserting the `div`) as I strongly believe users should have the ability to customise the title fully. Without the check for element if I pass
```
title: (
   <div>
      <Icon name='heart' />
      What a lovely title
   </div>
)
```
it will render two icons (dropdown and heart). On the other side when somebody passes text the dropdown icon should be rendered.

docs/app/Examples/modules/Accordion/Types/AccordionExamplePanelsProp.js
rkostrzewski: Done, let me now if there should be any description or any other title (writing good documentation is really really hard :smile:)

levithomason:
>I've fixed isValidElement issue and left the logic intact (apart from inserting the div) as I strongly believe users should have the ability to customise the title fully. Without the check for element if I pass

Hm, this now assumes that all elements will always include an Icon.  If I pass an element, such as a `<Header />` the icon will disappear.  I'd think most elements will not also contain a custom icon.  My intuition says most elements will be titles only.

I also think the shorthand API should always be consistent (always icon or never icon).  The majority use case would be to include the icon, presumably.  I fear confusing users if the accordion sometimes has an icon and sometimes doesn't with no explicit visible reason as to why that would be the case (vs something like an icon prop).

The idea with shorthand is to cover the majority of use cases while leaving the subcomponent API for the detailed use cases.  If users are importing Icon and building a complete title, I'd direct them to use `<Accordion.Title />` with their element as the child instead of the shorthand.

LMK your thoughts on this.

docs/app/Examples/modules/Accordion/Types/AccordionExamplePanelsPropWithCustomTitleAndContent.js
```diff
@@ -0,0 +1,38 @@
...
+  content: PropTypes.string.isRequired,
+}
+
+const panels = _.times(3, i => ({
+  key: `panel-${i}`,
+  title: (<CustomTitle title={faker.lorem.sentence()} />),
+  content: (<CustomContent content={faker.lorem.sentence()} />),
+}))
+
+const AccordionExamplePanelsProp = () => (
```
levithomason: The default export should match the filename.  Also, this would be a "Usage" example `.../Examples/modules/Accordion/Usage`.  We put all our SUIR specific examples here, for now.  We reserve the types/states/variation/etc. for SUI core parity examples.

rkostrzewski: Corrected :smile:

rkostrzewski:
> I also think the shorthand API should always be consistent (always icon or never icon). The majority use case would be to include the icon, presumably. I fear confusing users if the accordion sometimes has an icon and sometimes doesn't with no explicit visible reason as to why that would be the case (vs something like an icon prop).

Yeah, you're probably right. It does sound confusing when you put it that way.

> The idea with shorthand is to cover the majority of use cases while leaving the subcomponent API for the detailed use cases. If users are importing Icon and building a complete title, I'd direct them to use <Accordion.Title /> with their element as the child instead of the shorthand.

Using Accordion.Title would be lovely, unfortunately you can't easily render them programatically unless there is Accordion.Panel (at least yet, tough any solution of implementing Accordion.Panel seems hacky).

Altogether, I've changed the component so that it always renders the icon.

levithomason:
I've just rebased this to the latest master and pushed a couple minor fixes.  Will merge on pass.

levithomason (approved):

levithomason:

PR #1282, opened 2017-02-05
style(Confirm): refactor, update typings and propTypes usage

### LINKED ISSUE #1290: feat(Confirm): use shorthands for content and buttons definition
### Steps
1. Render the `Confirm` addon with a content that is an object: `<Confirm content={<b>hi!</b>} {...rest} />`

### Expected Result
It should be rendered without any warning.

### Actual Result
It renders correctly, but gives a warning: **content prop have to be a string**.

See src/addons/Confirm/Confirm.js#L53

layershifter:
Fixes #1290.

---

**Warning**. This PR updates behavior of `onCancel` and `onConfirm`, now it handled inside `Confirm` component.

@levithomason I think we need to mark this change as _breaking_.  Before `data` was `Button`'s props while after this PR it will `Confirm`s props.

### DIFF
--- a/src/addons/Confirm/Confirm.js
+++ b/src/addons/Confirm/Confirm.js
@@ -1,68 +1,105 @@
 import _ from 'lodash'
-import React, { PropTypes } from 'react'
+import React, { Component, PropTypes } from 'react'
 
-import { getUnhandledProps, META } from '../../lib'
+import {
+  customPropTypes,
+  getUnhandledProps,
+  META,
+} from '../../lib'
 import Button from '../../elements/Button'
 import Modal from '../../modules/Modal'
 
 /**
- * A Confirm modal gives the user a choice to confirm or cancel an action
+ * A Confirm modal gives the user a choice to confirm or cancel an action/
  * @see Modal
  */
-function Confirm(props) {
-  const { open, cancelButton, confirmButton, header, content, onConfirm, onCancel } = props
-  const rest = getUnhandledProps(Confirm, props)
+class Confirm extends Component {
+  static propTypes = {
+    /** The cancel button text. */
+    cancelButton: customPropTypes.itemShorthand,
 
-  // `open` is auto controlled by the Modal
-  // It cannot be present (even undefined) with `defaultOpen`
-  // only apply it if the user provided an open prop
-  const openProp = {}
-  if (_.has(props, 'open')) openProp.open = open
+    /** The OK button text. */
+    confirmButton: customPropTypes.itemShorthand,
 
-  return (
-    <Modal {...openProp} size='small' onClose={onCancel} {...rest}>
-      {header && <Modal.Header>{header}</Modal.Header>}
-      {content && <Modal.Content>{content}</Modal.Content>}
-      <Modal.Actions>
-        <Button onClick={onCancel}>{cancelButton}</Button>
-        <Button primary onClick={onConfirm}>{confirmButton}</Button>
-      </Modal.Actions>
-    </Modal>
-  )
-}
+    /** The ModalContent text. */
+    content: customPropTypes.itemShorthand,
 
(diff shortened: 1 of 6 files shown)


src/addons/Confirm/Confirm.js
```diff
@@ -1,68 +1,103 @@
...
+  static propTypes = {
+    /** The cancel button text. */
+    cancelButton: PropTypes.string,
 
-Confirm._meta = {
-  name: 'Confirm',
-  type: META.TYPES.ADDON,
-}
+    /** The OK button text. */
+    confirmButton: PropTypes.string,
```
layershifter: I'm also thinking about shorthand there too.

```diff
-confirmButton: PropTypes.string,
+confirm: customPropTypes.itemShorthand,
```

levithomason: That would be great if you'd like to implement it. 👍

src/addons/Confirm/Confirm.js
```diff
@@ -1,68 +1,103 @@
...
+    if (_.has(this.props, 'open')) openProp.open = open
 
-Confirm.defaultProps = {
-  cancelButton: 'Cancel',
-  confirmButton: 'OK',
-  content: 'Are you sure?',
+    return (
+      <Modal {...rest} {...openProp} size='small' onClose={this.handleCancel}>
+        {createShorthand(Modal.Header, val => ({ children: val }), header)}
+        {createShorthand(Modal.Content, val => ({ children: val }), content)}
```
levithomason: I'm slowing trying to remove use of `createShorthand` in favor of adding `create()` methods to the component when needed.  We do plan to have shorthand for the Modal, which we currently don't, so having the `create()` methods for ModalHeader and ModalContent would be helpful.

Let's remove `createShorthand` and add the `create()` method for these components instead.

src/addons/Confirm/Confirm.js
```diff
@@ -1,68 +1,103 @@
...
 
-  /** The cancel button text */
-  cancelButton: PropTypes.string,
+    /** The ModalHeader text. */
+    header: customPropTypes.itemShorthand,
 
-  /** The OK button text */
-  confirmButton: PropTypes.string,
+    /**
+     * Called when the Cancel button is clicked.
```
levithomason: Let's correct this, "Called when the Modal is closed without clicking confirm."  Since, it is called on Escape and click outside as well.

src/addons/Confirm/index.d.ts
```diff
@@ -1,26 +1,35 @@
...
-  onConfirm?: () => void;
-
-  /** Whether or not the modal is visible */
+  content?: any;
+
+  /** The ModalHeader text. */
+  header?: any;
+
+  /**
+   * Called when the Cancel button is clicked.
```
levithomason: See my comment for the `onClick` propType.

levithomason (changes requested):
Couple changes related to shorthand.  If you'd like to do these in a separate PR, we can do that as well.  LMK.

levithomason:
>**Warning.** This PR updates behavior of onCancel and onConfirm, now it handled inside Confirm component.
>
>@levithomason I think we need to mark this change as breaking. Before data was Button's props while after this PR it will Confirms props.

Thanks, this is how it should've been written initially.  I don't think we'll need to make this one breaking since the previous behavior was a bug and this fixes it.  Also, the callback did not advertise a second `data` parameter, so it was not a supported feature.

src/addons/Confirm/Confirm.js
layershifter: Okay, I'll make it today :smile:

layershifter:
@levithomason I've updated PR, take a look 😄

levithomason (approved):

levithomason:

PR #1283, opened 2017-02-05
chore(coverage): remove docs from coverage reports

layershifter:
It looks very strange to me that we include our docs to coverage reports. This PR fixes it.


levithomason:

PR #1285, opened 2017-02-05
style(Feed): update typings and propTypes usage

layershifter:

### DIFF
--- a/src/views/Feed/index.d.ts
+++ b/src/views/Feed/index.d.ts
@@ -1,7 +1,8 @@
-import {SemanticSIZES} from '../..';
 import * as React from 'react';
 
 interface FeedProps {
+  [key: string]: any;
+
   /** An element type to render as (string or function). */
   as?: any;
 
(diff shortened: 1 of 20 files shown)


src/views/Feed/index.d.ts
```diff
@@ -128,18 +136,20 @@ interface FeedExtraProps {
   className?: string;
 
   /** Shorthand for primary content. */
-  content?: any;
+  content?: React.ReactNode;
 
   /** An event can contain additional information like a set of images. */
-  images?: any;
+  images?: boolean | Array<any>;
```
levithomason, line 142: Should this be array of React.ReactNode?

levithomason (commented):
Just one comment regarding the image shorthand array type.

src/views/Feed/index.d.ts
layershifter: We set `React.ReactNode` when we deal with `contentShorthand`, but in this case we have:
```js
PropTypes.oneOfType([
  PropTypes.bool,
  customPropTypes.collectionShorthand,
]),
```

Which represents array of `itemShorthand`, we mark it as `any` for now.

I want to add new TS types for `customPropTypes` in future.

layershifter: #1289

levithomason:
Looks great, we'll follow up on types for customPropTypes in #1289 .

levithomason:

PR #1292, opened 2017-02-06
feat(Advertisement): add new component

### LINKED ISSUE #185: <Advertisement /> Component
http://semantic-ui.com/views/advertisement.html

tarang9211:
Fixes #185


tarang9211:
Can someone giving feedback on this? I have 4 tests that need to pass, also need more info about the workings of this component.

@layershifter or @levithomason could you give some feedback?

src/index.js
```diff
@@ -139,6 +139,8 @@ export { default as SidebarPushable } from './modules/Sidebar/SidebarPushable'
 export { default as SidebarPusher } from './modules/Sidebar/SidebarPusher'
 
 // Views
+export { default as Advertisement } from './views/Advertisement/Advertisement'
+
```
layershifter: Unnecessary blank line.

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,40 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
+import {
+  customPropTypes,
+  getElementType,
+  META,
+} from '../../lib'
+
+/**
+*An ad displays third-party promotional content
```
layershifter: Missing gap and dot:

```diff
- *An ad displays third-party promotional content
+* An ad displays third-party promotional content.
```

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,40 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
```
layershifter: We sort imports in last PRs:

```diff
-import React, { PropTypes } from 'react'
-import cx from 'classnames'
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
```

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,40 @@
...
+  } = props
+
+  const classes = cx(
+    'ui',
+    'ad',
+    className
+  )
+  const ElementType = getElementType(Advertisement, props)
+
+  return <ElementType className={classes} />
```
layershifter: I think that this component will have `children`?

layershifter (changes requested):
@tarang9211 I've left some comments, but all of them is about styling. You're on right way :+1: 

I could say more on the case, after you will add the some examples in the docs.

src/views/Advertisement/Advertisement.js
layershifter: Also, to get test working you need use there `getUnhandledProps` function and spread `rest` to `ElementType`.

tarang9211: I added the meta property here: https://github.com/tarang9211/Semantic-UI-React/blob/5e34b354fcbd0a9da6847893c96172bac4564dee/src/views/Advertisement/Advertisement.js#L27 I was referring to the Feed component and structuring it that way. I'll push a new commit with your changes.

tarang9211: What should be the `children` of this component? If I had the line below, about 11 tests fail
```
if (!_.isNil(children)) {
    return <ElementType {...rest} className={classes}>{children}</ElementType>
  }
```

tarang9211:
@layershifter pushed a new commit with your changes. Here's a screenshot of the test that fails.

screen shot 2017-02-07 at 2 58 51 pm

Im defining the `_meta` property as per this - https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/views/Statistic/Statistic.js#L58

layershifter:
I've pushed some commits, tests are passed :+1:

tarang9211:
@layershifter checked them out right now. Don't we also need a nil check for `children`?

tarang9211:
@layershifter could we create like a to-do for this task? That would help knocking things off. Let me know your thoughts.

layershifter:
> Don't we also need a nil check for children?

Nope.

> could we create like a to-do for this task?

Let's implement examples from [SUI docs](http://semantic-ui.com/views/advertisement.html).

tarang9211:
@layershifter hmm alright. So are we handling the various sizes via `props`?

tarang9211:
Let me know if you think these are on point and what you're looking for!

layershifter:
@tarang9211 I'll take a look today.

layershifter:
Okay, as I see we [have following units](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/views/ad.less):
- (mobile, large) leaderboard;
- (medium, large) rectangle;
- half page;
- (small) square;
- (small, vertical) rectangle;
- (small, square) button;
- (wide) skyscrape;
- (vertical, top, half, large mobile) banner;
- billboard;
- panorama;
- netboard.

Other props:
- centered;
- test.

### Thoughts

All unit props will be separate:
```js
leaderboard: PropTypes.oneOfType(
  PropTypes.bool,
  PropTypes.oneOf(['mobile', 'large'])
),
panorama: PropTypes.bool,
```

Also, as I see we cannot combile units, so they must also be single:

```js
panorama: customPropTypes.every([
  customPropTypes.disallow(['leaderboard']),
  customPropTypes.disallow(['button']),
  // ...
  PropTypes.node,
])
```

Test will be combo-prop:

```js
test: PropTypes.oneOfType(
  PropTypes.bool,
  PropTypes.number,
  PropTypes.string,
)
```

```jsx 
<Ad test/> // => <div class="ui test ad"></div>
<Ad test='hi' /> // => <div class="ui test ad" data-text="Ad Unit 2"></div>
```

@levithomason Other ideas?

tarang9211:
Thanks for the input I'll get to this in a few

jeffcarbs:
@layershifter - Given that the different "units" only affect height/width and thus can't be combined,  I consider "unit" a single prop with different possible values. I think it should be similar to https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Grid/Grid.js#L121-L123

I took the values from the CSS file you linked, so it'd be:
```js
unit: PropTypes.oneOf([
  'banner', 'half banner', 'large mobile banner' 'top banner', 'vertical banner'
  'billboard',
  'button', 'square button', 'small button',
  'half page',
  'small rectangle', 'medium rectangle', 'large rectangle', 'vertical rectangle',
  'leaderboard', 'large leaderboard', 'mobile leaderboard',
  'netboard',
  'panorama',
  'skyscraper', 'wide skyscraper',
  'square', 'small square'
])
```

layershifter:
@jcarbo good suggestion, I like it :+1:

tarang9211:
Pushing a commit with @jcarbo's suggestions!

Could someone explain what `test` prop is really?  Right now I'm seeing this on my local doc site examples and I don't know where it gets the **Ad** text from.

screen shot 2017-02-09 at 12 24 32 am

jeffcarbs:
There's a little bit of CSS cleverness going on here:
screen shot 2017-02-08 at 3 04 31 pm

When the `test` class is present, there's an `after` pseudo element that looks at the `data-text` attribute and uses that as it's content. So if we want to support this as SUI-core does:

```jsx
<Advertisement unit='medium rectangle' test='Hello' />

// Should result in
<div class="ui medium rectangle test ad" data-text="Hello"></div>
```

So the presence of the `test` prop should add the class and if it happens to be a string then it should also set the `data-text` prop.

tarang9211:
> There's a little bit of CSS cleverness going on here

Ah that makes sense now! I think I have this working!

tarang9211:
@layershifter let me know what you think of this so far!

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,76 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+import {
```
layershifter: New line after externals imports.

layershifter: New line after externals imports.

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,76 @@
...
+  customPropTypes,
+  getElementType,
+  useKeyOnly,
```
layershifter: Sort in alphabet order, please.

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * An ad displays third-party promotional content.
+ */
+function Advertisement(props) {
+  const {
+    className,
+    children,
+    unit,
+    test,
+    centered,
```
layershifter: Sort in alphabet order, please.

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,76 @@
...
+  } = props
+
+  const classes = cx(
+    'ui',
+    'ad',
+    unit,
+    useKeyOnly(test, 'test'),
+    useKeyOnly(centered, 'centered'),
+    className
```
layershifter: Sort in alphabet order, please:

```diff
-    'ui',
-    'ad',
-    unit,
-    useKeyOnly(test, 'test'),
-    useKeyOnly(centered, 'centered'),
-    className
-    'ui',
+    unit,
+    useKeyOnly(centered, 'centered'),
+    useKeyOnly(test, 'test'),
+    'ad',
+    className
```

src/views/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /** Text to be displayed on the advertisement. */
+  test: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.number,
+    PropTypes.string,
+  ]),
+
+  /** Center the advertisement. */
+  centered: PropTypes.bool,
```
layershifter: Sort props in alphabet order, please.

test/specs/views/Advertisement/Advertisement-test.js
```diff
@@ -0,0 +1,21 @@
...
+  common.propKeyOnlyToClassName(Advertisement, 'centered')
+  common.propKeyOnlyToClassName(Advertisement, 'test')
+
+  it('renders a <div> by default', () => {
+    shallow(<Advertisement />).should.have.tagName('div')
+  })
+
+  it('should have appropriate classes for unit prop', () => {
+    shallow(<Advertisement unit='medium rectangle' />).should.have.className('medium rectangle')
+  })
```
layershifter: Please replace this test with `commonTest`:

```diff
  common.propKeyOnlyToClassName(Advertisement, 'centered')
  common.propKeyOnlyToClassName(Advertisement, 'test')

+  common.propValueOnlyToClassName(Advertisement, 'unit', [
+    'medium rectangle', 'large rectangle', 'vertical rectangle', 'small rectangle',
+    'mobile banner', 'banner', 'vertical banner', 'top banner', 'half banner',
+    'button', 'square button', 'small button',
+    'skyscraper', 'wide skyscraper',
+    'leaderboard', 'large leaderboard', 'mobile leaderboard', 'billboard',
+    'panorama',
+    'netboard',
+    'half page',
+    'square', 'small square',
+  ])

**Review by layershifter — CHANGES_REQUESTED** [review]

**Comment by layershifter** [comment]
I've added some comments about styling. Component looks ready, lets add all left examples :+1:

**Comment by tarang9211** [comment]
@layershifter or @jcarbo updated the PR with doc site example for the **Type** of advertisements. There are some padding issues between the `skyscraper, wide skyscraper` and the `square, small square` components. Could you verify those or let me know how to rectify those?

**Comment by tarang9211** [comment]
Everything seems to be in place for this PR. Just the small CSS related issue I mentioned above. Let me know your thoughts! Thanks for the help, and can't wait for the v1.0.0 of this awesome library!

File: docs/app/Examples/views/Advertisement/Types/AdvertisementExampleBanner.js [hunk 11]
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Advertisement } from 'semantic-ui-react'
+
+const AdvertisementExampleBanner = () => (
+  <div>
+    <Advertisement unit='banner' test='Banner' />
+    <Advertisement unit='vertical banner' test='Vertical Banner' />
+    <Advertisement unit='top banner' test='Top Banner' />
+    <Advertisement unit='half banner' test='Half Banner' />
```
layershifter: Let's split these examples (another, too) to separate files.

```diff
-<ComponentExample
-  title='Banner'
-  description='An ad can use banner ad unit size'
-  examplePath='views/Advertisement/Types/AdvertisementExampleBanner'
-/>
+ <ComponentExample
+  title='Banner'
+  description='An ad can use banner ad unit size'
+  examplePath='views/Advertisement/Types/AdvertisementExampleBanner'
+/>
+<ComponentExample examplePath='views/Advertisement/Types/AdvertisementExampleBannerVertical' />
+<ComponentExample examplePath='views/Advertisement/Types/AdvertisementExampleBannerTop' />
```

layershifter: It will also solve issues with padding.

docs/app/Examples/views/Advertisement/Types/AdvertisementExampleMobile.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { Advertisement } from 'semantic-ui-react'
+
+const AdvertisementExampleMobile = () => (
+  <div>
+    <div className='ui info ignored message'>
+      Mobile ads will automatically only appear on mobile browser viewports
```
layershifter: Trailing dot, please.

docs/app/Examples/views/Advertisement/Types/index.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+
+import ComponentExample from 'docs/app/Components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
+
+const Types = () => (
+  <ExampleSection title='Types'>
+    <ComponentExample
+      title='Common Units'
+      description='An advertisement can appear in common ad unit sizes'
```
layershifter: Trailing dot on each description, please.

test/specs/views/Advertisement/Advertisement-test.js
```diff
@@ -0,0 +1,28 @@
...
+import * as common from 'test/specs/commonTests'
+
+describe('Advertisement', () => {
+  common.isConformant(Advertisement)
+  common.hasUIClassName(Advertisement)
+  common.rendersChildren(Advertisement)
+
+  common.propKeyOnlyToClassName(Advertisement, 'centered')
+  common.propKeyOnlyToClassName(Advertisement, 'test')
+  common.propValueOnlyToClassName(Advertisement, 'unit', [
```
layershifter: Empty line, please.

layershifter (changes requested):
Let's split examples to separate files.

layershifter:
Also, let's add typings for this component.

tarang9211:
> Also, let's add typings for this component.

what's that? unfamilair with that

test/specs/views/Advertisement/Advertisement-test.js
tarang9211: you want me to remove the empty line, right? (line 10) ?

layershifter:
Type definitions for typescript, you can use [Rail](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Rail/index.d.ts) component as reference.

test/specs/views/Advertisement/Advertisement-test.js
layershifter: No, add :) Between 12 and 13.

```diff
  common.propKeyOnlyToClassName(Advertisement, 'test')
+
  common.propValueOnlyToClassName(Advertisement, 'unit', [
```

docs/app/Examples/views/Advertisement/Types/AdvertisementExampleBanner.js
tarang9211: separate files for all props? So banner will have 4, square will have 2?

tarang9211: I think the `<ComponentExample examplePath='views/Advertisement/Types/AdvertisementExampleBannerVertical' />` needs the `title` and `description` props. The doc site crashes when I try the code above

layershifter: Yes, you can check [Breadcrumb](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/collections/Breadcrumb/Types/index.js)'s Types examples.

layershifter: > needs the title and description props

No, if docs crashed then the problem is in something else.

tarang9211: this introduced way too much spacing @layershifter 
i just pushed a commit with those changes for banner, let me know what you think. i like the old way better, just a suggestionn.

layershifter: > i just pushed a commit with those changes for banner, let me know what you think.

Let's update all left examples to the described pattern.

tarang9211: Will do.

layershifter:
- [x] Split components examples to separate files.
- [x] Add typings for the component.
- [x] Add the first example (A standard ad).
- [x] Common Units section missing  `Large Rectangle` example.
- [x] Fix order of `Mobile` section, it goes after `Common Units`.
- [x] Fix order of examples in `Button` section (must be like in SUI docs).
- [x] Fix order of `Banner` section, it goes after `Skyscraper`.
- [x] Fix items in `Test` section, it contains another examples in SUI.
- [x] Fix description:
```diff
- You can adjust the text displayed for your test ad placement by changing the value of data-text.
+ You can adjust the text displayed for your test ad placement by changing the value of <code>text</code> prop.

**Comment by tarang9211** [comment]
@layershifter i've gone through the list. I have two concerns. Could you verify the `index.d.ts` file? 
Second question, where do I get the ad from for the standard ad example? Just copy it from [here](https://support.google.com/adsense/answer/181950?hl=en&ref_topic=28893)?

**Comment by tarang9211** [comment]
@layershifter Pasting the Google Adsense code directly into the `Advertisement` component doesn't seem to work.

This is what I have currently
```
  <Advertisement unit='medium rectangle'>
    <script async src='//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js'></script>
      <ins className='adsbygoogle'
        data-ad-client='ca-pub-4591861188995436'
        data-ad-slot='4640466102'
      >
      </ins>
      <script>
      (adsbygoogle = window.adsbygoogle || []).push({});
      </script>
  </Advertisement>
```

**Comment by levithomason** [comment]
Not sure what is missing or if this will help, but the SUI core example includes an inline style tag on the `ins` element:

```html
<script async="" src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
<ins
  class="adsbygoogle"
  style="display:inline-block;width:300px;height:250px"
  data-ad-client="ca-pub-4591861188995436"
  data-ad-slot="4640466102"
></ins>
<script>
  (adsbygoogle = window.adsbygoogle || []).push({});
</script>
```

tarang9211:
> Not sure what is missing or if this will help, but the SUI core example includes an inline style tag on the ins element

@layershifter  I tried that. After some research, it seems like a well known problem coupling Adsense and React. 
The link explains one way to tackle it. Your thoughts?
http://brandonlehr.com/how-to-incorporate-google-adsense-react-js/

tarang9211:
@layershifter my last commit got it working, but I had to wrap some of the google adsense code in the `window.onload` callback

layershifter:
@tarang9211 I've pushed multiple fixes, making a final review.

tarang9211:
@layershifter got it, thanks for the cleanup. So is this ready to be merged?

layershifter (approved):
@tarang9211 Thanks for your work there :+1: 

@levithomason We need a final review from you there :)

tarang9211:
> Thanks for your work there 👍

You too, learning a ton about react from this! Thanks for the guidance!

docs/app/Examples/views/Advertisement/Types/AdvertisementExampleStandard.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from 'react'
+import { Advertisement } from 'semantic-ui-react'
+
+export default class AdvertisementExampleStandard extends Component {
+  componentDidMount() {
+    (window.adsbygoogle = window.adsbygoogle || []).push({})
```
levithomason, line 6: Example components are re-mounted on every change in the editor, this will be called several times.

Also, we should check if the `adsbygoogle` object is loaded, then push an ad after it is loaded.  Let's move this to a script tag in `index.ejs` immediately after the google ads script is loaded.

Once this is done, this should be a functional component rather than a class, since, there will be no more lifecycle methods in use.

docs/app/index.ejs
```diff
@@ -67,6 +67,7 @@
   <script src="//cdnjs.cloudflare.com/ajax/libs/react/<%= htmlWebpackPlugin.options.versions.react %>/react<%= __PROD__ ? '.min' : '' %>.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/react/<%= htmlWebpackPlugin.options.versions.reactDOM %>/react-dom.min.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/react/<%= htmlWebpackPlugin.options.versions.reactDOM %>/react-dom-server.min.js"></script>
+  <script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
```
levithomason: This script tag is in the body so there is no need to make it async since it cannot block body content from loading.  Let's remove the `async` attribute.

Regarding [this comment](https://github.com/Semantic-Org/Semantic-UI-React/pull/1292/files#r100858877), the `adsbygoogle` object's custom push method should be called only once in another script tag immediately following this one.  I've used a recursive timeout to wait until the library has loaded:

```html
<script>
  var pushGoogleAd = (function() {
    setTimeout(function () {
      if (window.adsbygoogle && window.adsbygoogle.loaded) window.adsbygoogle.push({})
      else pushGoogleAd()
    }, 1000)
  }())
</script>
```

levithomason (changes requested):
Fantastic work guys!  I've left a few comments for addressing.  We're very close to a merge here 😄

tarang9211:
@levithomason & @layershifter do take a look. pushed commits related to your comments. 👍 

One thing I noticed was the `medium rectangle` ad for the standard ad example isn't displaying properly.
screen shot 2017-02-14 at 12 56 37 am

tarang9211:
@levithomason let me know what you think of the changes I made.

levithomason (approved):

tarang9211:
Any idea about my concern regarding the sizing issue?

levithomason:
Checking into it, I think it is `style` related.

Also, the ad only loads on page refresh.  If you navigate from, say the button docs, to the advertisement docs, the ad doesn't load (at least for me)

tarang9211:
Same for me. I noticed the same thing

levithomason:
I think the size issue is due to the style prop, using this style prop from the SUI core [example](http://semantic-ui.com/views/advertisement#ad):
```jsx
style={{ display: 'inline-block', width: 300, height: 250 }}
```

I get these results:



I am not sure on the mounting problem.

levithomason:
Unfortunately, that's all the time I have for now on this one.  I'll check back here later and see where we are at with the mounting issue.  That is the final issue to solve I believe, everything else is looking great.

tarang9211:
Interesting. Sounds good, let me know your findings. Let's try to get this merged soon!

tarang9211:
@layershifter Could you help look into this? We're pretty close to a merge

tarang9211:
@levithomason I suggest we revert the changes we made regarding the Adsense and have it be like what @layershifter did [here](https://github.com/tarang9211/Semantic-UI-React/blob/7d303a0fd9a38b8d2cb03906ec42c6a98ba90264/docs/app/Examples/views/Advertisement/Types/AdvertisementExampleStandard.js#L4). That's how everyone seems to be implementing the Adsense with React. Thoughts?

I don't think we faced any issues with that method of implementation.

levithomason:
OK, let's give it a try.  Whatever solves the problem is the correct answer :)

tarang9211:
@levithomason Could you take a look at it? I'm looking at the Checkbox issue(s) currently. Let me know if that's not feasible.

layershifter (approved):
@levithomason I've updated typings structure and reverted standart example. LGTM, I think we ready to go :speedboat:

tarang9211:
@layershifter Thanks for getting to this! Appreciate the help

levithomason:
Nice work fellas! 🎉

levithomason:

PR #1297, opened 2017-02-08
chore(eslint): disable aria rules for docs

layershifter:
I think that our docs will never comply a11y rules, so let's disable them :smile:


levithomason:
Agreed.

levithomason:

PR #1306, opened 2017-02-11
fix(package): add babel-runtime for commonjs builds

### LINKED ISSUE #1260: Uncaught Error: Cannot find module 'babel-runtime/helpers/get' from 'semantic-ui-react/dist/commonjs/modules/Dropdown/Dropdown.js'
Semantic-UI-React Version: ^0.64.5 (Latest)

I get the following error when trying to import "semantic-ui-react" into any of my React components. I have tried installing babel-runtime to my dev packages, but still no luck. I am not even trying to use dropdown menus.

https://i.gyazo.com/34702d671ae499a2e419d62af995bfd4.png

levithomason:
Fixes #1260 

This PR adds `babel-runtime` as a package dep.

I've opted for a regular dependency over a peer dependency here as I don't want users to have to install more modules to get up and running.  Also, our package only requires the necessary modules from runtime, so it is cherry picked, similar to how we handle lodash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   },
   "homepage": "https://github.com/Semantic-Org/Semantic-UI-React#readme",
   "dependencies": {
+    "babel-runtime": "^6.22.0",
     "classnames": "^2.1.5",
     "debug": "^2.4.5",
     "lodash": "^4.17.2"


levithomason:

PR #1307, opened 2017-02-12
fix(Dropdown): Adjusts search input join() to allow for hashed params compatibility

hybridindie:
Dropdowns with search suffixed the name attribute creating an error with frameworks/servers that use hashed params like Ruby on Rails. This moves the suffix to a prefix for compatibility and updates the test to reflect this change.

```html
<!-- before / causes server errors -->
<input type="text" value="" aria-autocomplete="list" class="search" name="profile[city]-search" autocomplete="off" tabindex="0">

<!-- after / fixes errors -->
<input type="text" value="" aria-autocomplete="list" class="search" name="search-profile[city]" autocomplete="off" tabindex="0">
```

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1060,7 +1060,7 @@ export default class Dropdown extends Component {
         aria-autocomplete='list'
         onChange={this.handleSearchChange}
         className='search'
-        name={[name, 'search'].join('-')}
+        name={['search', name].join('-')}
         autoComplete='off'
         tabIndex={computedTabIndex}
         style={{ width: searchWidth }}
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -1530,7 +1530,7 @@ describe('Dropdown Component', () => {
       wrapperShallow(<Dropdown name='foo' options={options} selection search />)
         .should.have.descendants('input.search')
 
-      wrapper.find('input.search').should.have.prop('name', 'foo-search')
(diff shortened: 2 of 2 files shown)


levithomason:
>creating an error with frameworks/servers that use hashed params like Ruby on Rails.

Could you specify the error here?  I've had a task on my list for changing how the search name is built-up but would like to understand this issue before commenting further.

hybridindie:
The error is a fatal crash from the app server caused by malformed parameters. Ruby on Rails (and I believe other frameworks) nests parameters in a hash using the name attribute. So a param like client[address][name] gets encoded to {"address" => {"name" => "John"}} but when you suffix a string to it the params get malformed to something like {"address" => {"name" => "John"}}-string. By prefixing it you can white/blacklist the parameter and throw it away accordingly while it still remains valid.

levithomason:
@davezuko any insight here?

levithomason:
Looks like this PR has gone stale.  Closing for housekeeping.

allthesignals:
This is also an issue for me - it sent us spiraling on a project for a few days. It's a non-starter for us now because the only reason we are using Dropdown search is so that users can use keyboard input to select (which isn't possible without it, see issue here: #1935). 

The error in particular in server frameworks is:
`Rack::QueryParser::ParameterTypeError (expected Hash (got String) for param '***'):`

The "name" attribute in HTML5 forms is fairly sensitive to standards in the server-side framework ecosystem - I know many folks use React for AJAX type data submission, but some of us still have to use regular old form-data. I think at least the ability to override the naming convention here is desirable - it doesn't even seem possible without extending the existing module. 

If you are able to merge this somehow, that would be great - thank you!

PR #1320, opened 2017-02-13
style(Form): update typings and propTypes usage

### LINKED ISSUE #1304: bug(Form): incomplete typings for Form.TextArea
version 0.64.7
using TypeScript

**This works ok**
`import { Form, TextArea } from 'semantic-ui-react'`
`<Form.Field>
                    <label>Comments</label>
                    <TextArea value='Text In Here' />
</Form.Field>`

**This doesn't compile**
`<Form.TextArea label='Comments' value='Text In Here' />`

**Error:**

> Property 'value' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<FormTextAreaProps, ComponentState>> & Re...'.

**Checking [FormTextAreaProps] in index.d.ts it does not include [value]** 

Also not sure if this is the proper format but this doesn't work either

`<Form.Field control={TextArea} value='Text In Here' />`

layershifter:
Fixes #1304.


src/collections/Form/FormGroup.js
```diff
@@ -49,24 +53,20 @@ FormGroup.propTypes = {
...
     customPropTypes.disallow(['grouped']),
     PropTypes.bool,
   ]),
 
-  /** Fields Groups can specify their width in grid columns or automatically divide fields to be equal width */
-  widths: PropTypes.oneOf(FormGroup._meta.props.widths),
-}
-
-FormGroup.defaultProps = {
-  as: 'div',
```
layershifter, line 69: `ElementType` is `div` by default.

src/index.d.ts
```diff
@@ -1,5 +1,3 @@
-export type InputType = 'color' | 'date' | 'datetime' | 'datetime-local' | 'email' | 'number' | 'range' | 'search' | 'select' | 'password' | 'tel' | 'text' | 'time' | 'url' | 'week';
-
```
layershifter, line 2: Cruft.

src/index.d.ts
```diff
@@ -25,8 +23,6 @@ type SemanticWIDTHSSTRING = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' |
 
 export type SemanticWIDTHS = SemanticWIDTHSNUMBER | SemanticWIDTHSSTRING;
 
-export type SemanticGenericOnClick = (event: any, value: any) => void;
-export type SemanticFormOnClick = (event: any, object: Object) => void;
```
layershifter: Cruft.

src/collections/Form/index.d.ts
```diff
@@ -163,69 +145,34 @@ interface FormGroupProps {
...
-  /** Fields Groups can specify their width in grid columns or automatically divide fields to be equal width */
+  /** Fields Groups can specify their width in grid columns or automatically divide fields to be equal width. */
   widths?: SemanticWIDTHS | 'equal';
 }
 
-export const FormGroup: React.ComponentClass<FormGroupProps>;
+export const FormGroup: React.StatelessComponent<FormGroupProps>;
 
-interface FormInputProps extends InputProps {
+interface FormInputProps extends FormFieldProps, InputProps {
```
levithomason, line 161: Hm, is this going to cause the duplicate property error again?  Both FormFieldProps and InputProps define a `label` prop in their interfaces, each with different types:


**FormFieldProps**
```ts
   /** Mutually exclusive with children. */
  label?: any | Object;
```

**InputProps**
```ts
  /** Optional Label to display along side the Input. */
  label?: any;
```

This may also be the case with Checkbox/Radio/Button, since they all accept a label prop as well.

levithomason (commented):
One question before merging.

src/collections/Form/index.d.ts
layershifter: Seems, you're right, I'll check it.

layershifter:
I've checked `Form` typings with compiler and fixes issues.

layershifter:
Rebased to `master` and added some typings cleanups.

levithomason:
What about [this comment](https://github.com/Semantic-Org/Semantic-UI-React/pull/1320/files#r100992034), do we still have the conflicting input prop typing issue due to the FormFieldProps extending the InputProps?

layershifter:
[Nope](https://github.com/Semantic-Org/Semantic-UI-React/pull/1320#issuecomment-279756802).

levithomason:
Thanks, that was obvious :P

levithomason:

PR #1239, opened 2017-01-27
tests(mixed): remove usage of `meta.props`

layershifter:
This PR completely removes usage of  `meta.props` from tests.

PR #1322, opened 2017-02-13
style(Popup): fix tests, update typings and propTypes usage

layershifter:
Also, fixes `positions` test.


src/index.d.ts
```diff
@@ -11,7 +11,6 @@ export type SemanticSIZES = 'mini' | 'tiny' | 'small' | 'medium' | 'large' | 'bi
...
-export type SemanticPOSITIONING = 'top left' | 'top right' | 'bottom right' | 'bottom left' | 'right center' | 'left center' | 'top center' | 'bottom center';
```
layershifter, line 14: This type used only by `Popup`.

src/modules/Popup/Popup.js
```diff
@@ -17,25 +19,16 @@ import PopupHeader from './PopupHeader'
...
+export const POSITIONS = [
+  'top left',
+  'top right',
+  'bottom right',
+  'bottom left',
+  'right center',
+  'left center',
+  'top center',
+  'bottom center',
+]
```
layershifter, line 31: Positions are used in multiple places, so we need it as const.

src/modules/Popup/Popup.js
```diff
@@ -349,8 +349,8 @@ export default class Popup extends Component {
     const popupJSX = (
       <ElementType {...rest} className={classes} style={style} ref={this.popupMounted}>
         {children}
-        {!children && PopupHeader.create(header)}
-        {!children && PopupContent.create(content)}
+        {_.isNil(children) && PopupHeader.create(header)}
+        {_.isNil(children) && PopupContent.create(content)}
```
layershifter, line 353: Our problem with '0' fixed there 😄

test/specs/modules/Popup/Popup-test.js
```diff
@@ -107,23 +107,20 @@ describe('Popup', () => {
   })
 
   describe('positioning', () => {
-    it('is always within the viewport', () => {
-      _.each(Popup._meta.props.positions, position => {
+    POSITIONS.forEach(position => {
+      it('is always within the viewport', () => {
```
layershifter, line 111: This test work now and #1239 will work.

There is simple explanation why it worked before: test used `_meta.props.positions` while they were defined in `_meta.props.positioning`, so it doesn't really work. But, when I add real positions, it failed. `beforeEach` handler applied to `wrapperMount` once, now it applies on each `position` and everything works fine.

test/specs/modules/Popup/Popup-test.js
```diff
@@ -265,13 +262,10 @@ describe('Popup', () => {
   })
 
   describe('size', () => {
-    it('defines prop options in _meta', () => {
-      Popup._meta.props.should.have.any.keys('size')
-      Popup._meta.props.size.should.be.an('array')
-    })
```
layershifter, line 271: Cruft.

test/specs/modules/Popup/PopupContent-test.js
```diff
@@ -4,4 +4,6 @@ import * as common from 'test/specs/commonTests'
 describe('PopupContent', () => {
   common.isConformant(PopupContent)
   common.rendersChildren(PopupContent)
+
+  common.implementsCreateMethod(PopupContent)
```
layershifter, line 8: We need `commonTest` there :)

test/specs/modules/Popup/PopupHeader-test.js
```diff
@@ -4,4 +4,6 @@ import * as common from 'test/specs/commonTests'
 describe('PopupHeader', () => {
   common.isConformant(PopupHeader)
   common.rendersChildren(PopupHeader)
+
+  common.implementsCreateMethod(PopupHeader)
```
layershifter, line 8: And there :)

levithomason (approved):
:boom:

levithomason:

PR #1323, opened 2017-02-13
feat(Shape): add Shape component

### LINKED ISSUE #196: <Shape /> Component
http://semantic-ui.com/modules/shape.html

tarang9211:
Fixes #196

### DIFF
--- a/src/modules/Shape/Shape.js
+++ b/src/modules/Shape/Shape.js
@@ -9,6 +9,9 @@ import {
   META,
 } from '../../lib'
 
+/**
+* A shape is a three dimensional object displayed on a two dimensional plane.
+*/
 export default class Shape extends Component {
   static propTypes = {
     /** An element type to render as (string or function). */
@@ -28,14 +31,14 @@ export default class Shape extends Component {
 
   render() {
     const {
-      className,
       children,
+      className,
     } = this.props
 
     const classes = cx(
       'ui',
+      'shape',
       className,
-      'shape'
     )
 
     const rest = getUnhandledProps(Shape, this.props)
--- a/src/modules/Shape/Shape.js
+++ b/src/modules/Shape/Shape.js
@@ -1,8 +1,7 @@
 import cx from 'classnames'
-import React, { PropTypes } from 'react'
+import React, { Component, PropTypes } from 'react'
 
 import {
-  AutoControlledComponent as Component,
   customPropTypes,
   getElementType,
   getUnhandledProps,
--- a/src/modules/Shape/Shape.js
+++ b/src/modules/Shape/Shape.js
@@ -1,0 +1,48 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import {
+  AutoControlledComponent as Component,
+  customPropTypes,
+  getElementType,
+  getUnhandledProps,
+  META,
+} from '../../lib'
+
+export default class Shape extends Component {
+  static propTypes = {
+    /** An element type to render as (string or function). */
+    as: customPropTypes.as,
+
+    /** Primary content. */
+    children: PropTypes.node,
+
+    /** Additional classes. */
+    className: PropTypes.string,
+  }
+
+  static _meta = {
+    name: 'Shape',
+    type: META.TYPES.MODULE,
+  }
+
+  render() {
+    const {
+      className,
+      children,
+    } = this.props
+
+    const classes = cx(
+      'ui',
+      className,
+      'shape'
+    )
+
+    const rest = getUnhandledProps(Shape, this.props)
+    const ElementType = getElementType(Shape, this.props)
+
+    return (
+      <ElementType {...rest} className={classes}>{children}</ElementType>
+    )
+  }
+}
--- a/src/index.js
+++ b/src/index.js
@@ -132,6 +132,8 @@ export { default as Progress } from './modules/Progress'
 export { default as Rating } from './modules/Rating'
 export { default as RatingIcon } from './modules/Rating/RatingIcon'
 
+export { default as Shape } from './modules/Shape'
+
 export { default as Search } from './modules/Search'
 export { default as SearchCategory } from './modules/Search/SearchCategory'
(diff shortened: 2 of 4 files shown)


tarang9211:
@layershifter Could you guide me on this one, if possible? Thanks!

tarang9211:
@layershifter Could you give me some feedback when you get a chance?

src/modules/Shape/Shape.js
```diff
@@ -0,0 +1,51 @@
+import cx from 'classnames'
+import React, { PropTypes } from 'react'
+
+import {
+  AutoControlledComponent as Component,
```
layershifter: Let's don't use `AutoControlledComponent` at now, let's back to `React.Component`.

layershifter (changes requested):

layershifter:
Let's start from `Shape` sub-components and some props:

```jsx
<Shape.Side active header>Did you know? This side starts visible.</Shape.Side>
//<div class="active ui header side">Did you know? This side starts visible.</div>
```

```jsx
<Shape.Side>
  <Shape.Content center>4</Shape.Content>
</Shape.Side>

<Shape.Side>
  <Shape.Content center='4' />
</Shape.Side>

<Shape.Side center content='4' />

//<div class="side">
//  <div class="content">
//    <div class="center">4</div>
//  </div>
//</div>
```

```jsx
<Shape cube />
//<div class="ui cube shape"></div>
```

```jsx
<Shape type='auto' />
//<div class="ui auto shape"></div>
<Shape type='square' />
//<div class="ui square shape"></div>
<Shape type='auto' />
//<div class="ui irregular shape"></div>
```

tarang9211:
@layershifter Can we come up with a list of sub-components and props? I think that would make this easier!

ch3lo:
I suggest you an update in readme.md marking shape as supported.

levithomason:
@tarang9211 please see https://github.com/Semantic-Org/Semantic-UI-React/pull/1323#issuecomment-281065710 for subcomponents and props.

tarang9211:
@levithomason yup, will have a look. I'm just swamped in the week, will be closing out my open issues by Saturday.

cemremengu:
Any updates on this ? :eyes:

madwill:
Hello everyone! Would love to see this component make it into the react implementation! Thank you all for your hard work ! Making the world a prettier place!

levithomason:
Note, I've just rebased this branch to the latest master to keep it from getting to stale.

madwill:
Has this been forgotten ?

levithomason:
It appears @tarang9211 has gotten busy.  We'll gladly take another PR picking up where this one left off.

madwill:
Thank you for such a quick response, I pulled the fork from tarang but unless i'm not understanding something i don't think we were that close to completion. 

For better understanding of the library and its component, i'll try and give it a shot. Really just at initial stage but i got it to show in the docs with example having shape components and 2 sub components being side and content. 

https://github.com/madwill/Semantic-UI-React/commit/12dc99dd09136a64102d82ac7527bda9dc9ffa25

I confess having currently no idea how to manipulate it to get the semantic ui shape behavior. I'll dig into the official code for it .shape() and .shape('flip X'); as well as other components using javascript for behavior within Semantic-UI-React to see how it already handle such cases. 

I can not make promises as i don't feel confident enough right now but if you could point to a component you feel would be a good example of javascript manipulation without jquery it it may accelerate / make it possible. 

Thank you very much

tarang9211:
Hey guys, I'm really sorry for being away! Been extremely busy with work and personal commitments. Happy to help whenever needed.

madwill:
Turns out is a surprisingly complex component :

https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/shape.js

The way it invokes methods, using query arguments in order to allow for a .shape('flip up') makes it so that its not just re-implementation, it would require some re-engineering. 

With the 6 types of animations requiring adjustments based on component size, its a longer task than expected. 

I get why it stalled. Jquery less semantic UI is an ambitious project considering its implementation. It would even say it has to lose a bit of its "semantic-ism". 

This feels like a lot for a newcomer so i'll let my head cool down.   I don't think i'll be the one creating the pull request. 

Perhaps the flip / stage section will become a bit more clear as i take a step back. 

Thanks for such big amount of work.

madwill:
What if we simplify it a bit ? for the sake of quick implement-ability.

Top component (shape) would have 2 props being side and direction

```

    /** Currently displayed side, */
    side: PropTypes.number,

     /** Direction the shape should rotate. */
    direction: PropTypes.oneOf(['right', 'left', 'up', 'down', 'over', 'back']),
```
Usage would be something like this 

```
       <Shape type='auto' side={1} direction='left'>
          <Shape.Side >
            <Image src='https://react.semantic-ui.com/assets/images/avatar/large/matthew.png' />
          </Shape.Side>
          <Shape.Side>
            <Image src='https://react.semantic-ui.com/assets/images/avatar/large/elliot.jpg' />
          </Shape.Side>
        </Shape>
```

User would have to change side by himself with his own controls and state.

We would trigger animation in a life cycle function such as 

```
    if (nextProps.side !== this.props.side) {
      this.startAnimating()
    }
  }

```
User would still have the liberty to change the "direction" props if that's something someone would want and i would have to make a big "switch" statement in startAnimating to create the correct animation depending on our props and type 

What would be the point of 

`<Shape.Content center='4' />`

It feels like an extra layer that could be avoided, if you look at usage within semantic-ui itself there is no side-content divs. Content is directly written in the side div. 

```
<div class="ui shape">
  <div class="sides">
    <div class="active side">This side starts visible.</div>
    <div class="side">This is yet another side</div>
    <div class="side">This is the last side</div>
  </div>
</div>

```
Let me know what you think

layershifter:
@madwill Thanks for a detailed suggestions :+1:

My initial idea was about `componentWillReceiveProps`, too. I'll be glad to see changes that realizing this idea.

layershifter:
I take the PR on myself, #1751

Issue #1324, opened 2017-02-13
Popups and Tooltips steal focus from input fields


dgolds:
### Steps
1. A page with inputs and Popups. Focus on an input then mouseover an item with a popup.tooltip

### Expected Result
1. Mousing over items with popups/tooltips should not steal focus from the input since the user can easily knock a mouse/trackpad during typing
1. Note that this is what happens with semantic-ui

### Actual Result
1. The input loses focus

### Version
0.64.8

### Testcase
Fork this to show the issue: http://codepen.io/levithomason/pen/ZpBaJX

levithomason:
This was introduced in #1154, setting focus on the Portal when opened to allow keyboard navigation.  We've had other issues with this approach as well:

- https://github.com/Semantic-Org/Semantic-UI-React/pull/1215 focus stolen on re-render
- https://github.com/Semantic-Org/Semantic-UI-React/issues/1271 focus stolen on close

Given that we now also have the issue of focus stolen on open, I think it is clear that we need to nix this feature.  There is no possible way to know when it is appropriate to set/restore focus to/from a Portal.

If user's need focus onMount, they will have to set it manually.  The use cases for this are far fewer than the use case of not wanting focus to change.

### TL;DR
This feature should be removed.

iamdanthedev:
The problem is, that if let's say a popup should render a component which sets focus on it's child in componentDidMount, the popup steals focus almost immediately :( At the same time onOpen and onMount popup events happen earlier that the popup content is rendered..

```
browser.js?13f9:123 semanticUIReact:portal handleTriggerClick() - open +2s
browser.js?13f9:123 semanticUIReact:portal open() +1ms
browser.js?13f9:123 semanticUIReact:popup handleOpen() +1ms

///// onOpen
browser.js?13f9:123 Eatalizer:RecipeEditor onOpen() +2ms

browser.js?13f9:123 semanticUIReact:portal componentDidUpdate() +2ms
browser.js?13f9:123 semanticUIReact:portal renderPortal() +1ms
browser.js?13f9:123 semanticUIReact:portal mountPortal() +1ms
browser.js?13f9:123 semanticUIReact:popup handlePortalMount() +2ms

///// onMount
browser.js?13f9:123 Eatalizer:RecipeEditor onMount() +1ms

browser.js?13f9:123 semanticUIReact:dropdown componentWillMount() +37ms
browser.js?13f9:123 semanticUIReact:dropdown setValue() +2ms
[...]

PR #1326, opened 2017-02-14
Modal, Portal and Popup

iamdanthedev:
1) This addresses the problem of portal stealing the focus
https://github.com/Semantic-Org/Semantic-UI-React/issues/1324
**Popup**: added prop focusOnRender (bool)
**Portal**: added prop focusOnRender (bool, default true)

2) This addresses the issue, which allows user to close modal (by clicking outside or hitting esc) and we should be able to check if closing the modal is possible (for example changes inside the modal haven't been saved)
**Modal**: onClose callback allows to cancel closing modal if false is returned

3) This solves the problem of opening a modal with a scroller inside a modal with a scroller. When the upper modal is closed it reverts the original modal's scroller function back (otherwise lower modal loses a scrollbar)
**Modal**: mountNode original classnames are saved in modal state and reverted when modal is unmounted

Please let me know if this goes inline with the concept of SUI-React, than I could polish this changes or something..

Thanks, with respect, Dan

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -57,6 +57,9 @@ class Portal extends Component {
     /** Initial value of open. */
     defaultOpen: PropTypes.bool,
 
+    /** Should portal focus on itself on render. Usefull if portal content sets it's own focus */
+    focusOnRender: PropTypes.bool,
+
     /** The node where the portal should mount. */
     mountNode: PropTypes.any,
 
@@ -121,6 +124,7 @@ class Portal extends Component {
     closeOnDocumentClick: true,
     closeOnEscape: true,
     openOnTriggerClick: true,
+    focusOnRender: true
   }
 
   static autoControlledProps = [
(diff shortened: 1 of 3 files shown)


levithomason:
For now, we only want to revert the focus on open behavior.  If users need to focus a portal component, they will implement their own focus mechanisms.

Would you like to make a PR doing only this and we can address these other items separately?

iamdanthedev:
Ok I will try to make a proper PR on the first issue, going through all portal-based components.
Regarding others it will probably take some time to dig through all components
Ping soon, thanks

levithomason:
Heads up, I needed the original focus fix so I fixed that and shipped it a couple hours ago.

iamdanthedev:
Oh, I got it and that's great. Anyway will try to contribute soon. The lib is really helpfull

layershifter:
@levithomason Seems, we can close this for housekeeping.

levithomason:
Agreed, looks like most, if not all, of these fixes are released.

PR #1327, opened 2017-02-14
feat(Comment): add size prop to Group

### LINKED ISSUE #1311: Add `size` prop to Comment.Group
It is missing in SUI core docs, but is available in the CSS.

```jsx
<Comment.Group className="small" />
```

Should be accomplished with 

```jsx
<Comment.Group size="small" />
```

tarang9211:
Fixed #1311

### DIFF
--- a/src/views/Comment/CommentGroup.js
+++ b/src/views/Comment/CommentGroup.js
@@ -1,3 +1,4 @@
+import _ from 'lodash'
 import cx from 'classnames'
 import React, { PropTypes } from 'react'
 
@@ -6,6 +7,7 @@ import {
   getElementType,
   getUnhandledProps,
   META,
+  SUI,
   useKeyOnly,
 } from '../../lib'
 
@@ -59,8 +61,8 @@ CommentGroup.propTypes = {
   /** Comments can hide extra information unless a user shows intent to interact with a comment. */
   minimal: PropTypes.bool,
 
-  /** Comments can have different sizes*/
-  size: PropTypes.oneOf(['mini', 'tiny', 'small', 'large', 'big', 'huge', 'massive']),
+  /** Comments can have different sizes. */
+  size: PropTypes.oneOf(_.without(SUI.SIZES, 'medium')),
 
   /** A comment list can be threaded to showing the relationship between conversations. */
   threaded: PropTypes.bool,
--- a/src/views/Comment/CommentGroup.js
+++ b/src/views/Comment/CommentGroup.js
@@ -1,5 +1,5 @@
-import _ from 'lodash'
 import cx from 'classnames'
+import _ from 'lodash'
 import React, { PropTypes } from 'react'
 
 import {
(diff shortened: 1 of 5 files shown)


src/views/Comment/CommentGroup.js
```diff
@@ -57,6 +59,9 @@ CommentGroup.propTypes = {
   /** Comments can hide extra information unless a user shows intent to interact with a comment. */
   minimal: PropTypes.bool,
 
+  /** Comments can have different sizes*/
+  size: PropTypes.oneOf(['mini', 'tiny', 'small', 'large', 'big', 'huge', 'massive']),
```
layershifter: We use `SIZES` const for definitions, [example](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Message/Message.js#L94).

test/specs/views/Comment/CommentGroup-test.js
```diff
@@ -8,4 +8,7 @@ describe('CommentGroup', () => {
   common.propKeyOnlyToClassName(CommentGroup, 'collapsed')
   common.propKeyOnlyToClassName(CommentGroup, 'minimal')
   common.propKeyOnlyToClassName(CommentGroup, 'threaded')
+  common.propValueOnlyToClassName(CommentGroup, 'size', [
+    'mini', 'tiny', 'small', 'large', 'big', 'huge', 'massive',
```
layershifter: ^

layershifter (changes requested):
And we need add `size` to typings.

tarang9211:
@layershifter alright, fixed up. let me know.

layershifter (approved):

levithomason:
Fantastic, we just need a doc example for "Group Size" now.  Here are some existing group size doc examples for comparison:

[Examples/elements/Button/GroupVariations/ButtonExampleGroupSize.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/elements/Button/GroupVariations/ButtonExampleGroupSize.js)
[Examples/elements/Label/Groups/LabelExampleGroupSize.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/elements/Label/Groups/LabelExampleGroupSize.js)
[Examples/views/Statistic/Types/StatisticExampleGroupSize.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/views/Statistic/Types/StatisticExampleGroupSize.js)

The comment group size should be listed in `Comment/Variations/CommentExampleGroupSize.js`

tarang9211:
@levithomason I've added doc site examples as per your request. Only thing I noticed was that the props for size was listed this way:

screen shot 2017-02-15 at 1 24 14 am

tarang9211:
@levithomason I think this should wrap up this fix!

levithomason (approved):

levithomason:
Thanks again guys!

levithomason:

Issue #192, opened 2016-03-16
<Nag /> Component


levithomason:
http://semantic-ui.com/modules/nag.html

dcfranca:
Hi guys, is there any prediction about when Nag will be available?

levithomason:
There have been no takers on it to date.  PRs welcome!

jeffcarbs:
@levithomason - correct me if I'm wrong, but it doesn't seem like this component is fully baked in SUI-core. The docs (http://semantic-ui.com/modules/nag.html) say it's only 60% complete and need documentation (which is fine) and re-coding (which is a red flag, imho). I think we should consider removing from the v1 milestone.

@jlukic - can you confirm whether this is a fully-fleshed component or just an idea still?

levithomason:
Looks correct, going to close this for now since we don't want to chase development versions.  It would be great to have something like this or a toastr type component.

PR #1329, opened 2017-02-15
Update README.md with missing module Nag

AaronHarris:


layershifter:
There are no plans for this component before it will be fully implemented in SUI, more details in #192.

PR #1332, opened 2017-02-15
style(Search): update typings

layershifter:
Cleanups and updates typings for #1072.


levithomason:
Rebuilding...

levithomason:

PR #1334, opened 2017-02-15
style(Dropdown): update typings and propTypes usage

layershifter:


levithomason:

wcatron:
The typescripts in this PR are too strict. 

Property 'onClick' DropdownItemProps
Select uses the DropDown props which require onClose, onFocus, onMouseDown, etc.

There are others. Should I open an issue addressing all of these or do you they need to be individual?

levithomason:
Please open a new issue or PR.

PR #1337, opened 2017-02-16
isEmpty for image module instead of isNil

hiteshjoshi:
it should skip creating img tag even if children are null or empty array. isNil returns false for empty array.


levithomason:
Children can only be an object if there is an element child present:

http://www.react.run/SJsvtEXtl/3
***
image

PR #1345, opened 2017-02-18
fix(Popup): fix in propTypes

layershifter:
Fixes #1344.

PR #1354, opened 2017-02-20
Fix Popup's PropTypes to use array.

cayter:
The invalid definition is showing the warning `Warning: Invalid argument supplied to oneOfType, expected an instance of array.` on developer console.


layershifter:
@cayter Thanks for PR. However, it's duplicate #1345

PR #1355, opened 2017-02-20
breaking(Progress): control progress solely with progress

### LINKED ISSUE #1352: Progress bar always shows progress label when using value/total
The value inside the bar indicating the current progress is optional:

image

However, when the `value` and `total` props are used the progress is always shown:

```jsx
<Progress value={2} total={3}  />
```
image

We should only show the progress value when the `progress` prop is present or the `label='ratio|percent'` prop is present.

BrainMaestro:
## Breaking Changes!
The `label` prop no longer accepts `ratio|percent` as a means to control the progress readout on the bar. It now refers to the text label applied to the element. The `children` prop takes precedence however.
The `progress` prop now handles this by either accepting a `boolean` which defaults to `percent` or a string with either `ratio` or `percent`. 

### Before 
```jsx
<Progress total={2} value={1} label='ratio' />
<Progress total={2} value={1} label='percent' />
```

### After
```jsx
<Progress total={2} value={1} progress='ratio' />
<Progress total={2} value={1} progress='percent' />
<Progress total={2} value={2} label='Dragon ISS Berthing' />
```

### Unchanged
```jsx
<Progress total={2} value={2}>Dragon ISS Berthing</Progress>
```

Fixes #1352

### DIFF
--- a/src/modules/Progress/Progress.js
+++ b/src/modules/Progress/Progress.js
@@ -3,6 +3,7 @@ import _ from 'lodash'
 import React, { Component, PropTypes } from 'react'
 
 import {
+  createShorthand,
   customPropTypes,
   getElementType,
   getUnhandledProps,
@@ -51,16 +52,7 @@ class Progress extends Component {
     inverted: PropTypes.bool,
 
     /** Can be set to either to display progress as percent or ratio. */
-    label: customPropTypes.every([
-      customPropTypes.some([
-        customPropTypes.demand(['percent']),
-        customPropTypes.demand(['total', 'value']),
-      ]),
-      PropTypes.oneOfType([
-        PropTypes.bool,
-        PropTypes.oneOf(['ratio', 'percent']),
-      ]),
-    ]),
+    label: customPropTypes.itemShorthand,
 
     /** Current percent complete. */
     percent: customPropTypes.every([
@@ -75,7 +67,10 @@ class Progress extends Component {
     precision: PropTypes.number,
 
     /** A progress bar can contain a text value indicating current progress. */
-    progress: PropTypes.bool,
+    progress: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.oneOf(['ratio', 'percent']),
+    ]),
 
     /** A progress bar can vary in size. */
     size: PropTypes.oneOf(_.without(SUI.SIZES, 'mini', 'huge', 'massive')),
@@ -139,29 +134,41 @@ class Progress extends Component {
     return autoSuccess && (percent >= 100 || value >= total)
   }
 
-  showProgress = () => {
-    const { label, precision, progress, total, value } = this.props
+  renderProgress = (percent) => {
+    const { precision, progress, value, total } = this.props
 
-    if (label || progress || !_.isUndefined(precision)) return true
-    return !(_.every([total, value], _.isUndefined))
+    if (!progress && _.isUndefined(precision)) return
+    return (
+      <div className='progress'>
+        { progress !== 'ratio' ? `${percent}%` : `${value}/${total}` }
+      </div>
+    )
+  }
+
+  renderLabel = () => {
+    const { children, label } = this.props
+
+    if (!_.isNil(children)) return <div className='label'>{children}</div>
+    return createShorthand(
+      'div',
+      val => ({ children: val }),
+      label,
+      { className: 'label' }
+    )
   }
 
   render() {
     const {
       active,
       attached,
-      children,
       className,
       color,
       disabled,
       error,
       indicating,
       inverted,
-      label,
       size,
       success,
-      total,
-      value,
       warning,
     } = this.props
 
@@ -188,13 +195,9 @@ class Progress extends Component {
     return (
       <ElementType {...rest} className={classes}>
         <div className='bar' style={{ width: `${percent}%` }}>
-          {this.showProgress() && (
-            <div className='progress'>
-              { label !== 'ratio' ? `${percent}%` : `${value}/${total}` }
-            </div>
-          )}
+          {this.renderProgress(percent)}
         </div>
-        {children && <div className='label'>{children}</div>}
+        {this.renderLabel()}
       </ElementType>
     )
   }
--- a/src/modules/Progress/Progress.js
+++ b/src/modules/Progress/Progress.js
@@ -142,7 +142,7 @@ class Progress extends Component {
   }
 
   renderProgress = percent => {
-    const { 
+    const {
       precision,
       progress,
       total,
--- a/src/modules/Progress/Progress.js
+++ b/src/modules/Progress/Progress.js
@@ -69,7 +69,7 @@ class Progress extends Component {
     /** A progress bar can contain a text value indicating current progress. */
     progress: PropTypes.oneOfType([
       PropTypes.bool,
-      PropTypes.oneOf(['ratio', 'percent']),
+      PropTypes.oneOf(['percent', 'ratio']),
     ]),
 
     /** A progress bar can vary in size. */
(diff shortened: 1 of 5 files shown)


src/modules/Progress/Progress.js
```diff
@@ -140,10 +140,9 @@ class Progress extends Component {
   }
 
   showProgress = () => {
-    const { label, precision, progress, total, value } = this.props
+    const { label, precision, progress } = this.props
 
     if (label || progress || !_.isUndefined(precision)) return true
```
layershifter: Let's make this condition more simple:

```diff
-if (label || progress || !_.isUndefined(precision)) return true
+return label || progress || !_.isUndefined(precision)
```

layershifter (changes requested):

src/modules/Progress/Progress.js
BrainMaestro: Sure. I noticed it, but I didn't want to make changes that were not directly related to the task. I'll do that now

layershifter (approved):
@BrainMaestro Thanks for PR 👍

levithomason:
# Issue
I think we need to rework the relationship of the `progress` and `label` props altogether.

1. SUI core offers a `label` setting, which takes a value of `percent` or `ratio`.  This is where we got our prop `label`.  However, this is super confusing since the `label` prop controls the `progress` readout, and the `children` control the actual `label` component.

2. It also makes for this very confusing case where the `label` boolean prop does exactly the same thing as the boolean`progress` prop:

    

# Fix Proposal
I think we need to make the `progress` prop control the progress readout and make the `label` prop control the label component.

image

This would be more inline with our prop-to-subcomponent relationship we have across the entire library as well.
***
Thoughts?

BrainMaestro:
@levithomason I agree. That makes it clearer. The other way was slightly confusing.

levithomason:
Awesome.  Here are some pointers are how to handle the new propTypes and values.  The `label` prop is now a "shorthand" prop used to create the label element.  We should start with propTypes like so:

```jsx
    /** Can be set to either to display progress as percent or ratio. */
    label: customPropTypes.itemShorthand,

    /** A progress bar can contain a text value indicating current progress. */
    progress: PropTypes.oneOfType([
      PropTypes.bool,
      PropTypes.oneOf(['ratio', 'percent']),
    ]),
```

The `itemShorthand` allows users to pass a number, string, props object, or an element from which we will then create an element for them.  In this case we'll create the `className='label'` div.

We have a factory function for doing this called, `createShorthand`.  You get it by adding it to the `lib` import:

```jsx
import { createShorthand } from '../../lib'
```

This function takes an element class (string or function), a function to map number/string values to props, and finally the value you want to create an element with (number, string, props object, or another element).  Create a `labelElement` in the render method (comments can be left out):

```jsx
const labelElement = createShorthand(
  'div',                                // create a div
  val => ({ children: val }),           // map string/numbers to children
  label,                                // use the label prop as the value
  { className: 'label' }                // default props
)
```

Now, we can conditionally display the `children` or the `labelElement`, whichever the user passed.

```diff
-{children && <div className='label'>{children}</div>}
+{children ? <div className='label'>{children}</div> : labelElement}
```

BrainMaestro:
Hey @levithomason. Thanks for the task breakdown. I implemented it exactly as you suggested, but the check `{children ?` fails for any falsy value which includes `0` that is used in the `renders child number with 0 value` test. Should I use `!_.isUndefined` in the render function which is how I solved the problem? Or do you have a cleaner approach?

levithomason:
We've updated the code base to use `_.isNil(children)` so as to capture `null` and `undefined`.

BrainMaestro:
Okay has it been merged in?

src/modules/Progress/Progress.js
```diff
@@ -184,17 +189,18 @@ class Progress extends Component {
...
         <div className='bar' style={{ width: `${percent}%` }}>
           {this.showProgress() && (
             <div className='progress'>
-              { label !== 'ratio' ? `${percent}%` : `${value}/${total}` }
+              { progress !== 'ratio' ? `${percent}%` : `${value}/${total}` }
             </div>
           )}
         </div>
-        {children && <div className='label'>{children}</div>}
+        {!_.isUndefined(children) ? <div className='label'>{children}</div> : labelElement}
```
BrainMaestro: This is the line I was talking about. Can I safely change it to `{children ? <div className='label'>{children}</div> : labelElement}` without the false positives of `0` failing the truthy check?

layershifter:
```diff
- {!_.isUndefined(children) ? <div className='label'>{children}</div> : labelElement}
+ {_.isNil(children) ? this.getLabel() : <div className='label'>{children}</div>}
```

BrainMaestro: okay the label should take priority over the children? I thought it was the reverse. I will make the change

layershifter:
```jsx
{_.isNil(children) ? this.getLabel() : <div className='label'>{children}</div>}
{!_.isNil(children) ? <div className='label'>{children}</div> : this.getLabel()}
```

There conditions do same, but first looks more clear to me.

src/modules/Progress/Progress.js
```diff
@@ -190,11 +195,11 @@ class Progress extends Component {
         <div className='bar' style={{ width: `${percent}%` }}>
           {this.showProgress() && (
             <div className='progress'>
-              { label !== 'ratio' ? `${percent}%` : `${value}/${total}` }
+              { progress !== 'ratio' ? `${percent}%` : `${value}/${total}` }
             </div>
           )}
         </div>
-        {children && <div className='label'>{children}</div>}
+        {!_.isNil(children) ? <div className='label'>{children}</div> : this.getLabel()}
```
layershifter: I think we can move this condition to `getLabel`:

```diff
-  getLabel = () => {
-    const { label } = this.props
-
-    return createShorthand(
-      'div',
-      val => ({ children: val }),
-      label,
-      { className: 'label' }
-    )
-  }
+ renderLabel = () => {
+   const { children, label } = this.props
+   
+   if(!_.isNil(children)) return <div className='label'>{children}</div>

+    return createShorthand(
+      'div',
+      val => ({ children: val }),
+      label,
+      { className: 'label' }
+    )
+ }

File: src/modules/Progress/Progress.js [hunk 4]
```diff
@@ -190,11 +195,11 @@ class Progress extends Component {
         <div className='bar' style={{ width: `${percent}%` }}>
           {this.showProgress() && (
             <div className='progress'>
-              { label !== 'ratio' ? `${percent}%` : `${value}/${total}` }
+              { progress !== 'ratio' ? `${percent}%` : `${value}/${total}` }
             </div>
```
layershifter: And this to `renderProgress`:

```diff
- {this.showProgress() && (
-  <div className='progress'>
-    { progress !== 'ratio' ? `${percent}%` : `${value}/${total}` }
-  </div>
- })
+ {this.renderProgress()}
```

```diff
+ renderProgress = () => {
+   const { precision, progress } = this.props
+
+   if(!progress && _.isUndefined(precision)) return
+   return (
+     <div className='progress'>
+      { progress === 'ratio' ? `${value}/${total}` : `${percent}%` }
+    </div>
+   )
+}
```

layershifter (changes requested):

layershifter (approved):

levithomason:
Excellent!

levithomason:
This is a breaking change, so we need to update the PR title and description to include upgrade notes.  Reference #1367 or #989 for good examples.

levithomason:
Thanks for the great notes @BrainMaestro 👍

levithomason:

PR #1356, opened 2017-02-20
docs(Progress): added example for progress label prop

### LINKED ISSUE #1351: Add `label` docs usage example for Progress
The Progress component can take a `label` enum prop to control the type of label displayed within the Progres bar:

## ratio
```jsx
<Progress value='3' total='5' label='ratio' />
```

image

## percent
```jsx
<Progress value='3' total='5' label='percent' />
```

image

tarang9211:
Fixes #1351


docs/app/Examples/modules/Progress/Content/ProgressExampleLabel.js
```diff
@@ -2,7 +2,10 @@ import React from 'react'
 import { Progress } from 'semantic-ui-react'
 
 const ProgressExampleLabel = () => (
-  <Progress percent={55}>Label</Progress>
+  <div>
+    <Progress value='3' total='5' label='ratio' />
+    <Progress value='4' total='5' label='percent' />
```
layershifter: Let's split these examples to separate files.

layershifter (changes requested):

layershifter:
@tarang9211 Thanks for PR.

Let's point these examples in `Content` section under `Progress` example. Small description for these cases will be awesome.

tarang9211:
@layershifter Have a look at the recent commit and let me know if you still would like changes.

docs/app/Examples/modules/Progress/Content/ProgressExampleLabel.js
```diff
@@ -1,8 +0,0 @@
-import React from 'react'
-import { Progress } from 'semantic-ui-react'
-
-const ProgressExampleLabel = () => (
-  <Progress percent={55}>Label</Progress>
```
layershifter: Let's return this example, it shows another behaviour. However, there is #1355 that fixes bug, so we need a small update there:

```diff
-<Progress percent={55}>Label</Progress>
+<Progress percent={55} label>Label</Progress>
```

docs/app/Examples/modules/Progress/Content/index.js
```diff
@@ -17,7 +17,10 @@ const ProgressContentExamples = () => (
     <ComponentExample
       title='Label'
       description='A progress element can contain a label.'
-      examplePath='modules/Progress/Content/ProgressExampleLabel'
+      examplePath='modules/Progress/Content/ProgressExampleLabelPercent'
+    />
+    <ComponentExample
+      examplePath='modules/Progress/Content/ProgressExampleLabelRatio'
```
layershifter: Let's add descriptions:
- for `LabelPercent`: A progress element can contain a percent label.
- for `LabelRatio`: A progress element can contain a ratio label.

layershifter (changes requested):

docs/app/Examples/modules/Progress/Content/index.js
tarang9211: What should be the title of these sections?

layershifter: `description='A progress element can contain a percent label.'`

tarang9211: Sounds good, got it.

tarang9211:
Oops sorry for that, I copied the wrong thing.

layershifter (approved):

docs/app/Examples/modules/Progress/Content/ProgressExampleLabel.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Progress } from 'semantic-ui-react'
 
 const ProgressExampleLabel = () => (
-  <Progress percent={55}>Label</Progress>
+  <Progress percent={55} label>Label</Progress>
```
levithomason: The label prop is not required here.  This is confusing on SUI core's part, however, the label prop controls the label **inside of the bar** while the `Label` child here is the label **below the progress bar**:

image


I feel like these should almost have separate names so it is more clear that there are two different label positions (in the bar, under the bar).

levithomason (changes requested):

levithomason (approved):

PR #1359, opened 2017-02-21
feat(Card): add link prop

### LINKED ISSUE #1350: feat(Card): Add `link` prop to Card
This should be a boolean prop that adds the `link` className.  See SUI core [docs](http://semantic-ui.com/views/card#link-card) for more.

```jsx
<Card link />
// => <div className="ui link card"></div>
```

BrainMaestro:
Fixes #1350 

Does this need extra tests to check that the correct styles were applied to the card if it has the `link` prop?

### DIFF
--- a/src/views/Card/Card.js
+++ b/src/views/Card/Card.js
@@ -55,6 +55,9 @@ export default class Card extends Component {
     /** A card can contain an Image component. */
     image: customPropTypes.itemShorthand,
 
+    /** A card can be formatted to link to other content. */
+    link: PropTypes.bool,
+
     /** Shorthand for CardMeta. */
     meta: customPropTypes.itemShorthand,
 
@@ -100,6 +103,7 @@ export default class Card extends Component {
       header,
       href,
       image,
+      link,
       meta,
       onClick,
       raised,
@@ -110,6 +114,7 @@ export default class Card extends Component {
       color,
       useKeyOnly(centered, 'centered'),
       useKeyOnly(fluid, 'fluid'),
+      useKeyOnly(link, 'link'),
       useKeyOnly(raised, 'raised'),
       'card',
       className,
--- a/docs/app/Examples/views/Card/Content/CardExampleLinkCardProp.js
+++ b/docs/app/Examples/views/Card/Content/CardExampleLinkCardProp.js
@@ -6,8 +6,10 @@ const CardExampleLinkCardProp = () => (
     link
     header='Rick Sanchez'
     meta='Scientist'
-    description='Rick is a genius scientist whose alcoholism and reckless,
-    nihilistic behavior are a source of concern for his family'
+    description={[
+      'Rick is a genius scientist whose alcoholism and reckless,',
+      ' nihilistic behavior are a source of concern for his family',
+    ].join('')}
   />
 )
 
--- a/docs/app/Examples/views/Card/Content/CardExampleLinkCardProp.js
+++ b/docs/app/Examples/views/Card/Content/CardExampleLinkCardProp.js
@@ -1,0 +1,14 @@
+import React from 'react'
+import { Card } from 'semantic-ui-react'
+
+const CardExampleLinkCardProp = () => (
+  <Card
+    link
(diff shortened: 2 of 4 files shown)


layershifter:
Let's also add example for `link` prop like in SUI docs.

BrainMaestro:
@layershifter I added one

docs/app/Examples/views/Card/Content/CardExampleLinkCard.js
```diff
@@ -2,12 +2,20 @@ import React from 'react'
...
+      meta='Friend'
+      description='Elliot is a sound engineer living in Nashville who enjoys playing guitar and hanging with his cat.'
+    />
+    <Card
+      link
+      header='Rick Sanchez'
+      meta='Scientist'
+      description='Rick is a genius scientist whose alcoholism and reckless, nihilistic behavior are a source of concern for his family'
+    />
+  </Card.Group>
```
layershifter: Let's split these examples to separate files.

layershifter (changes requested):

docs/app/Examples/views/Card/Content/CardExampleLinkCard.js
BrainMaestro: Okay, but in the actual [docs](http://semantic-ui.com/views/card#link-card), they're next to each other. Since they're showing the same property functionality, shouldn't they be together?

levithomason: You can make these appear together while still having separate example files.  The advantage is that each example will have its own code editor.

To do this, create a new file and add it to the `index.js` in this directory just after the `CardExampleLinkCard.js` entry, but do not include a `title` nor `description`:

**docs/app/Examples/views/Card/Content/index.js**
```jsx
    <ComponentExample
      title='Link'
      description='A card can be formatted to link to other content.'
      examplePath='views/Card/Content/CardExampleLinkCard'
    />
    <ComponentExample
      examplePath='views/Card/Content/CardExampleLinkCardProp'
    />
```

The second example will render directly below the title/description of the first example.  This is common in our docs for grouping examples.

src/views/Card/Card.js
```diff
@@ -55,6 +55,9 @@ export default class Card extends Component {
     /** A card can contain an Image component. */
     image: customPropTypes.itemShorthand,
 
+    /** A card can be formatted to behave like a link (hover effects, pointer cursor) */
```
levithomason: Let's go with our doc site description:
`A card can be formatted to link to other content.`

Since, the specifics of the style can vary from theme to theme.

BrainMaestro:
I updated with your suggestions

levithomason (approved):

levithomason:
Changes look great, I'll merge as soon as tests pass:

```
/home/ubuntu/Semantic-UI-React/docs/app/Examples/views/Card/Content/CardExampleLinkCardProp.js
  9:1  error  Line 9 exceeds the maximum line length of 120  max-len
```

levithomason:
I've just pushed a fix for the description prop syntax error.

BrainMaestro:
Yeah I see it. Odd, didn't show up as an error locally

levithomason:
Looks like this is a lint error, so you'd only see it by running `npm run lint`.  We don't run lint automatically as it can take ~20 seconds to run and would hinder development workflow.  We rely heavily on the PR status checks instead.

Not as intuitive though, sorry about that :)

BrainMaestro:
Oh yeah I ran the lint check locally as well and it was fine. Oh well, is this good to merge now?

levithomason:
Oh, that is odd... perhaps outdated deps?  Not sure, but, we're good to go! Thanks much for the contribution, hope to have you around for some more 👍

BrainMaestro:
Thanks for being patient! I have had a great experience so far with this project, and I hope to contribute more.

levithomason:
Whoops, @layershifter we missed typings here :/

levithomason:

Issue #607, opened 2016-10-03
Performance: Update usage of stateless, functional components to class-based components when event handlers are used


jeffcarbs:
_See below for original conversation regarding "RFC: Memoize event handlers in stateless, functional components". These are the TODOs that came out of the discussion._

Functional components that need to be turned into a class (Found via grep: `const handle`):
- [x] #619 `AccordionTitle`
- [x] #619 `BreadcrumbSection`
- [x] #619 `Card`
- [x] #619 `DropdownItem`
- [x] #619 `Form`
- [x] #619 `Label`
- [x] #619 `MenuItem`
- [x] #619 `SearchResult`
- [x] #619 `Step`

Other places where functions are created within render (Found via grep: `const on[A-Z]`):
- [x] #640 `Rating`
- [x] #1375 `Accordion`
- [x] `Search` - `onMouseDown={e => e.preventDefault()}` that line can actually be removed I think. The analogous updates from [this PR](https://github.com/TechnologyAdvice/stardust/pull/500/files#diff-609214949026e594dbfd98481664c928L852) were carried over and this line should have been removed.
- [x]  #1360 `Modal` - Move `ref=` function to `handleRef` callback
- [x]  #1360 `Dropdown`
  - Move `ref=` function in `render` to `handleRef` callback
  - Move `ref=` function in `renderSearchInput` to `handleSearchRef` callback
  - Move `ref=` function in `renderSearchSizer` to `handleSizerRef` callback

There may be others so feel free to update this list!

---

_Original discussion below regarding "RFC: Memoize event handlers in stateless, functional components"_

---

Most of the conversation around this is in reference to actual components, not stateless functions. I've googled around a bit and the closest thing I've found to a discussion on this topic is: https://github.com/airbnb/javascript/issues/784. This is also more of a general React question for you guys. I've run into this in my projects and haven't really settled on a great way to handle it.

---

I think using stateless, functional components is ideal and something we should strive for. However, initializing new arrays or functions inside a render (which we do) is [considered an anti-pattern](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#no-bind-or-arrow-functions-in-jsx-props-jsx-no-bind) for performance reasons. Consider the [`DropdownItem` component](https://github.com/TechnologyAdvice/stardust/blob/master/src/modules/Dropdown/DropdownItem.js) (pasting a simplified version here):

``` js
function DropdownItem(props) {
[...]

PR #1360, opened 2017-02-21
fix(Dropdown|Modal|TextArea): update refs handlers

layershifter:
Addresses #607.


src/addons/TextArea/TextArea.js
```diff
@@ -60,6 +60,8 @@ class TextArea extends Component {
     this.updateHeight(e.target)
   }
 
+  handleRef = c => (this.rootNode = c)
```
levithomason: Nice, we should also standardize these ref variable names.  We currently use `*Node`, `_*` and `*Ref` in various places.

A ref can either be a node (DOM component) or a reference to a class instance (Composite component).  Due to this, I vote we go with `this.*Ref` for clarity and accuracy.

levithomason (changes requested):

src/addons/TextArea/TextArea.js
layershifter: Yep, it makes sense 👍

Do I have to change the names to these?
```diff
-handleRef = c => (this.rootNode = c)
+handleRef = c => (this.textAreaRef = c)
```
```diff
-handleSearchRef = c => (this._search = c)
+handleSearchRef = c => (this.searchRef = c)
```

levithomason: Looks right to my eyes 👍

levithomason (approved):

levithomason:
I've updated the refs per our agreement.  Refs at the root I simply used `this.ref` and `this.handleRef` for the naming convention.  Mirrors how we handle `onClick` `this.handleClick` vs `onItemClick` `this.handleItemClick`.

Give it a look over and let me know if we're good to go here.

layershifter:
I've updated `Checkbox` handler. LGTM :+1:

levithomason:

PR #1361, opened 2017-02-21
fix(Checkbox): onFocus and onBlur events

### LINKED ISSUE #1335: Checkbox: focus not properly activated on click
### Steps
1. Add a checkbox: `<Checkbox label="Label"/>`
2. Click the checkbox or the label.

### Expected Result
The checkbox should get focus. `onFocus` should be triggered.

### Actual Result
On Windows 7 64-bit:
Firefox 51.0.1 (32-bit): The checkbox does not get focus. `onFocus` is not triggered.
Chromium 58.0.3009.0 (64-bit): The checkbox does not get focus. `onFocus` is not triggered.
Internet Explorer 11.0.9600.18537: `onFocus` is triggered (as well as `onBlur`), but the focus is not visible.

In all three browsers the focus works as expected when using the tab key.

[Checkboxes in core Semantic UI](http://semantic-ui.com/modules/checkbox.html) work as expected.

### Version
0.64.8

### Testcase
http://codepen.io/anon/pen/apMgvO

tarang9211:
Fixes #1335


levithomason:
Rather than adding callbacks, we need to set focus and add tests, see https://github.com/Semantic-Org/Semantic-UI-React/issues/1335#issuecomment-281416074.

tarang9211:
@layershifter does this look right? what about the blur event?

levithomason:
I left a comment on the issue for the blur event.  Let's get tests passing as well.

tarang9211:
Sweet thanks! For the test, we're simulating a `onMouseDown` event right?

levithomason:
Indeed.

levithomason:
```
Checkbox
  ✖ handles events transparently
    PhantomJS 2.1.1 (Linux 0.0.0)
    <Checkbox onMouseDown={handleMouseDown} />
                           ^ was not called once on "mouseDown".
                             You may need to hoist your event handlers up to the root element.
```

Since the `onMouseDown` is now captured, we need to also call `this.props.onMouseDown` in our handler with the original event.  This way, the user's prop is also called.  Otherwise, we are intercepting it and swallowing it.

Since we're now handling the prop, it also needs added to `propTypes`.  In order to keep consistent with our callback signatures, we should pass `this.props` as the second callback parameter.  You can reference many other component's callbacks for examples on how to add the propType doc block and how to pass back the right arguments.

src/modules/Checkbox/Checkbox.js
```diff
@@ -196,7 +200,10 @@ export default class Checkbox extends Component {
     else computedTabIndex = disabled ? -1 : 0
 
     return (
-      <ElementType {...rest} className={classes} onClick={this.handleClick} onChange={this.handleClick}>
+      <ElementType {...rest} className={classes}
+        onClick={this.handleClick} onChange={this.handleClick}
```
levithomason: Need a line break between the props here.

tarang9211:
you mean something like  [this](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Button/Button.js#L171) @levithomason?

src/modules/Checkbox/Checkbox.js
```diff
@@ -149,6 +149,10 @@ export default class Checkbox extends Component {
     }
   }
 
+  handleMouseDown = (e) => {
+    _.invoke('focus', this.checkboxRef)
```
levithomason: This is correct, however, we also need to call the user's `onMouseDown` prop here if they provide one.  That is because we handling the prop, so it isn't auto spread on the rendered component.  Even if it were, we are also setting the onMouseDown event ourselves, so our handler is the only one that will ever be called.

So, if they pass:
```jsx
<Checkbox onMouseDown={() => console.log('hi')} />
```
Then nothing will happen onMouseDown.  To ensure they can also pass a `onMouseDown` handler, we need to also call it in our handler for them.   Since mouse down is called before the focus event, let's call it in the right order as well:

```jsx
handleMouseDown = (e) => {
  _.invoke('onMouseDown', this.props, [e])
  _.invoke('focus', this.checkboxRef)
}
```

src/modules/Checkbox/Checkbox.js
levithomason: Still need the line break here 😉 

```diff
-        onClick={this.handleClick} onChange={this.handleClick}
+        onClick={this.handleClick}
+        onChange={this.handleClick}
```

levithomason (changes requested):
I've clarified the changes necessary here.

tarang9211:
Sounds good @levithomason. I'll get to this in a bit.

tarang9211:
@levithomason Just pushed a fix that takes care of this issue, I believe. Let me know if we need anything else.

levithomason:
Changes look good, we just need some tests for this and we can merge 👍

tarang9211:
@levithomason I should also add `onMouseDown` to the component props, right? Similar to how onClick and onChange are?

tarang9211:
@levithomason without adding any test cases, this one seems to fail. are we not exposing the prop properly at the user level?

screen shot 2017-03-16 at 1 43 04 am

tarang9211:
@layershifter tests still fail locally.

tarang9211:
@layershifter pushed a new commit, it could be a potential fix. Let me know if it looks ok

levithomason:
This component looks good conformance as far as conformance tests go.  We now just need a test that asserts that:
```js
it('sets focuses the input onMouseDown', () => { ... })
```

tarang9211:
> it('sets focuses the input onMouseDown', () => { ... })

What exactly should be checked in this test? I was basing my implementation on the `onClick` and `onChange` tests. Is that a good approach?

levithomason:
The `mouseDown` event should be fired on the outer checkbox div and an assertion should be made that the `document.activeElement` is indeed the `input` element.

layershifter:
@levithomason I've made some cleanups and added test

levithomason:
💥 thanks much!

levithomason (approved):

levithomason:

Issue #1348, opened 2017-02-19
How to programatically submit a <Form> component


achimkoellner:
Please add documentation on how to programatically submit a <Form> component.
I guess it's about adding a ref and calling submit on it or is there another option?

levithomason:
Manually call the method that you'd otherwise pass to the `onSubmit` prop.  You should capture your form data in state using form control `onChange` handlers.

PRs welcome :)

dvdzkwsk:
Also, unless it's explicitly documented, you should avoid using internal methods of the form (or any other) component. This is because doing so means reaching through what should be an opaque abstraction into internal methods that may change at any time.

So, unless these methods become a formal part of the public API, it's likely best to follow @levithomason 's advice. Furthermore, it's generally bad practice to reach _downward_ through the React tree in order to initiate side effects.

levithomason:
FWIW, I actually plan to remove the onSubmit handler as it has caused lots of confusion and encouraged bad practice.  The current implementation is DOM based anyhow, which just makes it this much worse.

Issue #833, opened 2016-11-11
Submit Form with javascript does not trigger onSubmit


tomitrescak:
Is it possible to submit SUI form via javascript?
I tried pretty much everything I could but the form is not launching the onSubmit.
The only thing that works is to find the submit button and click() on it.

I tried

```
<Form name="myForm" .... />
and
document.myForm.submit();
```

But this does not trigger the onSubmit behaviour.

WHY I need it? I have very large forms and I have keyboard shortcuts that allow to save the form.

levithomason:
I've never tried submitting React forms like that, but apparently, React does not call `onSubmit` when you try to submit the form through the DOM.  This makes some sense, as you are trying to invoke a virtual DOM method (`onSubmit`) using the real DOM form and API.  I am betting this doesn't work because the real DOM is only the _output_ of the component, not the actual component itself.  Here is a vanilla React example that shows React simply does not support this:

http://codepen.io/levithomason/pen/JbGpwR

I think the more "React way" of doing this would be to use controlled form controls that hoist their values into the component state or app state using their `onChange` listener.  Then, your `onSubmit` handler just uses the data from state to do what it needs to do.  Finally, you can invoke what ever function you passed to onSubmit directly and achieve the same results as a submission.

Hope that helps!

PR #1367, opened 2017-02-22
breaking(Form): remove onSubmit serializer

### LINKED ISSUE #1362: fix(Form): Remove Form `onSubmit` prop and serializer
See #1348 and #833 for context.  We will remove the special behavior of the `onSubmit` prop in favor of the vanilla React `onSubmit` behavior.  This includes the serializer than passes the form data to the `onSubmit` callback.

More info:

We currently provide an `onSubmit` handler for the Form.  This handler includes a Form data serializer.   There are two issues with this:

1. The serializer is DOM based and was planned to be migrated to migrated to a React callback mechanism, but this never happened.  Interacting with the DOM is an antipattern in a React app.

1. The callback and its serializer have encouraged improper usage of Forms and events.  Users are encouraged to try to reach down the tree to invoke the `onSubmit` event so that they can retrieve the data out of their Form.  This is also an antipattern in React.  Data should be hoisted up the tree using controlled Form control components.  Then, when the form data is needed, it is readily available in some higher state.

In the long term, we may still provide a component that makes working with Forms easier.  However, this would be more akin to the DataTable #1133 which wraps a simple stateless Table and makes working with data state easier.  We would likely provide some kind of wrapper that could manage the parent-child relationship of the Form and its controls and provide simple top level callbacks for capturing data.  This might be similar to the AdvancedForm #527 experiment.

levithomason:
# Breaking Changes!

Form `onSubmit` no longer includes a second parameter with the serialized form values.  The serializer was accomplished by reading from the DOM, which is an anti-pattern in React.  The `onSubmit` prop itself remains in-tact and functions identically to a vanilla React form's `onSubmit` prop.

Fixes #1362.  See the related issues #1348 and #833 for more context.

## Before
```jsx
class OrderView extends React.Component {
  handleSubmit = (e, formData) => {
    e.preventDefault()
    console.log('Quantity:', formData.quantity)
  }

  render() {
    return (
      <Form onSubmit={this.handleSubmit}>
        <Form.Input name='quantity' defaultValue={this.state.quantity} />
      </Form>
    )
  }
}
```

## After
You should hoist your data up the tree by either setting state, or using a state management solution such as [Redux](http://redux.js.org/).

### setState Example
```jsx
class OrderView extends React.Component {
  state = {}

  handleChange = (e) => this.setState({ quantity: e.target.value })

  handleSubmit = (e) => {
    e.preventDefault()
    console.log('Quantity:', this.state.quantity)
  }

  render() {
    return (
      <Form onSubmit={this.handleSubmit}>
        <Form.Input name='quantity' value={this.state.quantity} onChange={this.handleChange} />
      </Form>
    )
  }
}
```


levithomason:

edmundtfy:
then where would placeholder be...?

qoalu:
Here is a handleToggle for Checkbox Groups state tracking compatible with the new API. Checkboxes have an unmanaged state and just pass current value up.

**JSX:**
```
<Form.Checkbox key={key} label={text} name='groupName' value={value1} onChange={this.handleToggle} />
<Form.Checkbox key={key} label={text} name='groupName' value={value2} onChange={this.handleToggle} />

<Form.Checkbox key={key} label={text} name='groupName' value={value3} onChange={this.handleToggle} />

```
**Action Function:**
```
handleToggle = (e, { name, value, checked }) => {
	this.setState({ [name]: {
		...this.state[name],
		[value]: checked
	}
	});
};
```
**Resulting State:**
```
this.state = {
	groupName: {
		"value1": false,
		"value2": false,
		"value3": false
	}
}
```

PR #1369, opened 2017-02-22
fix(Card): add link prop to typings

layershifter:
We've merged #1359, but forgot to add prop to typings.


levithomason:

PR #1372, opened 2017-02-22
fix(Icon): Icon component PropType update to FontAwesome 4.7.0

### LINKED ISSUE #1371: Icon component PropType update to FontAwesome 4.7.0
It appears the PropType validation for `name` on the Icon component needs to get updated to support the new FontAwesome 4.7.0 icons that were just added in the most recent SUI release.

### Steps
Set the `name` property to "address book outline" on an Icon component or on a `icon` property for a List.Item component

### Expected Result
The "address book outline" icon should show and there should be no console error for an invalid string passed.

### Actual Result
The icon is rendered correctly but there is a console error:
```
Warning: Failed prop type: Invalid prop `name` of value `address book outline` supplied to `Icon`.

Instead of `address book outline`, did you mean:
  - disk outline
  - bell outline
  - hdd outline
```

### Version
0.65.0 SUIR
2.2.9 SUI

### Testcase
http://codepen.io/anon/pen/gmYGXZ

kamdz:
fixes #1371

I took names and aliases from changelog on here http://fontawesome.io/whats-new/

### DIFF
--- a/src/lib/SUI.js
+++ b/src/lib/SUI.js
@@ -76,7 +76,7 @@ export const LITERAL_OBJECTS_ICONS = [
   'fire extinguisher', 'rocket', 'anchor', 'bullseye', 'sun', 'moon', 'fax', 'life ring', 'bomb', 'soccer',
   'calculator', 'diamond', 'sticky note', 'sticky note outline', 'law', 'hand peace', 'hand rock', 'hand paper',
   'hand scissors', 'hand lizard', 'hand spock', 'tv', 'thermometer empty', 'thermometer full', 'thermometer half',
-  'thermometer quarter', 'thermometer three quarters', 'bath', 'snowflake', 'snowflake outline',
+  'thermometer quarter', 'thermometer three quarters', 'bath', 'snowflake outline',
 ]
 export const SHAPES_ICONS = [
   'crosshairs', 'asterisk', 'square outline', 'certificate', 'square', 'quote left', 'quote right', 'spinner', 'circle',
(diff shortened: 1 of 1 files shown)


levithomason:
SUI doesn't use font-awesome css classes directly, they are aliased to more "semantic" names.  Example:

### FA
image

### SUI
image

See the [line comment](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/SUI.js#L30) at the top of the SUI icons arrays which indicates where the names should be retrieved from in SUI core, namely, `icon.css`:

See https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/icon.css#L486

Sidenote, looks like this file was updated 23hrs ago :)  Thanks for the quick response to the change!

levithomason:
This actually now has me wondering if/how we can achieve this programmatically.  Perhaps that is for another PR, but it would be handy to auto parse this data from the installed `semantic-ui-css` package.

kamdz:
Ok, I will change it. For now, manually...

kamdz:
class names are now 1:1 to this from SUI-css

levithomason:
Just one lint issue to fix
```
/home/ubuntu/Semantic-UI-React/src/lib/SUI.js
  59:1  error  Line 59 exceeds the maximum line length of 120  max-len
```

levithomason:
Also could you please put `fixes #1371` in the PR description in order to auto close the issue on merge?  Thanks!

https://github.com/blog/1506-closing-issues-via-pull-requests

layershifter:
Also, please update Icon typings file.

kamdz:
@levithomason Done
@layershifter But why and how? Now it is:
```javascript
/** Name of the icon. */
  name?: string;
```
and for me it's enough. If we change it to SUI.ICONS_AND_ALIASES then users can't set their own custom names for Icon component

layershifter:
@kamdz Nevermind, I forgot that we don't have icons list in typings.

levithomason:
```
/home/ubuntu/Semantic-UI-React/src/lib/SUI.js
  59:106  error  Trailing spaces not allowed  no-trailing-spaces
```

kamdz:
done

levithomason:
Thanks much!

levithomason:

Issue #1182, opened 2017-01-18
TypeScript: can't import components individually


dennari:
When I try to import an individual component, for example

```javascript
import { Icon } from 'semantic-ui-react/dist/commonjs/elements/Icon';
```
I get a runtime error:

```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of `Component`.
```

For some reason `Icon` is `undefined`. I added a PR which includes a test for this scenario: #1181.

layershifter:
I've investigated this issue, problem is quite simple.

---

When you write this in TS, it's correct because typings doesn't have default export.
```ts
import { Icon } from 'semantic-ui-react/dist/commonjs/elements/Icon';
import { IconGroup } from 'semantic-ui-react/dist/commonjs/elements/Icon';
```

However, when it's transpiled in JS, it's incorrect, `Icon/index.js` has only default export, so we need to write this:
```ts
import Icon from 'semantic-ui-react/dist/commonjs/elements/Icon';
import IconGroup from 'semantic-ui-react/dist/commonjs/elements/Icon/IconGroup';
```

But, it will not work with TS, because typings says another :thinking: So, we need update typings as:
```diff
-export const Icon: IconComponent;
+export default IconComponent;
// ...
-export const IconGroup: React.StatelessComponent<IconGroupProps>;
+const IconGroup: React.StatelessComponent<IconGroupProps>;
```

But, I'm not sure that this will work after these updates:
```ts
import { Icon, IconGroup } from 'semantic-ui-react'
```

dennari:
Ah now I can see the problem, thanks @layershifter. It seems that the `index.d.ts` typings files in `src/*/*` are not actually reflecting the corresponding `index.js` files. Take the `Icon` for example:

```js 
// src/elements/Icon/Icon.js
...
export default Icon
```
and
```js
// src/elements/Icon/index.js
export default from './Icon'
```
but
```js
// src/elements/Icon/index.d.ts
...
export const Icon: IconComponent;
...
export const IconGroup: React.StatelessComponent<IconGroupProps>;
```

There are two issues here:

1. `index.d.ts` doesn't specify a default export but `index.js` does (the `Icon` class)
2. `index.d.ts` exports `IconGroup`, but `index.js` doesn't (it only exports the default export)

There are a couple of ways to solve this:

[...]

PR #1395, opened 2017-02-28
style(typings): Update structure

layershifter:
Refence: #1182.

This PR changes typings structure to correspond JS-files.


layershifter:
@levithomason It's ready to review, however, I don't have idea how it can be really done :rabbit: 

I've checked with import from root and some components `index.d.ts` with compiler, it works.

levithomason:
>I don't have idea how it can be really done 🐰

Haha, yep.  Let's ship it and see what happens 💥

levithomason:

Issue #1169, opened 2017-01-16
Advertise all underlying propTypes and typings in wrapper components


levithomason:
See #1163 for the inspiration here.

Some components wrap others, like a Popup wrapping a Portal.  However, we do not advertise the Portal props as propTypes nor typings of the Popup.  This means there is no way for users to know what props are available.

### Add propTypes
I propose we explicitly list all known props in propTypes and typings of the underlying component when a component wraps another component.

### Add Examples
We should then also add doc site examples showing use of these props.  Example, Popup mouse enter/leave delays are not shown but are available as it composes the Portal which has mouse enter/leave delays.

### Task List
Here is a brain dump of wrapper components, there may be more.  @jcarbo / @layershifter feel free to update this list at will:

- [ ] Popup -> Portal
- [ ] Modal -> Portal
- [ ] Confirm -> Modal
- [ ] Select -> Dropdown
- [ ] Radio -> Checkbox
- [ ] Table.Footer -> Table.Header
- [ ] Table.HeaderCell -> Table.Cell 
- [ ] Form.Button -> FormField + Button
- [ ] Form.Checkbox -> FormField + Checkbox
- [ ] Form.Dropdown -> FormField + Dropdown
- [ ] Form.Input -> FormField + Input
- [ ] Form.Radio -> FormField + Radio
- [ ] Form.Select -> FormField + Select
- [ ] Form.TextArea -> FormField + TextArea

layershifter:
LGTM, I'll will make this updates in PRs that affects typings.

levithomason:
Great, thanks!  I've left Select / Radio unfinished on the list as I don't see all the props of their underlying components listed in their propTypes.  Example, Dropdown takes a `search` prop so the Select propTypes should also list `search`.

layershifter:
@levithomason I need there some clarification. Most of these pairs has a correct inheritance in typings, however, I need to make an additional check.

I'm also support an idea with docs, but there is problem with `propTypes`. I see two ways to deal with them.

```jsx
TableHeaderCell.propTypes = {
  ...TableCell.propTypes,
  as: customPropTypes.as,
  children: PropTypes.node,,
}
```

```
TableHeaderCell.propTypes = {
  as: customPropTypes.as,
  children: PropTypes.node,
  collapsing: PropTypes.bool,
  icon: customPropTypes.itemShorthand,
}
```

[...]

PR #1399, opened 2017-02-28
feat(tableheadercell): add collapsing

DomiR:
Collapsing is missing from headercells.

### DIFF
--- a/src/collections/Table/TableHeaderCell.js
+++ b/src/collections/Table/TableHeaderCell.js
@@ -13,9 +13,10 @@ import TableCell from './TableCell'
  * A table can have a header cell.
  */
 function TableHeaderCell(props) {
-  const { as, className, sorted } = props
+  const { as, className, sorted, collapsing } = props
   const classes = cx(
     useValueAndKey(sorted, 'sorted'),
+    useKeyOnly(collapsing, 'collapsing'),
     className
   )
   const rest = getUnhandledProps(TableHeaderCell, props)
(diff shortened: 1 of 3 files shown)


levithomason:
Thanks, we'll also need to update the `index.d.ts` file in this directory as well as add a test call in `TableHeader-test.js`.

levithomason:
Nice, just need a common test now.

layershifter:
@levithomason Wait a second there. `TableHeaderCell` is only simple wrapper around `TableCell`, so `collapsing` prop will be pass as unhandled.

So, I think it's job for #1169. It's already done in the [typings](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Table/index.d.ts#L201).

levithomason:
This is true, however, there is also a className and [CSS rule](https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/table.css#L452) for it:

```css
/*--------------
    Collapsing
---------------*/

.ui.table th.collapsing,
.ui.table td.collapsing {
  width: 1px;
  white-space: nowrap;
}
```

layershifter:
@levithomason It's not the single property that `th` has in CSS. I still think that it's not best idea to handle `collapsing` prop in `TableHeaderCell` while we handle it too in child `TableCell`.

levithomason:
I see what you're saying, the header cell composes the regular cell.

This plays into #1169.  There is no documentation visibility with the inheritance here.  We could handle this by only adding the propTypes for it, or, pending the babel plugin update for unhandled props, we could spead the cell prop types and update the docs to show "composes" propTypes as well.

levithomason:
Note that this prop is supported as it is passed to the underlying TableCell.  Typings extend the TableCell props as well, so there is support there.   Lastly, we're very close to finishing updates to the doc site to show inherited or "composed" propTypes as well, track #1169.

Appreciate the notice and PR here.  Closing this PR in favor of those efforts.

PR #1401, opened 2017-02-28
fix(Dropdown|Select|Form): fixes in typings

### LINKED ISSUE #1400: Modal, how to get access to onBlur method?
I'm using the modal in a slightly different way, controlling it's opening/closing using an event dispatcher. I'd like to use the native React event handler onBlur so that I can do so:
<Modal open={true} dimmer={false} onBlur={this.handleClose} ... >

private handleClose =  () => ModalWindowService.closeModal();

Of course, since I'm using TypeScript and I'm a n00b at it, I get this error:

ERROR in ./src/components/Modal/AssignBasketModal/AssignBasketModal.tsx
(34,55): error TS2339: Property 'onBlur' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ModalProps, ComponentState>> & { childre...'.

I've realized that setting closeOnDocumentClick to true would do the same thing, however, I noticed this functionality doesn't work if the Modal open={true}

layershifter:
Fixes #1400.

### DIFF
--- a/src/addons/Select/index.d.ts
+++ b/src/addons/Select/index.d.ts
@@ -8,7 +8,6 @@ import {
 } from '../../modules/Dropdown';
 
 export interface SelectProps extends DropdownProps {
-  selection: true;
 }
 
 interface SelectComponent extends React.StatelessComponent<SelectProps> {
--- a/src/collections/Form/index.d.ts
+++ b/src/collections/Form/index.d.ts
@@ -46,10 +46,6 @@ interface FormProps {
   widths?: 'equal';
 }
 
-interface FormOnSubmitData extends FormProps {
-  formData: {[key: string]: any};
-}
-
 interface FormClass extends React.ComponentClass<FormProps> {
   Field: typeof FormField;
   Button: typeof FormButton;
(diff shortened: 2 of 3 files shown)


src/addons/Select/index.d.ts
```diff
@@ -8,7 +8,6 @@ import {
 } from '../../modules/Dropdown';
 
 export interface SelectProps extends DropdownProps {
-  selection: true;
```
layershifter, line 11: `true` is not valid type, so it's cruft.

src/collections/Form/index.d.ts
```diff
@@ -46,10 +46,6 @@ interface FormProps {
   widths?: 'equal';
 }
 
-interface FormOnSubmitData extends FormProps {
```
layershifter, line 49: Cruft.

src/modules/Dropdown/index.d.ts
```diff
@@ -106,71 +106,71 @@ export interface DropdownProps {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onBlur: (event: React.KeyboardEvent<HTMLElement>, data: DropdownProps) => void;
+  onBlur?: (event: React.KeyboardEvent<HTMLElement>, data: DropdownProps) => void;
```
layershifter, line 109: Prop is optional

levithomason:
Whoops, I missed that one on review as well.  Thanks for the update.

levithomason:

PR #1407, opened 2017-03-01
fix(typings): Change some props of Search to optional

koenvg:
Some of the props of Search were wrongly labelled as mandatory.


levithomason:
Thanks, however, we just merged a major refactor to the typings structure.  The files are now 1:1 with the JS filenames.  Would you mind updating this PR to modify `Search.d.ts` instead?

koenvg:
Not a problem, I'll update it tonight.

levithomason:

PR #1419, opened 2017-03-06
chore(package): update babel-plugin-transform-react-handled-props

layershifter:
I've updated plugin (fix and test case) for #1169. Now, it ignores object's spread.

See [PR](https://github.com/layershifter/babel-plugin-transform-react-handled-props/pull/7) and [this comment ](https://github.com/Semantic-Org/Semantic-UI-React/issues/1169#issuecomment-283164599)for details.


levithomason:
❤️

PR #1425, opened 2017-03-08
style(Confirm): update typings

layershifter:
Updates typings of `Confirm` component for #1169


levithomason:

PR #1427, opened 2017-03-09
style(mixed): update typings for usage with UMD

### LINKED ISSUE #1392: index.d.ts file with semanticUIReact
How can I have typed `semanticUIReact` in a TypeScript file? I have tried:

    /// <reference path="../node_modules/semantic-ui-react/index.d.ts" />
    const { Menu } = semanticUIReact

But I get `cannot find name semanticUIReact` error.

layershifter:
Fixes #1392.


levithomason:
Agreed on the extra reviewers here. I'm not sure about exporting a single namespace versus individual exports.

levithomason:
Perhaps @czb could test this out as he is requesting it.  @dennari and @asvetliakov are also superb TS users and may have opinions here.

layershifter:
Needs rebase after #1467.

levithomason:
Fixed.

layershifter:
PR was merged, I've need make rebase and actualize this PR.

layershifter:
I want to close this for housekeeping. I don't see there any feedback from users while changes in this PR looks very ugly. I'm glad to see some feedback, possible it can be done more elegant.

levithomason:
Agreed, we can always revisit this but I would like to clean up the open issues and PR's as much as possible.

Issue #1424, opened 2017-03-07
Issue with props merging on Menu


layershifter:
Extracted from #1375.

### Steps
1. Open codepen test case.
2. Click on an item of menu.

### Expected Result
Call of `handleItem` and `handleMenu` on click on `MenuItem`.

### Actual Result
Only `handleItem` was called.
In fact, we don't call [`handleItemClick`](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Menu/Menu.js#L136) when `onClick` was passed to an `MenuItem`.

### Version
Latest.

### Testcase
http://codepen.io/layershifter/pen/aJBmwz

levithomason:
This is not a factory bug but the menu is incorrectly handling the items.

First, when creating shorthand items, the menu passes default props as an object.  This means the items' onClick handler will override the default prop onClick handler:

```
    return _.map(items, (item, index) => MenuItem.create(item, {
      active: activeIndex === index,
      index,
      onClick: this.handleItemClick,
    }))
```

The menu should use the callback form of defaultProps which receives the user's props as an argument. This way, the onClick handler can wrap the user's onClick handler as well. 

Second, the handleItemClick method assumes the items to be an array of objects from which it tries to lookup the clicked index and call the item's  onClick handler. Instead, we should leave this to the wrapped item handler. 

I'm not sure that we should auto wrap handlers in the factories as there may be cases where we want the override behavior.

layershifter:
I'll make PR that updates tests and Menu items handling.

levithomason:
I may have spoken prematurely on this as well.  It will take more effort to diagnose properly.  The defaultProps I believe are always overriden by the user's props.  It is possible that even wrapping the user's props in the default props will be of no use as the user's handler will override the wrapped default handler:

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/factories.js#L46
```jsx
const props = { ...defaultProps, ...usersProps }
```

Seems we need a way to define not `defaultProps` but some kind of `overrideProps`.

layershifter:
It was fixed in #1428 :v:

PR #1428, opened 2017-03-09
feat(factories): add overrideProps

layershifter:
Refence: #1424


layershifter:
@levithomason I've added `overrideProps` to handle situations like with `Menu`. It would be great to receive your initial feedback there.

levithomason:
I'd like to think about this one a bit.

layershifter:
@levithomason It will be great receive some feedback from you. Of course, when you will have enough time

levithomason:
Whoops, forgot about this thanks!  I'm working on the Tabs ATM and also need this feature.  Looking now...

src/lib/factories.js
```diff
@@ -105,8 +114,8 @@ export function createShorthandFactory(Component, mapValueToProps, generateKey)
     throw new Error('createShorthandFactory() Component must be a string or function.')
   }
 
-  return (val, defaultProps) => {
-    return createShorthand(Component, mapValueToProps, val, defaultProps, generateKey)
+  return (val, defaultProps, overrideProps) => {
+    return createShorthand(Component, mapValueToProps, val, defaultProps, overrideProps, generateKey)
```
levithomason: The only comment I have is that the arguments here are getting a little wild.  There are cases where you'd want to set `generateKey` but not `defaultProps` nor `overrideProps`.  That usage is ugly:

```jsx
createShorthand(
  Menu, 
  items => ({ items }),
  null, // defaultProps
  null, // overrideProps
  true, // generateKey
)
```

Perhaps we move these optional arguments to an options object?

```jsx
createShorthand(
  Menu, 
  items => ({ items }), 
  { generateKey: true }
)
```
**EDIT**
Fixed completely incorrect signatures.

layershifter: I don't like functions with more that 3 args, so I agree, it will better.

layershifter:
I'll rebase it, complete TODOs and update code with your suggestion tomorrow

src/lib/factories.js
levithomason: This would also apply to the returned `create` factory.  Instead of:

```jsx
Menu.create(menu, null /* defaultProps */, () => /* override props */)
```

We'd have something like:

```jsx
Menu.create(menu, { overrideProps: () => { ... } })
```

levithomason:
I'd like to try and simplify the factories as we also add more features.  There is only one component that currently requires usages of `defaultProps` as a function (others use a function but do not need to):

**Input.js**

```jsx
    const actionElement = Button.create(action, elProps => ({
      className: cx(
        // all action components should have the button className
        !_.includes(elProps.className, 'button') && 'button',
      ),
    }))
```

This is ensuring the `button` className is added, but not duplicated.  Currently, the factory merges className but does not check for duplicates, it could check first.  Then, we can reduce the `defaultProps` back to a simple object.  If we need a function, there is always `overrideProps` for all other use cases.

I think we should then also remove the option for `defaultProps` to be a function and instead use a simple object.  Then, we update the current merge className logic to only merge unique classNames.

Thoughts?

levithomason:
Just ran into an issue over at https://github.com/Semantic-Org/Semantic-UI-React/pull/1508/files#r108559120.  We currently only enable support for keys if a `create` method is instantiated with `generateKey` set to `true`.

Although we know which components we need keys for, consumers can also use `create` in their own code.  We should enable key support in all factories so that any user can map over any `create` method and have support for keys.  Otherwise, some factories will be impossible to use in a map.

levithomason:
Heads up, I'm gonna push a few changes here.

levithomason:
Heads up, I've got more pushing coming.  Working on some of the tests.

layershifter:
I've pushed some updates, so only `28 tests failed` now :+1:

levithomason:
`✔ 7118 tests completed`
`✖ 6 tests failed`

Seems we may have an issue with the BreadcrumbDivider test.  The common test for a shorthand prop expects the rendered element to `contain` the shorthand element.  However, the breadcrumb divider uses it's icon props to return an icon with the `divider` class added.

levithomason:
I'll leave this PR alone for now and come back when I can.  I also think we'll need to still make `overrideProps` a function.  It would received the merged `defaultProps` and `usersProps`.  This way, we can wrap an event handler and also still call the user's original event handler.

layershifter:
`✔ 7124 tests completed` 👍 

### TODO

- [x] make `overrideProps` a function
- [x] update event handlers where we use `overrideProps`
- [x] update tests, add tests for `overrideProps`

src/collections/Breadcrumb/Breadcrumb.js
```diff
@@ -34,35 +34,17 @@ function Breadcrumb(props) {
...
-      // Divider key: content=Home|link=true|onClick=handleClick
-      let key
-      if (section.key) {
-        key = `${section.key}_divider`
-      } else {
-        key = _.map(breadcrumbSection.props, (v, k) => {
-          return `${k}=${typeof v === 'function' ? v.name || 'func' : v}`
-        }).join('|')
-      }
-      childElements.push(BreadcrumbDivider.create({ content: divider, icon, key }))
```
layershifter: Cruft.

src/addons/Confirm/Confirm.js
```diff
@@ -59,16 +59,22 @@ class Confirm extends Component {
...
+  })
 
-    if (onConfirm) onConfirm(e, this.props)
-  }
+  handleConfirmOverrides = predefinedProps => ({
+    onClick: (e, buttonProps) => {
+      _.invoke(predefinedProps, 'onClick', e, buttonProps)
+      _.invoke(this.props, 'onConfirm', e, this.props)
+    },
+  })
```
layershifter, line 77: We handle there `onClick` on `Button`s, let's allow user to handle own `onClick` 👍

src/collections/Breadcrumb/BreadcrumbDivider.js
```diff
@@ -26,13 +26,14 @@ function BreadcrumbDivider(props) {
   const rest = getUnhandledProps(BreadcrumbDivider, props)
   const ElementType = getElementType(BreadcrumbDivider, props)
 
-  const iconElement = Icon.create(icon, { ...rest, className: classes })
-  if (iconElement) return iconElement
+  if (!_.isNil(icon)) return Icon.create(icon, { defaultProps: { ...rest, className: classes } })
```
layershifter, line 29: Simplified contidition

layershifter (commented):
@levithomason I've added some tests and performed some cleanups, seems we can go after review 😄

src/collections/Breadcrumb/Breadcrumb.js
levithomason: Darn, this is technically a breaking change.  Users who are using objects to define a breadcrumb divider and also not defining a key will now need to add the key.  Otherwise, they won't all render.

levithomason:
**EDIT**
I reverted the removal of `key`s in shorthand for breadcrumb dividers.  I had misread that code the first time.

**Original**
this looks good, I'm just going to revert the removal of generated keys for BreadcrumbDividers to prevent a breaking change, see https://github.com/Semantic-Org/Semantic-UI-React/pull/1428/files#r109309892.

We've got #1533 to remove all manually generated keys.  This way we can consolidate our breaking changes for `key` into one release.

src/collections/Breadcrumb/Breadcrumb.js
levithomason: I misread this initially, we are one using the factory as an object in this case.  The key is then required by us, otherwise, we're getting this in the docs for Breadcrumb:

`Each child in an array or iterator should have a unique "key" prop. Check the top-level render call using <div>.`

levithomason:
Rebased to master, fixed breadcrumb divider keys.  Also tossed in a fix the for "show HTML" button in examples.

Will merge on pass.

PR #1435, opened 2017-03-10
feat(Transition): add component

### LINKED ISSUE #200: <Transition /> Component

layershifter:
Fixes #200.

This PR is second try for `Transition` component. It uses `react-addons-transition-group` for deal with SUI's CSS animations.


levithomason:
:heart:

layershifter:
Main problem that I have there it's missing `cancelUnmount` event. So when component is unmouting there is no way to know that we need to cancel unmount animation if component changed state and is visible again.

I need to pull `transition-group` code and review it, possible I'm missing something.

bjmiller:
Is there anything we can do to help get this completed and merged?  My team is excited to have transitions.

layershifter:
I've pulled code from [react-transition-group#24](https://github.com/reactjs/react-transition-group/pull/24) as start point. I will try to spend all free time to finish it.

### TODO
- [x] combile `Transition` and `TransitionWrapper`;
- [x] wrap all children automatically on `TransitionGroup`
- [x] remove dependency on `dom-helpers`
- [x] replace original callbacks with callbacks that more closer to SUI
- [x] make an inventory in `propTypes`
- [x] typings
- [x] docs and tests
- [x] refactor `mergeChildMappings`

layershifter (commented):
@levithomason I'm finally ready to say that we are ready to review :tada: 

To all: this component is ready for some user testing and feedback.

### How?

If you could pull this branch `feat/transition-2`, `npm install`, `npm start`, and navigate to http://localhost:8080/modules/transition, you'll see the Transition component examples. Please have a look at the code, play with the editors, and let me know what you think.

brianespinosa:
Sweet! I will get some testing in on this branch this weekend. Great work @layershifter. I was sad when the old version had to be abandoned, but I am glad we are using the non-deprecated package that React team is supporting with this one.

john-osullivan:
Just wanted to drop by and say that y'all are MVPs for going through the legwork of porting Semantic-UI.  Particularly here when it gets really messy, hooking into React's animation systems.  I've been using React for years and that stuff still mystifies me.  Thanks for working on this!  I'll try to test it if my team finishes required UI soon, and if not, I look forward to using it once it's merged.  Thank you for contributing to OSS ❤️

docs/app/Examples/modules/Transition/Types/TransitionExampleGroup.js
```diff
@@ -0,0 +1,49 @@
...
+
+  render() {
+    const { items } = this.state
+
+    return (
+      <div>
+        <Transition.Group
+          as={List}
+          duration={1000}
+          divided size='huge'
```
levithomason: Whoops, two lines: `divided size='huge'`

src/modules/Transition/Transition.d.ts
```diff
@@ -0,0 +1,84 @@
...
+export interface TransitionProps {
+  [key: string]: any;
+
+  /** Named animation event to used. Must be defined in CSS. */
+  animation?: SemanticTRANSITIONS;
+
+  /** Primary content. */
+  children?: React.ReactNode;
+
+  /** Duration of the CSS transition animation in microseconds. */
```
levithomason: `microseconds` => `milliseconds`

src/modules/Transition/Transition.js
```diff
@@ -0,0 +1,288 @@
...
+ */
+export default class Transition extends Component {
+  static propTypes = {
+    /** Named animation event to used. Must be defined in CSS. */
+    animation: PropTypes.oneOf(SUI.TRANSITIONS),
+
+    /** Primary content. */
+    children: PropTypes.node,
+
+    /** Duration of the CSS transition animation in microseconds. */
```
levithomason: `microseconds` => `milliseconds`

src/modules/Transition/TransitionGroup.d.ts
```diff
@@ -0,0 +1,25 @@
...
+  /** An element type to render as (string or function). */
+  as?: any;
+
+  /** Named animation event to used. Must be defined in CSS. */
+  animation?: SemanticTRANSITIONS;
+
+  /** Primary content. */
+  children?: React.ReactNode;
+
+  /** Duration of the CSS transition animation in microseconds. */
```
levithomason: `microseconds` => `milliseconds`

src/modules/Transition/TransitionGroup.js
```diff
@@ -0,0 +1,126 @@
...
+    /** An element type to render as (string or function). */
+    as: customPropTypes.as,
+
+    /** Named animation event to used. Must be defined in CSS. */
+    animation: PropTypes.oneOf(SUI.TRANSITIONS),
+
+    /** Primary content. */
+    children: PropTypes.node,
+
+    /** Duration of the CSS transition animation in microseconds. */
```
levithomason: `microseconds` => `milliseconds`

levithomason:
Heads up, I'm going to push some refactors and updates here for a bit.

docs/app/Examples/modules/Transition/Explorers/TransitionExampleSingleExplorer.js
```diff
@@ -0,0 +1,57 @@
...
+            label={`Duration: ${duration}ms `}
+            min={100}
+            max={2000}
+            name='duration'
+            onChange={this.handleChange}
+            step={100}
+            type='range'
+            value={duration}
+          />
+          <Form.Button content={visible ? 'Hide' : 'Show'} onClick={this.handleVisibility} />
```
levithomason: When changing state during a transition the example does not behave correctly as the transition is interrupted.  I'd like to disable the form controls while the transition is in progress.

In trying to disable/enable the Show/Hide button in the explorer examples, I now think we need to fire the callbacks on `Transition.Group` for each transition as well.  I'm not sure there is a way to capture start/complete without this.

Would you mind updating the examples to show how this is possible?

levithomason:
This is looking fantastic, thanks!  I have a couple questions in addition to what I left inline.  It would be great to get your thoughts behind these items:

## Transition vs Transition.Group

I'm a little confused on the use of `Transition` vs `Transition.Group`.  The group seems to be used on single elements when they need to mount/unmount in sync with the transition.  The single `Transition` seems to be used for static transitions.  However, in the examples when trying to replace a `Transition.Group` around a single element with a `Transition` I cannot seem to figure out how to get it to work.

Intuitively, I would expect a `Transtion` to only allow a single child and a `Transition.Group` to accept multiple children.  I'd then think a group would handle group operations such as staggering:

https://semantic-ui.com/modules/transition.html#/examples



Could you give some info on what the differences are and why two components are needed to accomplish this?

## `into`

Could you expound a bit on this prop?  I'm not entirely sure why it is needed and what the intent behind it is.

***

Again, thanks for the massive amount of work here.  We'll surely get this merged soon 🎉

docs/app/Examples/modules/Transition/Explorers/TransitionExampleSingleExplorer.js
layershifter: Hm, thanks for attracting attention there. I'm that `Transition.Group` should also have `onComplete`, `onHide`, `onShow`, `onStart` and this should be illustrated in separate example.

layershifter: However, I want to do this in sepate PR, this one is already big enough.

layershifter:
I will start from the end. I took the idea from `react-transition-group` as the right way.

## Why don't simply use `react-transition-group`

- additional dependency
- multiple `classNames` aren't supported
- when I started this PR, `v2` wasn't released

## `into`

This prop is a switch, it shows whether the component is visible or not. When the value changes, the animation is called. This is the main idea that was in the first PR, too. It is possible that there it should be renamed into something more understandable, possible `visible`?

## Transition vs Transition.Group

This is the sweetest, this is the main idea. The main problem of the first iteration of this component was that it worked incorrectly with the lifecycle of React. As you know, there is no way to stop unmounting the component in React, and therefore there is no way to reproduce the dissapering animation. Short example:

```jsx
const visible = true // => switch to `false`

<Transition into={false} /> // will work, component still mounted
{visible && <Transition />} // will not work, component was unmounted
```

Now we can move on to the borrowed idea. `Transition.Group` is needed to handle the unmounting of components. As you can see, the rendering of children in it comes from the state, and they are in it until the animation is finished. That's why we can do this and are not afraid of the fact that the component is unmounted:

```jsx
const visible = true // => switch to `false`

<Transition.Group>
  {visible && <Image />}
</Transition.Group>
```

However, we can also use `Transition` if we are sure that the component will not be unmounted. I show static animations with `Transition`, because they basically don't require mounting or unmounting the component. BTW, `Transition` accepts a single child in all cases, possible we need to make propTypes more strict?

## Summary

`Transition` handles all actions with animations while `Transition.Group` cares about children. Possible we should rename components to something more understandable.

brianespinosa:
I like `Transition.Group` as the naming for the component controlling the `Transition` children below. Seems like it is consistent with the overall project and makes more sense now that you explain what it does and it's relation to `Transition` components as children.

I agree on the suggestion to rename the `into` prop as `visible`.

Again... great work on all this @layershifter

levithomason:
### into
> It is possible that there it should be renamed into something more understandable, possible `visible`?

This makes sense to me.  I think it conveys the idea that it is not unmounted as well.

### Transition vs Transition.Group

My previous concerns can probably be solved with documentation.  An updated description for each component may help:

- A `Transition` animates a single child by toggling the the `visible` prop.
- A `Transition.Group` animates children as they mount and unmount.

It may be worth repeating these descriptions for each example.  I'll update the docs accordingly.

>Transition accepts a single child in all cases, possible we need to make propTypes more strict?

Yep, `PropTypes.element.isRequired` is for [requiring a single child:](https://facebook.github.io/react/docs/typechecking-with-proptypes.html#requiring-single-child).   Since it should not be unmounted in this case, this sounds like the correct PropType.

***

I'll update this PR with the above conversation and see if we can get it merged 😄

docs/app/Examples/modules/Transition/Explorers/TransitionExampleSingleExplorer.js
levithomason: Separate PR sounds good.

levithomason:
I've pushed some changes here, no changes to the logic, only naming and docs:

1. rename `into` -> `visible`
2. rename enter/exit/appear terminology to show/hide/mount in keeping with SUI
1. rename "entire" transitions to "directional" (vs static)
3. refactored Transition/Transition Group docs to clarify differences and usage
4. refactored Explorer docs to clarify differences and usage
1. fixed missing typings

In the next PR(s) let's address:

- [ ] Implementing Transition in our own components
- [ ] Transition.Group callbacks
- [ ] Usage examples for various props (`unmountOnHide`, etc)

bjmiller:
@levithomason: Is there an ETA for a new release that includes this feature?

levithomason:

levithomason:
I do the best I can to make a release every weekend.

rutvijprimaseller:
Will there be animation in accordions @levithomason ?

rutvijprimaseller:
Is there a way to animate Accordion or shall we expect it in later releases? @levithomason

brianespinosa:
@rutvijprimaseller Currently there is no support for animating in accordions. This `Transition` component just barely got merged into SUIR and the components in the project have not yet been updated to use it.

@layershifter you might know... are there separate tasks set aside for converting each of these yet?

layershifter:
No, there are no separate tasks. But, plans for them exist of course. I want begin from `Dimmer` because it will be used by `Modal` (#1739). I want finish some of my open PRs and then I will start on it.

brianespinosa:
@layershifter That makes my day to hear.

Maxhodges:
can I track the status of Accordion transitions somewhere? It's too fast, so I want a slightly slower animation.

subodhpareek18:
Is it possible to run static animation transitions in a loop? Like a continuous tada animation on a call icon for an incoming call notification until it's clicked on and picked up.

levithomason:
@subodhpareek18 I don't see why not

PR #1442, opened 2017-03-11
feat(Search): add handling of html input props

### LINKED ISSUE #1437: Cannot pass `name` props to Search component
Passing prop `name` to `Search` should set it as an attribute to the `<input>`, not the `<div>`.

### Code
```
<Search name="foo" />
```

### Expected Result
```
<div class="ui fluid search">
  <div class="ui icon input">
    <input name="foo" type="text" placeholder="Name" value="" class="prompt" tabindex="0" autocomplete="off">
...
```

### Actual Result
```
<div name="foo" class="ui fluid search">
  <div class="ui icon input">
    <input type="text" placeholder="Name" value="" class="prompt" tabindex="0" autocomplete="off">
...
```

### Version
0.63.5

layershifter:
Fixes #1437.


levithomason:
I think we need to implement the mechanism that the Input itself implements, so that all HTML `<input>` props are forwarded to the input.

We can pull this list of HTML input props out of `Input.js` and into `/lib` inside a method like `getHTMLInputProps(props)` so that it can be used in both components:

```jsx
import { getHTMLInputProps } from '../lib'

const htmlInputProps = getHTMLInputProps(props)
```

layershifter:
Makes sense to me. Will implement it.

src/elements/Input/Input.js
```diff
@@ -148,10 +116,10 @@ class Input extends Component {
   }
 
   handleChange = (e) => {
+    const { onChange } = this.props
     const value = _.get(e, 'target.value')
 
-    const { onChange } = this.props
-    if (onChange) onChange(e, { ...this.props, value })
```
layershifter, line 154: Cruft. `handleChange` will be called only if `onChange` defined.

src/lib/htmlInputPropsUtils.js
```diff
@@ -0,0 +1,43 @@
...
+  // selection
+  'onSelect',
+
+  // touch
+  'onTouchCancel', 'onTouchEnd', 'onTouchMove', 'onTouchStart',
+]
+
+export const htmlInputPropNames = [...htmlInputProps, ...htmlInputEvents]
+
+export const omitHTMLInputProps = (props, htmlProps = htmlInputPropNames) => _.omit(props, htmlProps)
```
levithomason: Lodash has perf issues with `_.omit` and it is also planned to be removed in v5, #864 .  How about we move these two pick/omit methods into a single reduce that returns a pair of objects?  Something like:

```jsx
const [htmlInputProps, rest]  = partitionHTMLInputProps(props)
```
Given:
```jsx
const htmlInputProps = ['selected', ...]

const getHTMLInputProps = props => htmlInputProps.reduce((acc, next) => {
  acc[+!(next in props)] = props[next]
  return acc
}, [])
```

layershifter: Good suggestion, will make it :+1:

levithomason:

PR #1446, opened 2017-03-12
docs(callbacks): updates in typings

layershifter:
Typings update for #1445.


src/modules/Accordion/Accordion.js
```diff
@@ -52,7 +52,7 @@ export default class Accordion extends Component {
      * Called when a panel title is clicked.
      *
      * @param {SyntheticEvent} event - React's original SyntheticEvent.
-     * @param {index} - The index of the clicked panel.
+     * @param {number|array} index The index of the clicked panel.
```
levithomason: I believe this should be number only, correct?

layershifter: Updated 👍

PR #1450, opened 2017-03-13
docs(Progress): fix incorrect example names

yangshun:
This page is currently broken due to incorrectly named modules: http://react.semantic-ui.com/modules/progress


levithomason:
Thanks!

PR #1465, opened 2017-03-15
fix(Dropdown):make active dropdown item scroll into view on open

### LINKED ISSUE #1388: Active Dropdown item should be scrolled into view on open
### Steps

1. Set `selectOnBlur={false}`
2. Make an item active.
3. Scroll the dropdown so it is out of view.
4. Close the dropdown.
5. Open the Dropdown.

#### Actual

The Dropdown does not show the active item because it persists its scroll position, so the active item is still out of view.

#### Expected

Opening the dropdown should always show the active item.  Otherwise, users are confused by what they see.

#### Fix
There is already a method to scroll the active item into view.  We simply need to call it on open.

#### Version
Version 0.66.0

mchirkin:
Fix #1388

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -940,8 +940,11 @@ export default class Dropdown extends Component {
 
   scrollSelectedItemIntoView = () => {
     debug('scrollSelectedItemIntoView()')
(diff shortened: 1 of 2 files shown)


levithomason:

Issue #1458, opened 2017-03-14
Modal(Portal) not compatible with React Fiber (react@16)


cpunion:
### Version
semantic-ui-react@0.67.1
semantic-ui-css@2.2.9
react@16.0.0-alpha3
react-dom@16.0.0-alpha3

### Testcase
https://github.com/cpunion/TestSemanticUIWithFiber

See:
https://github.com/cpunion/TestSemanticUIWithFiber/blob/master/src/index.js

Errors:
```
Uncaught TypeError: Cannot read property 'addEventListener' of null
    at Portal.renderPortal (Portal.js:348)
    at Portal.componentDidUpdate (Portal.js:291)
    at commitLifeCycles (ReactFiberCommitWork.js:431)
    at commitAllLifeCycles (ReactFiberScheduler.js:349)
    at HTMLUnknownElement.boundFunc (ReactErrorUtils.js:81)
    at ReactErrorUtils.invokeGuardedCallback (ReactErrorUtils.js:95)
    at commitAllWork (ReactFiberScheduler.js:468)
    at completeUnitOfWork (ReactFiberScheduler.js:610)
    at performUnitOfWork (ReactFiberScheduler.js:637)
    at workLoop (ReactFiberScheduler.js:760)
renderPortal @ Portal.js:348
componentDidUpdate @ Portal.js:291
commitLifeCycles @ ReactFiberCommitWork.js:431
commitAllLifeCycles @ ReactFiberScheduler.js:349
boundFunc @ ReactErrorUtils.js:81
ReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:95
commitAllWork @ ReactFiberScheduler.js:468
completeUnitOfWork @ ReactFiberScheduler.js:610
performUnitOfWork @ ReactFiberScheduler.js:637
workLoop @ ReactFiberScheduler.js:760
boundFunc @ ReactErrorUtils.js:81
ReactErrorUtils.invokeGuardedCallback @ ReactErrorUtils.js:95
performWork @ ReactFiberScheduler.js:802
batchedUpdates @ ReactFiberScheduler.js:1245
performFiberBatchedUpdates @ ReactGenericBatching.js:32
stackBatchedUpdates @ ReactGenericBatching.js:23
batchedUpdates @ ReactGenericBatching.js:37
batchedUpdatesWithControlledComponents @ ReactGenericBatching.js:50
dispatchEvent @ ReactEventListener.js:174
[...]

PR #1470, opened 2017-03-16
fix(Portal): make compatible with React Fiber

cpunion:
Tested with react@15.4.2 and react@16.0.0-alpha4

Fix https://github.com/Semantic-Org/Semantic-UI-React/issues/1458

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -356,13 +356,14 @@ class Portal extends Component {
     ReactDOM.unstable_renderSubtreeIntoContainer(
       this,
       Children.only(children),
-      this.rootNode
+      this.rootNode,
+      () => {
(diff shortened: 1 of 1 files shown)


levithomason:
Excellent work!  I've checked the source and it appears the callback is indeed called on completion.

levithomason:

cpunion:
Thanks!

jpollard-cs:
you could have used v16's [unstable_createPortal](https://github.com/facebook/react/blob/7d2be2c9a7b4a976e20d2d7c312b43b1bee01a07/src/renderers/dom/fiber/ReactDOMFiber.js#L212) 😄

levithomason:
Guessing this would abandon React v15 users?

jpollard-cs:
move fast and break things! 

wasn't entirely a serious comment

Issue #1158, opened 2017-01-14
Closing sidebar by clicking on the page


andrewleith:
Is there a way to close a sidebar by clicking on the page? The official semantic-ui supports this out of the box and is configurable via a `closable` setting.

Is there something I can use in this library to do it?

andrewleith:
I modified your sidebar example to do the following and it seems to meet my needs.  For the 'ESC' keypress I am adding a keypress listener at the document level when the sidebar opens, and then removing it when it closes. For the click, I specifically target the pusher element.  Any clicks on it will close the sidebar.

Targeting the DOM like this doesn't feel particularly in line with react, so there probably is a cleaner way to do this.  In the meantime, this may help someone.

```js
class SidebarLeftOverlay2 extends Component {
  state = { visible: false }

  toggleVisibility = () => this.setState({ visible: !this.state.visible })

  componentWillUpdate(nextProps, nextState) {
    // when the menu becomes visible, setup some handlers so we can close the menu easily
    if (nextState.visible == true) {
      document.addEventListener('keydown', this.handleKeyPress);
      document.querySelector('.pusher').addEventListener('click', this.handleClick);
    }
    else {
      document.removeEventListener('keydown', this.handleKeyPress);
      document.querySelector('.pusher').removeEventListener('click', this.handleClick);
    }
    
  }

  handleClick = () => {
    if (this.state.visible) {
      this.setState({ visible: false });
    }
  }
  handleKeyPress = (e) => {
    if(e.keyCode === 27 && this.state.visible) {
      this.setState({ visible: false })
    }
  }

  render() {
    const { visible } = this.state
    return (
      <div>
        <Button onClick={this.toggleVisibility}>Toggle Visibility</Button>
        <Sidebar.Pushable as={Segment}>
          <Sidebar as={Menu} animation='overlay' width='thin' visible={visible} icon='labeled' vertical inverted>
            <Menu.Item name='home'>
              <Icon name='home' />
              Home
            </Menu.Item>
            <Menu.Item name='gamepad'>
              <Icon name='gamepad' />
              Games
            </Menu.Item>
            <Menu.Item name='camera'>
              <Icon name='camera' />
              Channels
            </Menu.Item>
          </Sidebar>
          <Sidebar.Pusher>
            <Segment basic>
              <Header as='h3'>Application Content</Header>
              <Image src='http://semantic-ui.com/images/wireframe/paragraph.png' />
            </Segment>
[...]

PR #1473, opened 2017-03-17
feat(Sidebar): closing sidebar by clicking outside

mchirkin:
Attempt to #1158


aikonplus:
this is not merged yet?

RhanCandia:
Hi! Any news on this?

levithomason:
I've merged the latest master here.  We'll work to get this reviewed and merged.

littlehome-eugene:
waiting for this as well.

test/specs/modules/Sidebar/Sidebar-test.js
```diff
@@ -20,4 +53,38 @@ describe('Sidebar', () => {
...
       .should.have.tagName('div')
   })
+
+  it('close Sidebar when clicking outside', () => {
+    const wrapper = mount(<TestComponent visible />)
+    const SidebarComponent = wrapper.find('Sidebar')
+    SidebarComponent.should.have.prop('visible', true)
+
+    const evt = document.createEvent('HTMLEvents')
+    evt.initEvent('click', false, true)
```
levithomason: `evt` is created but never used here.

src/modules/Sidebar/Sidebar.js
```diff
@@ -70,10 +77,40 @@ class Sidebar extends Component {
     this.stopAnimatingTimer = setTimeout(() => this.setState({ animating: false }), duration)
   }
 
+  componentDidMount() {
+    if (this.props.visible && this.props.closable) {
```
levithomason: Code style: please destructure props up top:

```jsx
componentDidMount() {
  const { closable, visible } = this.props
  
  if (visible && closable) { ... }
}
```

src/modules/Sidebar/Sidebar.js
```diff
@@ -70,10 +77,40 @@ class Sidebar extends Component {
...
+
+  handleDocumentClick = (e) => {
+    const sidebarDOMNode = ReactDOM.findDOMNode(this.ref)
+    // close Sidebar if e.target is outside of Sidebar
+    if (!this.checkIfSidebarContainsClickTarget(e.target, sidebarDOMNode)) {
+      this.props.toggleVisibility()
+    }
+  }
+
+  checkIfSidebarContainsClickTarget = (target, node) => {
```
levithomason: I believe method can be replaced with [`node.contains(otherNode)`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains).  It has been around for a while and has good browser support.

src/modules/Sidebar/Sidebar.js
```diff
@@ -101,7 +138,7 @@ class Sidebar extends Component {
     const rest = getUnhandledProps(Sidebar, this.props)
     const ElementType = getElementType(Sidebar, this.props)
 
-    return <ElementType {...rest} className={classes}>{children}</ElementType>
+    return <ElementType {...rest} className={classes} ref={(c) => { this.ref = c }}>{children}</ElementType>
```
levithomason: The ref handler needs to be hoisted here.  Otherwise, this creates a new function on every render.
```jsx
handleRef = c => { this.ref = c }

<ElementType ref={this.handleRef} />
```

test/specs/modules/Sidebar/Sidebar-test.js
```diff
@@ -20,4 +53,38 @@ describe('Sidebar', () => {
...
+    document.body.click()
+
+    SidebarComponent.should.have.prop('visible', false)
+  })
+
+  it('doesn\'t close when clicking on element inside Sidebar', () => {
```
levithomason: One more test would be great:

```jsx
it('does not close when clicking on the Sidebar', () => { ... })
```

test/specs/modules/Sidebar/Sidebar-test.js
```diff
@@ -20,4 +53,38 @@ describe('Sidebar', () => {
...
+
+  it('close Sidebar after opening and then clicking outside', () => {
+    const wrapper = mount(<TestComponent />)
+    const SidebarComponent = wrapper.find('Sidebar')
+    SidebarComponent.should.have.prop('visible', false)
+    wrapper.setState({ visible: true })
+    SidebarComponent.should.have.prop('visible', true)
+
+    const evt = document.createEvent('HTMLEvents')
+    evt.initEvent('click', false, true)
```
levithomason: It looks like `evt` is never used here either.

src/modules/Sidebar/Sidebar.js
```diff
@@ -70,10 +77,40 @@ class Sidebar extends Component {
...
+        document.removeEventListener('click', this.handleDocumentClick)
+      }
+    }
+  }
+
+  handleDocumentClick = (e) => {
+    const sidebarDOMNode = ReactDOM.findDOMNode(this.ref)
+    // close Sidebar if e.target is outside of Sidebar
+    if (!this.checkIfSidebarContainsClickTarget(e.target, sidebarDOMNode)) {
+      this.props.toggleVisibility()
```
levithomason: It looks like the Sidebar is already extending AutoControlledComponent, meaning it can `this.trySetState()` in an attempt to change it's state.  The state change will only be applied if there is no prop controlling the state keys.  Otherwise, it will delegate to the prop value.

In this case, `this.trySetState({ visible: false })` would attempt change the internal state, but the user's prop (if present) would win.  This allows us to rely only on `this.state.visible` internally and use `this.trySetState()` when we want to attempt to change the state.

The `toggleVisibility` prop should then be removed.

levithomason (changes requested):
I've left some comments for addressing, apologies it took so long on this one.

levithomason:
I think we just need a usage doc for this now.  I'll see about adding one.

src/modules/Sidebar/Sidebar.js
```diff
@@ -41,6 +42,12 @@ class Sidebar extends Component {
 
     /** Sidebar width. */
     width: PropTypes.oneOf(['very thin', 'thin', 'wide', 'very wide']),
+
+    /** Controls whether or not to close sidebar when click outside */
+    closable: PropTypes.bool,
+
+    /** Function to change "visible" prop */
+    toggleVisibility: PropTypes.func,
```
levithomason: Looking closer at the core API docs for Sidebar callbacks, I think we need to go with these:

image

https://semantic-ui.com/modules/sidebar#settings

src/modules/Sidebar/Sidebar.js
```diff
@@ -41,6 +42,12 @@ class Sidebar extends Component {
 
     /** Sidebar width. */
     width: PropTypes.oneOf(['very thin', 'thin', 'wide', 'very wide']),
+
+    /** Controls whether or not to close sidebar when click outside */
+    closable: PropTypes.bool,
```
levithomason: Again considering the core behavior, we should actually default this to `true`.  Since it is auto controlled state, it needs to be set in the constructor.

src/modules/Sidebar/Sidebar.js
```diff
@@ -67,9 +74,33 @@ class Sidebar extends Component {
...
+    }
+  }
+
+  handleRef = c => { this.ref = c }
+
+  handleDocumentClick = (e) => {
+    const sidebarDOMNode = ReactDOM.findDOMNode(this.ref)
+    // close Sidebar if e.target is outside of Sidebar
+    if (!sidebarDOMNode.contains(e.target)) {
+      this.trySetState({ visible: false })
```
levithomason: `visible` is destructured from `props` in the `render()` method.  It should, in fact, be from `state` since it is auto controlled.

jjjkkklll:
any news on this?

levithomason:
This branch needs to be updated to the latest master so CI can pass.

layershifter:
Closing in favor of #2845.

PR #1477, opened 2017-03-18
feat(Input): add focus() method

### LINKED ISSUE #1475: Add focus method to Input component
When Input component is used with [DatePicker](https://hacker0x01.github.io/react-datepicker/) it is expected to return a DOM node which has a focus method. @levithomason  had mention The ref is a reference to the Input class instance and we need to add a focus method to the Input component. This is also considered best practice.

```
Uncaught TypeError: this.refs.input.focus is not a function
    at Object.focus (bundle.js:86206)
    at Object.setFocus (bundle.js:86206)
    at bundle.js:86206
    at bundle.js:86212
```

levithomason:
Fixes #1475 

A `ref` on a composite component (like the Input) will return a reference to the class instance.  This PR adds a focus method to the Input that focuses the HTML input.  This allows users to call `focus()` on the Input ref and it will focus the internal HTML input element.

An example and test was also added.

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentExample.js
+++ b/docs/app/Components/ComponentDoc/ComponentExample.js
@@ -129,7 +129,7 @@ class ComponentExample extends Component {
   }
 
   copyJSX = () => {
-    copyToClipboard(this.getOriginalSourceCode())
+    copyToClipboard(this.state.sourceCode)
     this.setState({ copiedCode: true })
     setTimeout(() => this.setState({ copiedCode: false }), 1000)
   }
--- a/src/elements/Button/ButtonOr.js
+++ b/src/elements/Button/ButtonOr.js
@@ -34,10 +34,10 @@ ButtonOr.propTypes = {
   className: PropTypes.string,
 
   /** Or buttons can have their text localized, or adjusted by using the text prop. */
-  text: PropTypes.oneOfType(
+  text: PropTypes.oneOfType([
     PropTypes.number,
     PropTypes.string,
-  ),
+  ]),
 }
 
 export default ButtonOr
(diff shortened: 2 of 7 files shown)


docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -129,7 +129,7 @@ class ComponentExample extends Component {
   }
 
   copyJSX = () => {
-    copyToClipboard(this.getOriginalSourceCode())
+    copyToClipboard(this.state.sourceCode)
```
levithomason, line 132: Bug fix, an example's copy button was copying the original source code.  This update makes it copy the current state of the editor.

src/elements/Button/ButtonOr.js
```diff
@@ -34,10 +34,10 @@ ButtonOr.propTypes = {
   className: PropTypes.string,
 
   /** Or buttons can have their text localized, or adjusted by using the text prop. */
-  text: PropTypes.oneOfType(
+  text: PropTypes.oneOfType([
```
levithomason, line 37: Fixed a bad propType argument.

levithomason:

phhoef:
I have a smiliar issue with the Form.Input. I am building a small electron app.
I am quite new to the Electron/React stuff, so please be patient :-)

```
<Form>
                <Form.Input ref={i => this.searchTextField = i} label='Search' value={this.state.searchFor} onChange={this.handleSearchForChange} />
</Form>
```
I am setting the `ref` as shown above. I created a method, that is bound to a shortcut, that is calling the focus method of the input. The code below is executed in the constructor of the component.
```
ipcRenderer.on('focus-search', function () {
            this.searchTextField.focus()
        }.bind(this))
```

Unfortunately, the `this.searchTextField` is undefined.
When just using a `Input` (and not `Form.Input`) the focus is working, but my UI is destroyed.

Can you please explain what I am doing wrong?

PR #1481, opened 2017-03-20
feat(Dropdown): shorthand for search input

nhducit:
This is very helpful when developers want to add/override default props of search input field.
example: 
I want to change `type="text"` of search input to `type="number"`


levithomason:
Thanks for the PR, our pattern here would be to use our `createHTMLInput` lib factory to create the `input`.  We'd probably want to accept a `searchInput` for this case.  This would allow you to create the search input with type number like so:

```jsx
<Dropdown searchInput='number' />
```
>The value would default to the `type` of the input, per the factory.

Or, a more complicated input with a props object:
```jsx
<Dropdown searchInput={{ type: 'number', min: 1, max: 10 }} />
```

You can see this pattern in `Input.js` where we accept an `input` prop to configure the html input element.

layershifter (commented):
@levithomason I've added there initial functionality for `searchInput` shorthand.

However, I have an idea to extract `searchInput` to the `Dropdown.SearchInput` component. What you think about this?

levithomason:
Yes, please!

P.S. - Some thoughts on the bigger picture here

Currently, the Dropdown itself owns too much.  Much of its logic should be delegated to subcomponents.  Example, the Dropdown is currently responsible for creating all the `Dropdown.Items`, however, this should be the responsibility of the `Dropdown.Menu`.  The `active` className of the Menu is also controlled by the Dropdown, rather than by an `active` prop on the Menu itself.

As we break down the Dropdown and delegate its subcomponent logic, we'll find that we can more easily support the coming updates for more complicated menu markup and nested menus.

PR #1483, opened 2017-03-20
fix(Popup): use original position if none fit in viewport

### LINKED ISSUE #1482: Popup position is not respected when near corner of window
### Steps
1. Create a popup with a trigger at the top right corner of the window.
2. Make it a large popup so it can't fit on any side of the trigger without going out of the window

### Expected Result
It should be positioned according to the position prop.

### Actual Result
It always defaults to the last value in the `POSITIONS` enum excluding the value from the position prop (mostly "bottom center"). And it never uses the user specified value from the prop.

### Version
0.67.0

### Testcase
Just look at https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Popup/Popup.js#L234

MindFreeze:
Should fix #1482

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -228,7 +228,7 @@ export default class Popup extends Component {
 
     // Lets detect if the popup is out of the viewport and adjust
     // the position accordingly
-    const positions = _.without(POSITIONS, position).push(position)
+    const positions = _.without(POSITIONS, position).concat([position])
     for (let i = 0; !this.isStyleInViewport(style) && i < positions.length; i++) {
       style = this.computePopupStyle(positions[i])
       position = positions[i]
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -228,7 +228,7 @@ export default class Popup extends Component {
 
     // Lets detect if the popup is out of the viewport and adjust
     // the position accordingly
-    const positions = _.without(POSITIONS, position)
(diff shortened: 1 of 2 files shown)


MindFreeze:
I have no idea why the `is always within the viewport` test fails with this change. Some help with this would be appreciated.

src/modules/Popup/Popup.js
```diff
@@ -228,7 +228,7 @@ export default class Popup extends Component {
 
     // Lets detect if the popup is out of the viewport and adjust
     // the position accordingly
-    const positions = _.without(POSITIONS, position)
+    const positions = _.without(POSITIONS, position).push(position)
```
layershifter: This line doesn't make sense, you remove `position` from array and then push it again :confused:

MindFreeze: I add it to the end of the array, so that if no suitable position can be found, the selected (last) one is used. Otherwise it would just use the last one in POSITIONS, which is `bottom center`.

levithomason:
@MindFreeze any interest in pushing this one over the finish line?  I'd be OK merging it if tests were fixed and a new test added that asserts the new behavior.  Namely, we'd need a test that says if no other position fits in the viewport then the original position is used.

Since this is a very rare edge case, it probably will not get attention from the core team.

MindFreeze:
I'll give it a go

MindFreeze:
@levithomason All done.

levithomason:
You're awesome, thanks much!

levithomason:

PR #1489, opened 2017-03-21
fix(Button): update handling of floated prop

### LINKED ISSUE #1488: Button: floated doesn't work for Labeled Button
**Steps to Reproduce**
1. Add property "floated" with value "right" to a labeled button.

```
    <Button
      icon='fork'
      label={{ as: 'a', basic: true, content: '2,048' }}
      labelPosition='left'
      floated='right'
    />
```

**Expected**
The Button should be at the right side of the container.

**Result**
The Button doesn't move to the right side.

layershifter:
Fixes #1488.

`floated` class must be on wrapper if we have `Label` element, this PR update handling of this.

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -160,6 +160,20 @@ class Button extends Component {
   static Group = ButtonGroup
   static Or = ButtonOr
 
+  computeElementType = () => {
+    const { attached, label } = this.props
+
+    if (!_.isNil(attached) || !_.isNil(label)) return 'div'
+  }
+
+  computeTabIndex = ElementType => {
+    const { disabled, tabIndex } = this.props
+
+    if (!_.isNil(tabIndex)) return tabIndex
+    if (disabled) return -1
+    if (ElementType === 'div') return 0
+  }
+
   handleClick = (e) => {
     const { disabled, onClick } = this.props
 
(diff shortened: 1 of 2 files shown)


src/elements/Button/Button.js
```diff
@@ -229,18 +231,23 @@ class Button extends Component {
...
+    const wrapperClasses = cx(
+      useKeyOnly(disabled, 'disabled'),
       useValueAndKey(floated, 'floated'),
     )
-    const wrapperClasses = cx(useKeyOnly(disabled, 'disabled'))
+
     const rest = getUnhandledProps(Button, this.props)
-    const ElementType = getElementType(Button, this.props, () => {
-      if (!_.isNil(label) || !_.isNil(attached)) return 'div'
-    })
```
layershifter, line 238: I've moved this inline function to method, so it will not be created on each render.

test/specs/elements/Button/Button-test.js
```diff
@@ -109,6 +109,11 @@ describe('Button', () => {
       wrapper.find('Label').should.not.have.className('disabled')
       wrapper.find('button').should.not.have.className('disabled')
     })
+    it('contains children without floated class when floated attribute is set', () => {
```
levithomason, line 114: Would prefer we assert the expected location of the floated class in this test as well. It currently asserts where the className should _not_ be, but doesn't assert where it _should_ be.

levithomason (changes requested):

test/specs/elements/Button/Button-test.js
layershifter: I've updated tests as you requested :+1:

levithomason (approved):

levithomason:

PR #1490, opened 2017-03-21
fix(Dropdown): do not close on click when search is empty

### LINKED ISSUE #1451: Dropdown: Clicking on "No results found." leads to hard to remove text from input
### Steps
1. Type enough characters in a "fluid multiple selection search" dropdown until "No results found." is displayed.
2. Click "No results found."
3. Double click the empty space right of the typed text to select it. Do not double click the text itself.
4. Press backspace or delete

### Expected Result
The text is removed

### Actual Result
Text stays selected

### Notes

- Semantic-UI doesn't allow clicking on "No results found."
- Only with **fluid**, text selection works fine without fluid.
- Maybe click on "No results found." should just remove the text just as if you click outside the dropdown to cancel it.

### Version
Semantic-UI 2.2.9
React Semantic-UI 0.67.0

### Testcase
http://codepen.io/nreijmersdal/pen/JWJbbg

rajchourasia:
Fixes #1451 . Mimics the behaviour of SUI core in similar scenario.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1008,8 +1008,8 @@ export default class Dropdown extends Component {
       e.preventDefault()
       return
     }
+
     this.close(e)
-    return
   }
 
   // ----------------------------------------
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -995,7 +995,18 @@ export default class Dropdown extends Component {
     this.setState({ focus: hasFocus })
   }
 
-  toggle = (e) => this.state.open ? this.close(e) : this.open(e)
+  toggle = (e) => {
+    if (this.state.open) {
+      const { search } = this.props
+      const options = this.getMenuOptions()
+      if (search && _.isEmpty(options)) {
+        e.preventDefault()
+      } else {
+        return this.close(e)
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -995,7 +995,18 @@ export default class Dropdown extends Component {
...
+    if (this.state.open) {
+      const { search } = this.props
+      const options = this.getMenuOptions()
+      if (search && _.isEmpty(options)) {
+        e.preventDefault()
+      } else {
+        return this.close(e)
+      }
+    }
+    return this.open(e)
```
layershifter: Can we simplify logic there?

```js
toggle = (e) => {
  if (this.state.open) {
    this.open(e)
    return
  }

  const { search } = this.props

  if (search && _.isEmpty(options)) {
    e.preventDefault()
    return
  }

  this.close(e)
}
```

It looks more readable to me.

layershifter (changes requested):

layershifter (approved):

levithomason:
Thanks for writing a test for this as well.  This is a good stop-gap for now, however, the SUI core dropdown still allows toggling an empty search dropdown when the icon is clicked:



We should address that at some point.

levithomason:

PR #1492, opened 2017-03-21
fix(Modal): spread the user's style prop on the Modal

### LINKED ISSUE #1308: Modal window fixed height
Hello

Is there any simple way to open modal at fixed screen height (minus margins) ?
By default a modal window tries to center itself vertically which causes problematic UX when modal content changes

rajchourasia:
Fixes #1308


layershifter:
@rajchourasia It will be awesome if we add test there. And we also need to add prop to typings.

layershifter (approved):
@rajchourasia Thanks for PR!

levithomason:
Superb, thanks much!

levithomason:

PR #1497, opened 2017-03-22
docs(bundlers): add webpack 2 example

layershifter:
Ref: #524

-----

@levithomason I said earlier that tree-shaking doesn't work in Webpack 2 and you [asked](https://github.com/Semantic-Org/Semantic-UI-React/issues/524#issuecomment-281799647) for example, here it is. I'm think that I'm missing something, so it's will be awesome to get your feeback there.


examples/webpack2/src/main.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { render } from 'react-dom'
+import { Button, Container, Header } from 'semantic-ui-react'
```
levithomason, line 3: This cannot use tree shaking since it is importing commonjs modules. Use the ES build. 

https://github.com/Semantic-Org/Semantic-UI-React/issues/524#issuecomment-289218645

layershifter: This example contains bundle analyzer, so you can see same picture as [there](https://github.com/Semantic-Org/Semantic-UI-React/issues/524#issuecomment-289219955). It loads build from `es` dir.

levithomason: Hm, seems I'll have to put in more due diligence to investigate here.  I will do this as soon as I can.

pruhstal: @levithomason any update here?

layershifter:
Seems, that it's not our problem at all, [here](https://github.com/FormidableLabs/victory/issues/549) guys have same issues.

Also for tracking:
[webpack#2867](https://github.com/webpack/webpack/issues/2867)
[webpack#2899](https://github.com/webpack/webpack/issues/2899)
[webpack#3092](https://github.com/webpack/webpack/issues/3092)

I think, that we need update this PR and add usage of `babel-plugin-lodash` at now :confused:

levithomason:
Hm, I think in this case we show the proper config and include a note about the Webpack bug.  We can also include a note to refer to the Webpack 1 example as a workaround while the bug is being fixed.

Thoughts?

layershifter:
@levithomason Yes, I will update this PR soon

examples/webpack2/src/main.js
levithomason: @pruhstal you may have already noticed, but check the latest comments on the PR.  Appears to be a webpack bug.

layershifter (commented):
I've added a message with notice and lodash plugin to config.

levithomason:
Going to merge this for now and we can iterate on it.

levithomason:

PR #1502, opened 2017-03-24
chore(package): update to webpack 2

layershifter:
This PR updates webpack to 2 version.


config.js
```diff
@@ -18,12 +19,10 @@ let config = {
...
   dir_src: 'src',
   dir_dist: 'dist',
-  dir_docs_root: 'docs',
-  dir_docs_src: 'docs/app',
+  dir_dll: 'dll',
   dir_docs_dist: 'docs/build',
+  dir_docs_src: 'docs/app',
   dir_umd_dist: 'dist/umd',
-  dir_server: 'server',
-  dir_test: 'test',
```
layershifter, line 26: Cruft.

config.js
```diff
@@ -35,10 +34,10 @@ const paths = {
   base,
   src: base.bind(null, config.dir_src),
   dist: base.bind(null, config.dir_dist),
-  test: base.bind(null, config.dir_test),
```
layershifter, line 38: Cruft.

config.js
```diff
@@ -53,30 +52,27 @@ config = Object.assign({}, config, {
   // ----------------------------------
   // Compiler Configuration
   // ----------------------------------
-  compiler_devtool: __DEV__ && 'cheap-source-map'
-  || __TEST__ && 'cheap-source-map'
```
layershifter, line 57: [karma-webpack](https://github.com/webpack-contrib/karma-webpack) use `inline-source-map`

docs/app/Components/ComponentDoc/ComponentExample.js
```diff
@@ -146,7 +146,7 @@ class ComponentExample extends Component {
   getOriginalSourceCode = () => {
     const { examplePath } = this.props
 
-    if (!this.sourceCode) this.sourceCode = require(`!raw!docs/app/Examples/${examplePath}`)
```
layershifter, line 149: Update loader's usage

karma.conf.js
```diff
@@ -1,78 +1,79 @@
...
-        ui: 'bdd',
-      },
+      './test/tests.bundle.js': ['webpack', 'sourcemap'],
     },
+    singleRun: !argv.watch,
     webpack: {
       devtool: config.compiler_devtool,
-      module: Object.assign({}, webpackConfig.module, {
-        loaders: [
+      entry: './test/tests.bundle.js',
```
layershifter: [karma-webpack](karma-webpack) tells we don't need `entry`, however webpack didn't work without it

package.json
```diff
@@ -108,7 +108,8 @@
     "karma-mocha-reporter": "^2.2.1",
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-phantomjs-shim": "^1.4.0",
-    "karma-webpack-with-fast-source-maps": "^1.9.2",
```
layershifter: I've replaced it with `karma-webpack` and `karma-sourcemap-loader` because it looks quite outdated.

package.json
```diff
@@ -135,9 +136,9 @@
     "tslint-config-typings": "^0.3.1",
     "typescript": "^2.1.5",
     "watch": "^1.0.1",
-    "webpack": "^1.12.14",
-    "webpack-dev-middleware": "^1.5.1",
-    "webpack-hot-middleware": "^2.15.0",
+    "webpack": "^2.3.1",
+    "webpack-dev-middleware": "^1.10.1",
+    "webpack-hot-middleware": "^2.17.1",
```
layershifter: Updated versions

test/specs/lib/isBrowser-test.js
```diff
@@ -7,12 +7,12 @@ describe('isBrowser', () => {
   })
 
   it('should return false when there is no document', () => {
-    require('imports?document=>undefined!src/lib/isBrowser').default.should.be.false()
-    require('imports?document=>null!src/lib/isBrowser').default.should.be.false()
+    require('imports-loader?document=>undefined!src/lib/isBrowser').default.should.be.false()
+    require('imports-loader?document=>null!src/lib/isBrowser').default.should.be.false()
```
layershifter, line 11: Update loader's usage

webpack.config.js
```diff
@@ -72,34 +74,35 @@ webpackConfig.plugins = [
...
 if (__DEV__) {
   webpackConfig.plugins.push(
     new webpack.HotModuleReplacementPlugin(),
-    new webpack.NoErrorsPlugin()
+    new webpack.NoEmitOnErrorsPlugin()
   )
 } else if (!__TEST__) {
   webpackConfig.plugins.push(
-    new webpack.optimize.OccurrenceOrderPlugin(),
-    new webpack.optimize.DedupePlugin(),
```
layershifter: Cruft at now

config.js
levithomason: We use `karma-webpack-with-fast-source-maps` which allows for [faster file sourcemaps](https://github.com/aaronjensen/karma-webpack#file-faster).  Its main trick is that it only re-runs tests that have changed since the last run.  With vanilla `karma-webpack`, if you change one test suite, all 7k+ tests will re-run.

I'm open to changes here but I'd like to address the issue of too many tests taking too long to run while we're at it, or stick with `karma-webpack-with-fast-source-maps`.  Another option I've been considering is using something like the `dot` reporter to lighten up and speed up test output.

config.js
```diff
@@ -53,30 +52,27 @@ config = Object.assign({}, config, {
...
-  || __STAGING__ && 'source-map',
+  compiler_devtool: __DEV__ && 'cheap-source-map' || __TEST__ && 'inline-source-map' || __STAGING__ && 'source-map',
+  compiler_globals: {
+    process: {
+      env: {
+        NODE_ENV: JSON.stringify(env),
+      },
+    },
+    __DEV__,
+    __DEBUG__: !!argv.debug,
```
levithomason: `__DEBUG__` is also cruft, note, this build system was copied from one of our production apps and not thoroughly updated, so it has some extra junk that we don't need.

karma.conf.js
```diff
@@ -1,78 +1,79 @@
...
-      let haveSeenStack = false
-      return msg
-        .split('\n')
-        .reduce((list, line) => {
-          // filter out node_modules
-          if (/~/.test(line)) return list
+const formatError = (msg) => {
+  let haveSeenStack = false
+  return msg
+    .split('\n')
```
levithomason: This formatter is also borked and I'd like to clean it up with this PR.  It used to receive the full stack and filter out blackboxed scripts, however, some time ago it began receiving each line of the stack trace instead.  This is why the assertions have had so much extra whitespace in them.

I'll push a fix for this.

levithomason: Turns out Karma needs a fix first, https://github.com/karma-runner/karma/pull/2627

karma.conf.js
```diff
@@ -1,78 +1,79 @@
...
-      module: Object.assign({}, webpackConfig.module, {
-        loaders: [
+      entry: './test/tests.bundle.js',
+      module: {
+        rules: [
+          ...webpackConfig.module.rules,
           {
             test: /sinon\.js$/,
-            loader: 'imports?define=>false,require=>false',
+            loader: 'imports-loader?define=>false,require=>false',
```
levithomason: This is no longer required after the sinon 2 update.  We just need to rebase this branch.

package.json
levithomason: However, karma-webpack is very slow :(  I'll come back to this when I have some more time.

webpack.config.js
```diff
@@ -157,20 +153,20 @@ webpackConfig.module.loaders = [{
 // Local modules can still be enabled (ie for offline development)
 // in TEST we need local modules because karma uses a different index.html (no CDNs)
 if (__TEST__ || argv.localModules) {
```
levithomason: This is more complication that I've never made use of in 2 years (maybe once?).  Unless you use it, I say we drop it.  It allows for running the project without an internet connection, provided you've already installed node_modules.  You'll still get all the network errors due to CDNs failing to load, but, the local node_modules versions will be used instead.

However, this one feature alone requires us to have the style-loader, css-loader, sass-loader, file-loader, and yargs.  All these can go away if we drop this feature.

levithomason:
`2.3.2` is out, #1506

config.js
layershifter: Reverted, `karma-webpack-with-fast-source-maps` works perfectly as I see.

package.json
layershifter: Reverted

config.js
layershifter: Removed 👍

webpack.config.js
layershifter: Removed 👍

layershifter:
I've pushed a commit:
- changed `webpack` version to `2.3.2`
- `karma-webpack-with-fast-source-maps` returned 😄 
- removed logic with [Local Modules](https://github.com/Semantic-Org/Semantic-UI-React/pull/1502#discussion_r108031580)

layershifter:
@levithomason is there something that stops merging of this PR?

levithomason:
Testing this now, thanks for the ping.

webpack.config.js
```diff
@@ -125,76 +129,32 @@ webpackConfig.module.noParse = [
...
-      loaders: ['style', 'css', 'sass'],
-    }, {
-      //
-      // Files
-      //
-      test: /\.(eot|ttf|woff|woff2|svg|png)$/,
-      loader: 'file',
-    },
-  ]
-} else {
```
levithomason, line 176: We still need the else branch here as it is applicable when not in tests.  I'll review it and push an update.

levithomason: After restoring, the `docs` gulp task dropped from `1.8m` to `24s`, woop!

levithomason:
I've finished reviewing and updating this PR.  I've restored some build performance config we had previously and I also remove some deps that we don't need.  Will merge on pass.

PR #1505, opened 2017-03-24
Update getUnhandledProps.js

kulakowka:
Removed "babel-plugin-transform-react-handled-props" dependency.


levithomason:
This will fail in production as propTypes are often stripped as a package size optimization.

kulakowka:
@levithomason Thank you for your explanation.

levithomason:
No problem, lots more context here #524 and #731.

Issue #1390, opened 2017-02-26
Modal with passed trigger element, buttons can't close modal- usage question/feature request


josie11:
### Steps
Pass the Modal component a trigger element instead of using it as a controlled component.

### Expected/Desired Result
Clicking the trigger element will open the Modal, and you have the option of passing a button in the <Modal.Action /> section a prop that can trigger closing the Modal. Maybe have a <Modal.Button /> or another prop that can trigger a close?

### Actual Result
You can no longer use buttons in <Modal.Action /> or any other child elements to close the modal. You can only click outside the Modal to close it and/or use a closeIcon.

### Version
latest

### Testcase
Basic Modal example in the docs (the second example)

levithomason:
This is the intended usage.  In React, there are no implicit parent-child relationships.  So, the buttons cannot automatically close the Modal unless they are explicitly wired to do so (e.g. onClick -> set state).

The key is here:
>instead of using it as a controlled component

In order to do this, the Modal must be controlled.

In the future, we plan to support defining Modals with shorthand props.  This would allow you do accomplish what you are asking for, but, you wouldn't do it with children, you would do it with props that we provide.  This is possible because we can take the prop values and internally instantiate all the children and, while doing so, add the appropriate event listeners and state mutations.

It might look something like this (note, this is not a planned API, just an example):

```jsx
<Modal
  trigger={<Button>Basic Modal</Button>}
  basic
  size='small'
  header={{ icon: 'archive', content: 'Archive Old Messages' }}
  content='Your inbox is getting full, would you like us to enable automatic archiving of old messages?'
  actions={[
    { basic: true, color: 'red', inverted: true, icon: 'remove', content: 'No', },
    { color: 'green', inverted: true, icon: 'checkmark', content: 'Yes' },
  ]}
/>
```

As opposed to the current:
```jsx
<Modal trigger={<Button>Basic Modal</Button>} basic size='small'>
  <Header icon='archive' content='Archive Old Messages' />
  <Modal.Content>
    <p>Your inbox is getting full, would you like us to enable automatic archiving of old messages?</p>
  </Modal.Content>
  <Modal.Actions>
    <Button basic color='red' inverted>
      <Icon name='remove' /> No
    </Button>
    <Button color='green' inverted>
      <Icon name='checkmark' /> Yes
    </Button>
  </Modal.Actions>
</Modal>
```

levithomason:
PRs welcome!

[...]

PR #1508, opened 2017-03-25
feat(Modal): Create a Modal using shorthand props

josie11:
* can pass modal header, content, and actions via shorthand props [(#1390)](https://github.com/Semantic-Org/Semantic-UI-React/issues/1390)

* can pass actions props a triggerClose option that will allow buttons to control open/closing modal


src/modules/Modal/ModalActions.js
```diff
@@ -37,4 +38,6 @@ ModalActions.propTypes = {
   className: PropTypes.string,
 }
 
+ModalActions.create = createShorthandFactory(ModalActions, children => ({ children }))
```
levithomason: Because there can be many buttons, we need to allow the user to pass a `key` or our `childKey` prop.  The third parameter to a`createShorthandFactory ` function is a boolean telling the factory to accept keys:

```jsx
createShorthandFactory(ModalActions, children => ({ children }), true)
```

Next, this should probably map its value to an `actions` prop rather than `children`.  The render method will then map over the `actions` and create buttons:

```jsx
createShorthandFactory(ModalActions, actions => ({ actions }), true)
```

This enables users to use the `actions` prop on the `Modal.Actions` themselves as well:
```jsx
<Modal>
  <Modal.Actions actions={['No', { primary: true, content: 'Yes' }]} />
</Modal>
```

src/modules/Modal/Modal.js
```diff
@@ -265,6 +279,18 @@ class Modal extends Component {
...
+        {_.isNil(children) && content && ModalContent.create(content)}
+        {_.isNil(children) && actions && ModalActions.create(actions.map((action, i) => {
+          if (action.triggerClose) {
+            const onClick = (callback) => (e) => {
+              if (callback) callback(e, this.props)
+              this.handleClose()
+            }
+            action.onClick = onClick(action.onClick)
+          }
+          return Button.create(action, { key: i })
```
levithomason: Using index for a key is an [antipattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318#.cegqr4uf1).  Instead, we can configure the factory to allow the to pass a key.  See comment on `ModalActions.create` below.

src/modules/Modal/Modal.js
```diff
@@ -53,15 +57,21 @@ class Modal extends Component {
     /** Whether or not the Modal should close when the document is clicked. */
     closeOnDocumentClick: PropTypes.bool,
 
+    /** Simple text content for the Modal. */
+    content: customPropTypes.itemShorthand,
```
levithomason, line 61: As we update propTypes, we should also update the typings in the same directory.  Example, `Modal.d.ts`.

levithomason (changes requested):
Thanks for the work!  Here's an initial pass review on progress so far.

josie11:
I've made the changes.

src/modules/Modal/ModalActions.js
```diff
@@ -27,6 +35,9 @@ ModalActions._meta = {
 }
 
 ModalActions.propTypes = {
+  /** Elements to render as Modal action buttons. */
+  actions: PropTypes.arrayOf(customPropTypes.itemShorthand),
```
layershifter: `as` is always first in `propTypes`

src/modules/Modal/ModalActions.js
```diff
@@ -1,23 +1,31 @@
...
   META,
 } from '../../lib'
+import Button from '../../elements/Button'
 
 /**
  * A modal can contain a row of actions.
  */
 function ModalActions(props) {
-  const { children, className } = props
+  const { children, className, actions } = props
```
layershifter: Please sort props in alphabetical order

src/modules/Modal/ModalActions.js
```diff
@@ -1,23 +1,31 @@
...
   const rest = getUnhandledProps(ModalActions, props)
   const ElementType = getElementType(ModalActions, props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+  return (
+    <ElementType {...rest} className={classes}>
+      {children}
+      {_.isNil(children) && actions.map((action) => Button.create(action, true))}
+    </ElementType>
+  )
```
layershifter: I think we can make there a short condition:

```jsx
if(!_.isNil(children)) return <ElementType {...rest} className={classes}>{children}</ElementType>

return (
    {actions.map(action => Button.create(action, true))}
  </ElementType>
)

File: src/modules/Modal/Modal.js [hunk 7]
```diff
@@ -268,6 +281,17 @@ class Modal extends Component {
...
         {children}
+        {_.isNil(children) && ModalHeader.create(header)}
+        {_.isNil(children) && ModalContent.create(content)}
+        {_.isNil(children) && ModalActions.create(_.forEach(actions, (action) => {
+          if (action.triggerClose) {
+            const onClick = (callback) => (e) => {
+              if (callback) callback(e, this.props)
+              this.handleClose()
+            }
+            action.onClick = onClick(action.onClick)
```
layershifter: Can we move event handler to a class method?

layershifter (changes requested):

josie11:
I've made the changes, and I added an example.

docs/app/Examples/modules/Modal/Types/ModalExampleShorthand.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Button, Modal } from 'semantic-ui-react'
+
+const ModalShorthandExample = () => (
+  <Modal trigger={<Button>Show Modal</Button>}
```
levithomason: `trigger` on new line please.

src/modules/Modal/ModalActions.js
```diff
@@ -1,23 +1,32 @@
...
   const classes = cx('actions', className)
   const rest = getUnhandledProps(ModalActions, props)
   const ElementType = getElementType(ModalActions, props)
 
-  return <ElementType {...rest} className={classes}>{children}</ElementType>
+  if (!_.isNil(children)) return <ElementType {...rest} className={classes}>{children}</ElementType>
+
+  return (
+    <ElementType {...rest} className={classes}>
+      {actions.map((action) => Button.create(action, true))}
```
levithomason: Ah, this is a new use case.  We currently don't have a way to enable keys for a create method during runtime like this.  Instead, the create method itself is defined as supporting keys or not.  The button currently does not support keys.

The second param here is actually `defaultProps`.  In order to support keys, we need to pass `true` as the 3rd param to `createShorthandFactory` for the Button's `create` method:

```diff
-Button.create = createShorthandFactory(Button, value => ({ content: value }))
+Button.create = createShorthandFactory(Button, value => ({ content: value }), true)
```

We have a PR up for factory updates.  I'll enable keys for all factories in light of this issue.  We want users to be able to use the `create` method themselves as well, and, they may need keys.

src/modules/Modal/ModalActions.js
```diff
@@ -30,11 +39,20 @@ ModalActions.propTypes = {
...
 
   /** Additional classes. */
   className: PropTypes.string,
 }
 
+ModalActions.defaultProps = {
+  actions: [],
+}
+
+ModalActions.create = createShorthandFactory(ModalActions, actions => ({ actions }), true)
```
levithomason: Because there will only be one instance of Modal.Actions, the Modal.Actions component itself won't need to have keys.  We can remove `true` from here.

See above comment regarding `Button.create` which is what will need the keys, since, there is an array of button elements being created (within Modal.Actions).

levithomason (changes requested):
Here is another interim review.  We're getting closer here, thanks for continuing on this!

josie11:
Changes are made - I've enjoyed learning whats going on under the hood a bit more.

src/modules/Modal/ModalActions.js
```diff
@@ -30,11 +39,20 @@ ModalActions.propTypes = {
...
   /** Primary content. */
   children: PropTypes.node,
 
   /** Additional classes. */
   className: PropTypes.string,
 }
 
+ModalActions.defaultProps = {
+  actions: [],
+}
```
levithomason: Since shorthand is disallowed with `children`, we'll want to avoid setting default values for it.  Otherwise, when a user uses `children` they will also have conflicting `actions` array defined unawares.

Instead, in the render method, just use `_.map` which will gracefully handle undefined values:

```diff
-{actions.map((action) => Button.create(action))}
+{_.map(actions, action => Button.create(action))}
```

levithomason (commented):
Just one change on this review.  I think we're ready to get some tests in here as well 👍 

We have a common test for testing shorthand props.  Search for `implementsShorthandProp` in other tests to see how it works.  You'll have a bit of a tricky case with `actions` when using `triggerClose` as this is not a Button prop and needs to be removed from the object.

josie11:
I added some tests to Modal. Should the situation with triggerClose/Button be throwing an error? I could prevent triggerClose from being passed to ModalActions by using _.omit(action, 'triggerClose') and not using _.forEach.

I had to add _.isPlanObject(action) to prevent implementsShorthandProp from failing for actions:
```
{_.isNil(children) && ModalActions.create(_.forEach(actions, (action) => {
  if (_.isPlainObject(action) && action.triggerClose) action.onClick this.handleTriggerClose(action.onClick)
}))}
```

josie11:
Is there anything I should change/fix here?

layershifter (commented):
@josie11 Thanks for PR, again :+1:  I shipped some updates there. I'll finish review tomorrow, need more time on tests

levithomason:
Thanks, folks! As a heads up, I'll be out on travel for a week starting today. Expect silence from me during this time, however, I will try to catch some mobile updates as I am able.

layershifter (approved):
I've updated tests. @levithomason we need your review there 😄

layershifter:
@levithomason I've resolved merge conflicts, will be glad to receive your review there 👍

levithomason (approved):

levithomason:
Changes look great, merging!

levithomason:

Issue #1365, opened 2017-02-21
RFC: Support submenus in Menu `items` shorthand prop


levithomason:
# Problem

Currently, the `Menu` component allows defining `items` as a shorthand prop:

```jsx
const items = [
  'One',
  { key: 'two', content: 'Two' },  // props object usage shown here
  'Three',
]

<Menu vertical items={items} />
```
![image](https://cloud.githubusercontent.com/assets/5067638/23188391/652c147c-f843-11e6-8fbd-6042c4f75572.png)

A Menu can also have a submenu, by placing a `Menu.Menu` in a `Menu.Item`:

```jsx
<Menu vertical>
  <Menu.Item>Top Level Item 1</Menu.Item>
  <Menu.Item>
     Top Level Item 2
      <Menu.Menu>
        <Menu.Item>Submenu Item 1</Menu.Item>
        <Menu.Item>Submenu Item 2</Menu.Item>
      </Menu.Menu>
    </Menu.Item>
  <Menu.Item>Top Level Item 3</Menu.Item>
</Menu>
```
![image](https://cloud.githubusercontent.com/assets/5067638/23188293/078cedd2-f843-11e6-9c05-e12b2985bd92.png)

However, there is no way to add a submenu when using the `items` prop shorthand definition.

# Proposed Solution

It would be great to be able to define a submenu as a key on an item's shorthand object:

```jsx
const items = [
  'Top Level Item 1',
  {
    key: 'item-2',
    content: 'Top Level Item 2',
    menu: [
      'Sub Item 1',
      'Sub Item 2',
    ],
  },
  'Top Level Item 3',
]

<Menu vertical items={items} />
```

![image](https://cloud.githubusercontent.com/assets/5067638/23188293/078cedd2-f843-11e6-9c05-e12b2985bd92.png)

By default, this would enable support for `<Menu.Item menu={...} />`, as the `menu` config above would be passed to the menu item's `menu` prop.

# Handling Callbacks and Active Item State

What is unclear from here is how the `onItemClick` callback and `activeItemIndex` would be handled.

One idea is to use simply disregard the nested nature of the submenu's items when traversing indices and consider the index of an item to be its natural order in the menu.  The example above would have 5 menu items where:

- index 0 = `Top Level Item 1`
- index 1 = `Top Level Item` 
- index 2 = `Sub Item 1`
- index 3 = `Sub Item 2`
- index 4 = `Top Level Item 3`

[...]

PR #1509, opened 2017-03-25
fix(Dropdown): Close Dropdown when click on custom input in Dropdown-Menu

ali-master:


levithomason:
Could you provide some info on the use case please?  It is not possible to pass a custom input.

ali-master:
Thanks for the replies,

In **Dropdown** Component, User Can use the custom [**input**](http://react.semantic-ui.com/modules/dropdown#dropdown-example-input) in **Dropdown-Menu**, But when the user click on input, Dropdown **close** function is called! and input do not Focus for type text.

levithomason:
Ah, I see.  Apologies for the confusion, however, this example markup is not supported and will likely not be supported in the finalized v1.0 API.  Per the Message above the example:

>The shorthand props API ... needs to be extended to support the markup shown here.

This means, use of `children` will likely not be supported for this use case.  Instead, our shorthand props API will be extended to accomplish this usage.

You can see a little glimpse into how this might begin with more robust `menu` definitions in the RFC #1365, specifically, https://github.com/Semantic-Org/Semantic-UI-React/issues/1365#issuecomment-282119308.

I'll close this for now as the API is headed in a different direction that what is shown here, namely, shorthand props extension.

PR #1510, opened 2017-03-26
fix(Dropdown): Search + Mouse Selection

### LINKED ISSUE #1485: Dropdown: Search + Mouse Selection
### Description

When the dropdown component is used with `search={true}` and the mouse is used to select an option after performing a filtering action, the selected index corresponds to the filtered list, not the overall list.

This is exhibited in the documentation demo page https://react.semantic-ui.com/modules/dropdown#dropdown-example-search-selection

### Steps
1. Open the dropdown
1. Filter the dropdown by typing enough to reduce the number of items in the dropdown
1. Select an item in the dropdown using the mouse, not the keyboard
1. Re-open the dropdown.

### Expected Result

The dropdown placeholder text (and the selected item) should be set to the originally mouse-selected item.

### Actual Result

The dropdown placeholder text (and the selected item) is set to a completely different option. My guess is that the index from the filtered list was used to select the item from the unfiltered list on dropdown open.

### Version

0.67.1

### Testcase

Fork this to show the issue: http://codepen.io/anon/pen/ZerKvp

mchirkin:
Attempt to fix #1485


levithomason:
Thanks for this PR. I'm on mobile so please pardon any brevity/typos. 

I'm wondering if we ought to set the selected index when the value changes, such as in setValue. This way it is in one place and the selected index always matches the chosen value item's index no matter how it was set.

levithomason:
This PR looks to have a rebase issue.  Please rebase to our master branch.  Your master may be out of date as well.

levithomason:
We've received a few duplicate PRs for this bug.  I am working to get #2375 merged as it also addresses a few related issues I've been wanting to fix.  I've also ported the test you've written into that PR.

Thank you for the investigation and contribution!

PR #1515, opened 2017-03-27
docs(Form): add required and width examples

### LINKED ISSUE #1444: Add `required` Form.Field example
Our `Form Variations` section is missing the `required` form field example:
https://semantic-ui.com/collections/form#required

We should add this example in the same fashion as the current inline variation example.

layershifter:
Fixes #1444. Also adds example with withds.

### DIFF
--- a/docs/app/Examples/collections/Form/FieldVariations/FormExampleRequiredField.js
+++ b/docs/app/Examples/collections/Form/FieldVariations/FormExampleRequiredField.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Form, Input } from 'semantic-ui-react'
+
+const FormExampleRequiredField = () => (
+  <Form>
+    <Form.Field required>
+      <label>Last name</label>
+      <Input placeholder='Full name' />
+    </Form.Field>
+    <Form.Checkbox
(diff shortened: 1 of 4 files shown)


docs/app/Examples/collections/Form/FieldVariations/FormExampleRequiredField.js
```diff
@@ -0,0 +1,18 @@
...
+  <Form>
+    <Form.Field required>
+      <label>Last name</label>
+      <Input placeholder='Full name' />
+    </Form.Field>
+    <Form.Checkbox
+      inline
+      label='I agree to the terms and conditions'
+      required
+    />
```
levithomason: Could we split these into two examples with separate descriptions?
***
A Form.Field can be required.
```jsx
<Form.Field required>
```

Form.Field shorthand can also be required.
```jsx
<Form.Checkbox ... required />
```

layershifter:
@levithomason I've splitted examples as you asked 👍

levithomason:
Beautiful!  Will merge on pass 👍

levithomason:

Issue #1520, opened 2017-03-28
How to control text inside Form.Input (like grabbing and clearing after submit)?


zarnihpyoe:
I don't see any props and example of controlling text inside Form.Input in the [documentation](http://react.semantic-ui.com/collections/form) . It'd be great if there's an example in documentation.

levithomason:
Because our components are inline with vanilla React behavior, you can refer to React's docs for more info on how to work with Forms:

https://facebook.github.io/react/docs/forms.html#controlled-components

zarnihpyoe:
Thanks levi. That helps a lot.

levithomason:
I'm adding detailed usage examples for both these cases as well.  You can see them here for now:

https://github.com/Semantic-Org/Semantic-UI-React/pull/1521/files

filipetedim:
As per this issue here: https://github.com/Semantic-Org/Semantic-UI/issues/1870 , the default option where you can't edit values should not appear as editable, as it may confuse users.

PR #1521, opened 2017-03-28
docs(Form): add data usage examples

levithomason:
We get a lot of questions like #1520.  User's often want to know how to get the value from an input, or, how to reset form data on submit.

This PR adds `/Usage` examples to the `Form` and includes two examples.  One showing how to capture data and another showing how to reset on submit.


levithomason:

Issue #1301, opened 2017-02-09
Feature request: handler for <List selection/>


rokoroku:
List could be selectable (or clickable) with `selection` prop, but there is no way to catch the event.

I'll be very happy when we have a new prop like `onItemClick` from menu.
```js
    /**
     * onClick handler for ListItem. Mutually exclusive with children.
     *
     * @param {SyntheticEvent} event - React's original SyntheticEvent.
     * @param {object} data - All item props.
     */
    onItemClick: customPropTypes.every([
      customPropTypes.disallow(['children']),
      PropTypes.func,
    ]),
```

levithomason:
I would take a PR for this.

A workaround is to pass the handler to each item either as a component prop or shorthand object prop.  You can also let the event bubble and capture it.

http://codepen.io/levithomason/pen/WRYaxY

This does not give you the item's props, which is why I'd take a PR for this feature.

rokoroku:
For someone who wants props, here is another workaround currently I use. 
We can receive props in this way :)

http://codepen.io/anon/pen/WRYYdw

layershifter:
Was done in #1530

PR #1530, opened 2017-03-30
feat(List): handler for <List selection/>

josie11:
## To address the enhancement requested in [issue #1301](https://github.com/Semantic-Org/Semantic-UI-React/issues/1301). 

* onItemClick can be passed to List when there are no children; for click handling of list items
```<List content={['item1', 'item2', 'item3']} onItemClick={console.log} />```

* ListItem has click handler for passing list item props.

### CODE REVIEW

**Review by levithomason — COMMENTED** [review]
Changes look great.  We just need to add some tests.  Looking, there are a few things I'd like to fix with tests in this area, so I'll push those shortly.

**Comment by levithomason** [comment]
I've converted the List/ListItem components into classes.  This is due to the click handlers.  We don't want to re-create them on every render, so they need to live on the class instance.

I then added a couple tests for the click handler. I'm a little torn on the differences between `onItemClick` here and in `Menu.js` (where I added a TODO).

The menu preserves the user's menu item `onClick` handler and calls it along with the `onItemClick` handler.  I'll have to come back here and make some decisions on how this should work.

Thoughts?

**Comment by josie11** [comment]
Hm. It may be useful to preserve ```onClick``` while using ```onItemClick``` if you have a ListItem(s)/MenuItem(s) that you want to have any additional behavior beyond ```onItemClick```; I did not consider that. But perhaps logic inside the passed ```onItemClick``` can deal with customizing the behavior depending on the menu item that is clicked (if|else/switch).

File: src/elements/List/List.js [hunk 1]
```diff
@@ -23,124 +23,144 @@ import ListList from './ListList'
...
+  static Icon = ListIcon
+  static Item = ListItem
+  static List = ListList
+
+  handleItemOverrides = predefinedProps => ({
+    onClick: (e, itemProps) => {
+      _.invoke(predefinedProps, 'onClick', e, itemProps)
+      _.invoke(this.props, 'onItemClick', e, itemProps)
+    },
+  })
```
layershifter, line 111: I've updated handlers there like in #1428. I also removed `index` there, I think we don't really need it there.

test/specs/elements/List/List-test.js
```diff
@@ -34,9 +35,37 @@ describe('List', () => {
...
+      const itemProps = { key: 'notes', content: 'Notes', 'data-foo': 'bar', onClick }
+
+      shallow(<List items={[itemProps]} onItemClick={onItemClick} />)
+        .find('ListItem')
+        .first()
+        .shallow()
+        .simulate('click', event)
 
+      onClick.should.have.been.calledOnce()
+      onClick.should.have.been.calledWithMatch(event, callbackData)
```
layershifter, line 61: I also updated test, now it tests that `item` handler fired too

layershifter (approved):
@levithomason I've rebased with master and performed some updates in handlers and tests. Also added some missing tests for shorthand factories. I think we ready for review :+1: 

@josie11 Thanks for PR :+1:

src/elements/List/List.js
levithomason: Agreed, the Menu only needs it as there is an `activeIndex`.

test/specs/elements/List/List-test.js
```diff
@@ -34,9 +35,37 @@ describe('List', () => {
...
-    const items = ['Name', 'Status', 'Notes']
+  const items = ['Name', 'Status', 'Notes']
+
+  describe('onItemClick', () => {
+    it('can be omitted', () => {
+      const click = () => shallow(<List items={items} />).simulate('click')
+      expect(click).to.not.throw()
+    })
+
+    it('is called with (e, { index, ...itemProps }) when clicked', () => {
```
levithomason: No more index here.

levithomason:
Looks great, made one update to a test description.   Will merge tomorrow (3am here)!

levithomason:

PR #1542, opened 2017-04-03
chore(tests): fail on console activity

### LINKED ISSUE #1507: chore(tests): fail on console.error for propType warnings
We have a hard time tracking down failed propType warnings in tests simply because there are so many tests.  We should consider overloading `console.error` in tests to throw if there are propType warnings.

Note, we have the `consoleUtil` to silence tests that intentionally need to test failed propType cases.  This allows us to tell console functionality to `enable`, `disable`, and `disableOnce` for any given tests.

levithomason:
Fixes #1507 

This PR overloads console methods to throw.  This fails all tests that result in console activity.  This reveals several issues with props, keys, and other bad usages.

Karma now also serves the doc files.  This makes `404`s in tests relevant as it means the doc site itself is missing these assets as well.  I'd like to throw on `404` if possible.  This way, we also know all doc site assets exist.

### TODO

- [ ] add `routes-test.js` similar to `examples-test.js` to ensure all routes render with console activity.
- [x] fix remaining failing tests due to console activity
- [x] land #1599 as it is blocking this PR due to console messages


layershifter:
:heart:

layershifter:
#1599 was merged, unblocking.

layershifter (commented):
@levithomason I've merged with `master` and fixed some tests, really want to see it finished :+1:

levithomason:
Only 2 components left 😅 

```
Portal
  ✖ "after each" hook for "does not call this.setState() if portal is unmounted"
    Error: console.error should never be called but was called with:
    Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the WrapperComponent component.

Table
  shorthand
    ✖ renders the table
      Error: console.error should never be called but was called with:
      Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Table`. See https://fb.me/react-warning-keys for more information.
```

### Table (Factories breaking change)

Seems we need a breaking change to factories here.  Arrays cannot be used as shorthand values after all.  We use literal values as keys.  In the case of an array, there is no value to use as the key.

Table row shorthand allows passing an array as a shorthand value.  It is mapped to `cells`.  However, this then creates a prop object `{ cells: [] }` with no key.

### Portal

There is something setting state after unmount still.  Seems we may need the `setSafeState` approach that was taken in Transition.  Before adding that update, I'd like to know _why_ state is being called after mount in the first place.

Note, `Popup` is failing as it uses the Portal.

layershifter:
Unbeviable, tests are passed :v:

layershifter:
@levithomason I've reverted some style changes, marked as ready for review :+1:

levithomason:
Updating... hang tight

levithomason:
Over a year running, but we're actually going to merge this one 😅

layershifter:
Fixed last merge conflicts and merged 👍

levithomason:

PR #1543, opened 2017-04-04
chore(view): update cdn for faker

layershifter:
We recently [performed](https://github.com/Semantic-Org/Semantic-UI-React/pull/1418/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L90) update of `faker` to `4.1.0`. However, cdnjs doesn't contains this version.

I've updated `index.ejs` to use jsdelivr.


jimaek:
I wanted to let you know that our new backend is live. It pulls files directly from npm and github https://github.com/jsdelivr/jsdelivr#usage

Example https://cdn.jsdelivr.net/npm/jquery@3.1.1
Example directory listing https://cdn.jsdelivr.net/npm/jquery@3.1.1/

PR #1548, opened 2017-04-05
Backspace should delete items in multiselect dropdowns

### LINKED ISSUE #1547: Multiple Selection Dropdown backspace does not delete items
### Steps
Select any item from a multiple selection dropdown (not multiple search selection dropdown), press delete to remove that entry from the list

### Expected Result
Item is removed

### Actual Result
Item does not get removed

### Version
0.67.2

### Testcase
http://react.semantic-ui.com/modules/dropdown#dropdown-example-multiple-selection

It appears to work fine when search is turned on.

fracmak:
Fixes #1547 

Not sure why there was an if statement to prevent deletion of multiple items when not searching, but it means the component is not accessible to keyboard users. This removes that restriction.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -584,10 +584,10 @@ export default class Dropdown extends Component {
     debug(keyboardKey.getName(e))
     if (keyboardKey.getCode(e) !== keyboardKey.Backspace) return
 
-    const { multiple, search } = this.props
+    const { multiple } = this.props
     const { searchQuery, value } = this.state
 
-    if (searchQuery || !search || !multiple || _.isEmpty(value)) return
+    if (searchQuery || !multiple || _.isEmpty(value)) return
 
     e.preventDefault()
 
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -1330,16 +1330,23 @@ describe('Dropdown', () => {
 
       spy.should.not.have.been.called()
     })
-    it('does not remove items for multiple dropdowns without search', () => {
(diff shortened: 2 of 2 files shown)


levithomason:
Hm, I'm not sure about this one.  There is no `input` because there is no `search` prop, hence, there is nothing to press backspace in.  In order to remove labels on backspace, there needs to be a cursor to indicate to the user that they can do this.  Otherwise, I fear it will be confusing.

It would also deviate from SUI core as the core dropdown does not behave this way: https://semantic-ui.com/modules/dropdown#multiple-selection

fracmak:
would it be acceptable to add some kind of cursor to give that indication? Or are you disinclined to fix this issue because SUI core doesn't handle it?

levithomason:
The primary concern for me is the UX.  I think backspace behavior is best suited for controls which accept keyboard input and show an input cursor.  Although we could add a readonly input in order get a cursor, I think it'd be confusing to users as to why there is a cursor and backspace works but they are unable to type anything.

This does raise the question, rather than adding a faux cursor and new backspace functionality to a non-search Dropdown, why not use a `search` Dropdown?  This would add the cursor, allow backspace, and the user can type into the dropdown.

If there is a hard requirement here for not allowing search but still allowing backspace, you can always wrap the Dropdown, handle backspace, and control the options.  Something like:

```jsx
handleKeyDown = e => {
  const { options } = this.state
  if (e.key === 'Backspace') {
    this.setState({ options: options.slice(0, options.length - 1)
  }
}

<Dropdown onKeyDown={this.handleKeyDown} options={this.state.options} />
```

I will close this for now.  If we get more requests, I'll reopen this one and perhaps change my views on it.  I do appreciate your input and effort to help improve this library.

PR #1549, opened 2017-04-06
docs(Usage): add instructions for semantic-ui-css

kulakowka:
Added instruction for include styles from semantic-ui-css


levithomason:
Ideally, I think we should also show `<link>` tag usage as `import`ing CSS is not native functionality but unique to bundlers.

This is a step in the right direction so we'll merge this for now.  Thanks!

levithomason:

Issue #1289, opened 2017-02-06
RFC: New types for typings


layershifter:
We have `customPropTypes` for our shorthands and it works fine :smile: But, their represent in typings is not obvious.

My suggestion is to add new types for `customPropTypes` and update existing typings:

```ts
type ShorthandCollection = Array<ShorthandItem>; // for customPropTypes.collectionShorthand
type ShorthandContent = React.ReactNode; // for customPropTypes.contentShorthand
type ShorthandItem = any; // for customPropTypes.itemShorthand
```

@levithomason any feedback?

levithomason:
This sounds really sane.  It will make updates much easier to these special types.

layershifter:
I block this issue before we have the decision in #1364.

PR #1574, opened 2017-04-12
fix(Dropdown): simplify typings validation for renderLabel

### LINKED ISSUE #1568: Dropdown renderLabel expects a ReactElement<any> to be returned
According to the example, renderLabel returns an object, so I'm a little confused as to why the Type definition requires a ReactElement. I can't get the example to compile with TypeScript.

Any ideas?

Version 
0.67.2

```

ERROR in ./src/components/Menu/ComponentGroupMenu/ComponentGroupMenu.tsx
(37,11): error TS2322: Type '(item: DropdownItemProps, index: number, label: LabelProps) => { color: string; content: string; ...' is not assignable to type '((item: DropdownItemProps, index: number, defaultLabelProps: LabelProps) => ReactElement<any>) | ...'.
  Type '(item: DropdownItemProps, index: number, label: LabelProps) => { color: string; content: string; ...' is not assignable to type '(item: DropdownItemProps, index: number, defaultLabelProps: LabelProps) => ReactElement<any>'.
    Type '{ color: string; content: string; icon: string; }' is not assignable to type 'ReactElement<any>'.
      Property 'type' is missing in type '{ color: string; content: string; icon: string; }'.

```

code:

````
public render() {
     const options = [
      { key: 1, text: 'Green', value: ComponentGroup.GREEN },
      { key: 2, text: 'Red', value: ComponentGroup.RED },
      { key: 3, text: 'Blue', value: ComponentGroup.BLUE },
    ]

    const renderLabel = (item: DropdownItemProps, index: number, label: LabelProps) => ({
      color: 'blue',
      content: `Customized label - ${label.text}`,
      icon: 'check',
    });

    return (

        <Dropdown
          compact
          selection
          options={options}
          onChange={this.selectGroup} 
          placeholder='Choose an option'
          renderLabel={renderLabel}
        />
    );
  }`
```

layershifter:
Fixes #1568.

Simplifies validation for `renderLabel` because it uses `Label`'s factory that can accept element, object or string (ref #1289).

### DIFF
--- a/src/modules/Dropdown/Dropdown.d.ts
+++ b/src/modules/Dropdown/Dropdown.d.ts
@@ -201,7 +201,7 @@ export interface DropdownProps {
    * @param {object} defaultLabelProps - The default props for an active item Label.
    * @returns {*} Shorthand for a Label.
    */
-  renderLabel?: (item: DropdownItemProps, index: number, defaultLabelProps: LabelProps) => React.ReactElement<any>;
+  renderLabel?: (item: DropdownItemProps, index: number, defaultLabelProps: LabelProps) => any;
 
   /** A dropdown can have its menu scroll. */
   scrolling?: boolean;
@@ -225,7 +225,7 @@ export interface DropdownProps {
   tabIndex?: number | string;
 
   /** The text displayed in the dropdown, usually for the active item. */
-  text?: string|React.ReactNode;
+  text?: string;
 
   /** Custom element to trigger the menu to become visible. Takes place of 'text'. */
   trigger?: React.ReactNode;


src/modules/Dropdown/Dropdown.d.ts
```diff
@@ -225,7 +225,7 @@ export interface DropdownProps {
   tabIndex?: number | string;
 
   /** The text displayed in the dropdown, usually for the active item. */
-  text?: string|React.ReactNode;
+  text?: string;
```
layershifter, line 228: `text` prop can be only string.

levithomason:

PR #1576, opened 2017-04-12
chore(package): use prop-types package

### LINKED ISSUE #1564: Warning `Accessing PropTypes via the main React package is deprecated` in React 15.5+
### Steps
- npm install
  - react@15.5.3
  - react-dom@15.5.3
  - babel-core@6.24.1
  - babel-preset-latest@6.24.1
  - babel-preset-react@6.24.1
- Create an extended custom React.Component that imports Semantic UI React Icon Component
- Build and open app in Chrome Browser

### Expected Result
No warnings
### Actual Result
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
printWarning	@	warning.js?8a56:36
warning	@	warning.js?8a56:60
get	@	React.js?24ef:95
(anonymous)	@	IconGroup.js?f416:40

`
process.env.NODE_ENV !== "production" ? IconGroup.propTypes = {
  as: customPropTypes.as,

  /** Primary content. */
  children: PropTypes.node,

  /** Additional classes. */
  className: PropTypes.string,

  /** Size of the icon group. */
  size: PropTypes.oneOf(_without(SUI.SIZES, 'medium'))
`

### Version
0.67.2

### Testcase

layershifter:
Fixes #1564.

> Prop types are a feature for runtime validation of props during development. We've extracted the built-in prop types to a separate package to reflect the fact that not everybody uses them.

[Ref](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes).


package.json
```diff
@@ -56,7 +56,8 @@
     "babel-runtime": "^6.22.0",
     "classnames": "^2.1.5",
     "debug": "^2.6.3",
-    "lodash": "^4.17.2"
+    "lodash": "^4.17.2",
+    "prop-types": "^15.5.0"
```
layershifter: I'm not sure about this, however, I'm followed [recommendation](https://github.com/reactjs/prop-types#how-to-depend-on-this-package) there.

levithomason: Looks good, however, let's update to the latest:

>Note: there are known issues in versions before 15.5.7 so we recommend using it as the minimal version.

layershifter: Updated to `15.5.7` :+1:

webpack.umd.config.js
```diff
@@ -13,6 +13,7 @@ const webpackUMDConfig = {
     [pkg.name]: paths.src('umd.js'),
   },
   externals: {
+    'prop-types': 'PropTypes',
```
levithomason: I believe per their instructions we should bundle propTypes with the code as a dependency.  Otherwise, users will also have to make it globally available.  I'll update this.

Just as a heads up, the Codepen is powered by the UMD build as well so any UMD changes also require updates there.

levithomason (approved):
Looks great, made one minor change.  Will merge on pass.

levithomason:

PR #1579, opened 2017-04-13
[WIP/RFC] CSS-in-JS styling

### LINKED ISSUE #1009: RFC: Theming support
We get questions and requests for this a lot.  Some of which can be seen in #802 #775 #366 #851 and #472.  This is not an exhaustive list by any means and excludes Gitter conversations and conversations @TechnologyAdvice.  I'm opening this RFC to get feedback and consolidate the brain cycles being spent on this.

The goal here is to support the full [theming capabilities](http://semantic-ui.com/usage/theming.html) of Semantic-UI core (SUI) directly in Semantic-UI-React (SUIR).  This is extremely powerful as can be seen in [this GitHub remake example](http://semantic-org.github.io/example-github/), be sure to click the little paint bucket icon next to the notifications icon to open the theming sidebar.

# Problem Statement

Apps need themes.  Theming should be stupid simple and immediately accessible.  Currently, it is anything but.  Theming requires cloning a separate repo, with a separate build system, building a theme, hosting it somewhere, then adding it to your project.  This is not acceptable.

# Ideal Scenario

You should be able to choose a theme right in the project that is implementing SUIR.  You should also be able to change theme variables and add component style overrides directly in your project.

You should be able to add not only inline styles to individual components but theme-aware-inline-styles.  Example, when the background color is set on a Button, it should know how to handle the variations for `basic` and `inverted` and also states like `:hover` and `:active`.

# Possible Solutions

I've been experimenting with a few approaches to making this happen.  Just today, I was shown [Styletron](https://github.com/rtsao/styletron), a highly performant "universal CSS-in-JS engine."  This is almost identical to the approach I wanted to take.  I suggest reading the blog post and checking out the related projects linked in the README.  Something very similar to this, or Styletron itself, might be the tool we need to bring theme support to SUIR.

I don't think adding theme support is worth it if we were to consider doing so with a traditional CSS solution.  These include preprocessor builds (SCSS/LESS/etc), intelligently importing individual component stylesheets, modifying preprocessor variables, and the like.  All traditional approaches to CSS come with several issues such as globals, infinitely growing CSS sheets, no static analysis, no dead code elimination, etc.  If we do this, I think it is only worth it if we do so with a novel modern approach that solves these issues as well.

# How do we get there?

I don't know.  Here is a brain dump.

### It's a rewrite

Tools like Styletron bring us closer to be able to pull this off.  The daunting bit here is that no matter the tool, we're talking about implementing the SUI CSS spec in another language.  This is an undertaking comparable to SUIR itself, which re-implemented SUI HTML in React.

This leads to an obvious point that if this work were produced, it certainly should _not_ be dependent on React.  It should be a framework agnostic tool that React components can implement easily.  It also should not be dependent on a CSS preprocessor or CSS tool.  Again, Styletron has this covered with `styletron-react`.

The good news here is that we would no longer be limited by the sometimes awkward CSS and HTML required by SUI.  We could make changes to the styles along with the components as necessary.

### It should be in JS

This should be done with plain JS objects and a smart engine that can generate stylesheets.  JS gives you portability and static analysis.  Portability means you can easily share and use style objects anywhere, client side, server side, React, Elm, etc.  You can even request them as JSON from a server.  Static analysis means you can generate selectors, perform dead code elimination, optimize stylesheets, and do just about anything else you desire.

This is where I think Styletron really got it right.  It appears they are implementing many of these features based on plain objects.

### Bye, bye, `className` buildup

Yep, no component would ever use an HTML className.  Instead, they would compose style objects.  The style engine would be responsible for applying that style to the component.  Ideally, it would generate an optimized stylesheet, inject it into the page, and assign the required classNames to the component. In fact, this is precisely what Styletron does.

**_If we're not using SUI CSS nor SUI markup, is this even a SUI project anymore?!_**

Absolutely.  SUI is a language for web components, it is not an HTML, CSS, nor JS framework.  Qutoing @jlukic from a recent email, he is actually working toward 

>...a "W3C for UI", an organization which can create simple written specifications for UI components (what methods, options, behaviors) that developers can freely implement "to spec" in their chosen framework or methodology.

The SUI language would be expressed in our component props still just as it is today.  We just would not implement any HTML classes.  All our current utils for props-to-className buildup would become props-to-composed-style-object utils.  I can imagine this tech aligning far better with the Semantic UI language itself.

# Possible Implementation & API

Heads up!  These are completely fragile and undeveloped ideas.

### Composing Style Objects

SUI provides a common language across all components.  Consider "variations" such as `color`, `inverted`, `basic` and `fluid`.  _Most_ components implement _most_ variations.  Each variation would then have an associated style object. Changing styles from `basic` to `inverted` just means merging the `inverted` style object into the style after the `basic` variation object.  This is also true with theme information.

>This sounds identical to CSS until you consider what can be done in JS compared to CSS in terms of libraries, logic, and functional composition.  Color manipulation, context-aware CSS transitions, data-driven styles come to mind.

### Child Components

Forgive the extremely unruly code here, but in experiments in [this codepen](http://codepen.io/levithomason/pen/vKvOdm), I've shown some rough ideas on how you could generating themes from plain objects.  The important bit in that POC is the use of React context to pass cascading theme information to child components.

Example, there is a default Button style and at the bottom of the pen there is a Panel with a Button child.  When placed in the Panel, the Button merges a new style it obtained from the Panel.  A mechanism like this is needed to handle composing style objects in replacement of traditional cascading styles.  This way, when a Button is placed in an `inverted` Segment, the Button applies its own `inverted` style "variation".

A mechanism like this would need to be designed and generalized so that it works with all components.  It should be based on the SUI language.  If so, then most (hopefully all) nested styles could be rendered unnecessary.  Instead of a stylesheet including all possible permutations of component hierarchies, parent components would set the theme context for their domain and child components would apply their appropriate style objects.  This way, the theming is more or less "flat".

### Config and Provider

We would also need some ability to set configuration at the app level.  This would contain the theme settings.  I believe for this to work we may also need to include a root Provider component that would be in control the app's theme.

***

I'll leave this open-ended for comments, doubts, and other ideas.

maxdeviant:
This is a very early stage PR which aims to get us moving in the direction of a CSS-in-JS solution for styling SUIR. I've read over #1009 and am taking @levithomason's thoughts into account with the implementation.

The current plan is to use [glamor](https://github.com/threepointone/glamor) as the CSS-in-JS solution for React. The full list of features can be found [here](https://github.com/threepointone/glamor#features), but some important highlights:
- Framework agnostic
- Supports all the pseudo :classes/::elements
- Support for `@media` / `@font-face` / `@keyframes`
- SSR support

The base styles themselves will be distributed in a separate package, and will not be tied to React. This will allow for implementations for other JS frameworks. The one caveat here is that while the base JS style objects are generally the same between CSS-in-JS implementations, they start to vary when you get down to the pseudo-selectors, `@media` queries, etc. This means that the raw styles themselves would probably need to be glamor-specific, or at least something that looks a lot like glamor (such as [cxs](https://github.com/jxnblk/cxs)). Given that glamor is not tied to a particular framework, I do not see this as being a show stopper.

Currently the styles are located at [maxdeviant/semantic-ui-css-in-js](https://github.com/maxdeviant/semantic-ui-css-in-js) and have a corresponding [npm package](https://www.npmjs.com/package/semantic-ui-css-in-js). The approach here is going to look very similar to the LESS files in the main Semantic UI library. The primary difference will be that instead of exposing the styles directly, the package will expose style creators that take in the theme to allow for dynamic theming at runtime.

In way of theming, the theme configuration will be passed down using context via some sort of `ThemeProvider` component.

Landing this PR will close #1009.

@levithomason Would love to hear your thoughts on this :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,9 @@
     "babel-runtime": "^6.22.0",
     "classnames": "^2.1.5",
     "debug": "^2.6.3",
-    "lodash": "^4.17.2"
+    "glamor": "^2.20.24",
+    "lodash": "^4.17.2",
+    "semantic-ui-css-in-js": "0.0.1"
   },
   "devDependencies": {
     "@types/react": "^15.0.14",
@@ -67,8 +69,8 @@
     "babel-plugin-__coverage__": "^11.0.0",
     "babel-plugin-lodash": "^3.2.10",
     "babel-plugin-react-transform": "^2.0.2",
-    "babel-plugin-transform-react-remove-prop-types": "^0.3.2",
     "babel-plugin-transform-react-handled-props": "^0.2.3",
+    "babel-plugin-transform-react-remove-prop-types": "^0.3.2",
     "babel-plugin-transform-runtime": "^6.15.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.5.0",
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -214,23 +214,28 @@ class Button extends Component {
       size,
       toggle,
     } = this.props
+    const buttonStyles = makeButtonStyles()
+    const sx = css(
+      buttonStyles.base,
+      active && buttonStyles.active,
+      basic && buttonStyles.basic,
+      circular && buttonStyles.circular,
+      compact && buttonStyles.compact,
+      fluid && buttonStyles.fluid,
+      (icon === true || icon && (labelPosition || !children && !content)) && buttonStyles.icon,
+      inverted && buttonStyles.inverted,
+      loading && buttonStyles.loading,
+      negative && buttonStyles.negative,
+      positive && buttonStyles.positive,
+      primary && buttonStyles.primary,
+      secondary && buttonStyles.secondary,
+      toggle && buttonStyles.toggle
+    )
+    const classes = cx(`${sx}`, className)
 
     const baseClasses = cx(
       color,
       size,
-      useKeyOnly(active, 'active'),
-      useKeyOnly(basic, 'basic'),
-      useKeyOnly(circular, 'circular'),
-      useKeyOnly(compact, 'compact'),
-      useKeyOnly(fluid, 'fluid'),
-      useKeyOnly(icon === true || icon && (labelPosition || !children && !content), 'icon'),
-      useKeyOnly(inverted, 'inverted'),
-      useKeyOnly(loading, 'loading'),
-      useKeyOnly(negative, 'negative'),
-      useKeyOnly(positive, 'positive'),
-      useKeyOnly(primary, 'primary'),
-      useKeyOnly(secondary, 'secondary'),
-      useKeyOnly(toggle, 'toggle'),
       useKeyOrValueAndKey(animated, 'animated'),
       useKeyOrValueAndKey(attached, 'attached'),
     )
@@ -247,10 +252,6 @@ class Button extends Component {
     const tabIndex = this.computeTabIndex(ElementType)
 
     if (!_.isNil(children)) {
-      const buttonStyles = makeButtonStyles()
-      const sx = css(buttonStyles.base)
-      const classes = cx(`${sx}`, className)
-      // const classes = cx('ui', baseClasses, wrapperClasses, labeledClasses, 'button', className)
       debug('render children:', { classes })
 
       return (
@@ -266,7 +267,6 @@ class Button extends Component {
     } })
 
     if (labelElement) {
-      const classes = cx('ui', baseClasses, 'button', className)
       const containerClasses = cx('ui', labeledClasses, 'button', className, wrapperClasses)
 
       debug('render label:', { classes, containerClasses }, this.props)
@@ -283,7 +283,6 @@ class Button extends Component {
     }
 
     if (!_.isNil(icon) && _.isNil(label)) {
-      const classes = cx('ui', labeledClasses, baseClasses, 'button', className, wrapperClasses)
       debug('render icon && !label:', { classes })
 
       return (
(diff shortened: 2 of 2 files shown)


levithomason:
Very glad to see this. I've been thinking it's time to move forward on this endeavor.

I'm on vacation this week, but, will be back next week. Thanks for the PR and deep dive.

maxdeviant:
Made some more progress tonight!

The button on the left is rendered using Glamor, and the one on the right is
 using ye olde Semantic UI classnames 😮 

screen shot 2017-04-13 at 10 59 30 pm

I think the hardest part in all of this will probably be converting all of the Less over to JS. Some things that are making it easier:
- TypeScript for managing all of the variables in a type-safe manner
- [`✨ polished`](https://github.com/styled-components/polished) for replacing the built-in Less functions
  - The only hard part here is ensuring the same behavior when going from Less to `✨ polished`

I'm curious if there would be a way to generate TypeScript from Less, but I think that might be asking for too much 😅

levithomason:
SUI LESS is very minimal, mostly making use of variables only.  They do not make use of nesting, complex mixins, or LESS functions.  Because they are so simple, it may be possible to generate TypeScript and also style objects.

I do like what Styletron offers in terms of performance and approach over others, you can read more here https://ryantsao.com/blog/virtual-css-with-styletron.

@davezuko and I will investigate this endeavor a bit and offer some more ideas and feedback, hopefully soon.

maxdeviant:
@levithomason That blog post was a very good read. Seems like Styletron has a lot going for it.

levithomason:
@maxdeviant I've posted a high-level update regarding investigations and goals in https://github.com/Semantic-Org/Semantic-UI-React/issues/1009#issuecomment-295821011.  You may be interested.

layershifter:
I close this for housekeeping, there is no decision about CSS-in-JS library now (#1009).

levithomason:
Thanks for cleaning up.

So the community knows, this is still on my mind (along with a better theme customizer) but the time is not on my hands.  I would love to see a concerted community effort to pull this off, though.

PR #1581, opened 2017-04-13
fix(Input): add handling of input's ref

### LINKED ISSUE #1536: Can't get ref on <input> if it's a child of <Input>
Can't get React ref on `<input>` if it's a child of `<Input>`.

### Steps
```
<Input>
  <input type="text" ref={r => console.log(r)}/>
</Input>
```

### Expected Result
I would expect to see this following in the console, indicating that `r` is the ref to the `<input>` tag.
```
<input type="text">
```

### Actual Result
The ref function is never called. There is no console.log output at all.

### Version
0.67.2

### Testcase
http://codepen.io/darkadept/pen/mWazWa

layershifter:
Fixes #1536.

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -1,6 +1,6 @@
-import _ from 'lodash'
-import React, { Children, cloneElement, Component, PropTypes } from 'react'
 import cx from 'classnames'
+import _ from 'lodash'
+import React, { Children, Component, PropTypes } from 'react'
 
 import {
   createHTMLInput,
@@ -114,6 +114,8 @@ class Input extends Component {
     type: META.TYPES.ELEMENT,
   }
 
+  focus = () => (this.inputRef.focus())
+
   handleChange = (e) => {
     const { onChange } = this.props
     const value = _.get(e, 'target.value')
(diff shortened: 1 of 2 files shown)


src/elements/Input/Input.js
```diff
@@ -187,7 +192,10 @@ class Input extends Component {
       const childElements = _.map(Children.toArray(children), (child) => {
         if (child.type !== 'input') return child
 
-        return cloneElement(child, { ...htmlInputProps, ...child.props })
+        return createHTMLInput(child, {
+          defaultProps: htmlInputProps,
+          overrideProps: this.handleInputOverrides,
+        })
```
layershifter: Use factory instead of `cloneElement`

test/specs/elements/Input/Input-test.js
```diff
@@ -170,42 +186,39 @@ describe('Input', () => {
     })
   })
 
+  describe('ref', () => {
+    it('maintains ref on child node', () => {
+      const ref = sandbox.spy()
+      const wrapper = mount(<Input><input ref={ref} /></Input>)
+
+      // ref.should.have.been.calledOnce()
```
layershifter: @levithomason I can't get this test working, can you take a look?

levithomason: Refs are not passed through props. React makes them undefined. We'll have to accept the ref on a different prop name, or, update factories to include refs in the overrides. We should probably use a different prop name, such as, `inputRef`.

layershifter (commented):
I've reverted back to `cloneElement` and updated handling of `ref`, also updated tests now everything works as expected. @levithomason thanks for the point to right direction

levithomason (approved):
Perfect, thanks!

levithomason:

Issue #1565, opened 2017-04-10
Input does not get disabled attibute


cattermo:
### Steps
When putting disabled=true prop on Input component.

### Expected Result
Disabled attribute (or similar that screen reader can interpret) should be rendered on input html.

### Actual Result
I can't find anything in the html that tells a screenreader that the input is disabled.

### Version
0.66.0

### Testcase
See html rendered from docs example:
http://react.semantic-ui.com/elements/input#input-example-disabled 

For TextArea, Dropdown and Checkbox this works fine. They all get either disabled attribute or aria-disabled attribute.

dvdzkwsk:
Saw this myself the other day, I can pick it up.

dvdzkwsk:
Just got back to work today, will hopefully have this fixed before the day is over.

PR #1591, opened 2017-04-17
fix(Input): replace disabled class with disabled attribute

dvdzkwsk:
Resolves https://github.com/Semantic-Org/Semantic-UI-React/issues/1565.

@levithomason This, ironically enough, is a matter of semantics. Your call here, but in my opinion this improves the functionality of inputs as they now receive an actual `disabled` attribute instead of relying on custom styling with a `disabled` class above the input element itself. This allows disabled inputs to inherit standard browser behavior and conform to general user expectations.

I checked it out in the docs and by removing the `disabled` class, the styles still look correct.


src/lib/htmlInputPropsUtils.js
```diff
@@ -5,12 +5,8 @@ export const htmlInputAttrs = [
   'selected', 'defaultValue', 'defaultChecked',
 
   // LIMITED HTML PROPS
-  'autoCapitalize', 'autoComplete', 'autoCorrect', 'autoFocus', 'checked', 'form', 'max', 'maxLength', 'min',
+  'autoCapitalize', 'autoComplete', 'autoCorrect', 'autoFocus', 'checked', 'disabled', 'form', 'max', 'maxLength', 'min',
   'multiple', 'name', 'pattern', 'placeholder', 'readOnly', 'required', 'step', 'type', 'value',
-
-  // Heads Up!
```
dvdzkwsk: Handled this by not applying the `disabled` class. Potentially breaking change for users who relied on that as a CSS selector.

levithomason:
Per offline chats, let's keep both the top level className so child elements are disabled but also apply the attribute.

I'll PR the CSS to handle the double opacity issue.

levithomason:
Tracking https://github.com/Semantic-Org/Semantic-UI/pull/5284

levithomason:
This seems to have broken tests for `input shorthand prop (common)`.

dvdzkwsk:
SUI issue has no traction :(

src/elements/Input/Input.js
```diff
@@ -135,6 +142,23 @@ class Input extends Component {
...
+    const tabIndex = this.computeTabIndex()
+    const unhandled = getUnhandledProps(Input, this.props)
+    const [htmlInputProps, rest] = partitionHTMLInputProps(unhandled)
+
+    htmlInputProps.ref = this.handleInputRef
+    htmlInputProps.type = type
+
+    if (disabled) htmlInputProps.disabled = disabled
+    if (onChange) htmlInputProps.onChange = this.handleChange
+    if (tabIndex) htmlInputProps.tabIndex = tabIndex
```
layershifter, line 157: The problem is that the component receives `disabled` as `undefined`, same is for `tabIndex`

I made tests working, however, it's not the best idea :hankey:  May be better solution possible there?

levithomason: I believe if the `disabled` prop is `undefined`, then the `htmlInputProps.disabled` should also be undefined.  Perhaps  I'm overlooking the issue or use case here.

layershifter: @levithomason You're right, but if we replace current code with:

```js
if (onChange) htmlInputProps.onChange = this.handleChange

return [{...htmlInputProps, disabled, tabIndex}, rest]
```

But, we will have failing tests of shorhand.

layershifter: However, we can fix this in test:
```diff
  common.implementsHTMLInputProp(Input, {
    alwaysPresent: true,
-    shorthandDefaultProps: { type: 'text' },
+    shorthandDefaultProps: { disabled: undefined, tabIndex: undefined, type: 'text' },
  })
```

May be I overlooked something?

levithomason:
Thanks for the updates @layershifter !

levithomason:

PR #1596, opened 2017-04-18
docs(examples): fix in webpack1 example

layershifter:


examples/webpack1/.babelrc
```diff
@@ -5,7 +5,6 @@
     "stage-1"
   ],
   "plugins": [
-    "lodash",
-    ["lodash", { "id": "semantic-ui-react" }]
+    ["lodash", { "id": ["lodash", "semantic-ui-react"] }]
```
layershifter, line 8: We [can](https://github.com/lodash/babel-plugin-lodash#usage) run resolution updates in single run.

examples/webpack1/public/index.html
```diff
@@ -8,6 +8,6 @@
 </head>
 <body>
 <div id="root" style="height: 100%"></div>
-<script src="./dist/bundle.js"></script>
+<script src="./dist/app.js"></script>
```
layershifter, line 11: Fixed filename

examples/webpack1/src/main.js
```diff
@@ -17,4 +17,4 @@ const App = () => (
   </Container>
 )
 
-render(App, MOUNT_NODE)
+render(<App />, MOUNT_NODE)
```
layershifter, line 20: Ugly mistake :sob:

levithomason:
Whoops, I should have caught these issues in review!

levithomason:

PR #1599, opened 2017-04-19
breaking(Modal): update shorthand functionality

levithomason:
# Breaking Changes

Previously, shorthand Modal `actions` required passing button props objects with `triggerClose: true` on buttons that should close the Modal.

Now, all shorthand Modal `actions` buttons close the Modal automatically.  The `triggerClose` property has been removed.

## Before

```jsx
<Modal
  header='Reminder!'
  content='Call Benjamin.'
  actions={[
    { key: 'snooze', content: 'Snooze', triggerClose: true },
    { key: 'done', content: 'Done', positive: true, triggerClose: true },
  ]}
/>
```

## After

```jsx
<Modal
  header='Reminder!'
  content='Call Benjamin.'
  actions={[
    'Snooze',
    { key: 'done', content: 'Done', positive: true },
  ]}
/>
```

***

Blocking #1542 

This PR:

- [x] removes the `triggerClose` shorthand Modal `actions` property
- [x] makes all shorthand actions close the Modal
- [x] fix prop types / typings

### DIFF
--- a/docs/app/Examples/modules/Modal/Types/ModalExampleShorthand.js
+++ b/docs/app/Examples/modules/Modal/Types/ModalExampleShorthand.js
@@ -4,11 +4,11 @@ import { Button, Modal } from 'semantic-ui-react'
 const ModalShorthandExample = () => (
   <Modal
     trigger={<Button>Show Modal</Button>}
-    header='Delete Your Account'
-    content='Are you sure you want to delete your account'
+    header='Reminder!'
+    content='Call Benjamin regarding the reports.'
     actions={[
-      { key: 'no', content: 'No', color: 'red', triggerClose: true },
-      { key: 'yes', content: 'Yes', color: 'green', triggerClose: true },
+      'Snooze',
+      { key: 'done', content: 'Done', positive: true },
     ]}
   />
 )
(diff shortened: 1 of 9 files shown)


layershifter:
I've pushed some changes there:
- updated to latest master;
- reverted [change](https://github.com/Semantic-Org/Semantic-UI-React/pull/1599/commits/6e040233b1145817d879db63e7f62b750ce68667#diff-772940f95ec34befa3cce13afd06f144R16) of type in typings;
- changed `actions` to [collection shorthand](https://github.com/Semantic-Org/Semantic-UI-React/pull/1599/commits/6e040233b1145817d879db63e7f62b750ce68667#diff-e762b5cc3074c8997386b6de83cac7fbR36);
- added `onActionClick` [prop](https://github.com/Semantic-Org/Semantic-UI-React/pull/1599/commits/6e040233b1145817d879db63e7f62b750ce68667#diff-e762b5cc3074c8997386b6de83cac7fbR80);
- fixed broken tests.

### TODO
- [x] make a decision about `closeIcon` prop, it will be breaking change if we'll left it as `customPropTypes.itemShorthand`;
- [x] solve warnings with `implementsCreateMethod` test on `ModalActions`, I think we to update it with new option, that will skip non array inputs.

@levithomason I'll be glad to hear your thoughts.

a-x-:
ping

levithomason:
I've rebased to the latest master and resolved all issues:

`closeIcon` - Leaving this as is for minimal churn, allowing with children.

Updated `actions` propType to `itemShorthand` since it is not an array of buttons but the props to `Modal.Actions`.  You can pass an array of buttons because `mapValueToProps` for `ModalActions` maps the array to the `Modal.Actions` `actions` prop, which is an array of buttons.

levithomason:

Issue #1592, opened 2017-04-17
Search: Pressing up/down arrow in the search input with no results throws errors to console


maciej-gurban:
**Steps to Reproduce**
1. Start typing in the search input box (random characters, just make sure there's no results)
2. While still having the input focused, press arrow up or down on the keyboard

Reproducible in all [demos of search component](http://react.semantic-ui.com/modules/search)

**Expected**
No error should be thrown

**Result**
The following error is shown in the console:
```
Uncaught TypeError: Cannot read property 'offsetTop' of null
    at t.a.scrollSelectedItemIntoView (app.d0cf20c….js:7194)
    at t.a.moveSelectionBy (app.d0cf20c….js:7194)
    at HTMLDocument.a.moveSelectionOnKeyDown (app.d0cf20c….js:7194)
```

levithomason:
Fixed in #1600

maciej-gurban:
Damn, you guys are fast. Thanks!

Also, I've been working with the library for a few days now, and I have to say it's pretty damn well written and so nicely extensible.

levithomason:
Woot! :owl:

PR #1600, opened 2017-04-19
fix(Search): pressing up/down arrow in the search input causes error

josie11:
Attempts to fix the issue raised in [#1592](https://github.com/Semantic-Org/Semantic-UI-React/issues/1592) by adjusting how event listeners are added/removed based on search results.

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -228,9 +228,6 @@ export default class Search extends Component {
     // Do not access document when server side rendering
     if (!isBrowser) return
 
-    const isPreviousSearchResults = !_.isEmpty(prevProps.results)
-    const isSearchResults = !_.isEmpty(this.props.results)
-
     // focused / blurred
     if (!prevState.focus && this.state.focus) {
       debug('search focused')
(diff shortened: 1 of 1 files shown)


GregoryPotdevin:
Wouldn't it be safer to let the event listeners as usual (after all, the "no result" is still an open popup) but fix the listener functions to not throw an error if the results are empty? Is there a risk of keyboard events bubbling up if they aren't caught when there are no results?

levithomason:
The Dropdown is almost a copy of this component.  Ideally, it would share base logic and features, however, that work was begun but not completed.

It may also be worth looking at how the Dropdown solved this issue.  It has the same methods but over time has had a lot more attention and therefore fixes.

josie11:
Oops, I should have examined scrollSelectedItemIntoView closer instead of making all those other changes, sorry! 
I adjusted this to match to how I perceive Dropdown handles this, which is seemingly a one line fix?

levithomason:
On the surface, this sounds right.  There may is likely also a test we can copy over as well that ensures pressing arrow keys when there are no results actually works.

levithomason:
Going to merge this for now since it is such a minor fix resolves a known bug.

levithomason:

PR #1602, opened 2017-04-21
feat(TextArea): add focus() method

gnowoel:
I want to focus a `TextArea` element with a `ref`, but the `ref` refers to a component instance rather than a DOM element. We already solved this problem for `Input` by adding a `focus()` method, which will relay the method invocation to the underlying DOM element. So, in this PR, I'd like to add the same `focus()` method to the `TextArea` component.

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -54,6 +54,8 @@ class TextArea extends Component {
     }
   }
 
+  focus = () => (this.ref.focus())
+
   handleChange = (e) => {
     const { onChange } = this.props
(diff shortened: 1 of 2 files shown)


layershifter (commented):
@gnowoel Thanks for PR, I've updated a test, now lint passes. LGTM :+1:

layershifter (approved):

levithomason:
Thanks much guys!

gnowoel:
@layershifter Thanks much for the test update! I felt frustrated about the failed tests since they passed on localhost :)

levithomason:

Issue #720, opened 2016-10-21
feat(Dropdown): open to side with most space left


dylankiss:
When you have a dropdown at the bottom of your viewport, it opens below. That way the content of the dropdown is off screen. In Semantic-UI the dropdown opens to the side where there is the most space left, so in this case it opens to the top.

This would be a great UX improvement, since now users often have to scroll further down when opening a dropdown.

tamernasr:
العرب حياة

persocon:
👍

persocon:
I've added a specific class for direction on my project :v and changed myself the position on the CSS for those specific cases, but would be nice to have a automatic solution like in the `Semantic-ui` original dropdown

levithomason:
Agreed. This issue is labeled for help wanted and PRs are welcome.

henri-hulski:
I also need this feature.
Was just looking at the code, but no idea where to fix this.
In semantic-ui this was a quite easy fix:
https://github.com/Semantic-Org/Semantic-UI/commit/04d65e7483d1ac935ea9bdbb44996a4d0b876857

henri-hulski:
With `floating className='upward'` it actually works already and looks quite ok.
Without `floating` the round corners are only on the bottom which looks a little stange. They should be on the top.
With `floating` it's not really floating, but looks better as it has round corners all around.

cguedes:
This should go to the documentation for "className", also listing the ones that we can use (ex: icon, upward)

levithomason:
Semantic UI uses HTML classes to trigger functionality in their modules.  Semantic UI React provides a component API instead.  Our component API provides both the HTML classes and necessary React component logic.

Instead of adding the className to the docs, we need to add the prop `upward` to the Dropdown:

```jsx
<Dropdown upward />
```

We should also auto-apply this className when there is not enough room for the menu on the screen, just as we do for the Popup.

cguedes:
@levithomason 100% agree with you.

levithomason:
Great, I'm halfway through a feature branch already for adding this prop :)  PR coming soon.  

Initially, it will only be set manually.  It will take another PR to auto apply the className.  Would be great to have someone pick that up.

cguedes:
Let me know, I think I can help.

levithomason:
See https://github.com/Semantic-Org/Semantic-UI-React/pull/1603 for the new `upward` prop and 3 new doc examples.  I will merge this on CI pass.

[...]

PR #1603, opened 2017-04-21
feat(Dropdown): add upward prop

levithomason:
#720 raised the issue of opening a dropdown to the side with the most space available.  This PR adds a manually configurable prop for doing this.  A subsequent PR should measure the menu and apply this automatically.


levithomason:

PR #1606, opened 2017-04-21
docs(readme): Add Netflix EDX as a user

jayphelps:


levithomason:
Awesome, thanks much :)

PR #1609, opened 2017-04-22
fix(Modal): make content and header props optional

layershifter:
This PR makes new props introduced in #1508 optional. I missed this on the review 😿


layershifter:
Absorbed by #1611

levithomason:

PR #1611, opened 2017-04-23
test(typings): initial tests for typings

layershifter:
We have constant problems with typings (#1609 and more), it's time to solve them ✌️

-----

This PR replaces #1181 and offers initial tests for typings. What it does:
- tests that typings file `Component.d.ts` exists for each component;
- tests that definition file has `ComponentProps` interface and it's exported;
- tests that definition contains all props from `Component.propTypes` and props that are required in definitions are really required.

Also, it fixes all found problems :clinking_glasses: Our typiscript users will be satisfied 👍


index.d.ts
```diff
@@ -64,7 +64,7 @@ export { default as Flag, FlagProps } from './dist/commonjs/elements/Flag'
 
 export { default as Header, HeaderProps } from './dist/commonjs/elements/Header'
 export { default as HeaderContent, HeaderContentProps } from './dist/commonjs/elements/Header/HeaderContent'
-export { default as HeaderSubheader, HeaderSubHeaderProps } from './dist/commonjs/elements/Header/HeaderSubheader'
+export { default as HeaderSubheader, HeaderSubheaderProps } from './dist/commonjs/elements/Header/HeaderSubheader'
```
layershifter, line 67: Fixed interface name

src/collections/Breadcrumb/BreadcrumbSection.d.ts
```diff
@@ -15,6 +15,9 @@ export interface BreadcrumbSectionProps {
   /** Additional classes. */
   className?: string;
 
+  /** Shorthand for primary content. */
+  content?: any;
+
```
layershifter, line 20: Missing definition

test/specs/modules/Modal/Modal-test.js
```diff
@@ -45,6 +45,8 @@ describe('Modal', () => {
   })
 
   common.hasSubComponents(Modal, [ModalHeader, ModalContent, ModalActions, ModalDescription])
+  common.hasValidTypings(Modal)
```
levithomason, line 48: Looks like this might already be covered by the `common.isConformant()` test, true?

layershifter: `Modal-test.js` doesn't contain `isConformant` test

levithomason: Of course, thanks!

levithomason:

PR #1613, opened 2017-04-24
fix(acc): fix default prop dev warning

### LINKED ISSUE #1586: Auto-controlled props emit spurious warning when set to undefined (missing undefined check).
### Steps

For any auto-controlled property `x`, set both `x` and `defaultX` to `undefined`. For example, `open` and `value` in `Dropdown`.

### Expected Result

These values should be treated as if they are undefined/do not exist.

### Actual Result

Warning: `Dropdown prop "open" is auto controlled. Specify either defaultOpen or open, but not both.`

### Version
0.68.0

### Testcase

http://codepen.io/anon/pen/xdGPxz?editors=1111

### Suggested Fix

Here in [AutoControlledComponent.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/4ca4d03d0094d52489c8d492f0214f22e12dbe04/src/lib/AutoControlledComponent.js#L139), replace:

```js
  if (defaultPropName in this.props && prop in this.props) { ... }
```

by 

```js
  if (_.has(this.props, defaultPropName) 
      && !_.isUndefined(this.props[defaultPropName])  
      && _.has(this.props, prop) 
      && !_.isUndefined(this.props[props]) {
    ...
  }

jonmajorc:
fixed #1586 with the recommended change by @levithomason .

### DIFF
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -136,7 +136,7 @@ export default class AutoControlledComponent extends Component {
         const defaultPropName = getDefaultPropName(prop)
         const { name } = this.constructor
         // prevent defaultFoo={} along side foo={}
-        if (defaultPropName in this.props && prop in this.props) {
+        if (!_.isUndefined(this.props[defaultPropName]) && !_.isUndefined(this.props[prop])) {
           console.error(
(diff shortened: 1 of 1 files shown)


levithomason:
Perfect, thanks!

jonmajorc:
Thanks for Merging! This was my first open source PR

PR #1619, opened 2017-04-24
feat(Dropdown): add DropdownSearchInput component

layershifter:
Replaces #1481, discussion is [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/1481#pullrequestreview-33743808).


layershifter (commented):
@levithomason we are ready for review there :smile:

layershifter (commented):
I've rebased to `master` and fixed lint issue. Ready for review

layershifter (commented):
PR was updated to `master`,  merge conflicts were solved.

src/modules/Dropdown/DropdownSearchInput.js
```diff
@@ -0,0 +1,114 @@
...
+    type: PropTypes.string,
+
+    /** Stored value. */
+    value: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.string,
+    ]),
+
+    /** A width of an input. */
+    width: PropTypes.oneOfType([
```
levithomason: I would prefer we just use the `width` property within the `style` prop instead of breaking it out into a separate prop.  This way, all style props are always within `style` and there is no guessing about which style props are top-level props.

layershifter: Thanks for the feedback, will do :+1:

layershifter (commented):
@levithomason I've performed requested changes, `width` and `style` props were removed, unnecessary tests, too :+1:

levithomason:

PR #1622, opened 2017-04-24
fix typing failures with 0.68.2

seryl:
I noticed we had a couple of changes to this recently, this fixes the following issues:

```bash
../semantic-ui-react/dist/commonjs/views/Statistic/StatisticGroup.d.ts
(4,35): error TS2307: Cannot find module './Statictic'.
```

```bash
../semantic-ui-react/dist/commonjs/collections/Form/FormRadio.d.ts
(6,18): error TS2320: Interface 'FormRadioProps' cannot simultaneously extend types 'FormFieldProps' and 'RadioProps'.
  Named property 'type' of types 'FormFieldProps' and 'RadioProps' are not identical.
```

```bash
../semantic-ui-react/dist/commonjs/collections/Form/FormCheckbox.d.ts
(6,18): error TS2320: Interface 'FormCheckboxProps' cannot simultaneously extend types 'FormFieldProps' and 'CheckboxProps'.
  Named property 'type' of types 'FormFieldProps' and 'CheckboxProps' are not identical.
```

I've been catching all of these by just bundling with webpack, think it'd be worth adding that to the testsuite? I might add some extra time but it's been pretty reliable (and fast) at identifying these issues.

Or should we add cases for the collections (Form and friends) like #1611 did for Modal?

### DIFF
--- a/src/addons/Radio/Radio.d.ts
+++ b/src/addons/Radio/Radio.d.ts
@@ -11,7 +11,7 @@ export interface RadioProps extends CheckboxProps {
   toggle?: boolean;
 
   /** HTML input type, either checkbox or radio. */
-  type?: 'checkbox' | 'radio';
+  type?: string;
 }
 
 declare const Radio: React.StatelessComponent<RadioProps>;
--- a/src/collections/Form/FormCheckbox.d.ts
+++ b/src/collections/Form/FormCheckbox.d.ts
@@ -4,7 +4,6 @@ import { CheckboxProps } from '../../modules/Checkbox';
 import { FormFieldProps } from './FormField';
 
 export interface FormCheckboxProps extends FormFieldProps, CheckboxProps {
-  type?: 'checkbox' | 'radio';
   [key: string]: any;
 
   /** An element type to render as (string or function). */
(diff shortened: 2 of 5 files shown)


levithomason:
I completely agree on the tests, let's get that into this PR if we can.  I'd rather get tests proving it works before shipping more fixes that may or may not fix everything.

seryl:
Alright, so after looking into it more we actually have reasonable tests, but it means we need to be careful about how we do extensions.

So in this example, we extend from two pre-defined and relatively constrained constructs.

Since the current validator just checks a 1:1 mapping of typescript interfaces to proptypes, we basically have to match the same structure that we have in javascript.

So; unfortunately for compile-time type validation, we can't be lazy with the definitions, we'd have to have a third type which would break that test pattern we've put together.

We have to choose here whether we want to (as we currently do for everything else) match mostly 1:1 and relax the typescript compile-time enforcement, or, we have to revisit how that test is done entirely to allow this extra set of cases.

For now, I suggest we just relax these constraints; I can follow-up in a bit with another commit adding a type-script compiler to test that the thing compiles -- this should catch the errors that we have above like the `Statictic` mispellings.

seryl:
I relaxed the constraints on `checkbox` and `radio` to match `formfield` constraints so that when we extend from a combo of either:

* checkbox + formfield

or

* radio + formfield

we can merge the types since they are the same (weaker) constraints.

seryl:
Closing in preference of #1624 and #1625

layershifter:
@seryl I don't have time to leave a feedback there. Thanks for PR and suggestions :+1:

I'll be glad to see new PRs that adds more tests and makes typings more structured and actual.

Issue #1621, opened 2017-04-24
TypeScript compile errors


levithomason:
The latest bug fix broke further typings.  Ref #1612.

![https://cloud.githubusercontent.com/assets/11244106/25345192/b18c5924-2914-11e7-99e9-43ed3ad15da5.png](https://cloud.githubusercontent.com/assets/11244106/25345192/b18c5924-2914-11e7-99e9-43ed3ad15da5.png)

mihai-dinculescu:
Yep, I get the same errors.

layershifter:
@levithomason Working on a new PR

layershifter:
I've opened #1624, however, we still need rendering tests there

noah79:
Same.

layershifter:
@levithomason There is #1624 and #1625 that adds new tests that validate typings. We need your attention.

vcardins:
Not ideal, but for the time being I hacked by doing this:
```js
loaders: ['react-hot-loader/webpack', `awesome-typescript-loader?${JSON.stringify({ignoreDiagnostics:[2320, 2307]})}`],
```

tomitrescak:
Thanks for the hard work! Looking forward to the fix!

PR #1624, opened 2017-04-24
test(typings): additional tests for typings and fixes

layershifter:
Ref #1621

This PR adds two features:
- adds test for `[key: string]: any;` signature in `ComponentProps`
- adds `ignoredProps` option, with can skip props from typings (like in `FormRadio.d.ts`)

Also, fixes known issues and adds tests for `Portal` and `Popup` like in a `Modal` component

### DIFF
--- a/test/specs/collections/Form/FormFieldCheckbox-test.js
+++ b/test/specs/collections/Form/FormFieldCheckbox-test.js
@@ -5,7 +5,9 @@ import Checkbox from 'src/modules/Checkbox/Checkbox'
 import * as common from 'test/specs/commonTests'
 
 describe('FormCheckbox', () => {
-  common.isConformant(FormCheckbox)
+  common.isConformant(FormCheckbox, {
+    ignoredProps: ['type'],
+  })
 
   it('renders a FormField with a Checkbox control', () => {
     shallow(<FormCheckbox />)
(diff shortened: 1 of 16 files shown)


test/specs/collections/Form/FormFieldCheckbox-test.js
```diff
@@ -5,7 +5,9 @@ import Checkbox from 'src/modules/Checkbox/Checkbox'
 import * as common from 'test/specs/commonTests'
 
 describe('FormCheckbox', () => {
-  common.isConformant(FormCheckbox)
+  common.isConformant(FormCheckbox, {
+    ignoredProps: ['type'],
```
levithomason: Nit: given that the conformance test covers many things, can we rename this option to specify that it only ignores the prop in typings?

Perhaps `ignoreTypingsProps` or similar?

levithomason:
Just one comment regarding the option name.

test/specs/collections/Form/FormFieldCheckbox-test.js
layershifter: Yes, it's a good suggestion. I will fix it in evening

layershifter (commented):
@levithomason I've updated option name, so I think that we're ready to go 🚌

levithomason:

PR #1626, opened 2017-04-25
docs(README): add React-Semantic.UI-Starter

layershifter:
Add link to [react-semantic.ui-starter](https://github.com/Metnew/react-semantic.ui-starter)


levithomason:
I had considered this previously, but the project had some inconsistencies I'd like to have ironed out.  I suppose it is better to add it now and work those out as we go.

PR #1627, opened 2017-04-25
Fix statistic typings import

Jameskmonger:


layershifter:
Duplicate, will be fixed by #1624

Jameskmonger:
Thanks, do you know when that will go live?

layershifter:
I think it will be today after review

PR #1635, opened 2017-04-27
fix(ComponentExample): typo "loose" to "lose"

mikelyons:
Just for the pedants out there.


levithomason:
Thanks!

levithomason:

PR #1638, opened 2017-04-27
fix(Search): fix handling of fluid prop

### LINKED ISSUE #1632: Search: fluid search doesn't work properly
### Steps

Use `<Search fluid />`.

### Expected Result

The input gets a class of `ui fluid icon input`.

### Actual Result

The search wrapper gets a class of `ui fluid search` and the input does not become fluid.

### Version
0.67.2 and whichever is in the pen.

### Testcase
http://codepen.io/anon/pen/wdojYw

layershifter:
Fixes #1632

#### Before
image

#### After

image

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -516,11 +516,12 @@ export default class Search extends Component {
   // ----------------------------------------
 
   renderSearchInput = rest => {
-    const { icon, input } = this.props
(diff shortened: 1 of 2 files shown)


levithomason:
The API is a bit confusing at first, but there is actually no bug here.  See https://github.com/Semantic-Org/Semantic-UI-React/issues/1632#issuecomment-297906761.

PR #1639, opened 2017-04-28
fix(Dropdown): fix key handling for options

### LINKED ISSUE #1634: Dropdown: Should work with keys instead of values
The Dropdown component is working with each option's value internally. Whereas using the key, if supplied, would be correct.

### Steps
Throw a set of options into a Dropdown with two identical values.

### Expected Result
Show both options.

### Actual Result
1. Only the first option will be displayed.
2. Additionally a React warning is thrown in console: "Warning: flattenChildren(...): Encountered two children with the same key..."

### Version
0.68.1

### Testcase
http://codepen.io/karlludwigweise/pen/xdRvYL

And yes, this is a very very awkward use-case.

dvdzkwsk:
Fixes #1634

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1124,7 +1124,7 @@ export default class Dropdown extends Component {
       const defaultProps = {
         active: item.value === selectedLabel,
         as: 'a',
-        key: item.value,
+        key: item.key || item.value,
         onClick: this.handleLabelClick,
         onRemove: this.handleLabelRemove,
         value: item.value,
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1058,8 +1058,8 @@ export default class Dropdown extends Component {
     return (
       <select type='hidden' aria-hidden='true' name={name} value={value} multiple={multiple}>
         <option value='' />
-        {_.map(options, option => (
-          <option key={option.value} value={option.value}>{option.text}</option>
+        {_.map(options, (option, i) => (
+          <option key={option.key || option.value} value={option.value}>{option.text}</option>
         ))}
       </select>
     )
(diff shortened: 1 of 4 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -1058,8 +1058,8 @@ export default class Dropdown extends Component {
     return (
       <select type='hidden' aria-hidden='true' name={name} value={value} multiple={multiple}>
         <option value='' />
-        {_.map(options, option => (
-          <option key={option.value} value={option.value}>{option.text}</option>
+        {_.map(options, (option, i) => (
+          <option key={option.key || option.value} value={option.value}>{option.text}</option>
```
dvdzkwsk, line 1062: To retain parity with the generated `DropdownItem` this should likely include the `-${index}` suffix when using the value as the key. I originally included that fix in this change, but then redacted it since I'd prefer we stop using indexes in keys and require users to provide their own when values are non-unique.

levithomason: Yep.  In fact, this should use a factory which already requires users to use keys when passing objects or elements.  We use literal values for the key otherwise.

See `factories.js` where we should likely add an HTML factory for `option`:
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/factories.js#L125

src/modules/Dropdown/Dropdown.js
```diff
@@ -1152,7 +1152,7 @@ export default class Dropdown extends Component {
 
     return _.map(options, (opt, i) => (
       <DropdownItem
-        key={`${opt.value}-${i}`}
+        key={opt.key || `${opt.value}-${i}`}
```
levithomason: Again here, we should add the `create` method to the DropdownItem.  We have a factory creator for this:

**DropdownItem.js**
```jsx
Dropdown.create = createShorthandFactory(Dropdown, val => ({ text: val, value: val }))
```

Then:
```jsx
_.map(options, opt => Dropdown.create(opt)) // PLUS all the other props here...
```

Tests:
**DropdownItem-test.js**
```jsx
common.implementsCreateMethod(Dropdown)
```

dvdzkwsk: Did not know, will take a look. You mean `DropdownItem.create` though, right?

levithomason: Whoops. Yes, sir.`DropdownItem.create`.

levithomason (approved):
Looks great, thanks @davezuko

levithomason:

PR #1649, opened 2017-05-02
docs(Search): add example of input shorthand 

### LINKED ISSUE #1628: Search component doesn't accept iconPosition prop
While most of the props are spread onto Input component, including being able to pass a custom Icon - all of these making Search component very extensible, it doesn't seem to be possible to change the positioning of icon to appear on the left side instead of the right.

Is that something that could be supported? Or is there any other way to achieve that?

josie11:
Adding a simple example of passing input shorthand prop to ```Search``` mentioned in [#1628](https://github.com/Semantic-Org/Semantic-UI-React/issues/1628).

Fixes #1628

### DIFF
--- a/docs/app/Examples/modules/Search/Variations/SearchExampleInput.js
+++ b/docs/app/Examples/modules/Search/Variations/SearchExampleInput.js
@@ -1,5 +1,4 @@
 import React from 'react'
-
 import SearchExampleStandard from '../Types/SearchExampleStandard'
 
 const SearchExampleInput = () => (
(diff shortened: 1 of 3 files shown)


layershifter (approved):
@josie11 Thanks for PR! I think it's ready to merge, hovewer we need to wait for an opinion of @levithomason

levithomason:
Perhaps this could have been in `Usage`, but will ship it for now.  Thanks for the PR!

levithomason:

PR #1625, opened 2017-04-24
test(typings): add rendering test

layershifter:
Fixes #1621

Adds rendering tests for typescript typings. Now tests are failing, but #1624 will be merged, everything they will pass.

PR #1658, opened 2017-05-06
ts compiler found three problems in d.ts

smulas-pioneer:
Hello,
Didn't find a behavior to follow in the guidelines for type definitions, apoligies if this is not the correct way to contribute.


layershifter:
Dupe, #1624 and #1625

layershifter:
However, thanks for PR. Contributions are particularly welcome.

PR #1659, opened 2017-05-07
fix(docs): fix compatibility with IE

layershifter:
This fixes docs compatibility with IE.

### DIFF
--- a/docs/app/utils.js
+++ b/docs/app/utils.js
@@ -15,6 +15,14 @@ export const parentComponents = _.flow(
   _.sortBy('_meta.name')
 )(semanticUIReact)
 
+const mathSign = Math.sign || function (x) {
+  x = +x
+  if (x === 0 || isNaN(x)) {
+    return x
+  }
+  return x > 0 ? 1 : -1
+}
+
 /**
  * Get the Webpack Context for all doc site examples.
  */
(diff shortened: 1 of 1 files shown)


docs/app/utils.js
```diff
@@ -27,6 +35,7 @@ export const semanticUICSSRepoURL = 'https://github.com/Semantic-Org/Semantic-UI
 
 export const scrollToAnchor = () => {
   const anchor = location.hash && document.querySelector(location.hash)
+  const offsetY = window.scrollY || window.pageYOffset
```
layershifter, line 38: `pageYOffset` is alias for `scrollY`, but it's [supported](https://developer.mozilla.org/en/docs/Web/API/Window/pageYOffset) from IE 9

docs/app/utils.js
```diff
@@ -37,9 +46,9 @@ export const scrollToAnchor = () => {
   if (elementTop === 0) return
 
   // hit max scroll boundaries, stop
-  const isScrolledToTop = scrollY === 0
-  const isScrolledToBottom = scrollY + document.body.clientHeight === document.body.scrollHeight
-  const scrollStep = Math.ceil((Math.abs(elementTop / 8))) * Math.sign(elementTop)
+  const isScrolledToTop = offsetY === 0
+  const isScrolledToBottom = offsetY + document.body.clientHeight === document.body.scrollHeight
+  const scrollStep = Math.ceil((Math.abs(elementTop / 8))) * mathSign(elementTop)
```
layershifter, line 51: `Math.sign` is [not supported](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) in IE at all.

levithomason:
Thanks for the updates.

levithomason:

PR #1663, opened 2017-05-09
chore(es-modules): fix usage of require

### LINKED ISSUE #1238: Using rollup.js for web builds instead of webpack has unexpected results.
I observed that rollup tree shaking did not work using rollup.js and the following import:
`import Menu from 'semantic-ui-react'`
`import Header from 'semantic-ui-react'`

It also seems that rollup is breaking a subset of the mass imports where the 'react.createElement(ElementType' first parameter string was being substituted to '_default'; (for example Radio.js)  which must be a consequence of pattern
index.js: 
`import _default from './Radio';`
`export { _default as default };`

I've worked around the issue by using:
`import Menu from '../../node_modules/semantic-ui-react/dist/es/collections/Menu/Menu'`
`import Header from '../../node_modules/semantic-ui-react/dist/es/elements/Header/Header'`

instead.

It's certainly possible that this is a issue with rollup, rather than es target.  But I'm leaving the bug here to encourage investigation on why the es target does not work; since my workaround is sufficient to unblock developers, but it's still an issue that others will likely run into.

layershifter:
Fixes #1238.
Replaces #1237.

----

### What it does?

Stipes our debug functionality to `debug.cjs` and `debug.es` files, who use `require` and `import` to fetch `debug` module. Plugin for `babel` replaces import of module when build is for ES modules. This will fix our ES build for rollup and will left everything as is.

@levithomason I will be glad to hear your opinion there.

### DIFF
--- a/build/preset-es2015.js
+++ b/build/preset-es2015.js
@@ -1,6 +1,13 @@
-const options = process.env.BABEL_ENV === 'es' ? { modules: false } : {}
+let options = {}
+let plugins = []
+
+if (process.env.BABEL_ENV === 'es') {
+  options = { modules: false }
+  plugins = [['transform-rename-import', { original: './debug.cjs', replacement: './debug.es' }]]
+}
 
 module.exports = {
+  plugins,
   presets: [
     ['es2015', options],
   ],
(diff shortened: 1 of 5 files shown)


levithomason:
Though this fixes the issue, it adds quite a bit of complexity for one module.

We use the debugger mostly for development on the library itself, I don't think it is used my by others during development.  What if we instead just strip these statements entirely from the built version?  Perhaps a fork of [babel-plugin-transform-remove-console](https://www.npmjs.com/package/babel-plugin-transform-remove-console) would allow us to do this simply?  Possibly also [babel-plugin-remove-code](https://github.com/Sendoushi/babel-plugin-remove-code) could offer some tips.

PR #1678, opened 2017-05-14
docs(types): fix type definitions

### LINKED ISSUE #1563: Docs -> Form -> Props -> Size property - Showing lodash 'without' function
### Steps
Go to Form documentation page. Look at Props table, size property.

### Expected Result
Show the all possible sizes.

### Actual Result
A lodash function: _.without(...)

image

### Version
0.67.2

I would love to work on this. It seems to affect another pages as well. Can I have a try? Someone can point me to the right direction?

Thanks.

layershifter:
Fixes #1563.

### Before

image

### After

image

### DIFF
--- a/gulp/plugins/util/parseType.js
+++ b/gulp/plugins/util/parseType.js
@@ -2,16 +2,11 @@ require('babel-register')
 const _ = require('lodash')
 const SUI = require('../../../src/lib/SUI') // eslint-disable-line no-unused-vars
 
-const evalValue = (values) => _.uniq(eval(values)).map(value => ({ // eslint-disable-line no-eval
-  computed: true,
-  value: `'${value}'`,
-}))
+const evalValue = value => _.uniq(eval(value)) // eslint-disable-line no-eval
 
-const transformValues = (items) => _.flatMap(items, item => {
-  const { value } = item
-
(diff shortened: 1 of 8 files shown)


gulp/plugins/util/parseType.js
```diff
@@ -0,0 +1,35 @@
...
+}))
+
+const transformValues = (items) => _.flatMap(items, item => {
+  const { value } = item
+
+  if (_.startsWith(value, '...SUI')) return evalValue(value.substring(3))
+  return item
+})
+
+const parseEnum = type => {
```
levithomason: It looks like this reduces the need for some of the logic in `ComponentProps.js`:

[expandEnums()](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Components/ComponentDoc/ComponentProps.js#L116)
[renderEnums()](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Components/ComponentDoc/ComponentProps.js#L124)

layershifter (commented):
I've updated parser functions. `renderEnums` method was simpilified, too :+1:

levithomason:
Awesome, thanks much sir!  This will also help with the speed a bit, there's a pretty bad performance bug I need to trace down in the component doc page.

levithomason:

Issue #1647, opened 2017-05-01
TextArea: add minHeight property with autoHeight


rijk:
It would be cool if we'd have a `minHeight` property to combine with the `autoHeight` property. This way, we can make a textarea to start at e.g. 4 rows, and grow from there.

Right now, this is impossible, due to the inline styles applied in the [updateHeight](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/addons/TextArea/TextArea.js#L85-L89) method.

droms:
It would be my second contribution but I would like to give it a go.

Let me know if thats Ok and i'm the right track with the below change.

A prop minHeight should be added with a default value of zero and `this.ref.style.minHeight = '0'` in the method mention above should be changed to `this.ref.style.minHeight = minHeight`

levithomason:
Merged in #1679

PR #1679, opened 2017-05-14
feat(TextArea): add minHeight property, docs example

josie11:
Adds minHeight property to TextArea that can be used with autoHeight property to specify a minimum number of rows a TextArea should have. Referenced here: [#1647](https://github.com/Semantic-Org/Semantic-UI-React/issues/1647).

I've implemented this by setting ```this.ref.rows = minHeight```, but I'm not sure if this is the ideal way. I'd have to adjust ```assertHeight``` in the [tests](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/addons/TextArea/TextArea-test.js#L68) in order to add a test for it.


src/addons/TextArea/TextArea.js
```diff
@@ -75,14 +79,14 @@ class TextArea extends Component {
...
-    const { autoHeight } = this.props
+    const { autoHeight, minHeight } = this.props
     if (!autoHeight) return
 
     let { borderTopWidth, borderBottomWidth } = window.getComputedStyle(this.ref)
     borderTopWidth = parseInt(borderTopWidth, 10)
     borderBottomWidth = parseInt(borderBottomWidth, 10)
 
-    this.ref.rows = '1'
+    this.ref.rows = minHeight
```
layershifter: `minHeight` prop should affect next line,  `rows` should be also customizable:

```diff
static defaultProps = {
   as: 'textarea',
+  minHeight: 0,
+  rows: 5,
}
```

```diff
-this.ref.rows = minHeight
-this.ref.style.minHeight = '0'
+this.ref.style.minHeight = minHeight + 'px'
```

```diff
render() {
-  const { value } = this.props
+  const { rows, value } = this.props
// ...
- return <ElementType{...rest} onChange={this.handleChange} ref={this.handleRef} value={value} />
+  return (
+    <ElementType
+      {...rest} 
+      onChange={this.handleChange}
+      ref={this.handleRef}
+      rows={rows}
+      value={value}
+    />
+  )
}
```

layershifter (changes requested):
Thanks for PR. I think that `rows` should be also customizable:

```diff
static defaultProps = {
   as: 'textarea',
+  minHeight: 0,
+  rows: 5,
}
```

```diff
-this.ref.rows = minHeight
-this.ref.style.minHeight = '0'
+this.ref.style.minHeight = minHeight + 'px'
```

```diff
render() {
-  const { value } = this.props
+  const { rows, value } = this.props
// ...
- return <ElementType{...rest} onChange={this.handleChange} ref={this.handleRef} value={value} />
+  return (
+    <ElementType
+      {...rest} 
+      onChange={this.handleChange}
+      ref={this.handleRef}
+      rows={rows}
+      value={value}
+    />
+  )
}
```

josie11:
I added the ```rows``` prop but I left it in ```updateHeight``` & usable with ```autoHeight``` for now - otherwise it will break tests. Also, when using autoHeight it seems the intended default rows should be 1, so there is a conflict there.

layershifter (approved):
I've pushed some changes there, LGTM.

However, there is one breaking change with `autoHeight`. Before these changes `rows` prop had `1` value, now it's has `3` by default.

levithomason:
Reviewing the code changes, it strikes me odd that we're pulling `autoHeight` up to a prop.  Perhaps it makes more sense to leave this in the style prop:

```jsx
<TextArea style={{ minHeight: 100 }} />
```

layershifter:
@levithomason Thanks for the feedback, I'll make these changes.

layershifter (approved):
I've performed there some changes. `minHeight` was back to style :+1:

levithomason (approved):
Thanks much @josie11 and @layershifter!

levithomason:

PR #1680, opened 2017-05-16
feat(typings): expose FormComponent in typings

tomitrescak:
In order to generate typings for custom controls based on your react controls a FormControl needs to be exposed. This is only a change to typings.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,7 +16,7 @@ export {
   BreadcrumbSectionProps
 } from './dist/commonjs/collections/Breadcrumb/BreadcrumbSection';
 
-export { default as Form, FormProps } from './dist/commonjs/collections/Form';
+export { default as Form, FormProps, FormComponent } from './dist/commonjs/collections/Form';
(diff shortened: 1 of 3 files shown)


layershifter:
Can you provide a code example? I want to understand your use case.

levithomason:
Bump

tomitrescak:
I have created several mappings of semanti-ui-react components to front-end frameworks. The latest one is "semantic-ui-mobx" that uses mobx-state-tree functionality to map form elements to app state. This gives you the smae control as redux store with browsing history and more.

Here is the file: https://github.com/tomitrescak/semantic-ui-mobx/blob/master/src/index.tsx

To make these components type safe, I needed those exposed properties.

layershifter:
@tomitrescak Thanks for the feedback. I think we can merge this, possible we will need expose all components in future.

tomitrescak:
Thanks!

levithomason:

Issue #1664, opened 2017-05-09
Search: invalid PropType noResultMessage


jptissot:
### Steps
When using the Search component, I want to be able to pass a react component as a noResultMessage to allow for localization / links. This works as expected (It displays fine) but i get a proptype warning in the console.
### Expected Result
No warning in console
### Actual Result
 Invalid prop `noResultsMessage` of type `object` supplied to `Search`, expected `string`.
### Version
0.68.3

### Testcase

```
<Search
  noResultsMessage={<FormattedMessage {...messages.NoSurveysFound} values={{ link: <Link to="/" onClick={() => clear()}>{formatMessage(messages.ViewAllSurveys)}</Link> }} />}
/>
```

jptissot:
Just a side note: I did try this 

``` JavaScript
<Search
  noResultsMessage={formatMessage(messages.NoSurveysFound, {link: <Link to="/" onClick={() => clear()}>{formatMessage(messages.ViewAllSurveys)}</Link> })}
/>
```

Doing this makes the `<Link />` element display as `Object [object]` instead of displaying a link.

levithomason:
I believe we made the same updates to the Dropdown.  We simply need to relax the prop type validation here.  PRs welcome.

levithomason:
Fixed in #1683

lazazivkovic:
> I believe we made the same updates to the Dropdown. We simply need to relax the prop type validation here. PRs welcome.

@levithomason Could you please take a look at #3177 ? Thanks 😃

PR #1683, opened 2017-05-16
feat(Search): noResultsMessage prop can be a node

b0gok:
PR for this issue https://github.com/Semantic-Org/Semantic-UI-React/issues/1664

### DIFF
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -27,7 +27,7 @@ export interface SearchProps {
   minCharacters?: number;
 
   /** Additional text for "No Results" message with less emphasis. */
-  noResultsDescription?: string;
+  noResultsDescription?: React.ReactNode;
 
   /** Message to display when there are no results. */
   noResultsMessage?: React.ReactNode;
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -30,7 +30,7 @@ export interface SearchProps {
   noResultsDescription?: string;
 
   /** Message to display when there are no results. */
-  noResultsMessage?: string;
+  noResultsMessage?: React.ReactNode;
 
   /** Controls whether or not the results menu is displayed. */
   open?: boolean;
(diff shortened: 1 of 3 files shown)


src/modules/Search/Search.js
```diff
@@ -57,7 +57,10 @@ export default class Search extends Component {
     noResultsDescription: PropTypes.string,
 
     /** Message to display when there are no results. */
-    noResultsMessage: PropTypes.string,
+    noResultsMessage: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.object,
+    ]),
```
levithomason: I believe this should just be `PropTypes.node` only.  `PropTypes.node` includes any renderable value: string, number, element, or an array (or fragment) of any of these.  However, objects are not renderable.

Let's also update the typings in `Search.d.ts` to reflect the type change as well.

levithomason (changes requested):
Just one comment regarding propTypes and typings.

src/modules/Search/Search.js
b0gok: > PropTypes.node includes any renderable value: string, number, element, or an array (or fragment) of any of these. 

Didn't know about it. Thanks for useful information.
Made the necessary changes.

src/modules/Search/Search.d.ts
```diff
@@ -30,7 +30,7 @@ export interface SearchProps {
   noResultsDescription?: string;
```
layershifter: @b0gok Thanks for a PR. I think that we need make same changes for `noResultsDescription`.

layershifter (changes requested):

src/modules/Search/Search.d.ts
b0gok: @layershifter Should I add specific test for case when noResultsDescription isn't a string?

layershifter: Yes, it will be awesome.

levithomason (approved):

levithomason:
Thank you for the PR and updates!

levithomason:

PR #1690, opened 2017-05-21
feat(Dropdown): Add minCharacters prop

### LINKED ISSUE #1124: minCharacters and saveRemoteData for Dropdown setting missing?
Hello,
is `minCharacters` and/or `saveRemoteData` property implemented for React Semantic UI Dropdown element? Can't be found in documentation and error such as this appears:

> **"Warning: Unknown prop `saveRemoteData` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop**

dyesseyumba:
I've added minCharacters props in Dropdown as requested in #1124. Now we can set the minimum characters for a search to begin showing results.

I also would like to add saveRemoteData but I'm wondered if it's not useless. Because someone will use Flux' store and someone else the Redux' store and the other the locale storage. 

So should I use locale storage to implement saveRemoteData?

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -712,13 +712,13 @@ export default class Dropdown extends Component {
     debug(e.target.value)
     // prevent propagating to this.props.onChange()
     e.stopPropagation()
-    const { search, onSearchChange } = this.props
+    const { search, onSearchChange, minCharacters } = this.props
     const { open } = this.state
     const newQuery = e.target.value
 
     if (onSearchChange) onSearchChange(e, newQuery)
 
-    if (newQuery.length >= this.props.minCharacters) {
+    if (newQuery.length >= minCharacters) {
       // open search dropdown on search query
       if (search && newQuery && !open) this.open()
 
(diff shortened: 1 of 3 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -714,13 +718,15 @@ export default class Dropdown extends Component {
 
     if (onSearchChange) onSearchChange(e, newQuery)
 
-    // open search dropdown on search query
-    if (search && newQuery && !open) this.open()
+    if (newQuery.length >= this.props.minCharacters) {
```
levithomason: In keeping with the code style, please destructure `minCharacters` on line 715.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1604,6 +1604,19 @@ describe('Dropdown', () => {
       dropdownMenuIsOpen()
     })
 
+    it('Don\'t open the menu on change if query\'s length is less than minCharacters', () => {
```
levithomason: When a string includes single quote characters, please quote it with double quotes to prevent escaping.

levithomason (approved):
Changes look great.  Just a couple code style comments for addressing please.

levithomason:
Regarding `saveRemoteData`, I agree completely.  I do not think this prop is applicable in React apps.  I think we just skip this prop entirely.  The best practice today is to use some data flow solution, such as Redux.  Almost all of these solutions have some form of interoperability with localStorage.

levithomason:
Thanks much, will merge on pass.

dyesseyumba:
You're welcome

levithomason:

PR #1693, opened 2017-05-22
fix(Button): fix handling of icon class

### LINKED ISSUE #1691: Button: Icon button has different width if content is 0 (number)
When icon button content is `0` (number), it's rendered with different width. Should be the same as with `'0'` (string).

### Steps
```javascript
  <Button icon='world' content={0} />
  <Button icon='world' content='0' />
```
### Expected Result
expected

### Actual Result
actual

### Version
0.68.3

### Testcase
[codepen](http://codepen.io/kirillku/pen/gWZYaM)

layershifter:
Fixes #1691.

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -186,6 +186,13 @@ class Button extends Component {
     if (onClick) onClick(e, this.props)
   }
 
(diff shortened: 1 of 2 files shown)


levithomason:

PR #1694, opened 2017-05-22
fix(Checkbox): fix value prop types

fabien0102:
Just a little fix, `PropTypes` doesn't match typescript definition into Checkbox.js.

```typescript
// Checkbox.d.ts:79

/** The HTML input value. */
  value?: number|string;
```


layershifter (approved):
I fixed lint issues. LGTM, thanks for PR

levithomason:

PR #1695, opened 2017-05-22
docs(Dropdown): fix world icon in search example

aabustamante:
the example have a note where it says that needs a world icon


aabustamante:
I renamed "languageOptions" to "languages" because the row had more than 120 characters and tests will fail

docs/app/Examples/modules/Dropdown/common.js
```diff
@@ -337,7 +337,7 @@ export const countryOptions = [
   { key: 'zw', value: 'zw', flag: 'zw', text: 'Zimbabwe' },
 ]
 
-export const languageOptions = [
+export const languages = [
```
levithomason: For consistency with all the other exports in this file, let's restore this const name.

levithomason (approved):
Just one comment regarding naming conventions.

docs/app/Examples/modules/Dropdown/Types/DropdownExampleSearchDropdown.js
```diff
@@ -4,15 +4,9 @@ import { Dropdown } from 'semantic-ui-react'
...
- * NOTE: In this example, the dropdown should contain a label on the
- * left of the text:
- * <i class="world icon"></i>
- *
- * @returns {Element}
- */
 const DropdownExampleSearchDropdown = () => (
-  <Dropdown text='Select Language' search floating labeled button className='icon' options={languageOptions} />
+  <Dropdown text='Select Language' 
+  search floating labeled button className='icon' icon='world' options={languageOptions} />
```
levithomason: If props need to be on separate lines, one prop per line please.  The closing tag should also be on its own line.

```jsx
<Dropdown
  text='Select Language'
  search
  floating
  labeled
  button
  className='icon'
  icon='world'
  options={languageOptions}
/>
```

aabustamante: ok, thanks. I will change it.

aabustamante: @levithomason  thanks for helping me. I already made the changes you suggested me and now test passes successfully

levithomason:
Thanks for the update.

levithomason:

PR #1699, opened 2017-05-23
docs(Button): remove redundant prop in Vertical Group example

### LINKED ISSUE #1697: Button: redundant prop in Vertical Group example
**Steps to Reproduce**
1. Look at the docs
2. https://react.semantic-ui.com/elements/button#button-example-group-vertical

**Expected**
The last button (bottom) from the vertical group should have the top-left corner not rounded.

**Result**
Instead, it is rounded.

**Testcase**
https://react.semantic-ui.com/elements/button#button-example-group-vertical

**Possible fix**
The problem seams to be related to the fact that the parent has the classes `top attached`.
Those classes are used apparently by the __Vertically Attached__ examples.
If you remove `attached='top'` from the example, it looks fine.

andreipfeiffer:
Fixes #1697.


levithomason:
Thanks for the report and fix.

levithomason:

PR #1706, opened 2017-05-26
feat(Item): add unstackable prop to ItemGroup

layershifter:
Rel: [#2901](https://github.com/Semantic-Org/Semantic-UI/issues/2901)

Adds missing `unstackable` prop.


levithomason:
Thanks for the careful eye.

levithomason:

Issue #1710, opened 2017-05-28
Dropdown: items missing "key" prop, React warning displayed


mcrawshaw:
### Steps
 - Update from version 0.68.3 to 0.68.4.
 - Introduce a ```Dropdown``` element with more than one ```option``` specified.  ```selection``` needs to be specified also.
 - Run.

### Expected Result
No warning from React.

### Actual Result
React error:
``` Warning: Each child in an array or iterator should have a unique "key" props.  Check the render method of 'Dropdown'.```

### Version
0.68.4

**Comment by mcrawshaw** [comment]
Note, adding a ```key``` to each option fixes the issue.  For example:
```
<Dropdown selection options={[key: "1", value: "1", name: "Option 1"]} />
```
This wasn't the case for 0.68.3.  I see a reference to ```key``` in the docs, but it is not consistently specified in the examples.

mcrawshaw:
Looks to be an intended change.  This issue is mentioned in the latest release documentation:
https://github.com/Semantic-Org/Semantic-UI-React/issues/1634

So does this mean we need ```key```, ```value```, and ```text``` now?

@karlludwigweise @davezuko

dvdzkwsk:
From my understanding, `key` is definitely the _preferred_ way, but it should have been nearly backwards compatible due to [createShorthandFactory](https://github.com/Semantic-Org/Semantic-UI-React/blob/754416e7b37b260fcb625463afc6bcae5a41475e/src/lib/factories.js) using `value` as the default if no key exists.

@levithomason I am, however, curious about this line: https://github.com/Semantic-Org/Semantic-UI-React/blob/754416e7b37b260fcb625463afc6bcae5a41475e/src/lib/factories.js#L87. This looks to be checking the wrong value, at least from my initial read through of the code. Are we sure that `value` is being used correctly as the default? Maybe I missed something in the DropdownItem change?

jhartma:
I have a small problem, though: react throws the unique key prop error when ```allowAdditions``` is defined and I am not sure how to add a key prop there. Seems like the ```Add``` entry has no key?

karlludwigweise:
This is the reason I did not want to make the change myself. The Dropdown is too big of a component to oversee for a first time change; for me at least. The component should ideally generate its own keys and use those.

mcrawshaw:
@karlludwigweise, no stress, we have a work around.

Let's see what @levithomason says about the factory before we continue.

layershifter:
[...]

PR #1727, opened 2017-06-01
fix(Dropdown): add addition item key

levithomason:
Per https://github.com/Semantic-Org/Semantic-UI-React/issues/1710#issuecomment-304596090, we forgot to add a `key` to the addition item before pushing it into the filtered options.


levithomason:

PR #1729, opened 2017-06-01
fix(factories): handle falsy `key` values

### LINKED ISSUE #1710: Dropdown: items missing "key" prop, React warning displayed

### LINKED ISSUE #1720: Dropdown: option keys fallback to values incorrectly
### Steps

if we define our `Dropdown` component like this:

```jsx
const opts = [
  {
    key: 0,
    value: 1,
    text: 'Foo',
  },
  {
    key: 1,
    value: 2,
    text: 'Bar',
  },
]

const MyDropdown = () => <Dropdown selection options={opts} />
```

*Note*: this only happens with `selection` prop `true`, because only in that case, `select` is rendered behind the scenes

### Expected Result

This should render `<select>` with two `<option>`s with keys `0` and `1`

### Actual Result
image

This tries to render `<select>` with two `<option>`s with keys `1` and `1` because of `{ option.key || option.value }` (see this line of the code https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L1074)

I think this line should be rewritten to only fallback to `option.value` when `option.key` is either `undefined` or empty string. All other falsy values are valid keys, react converts them to strings.

### Version
last

### Testcase
https://codepen.io/notgiorgi/pen/ZKgBob?editors=0010 (you cannot use devtools on codepen)

layershifter:
Fixes #1710.
Fixes #1720.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -25,6 +25,8 @@ import DropdownMenu from './DropdownMenu'
 
 const debug = makeDebugger('dropdown')
 
+const getKeyOrValue = (key, value) => _.isNil(key) ? value : key
+
 /**
  * A dropdown allows a user to select a value from a series of options.
  * @see Form
@@ -1072,7 +1074,7 @@ export default class Dropdown extends Component {
       <select type='hidden' aria-hidden='true' name={name} value={value} multiple={multiple}>
         <option value='' />
         {_.map(options, (option, i) => (
-          <option key={option.key || option.value} value={option.value}>{option.text}</option>
+          <option key={getKeyOrValue(option.key, option.value)} value={option.value}>{option.text}</option>
         ))}
       </select>
     )
@@ -1137,7 +1139,7 @@ export default class Dropdown extends Component {
       const defaultProps = {
         active: item.value === selectedLabel,
         as: 'a',
-        key: item.key || item.value,
+        key: getKeyOrValue(item.key, item.value),
         onClick: this.handleLabelClick,
         onRemove: this.handleLabelRemove,
         value: item.value,
@@ -1168,6 +1170,7 @@ export default class Dropdown extends Component {
       onClick: this.handleItemClick,
       selected: selectedIndex === i,
       ...opt,
+      key: getKeyOrValue(key, value),
       // Needed for handling click events on disabled items
       style: { ...opt.style, pointerEvents: 'all' },
     }))
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1170,7 +1170,7 @@ export default class Dropdown extends Component {
       onClick: this.handleItemClick,
       selected: selectedIndex === i,
       ...opt,
-      key: getKeyOrValue(key, value),
+      key: getKeyOrValue(opt.key, opt.value),
       // Needed for handling click events on disabled items
       style: { ...opt.style, pointerEvents: 'all' },
     }))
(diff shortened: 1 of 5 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -1071,7 +1073,7 @@ export default class Dropdown extends Component {
       <select type='hidden' aria-hidden='true' name={name} value={value} multiple={multiple}>
         <option value='' />
         {_.map(options, (option, i) => (
-          <option key={option.key || option.value} value={option.value}>{option.text}</option>
+          <option key={getKeyOrValue(option.key, option.value)} value={option.value}>{option.text}</option>
```
layershifter, line 1077: As described in #1720, `0` will break the previous condition while it's a valid key.

src/modules/Dropdown/Dropdown.js
```diff
@@ -1136,7 +1138,7 @@ export default class Dropdown extends Component {
       const defaultProps = {
         active: item.value === selectedLabel,
         as: 'a',
-        key: item.key || item.value,
+        key: getKeyOrValue(item.key, item.value),
```
layershifter, line 1142: ^

src/modules/Dropdown/Dropdown.js
```diff
@@ -1167,6 +1169,7 @@ export default class Dropdown extends Component {
       onClick: this.handleItemClick,
       selected: selectedIndex === i,
       ...opt,
+      key: getKeyOrValue(opt.key, opt.value),
```
layershifter, line 1173: ## Problem

Before #1639, `key` [was captured](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.68.3/src/modules/Dropdown/Dropdown.js#L1155) from `value`. In #1639 we refactored this to factory usage, but as we [know](https://github.com/Semantic-Org/Semantic-UI-React/issues/1710#issuecomment-304513318) factory will not create `key` automatically when it recieves an object and it's expected behaviour.

But, it's **breaking** change that I think should be reverted, because our users rely on that `key` can be automatically generated from a `value` prop (#1710).

levithomason: You are correct, however, this is a semver fail on my part.  I should have incremented the minor (since we're pre 1.0).

I do not think we should ship this feature moving forward.  In `1.0` we will not automatically generate keys when using shorthand objects or elements.  Users will have to pass a `key` manually.  If we add this now, it will encourage users to rely on `value` being used as the `key` (at least in the Dropdown only).

If we feel very strongly about fixing this regression, I am willing to ship this as a _patch_ right now to take care of current users as long as we immediately follow it up with a `breaking(): ...` PR that removes it for `0.69.0`.

levithomason:
I've left https://github.com/Semantic-Org/Semantic-UI-React/pull/1729#discussion_r119653475, I'd like your thoughts there before merging this.

levithomason:
After reading up more on the other issues, I'll update this PR to fix the factory instead.  See https://github.com/Semantic-Org/Semantic-UI-React/issues/1710#issuecomment-305540320.

levithomason:

src/modules/Dropdown/Dropdown.js
layershifter: Yes, I completely agree :+1:  Let's ship this fix as `0.68.5` and remove this line in `0.69.0`, and mark PR that makes this change as `breaking`.

PR #1730, opened 2017-06-01
breaking(Dropdown): remove hidden select

levithomason:
# BREAKING

The hidden `select` element has been removed from the rendered Dropdown component.  A React application should never read nor write to the DOM, so you'll likely never know this was removed.

This was a unadvertised and nonsupported API, but we're marking it as a breaking change in case users were relying on it.  If you relied on it to "get the value" of a Dropdown, you should instead capture the value from the `data` argument in the `onChange(e, data)` handler and persist it for later access.  See the [docs](https://react.semantic-ui.com).

You can persist data in component state somewhere higher in the render tree.  You might also consider using something like redux.js.org as an app state solution.

***

Falsy factory keys were fixed in #1729, including the Dropdown's handling of keys on the hidden `option` elements. That fix will go out to 0.68.x to fix that release for users.

This PR removes the hidden `select` altogether.  It is an artifact from long ago. It will be released in `0.69.x` as **breaking** as there may be some users reading the values from the DOM, unfortunately.

### DIFF
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -1247,13 +1247,12 @@ describe('Dropdown', () => {
     })
     describe('removing items', () => {
       it('calls onChange without the clicked value', () => {
-        const name = 'my-dropdown'
         const value = _.map(options, 'value')
         const randomIndex = _.random(options.length - 1)
         const randomValue = value[randomIndex]
         const expected = _.without(value, randomValue)
         const spy = sandbox.spy()
-        wrapperMount(<Dropdown options={options} selection name={name} value={value} multiple onChange={spy} />)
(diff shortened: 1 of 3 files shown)


test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -211,10 +211,6 @@ describe('Dropdown', () => {
       wrapperMount(<Dropdown />)
       wrapper.find('div').at(0).should.have.prop('role', 'listbox')
     })
-    it('should label selection dropdown with aria-hidden=true', () => {
-      wrapperMount(<Dropdown selection />)
-      wrapper.find('select').at(0).should.have.prop('aria-hidden', 'true')
-    })
```
layershifter, line 217: @levithomason I removed this failing test, now they're passing.

levithomason:

PR #1733, opened 2017-06-02
feat(events): add eventStack

### LINKED ISSUE #1157: Multiple modals scrolling bug
`dimmable dimmed` are deleted from `body` when the second modal is closed
### Steps
1. Scroll modal
2. Open second modal
3. Scroll modal
4. Close second modal
5. Try to scroll modal

### Expected Result
The first modal should be scrollable
`<body class="dimmable dimmed scrolling">`

### Actual Result
First Modal is not longer scrollabel
`<body class="scrolling">`

### Version
tested in **0.64.0** & **0.63.5**

### Testcase
https://codepen.io/csxmedia/full/mRVgYX/

### LINKED ISSUE #1657: Modal: Enter ESC should only close the top layer modal
### Steps
1. open multiple modal
2. enter ESC

### Expected Result
only the top layer modal was closed

### Actual Result
all modals were closed

### Version
0.68.2

### Testcase
https://react.semantic-ui.com/modules/modal#modal-example-multiple

layershifter:
Rel #284.
Fixes #1657.

### What is it?

This PR introduces a new event handling system that will solve following problems:
- event priority, only top event will be emitted (will help with multiple `Modal`s and `Dropdown`s);
- solve performance problems with multiple `Dropdown`s;

### Example

```diff
-document.addEventListener('keydown', this.openOnArrow)
-document.addEventListener('keydown', this.openOnSpace)
+eventStack.sub('keydown', [this.openOnArrow, this.openOnSpace])
```

```diff
-document.removeEventListener('keydown', this.openOnArrow)
-document.removeEventListener('keydown', this.openOnSpace)
-document.removeEventListener('keydown', this.moveSelectionOnKeyDown)
-document.removeEventListener('keydown', this.selectItemOnEnter)
-document.removeEventListener('keydown', this.removeItemOnBackspace)
+eventStack.unsub('keydown', [
+  this.openOnArrow,
+  this.openOnSpace,
+  this.moveSelectionOnKeyDown,
+  this.selectItemOnEnter,
+  this.removeItemOnBackspace,
+])
```


layershifter:
@levithomason Would be awesome to hear your throughts.

However, I want to finish my other open PRs before I'll start there.

src/lib/eventPool.js
```diff
@@ -0,0 +1,63 @@
...
+  // ------------------------------------
+
+  listen = name => {
+    if(_.includes(this.queues, name)) return
+    document.addEventListener(name, this.emit(name))
+  }
+
+  unlisten = name => {
+    if(_.some(this.queues, name)) return
+    document.removeEventListener(name, this.emit(name))
```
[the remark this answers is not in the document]
layershifter: @davezuko Yep, you're fully right there. I pushed some changes and tests, now it works as I described.

src/modules/Dropdown/Dropdown.js
```diff
@@ -425,64 +422,57 @@ export default class Dropdown extends Component {
...
       }
       if (!this.state.open) {
-        document.addEventListener('keydown', this.openOnArrow)
-        document.addEventListener('keydown', this.openOnSpace)
-      } else {
-        document.addEventListener('keydown', this.moveSelectionOnKeyDown)
-        document.addEventListener('keydown', this.selectItemOnEnter)
+        eventPool.sub('keydown', [this.openOnArrow, this.openOnSpace])
       }
-      document.addEventListener('keydown', this.removeItemOnBackspace)
```
layershifter: `moveSelectionOnKeyDown`, `selectItemOnEnter` and `removeItemOnBackspace` should be added only if the `Dropdown` is open

layershifter (commented):
@davezuko @levithomason I've performed final updates there, will be awesome to get a feedback there.

levithomason:
I believe this also solves #1157, correct?

layershifter:
No, #1157 will need an another solution, however this PR opens the road to it.

src/modules/Dropdown/Dropdown.js
levithomason: `removeItemOnBackspace` works when a `multiple search selection` Dropdown is focused and `openOnFocus={false}`.  The cursor focusses the Dropdown, it does not open, but the backspace listener is added so the user can remove labels:



I'd prefer we don't change any logic in this PR except the adding/removing events.  That said, it looks like we're missing a test for this as well.  Changing that logic should have broken a test.

layershifter: @levithomason Thanks for clarification, I've restored event handlers and added a fix. Now `eventPull` [handles only](https://github.com/Semantic-Org/Semantic-UI-React/pull/1733/commits/ffa4a434373c57efc59fdc3f27d03a4b21ede4bb) unique events

levithomason:
I have not gone into this in detail yet, but I don't want to hold it up any longer so I'd like to merge it.

One nit I have is that a pool, in my experience, is a set of shared resources where any available item may be pulled form the pool (like a db connection).  I'd like to call this a "stack", which indicates they will be processed one at a time ~~in the order they were added~~ _from the end_.  Is this accurate with the implementation here?

If so, let's go with `eventStack` over `eventPool`, merge, and move forward 👍 

**EDIT** Fix stack description, thanks David.

layershifter:
I will change tomorrow and will sync with `master` :+1:

layershifter:
I merged with `master`, renamed to `eventStack` and fixed lint issues.

dvdzkwsk:
> I'd like to call this a "stack", which indicates they will be processed one at a time in the order they were added. Is this accurate with the implementation here?

I think you mean a `queue` ;) <3

kyrasteen:
Hey there. Thanks for fixing this issue. I was wondering if this will be merged and released in the near future? It would help my team out with a current production bug.  If there is anything I can contribute to get this merged, happy to help. :)

levithomason:
>I think you mean a queue ;) <3

Whoops, I still mean a stack, however, I was lazy and inaccurate in the description.  I meant to say "processed one at a time, from the end" rather than "in the order added".

When I imagine a Popup opening inside of a Modal, I think of a stack of handlers. The Popup handlers should be processed first even though they were added last.  In a queue, I'd think the Modal handlers would be processed first, then the Popup (in the order they were added).

Thanks for pointing out the error @davezuko.

levithomason:

PR #1734, opened 2017-06-03
fix(Input): add missing minLength prop

traverse:
Fixes `minLength` not being passed down to input elements by adding it to `htmlInputAttrs`.


levithomason:
Thanks much, Tom.

levithomason:

PR #1742, opened 2017-06-07
fix(Search): Allow default action if there is no selected result

nanego:
Currently, in the Search component, when there is no selected result, users cannot use the Enter key due to a misplaced call to "e.preventDefault()".

We should not stop the default action unless we actually do something.
This tiny PR will allow users to specify a custom onKeyPress action when there is no selected result.


levithomason:
Ah, nice catch.  Thanks much.

levithomason:

PR #1746, opened 2017-06-08
fix(htmlInputProps): fix handling of falsy values

### LINKED ISSUE #1743: Missing attributes in html
### Steps
```javascript
const props = {
      autoFocus:false,
      autoComplete: "username",
      disabled:false,
      error:false,
      id:"username",
      onBlur: () => {},
      onChange: () => {},
      placeholder:"",
      type:"text",
      value:"",
      name:"username"
}
<Input {...props} />
```
### Expected Result
props autofocus, name, value and some others are passed to html
### Actual Result
only type (by default I guess) passed
### Version
0.68.5

### Testcase
[CodePen](https://codepen.io/rmakarovWithPen/pen/ZybQaO)

layershifter:
Fixes #1743.

### DIFF
--- a/src/lib/htmlInputPropsUtils.js
+++ b/src/lib/htmlInputPropsUtils.js
@@ -44,7 +44,10 @@ export const partitionHTMLInputProps = (props, htmlProps = htmlInputProps) => {
   const inputProps = {}
   const rest = {}
 
-  _.forEach(props, (val, prop) => _.includes(htmlProps, prop) ? (inputProps[prop] = val) : (rest[prop] = val))
+  _.forEach(props, (val, prop) => {
+    const target = _.includes(htmlProps, prop) ? inputProps : rest
+    target[prop] = val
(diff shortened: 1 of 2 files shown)


src/lib/htmlInputPropsUtils.js
```diff
@@ -44,7 +44,10 @@ export const partitionHTMLInputProps = (props, htmlProps = htmlInputProps) => {
   const inputProps = {}
   const rest = {}
 
-  _.forEach(props, (val, prop) => _.includes(htmlProps, prop) ? (inputProps[prop] = val) : (rest[prop] = val))
```
layershifter, line 47: Statement `_.includes(htmlProps, prop) ? (inputProps[prop] = val) : (rest[prop] = val)` will return prop's value, when value will be `false` the `forEach` will stop execution.

levithomason:

PR #1747, opened 2017-06-08
feat(Sticky): add component

### LINKED ISSUE #198: <Sticky /> Component
http://semantic-ui.com/modules/sticky.html

mariolamacchia:
Fixes #198.


mariolamacchia:
@levithomason @layershifter Ready for review :)

layershifter:
@mariolamacchia Thanks for the huge work, I will review this as soon as possible

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,251 @@
...
+    )) {
+      parent = parent.parentElement
+
+      if (!parent) {
+        parent = document.body
+      }
+    }
+
+    return parent
+  }
```
layershifter: I think we can that use there [closest()](https://developer.mozilla.org/ru/docs/Web/API/Element/closest) with polyfill if we will choose a way with DOM.

layershifter (commented):
@mariolamacchia I've pushed some changes, I will continue when have time.

In general, it will be awesome to go away from the DOM, need to think about better solution.

layershifter:
```jsx
class App extends Component {
  handleContextRef = c => (this.contextRef = c)

  render() {
    return (
      <div>
        <Sticky.Context contextRef={this.handleContextRef}>
           { // .... }
          <Sticky context={this.contextRef} />
           { // ... }
        </Sticky.Context>
      </div>
    )
  }
}
```

@mariolamacchia I thought about `Sticky` on weekend and I think that need to keep away from DOM and use a more React way solution with refs. It is not so elegant, but it eliminates the need to traverse the tree. Let me know what you think about this.

mariolamacchia:
Yes that looks way better, let me implement it :)

mariolamacchia:
@layershifter I just relized... at this point we don't need the `StickyContext` component at all, right? It would just add the `ref` attribute to the children. This would be exactly the same:
```
class App extends Component {
  handleContextRef = c => (this.contextRef = c)

  render() {
    return (
      <div ref={handleContextRef}>
         { // .... }
        <Sticky context={this.contextRef} />
         { // ... }
      </div>
    )
  }
}
```

layershifter:
Yep, you're right :+1: So, `StickyContext` can be safely removed.

mariolamacchia:
@layershifter I'm done :)

docs/app/Examples/modules/Sticky/Types/StickyAdjacentContextExample.js
```diff
@@ -0,0 +1,56 @@
...
+  constructor(props) {
+    super(props)
+    this.state = { contextRef: null }
+  }
+
+  handleContextRef = contextRef => {
+    if (!this.state.contextRef) {
+      this.setState({ contextRef })
+    }
+  }
```
layershifter: @levithomason I'm not happy about this, possible there is better solution?

docs/app/Examples/modules/Sticky/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import Types from './Types'
+import Variations from './Variations'
+import Usage from './Usage'
+
+const AccordionExamples = () => (
```
levithomason: `AccordionExamples` Looks like a few of these need updated.

mariolamacchia: Oops! Sorry about that! 😅

docs/app/Examples/modules/Sticky/Types/StickyAdjacentContextExample.js
mariolamacchia: I cannot find a better way to pass the `ref` to a child. Any suggestions?

layershifter: @mariolamacchia I understand why you did this and I don't see a better way, too :cry: However, possible @levithomason knows it.

docs/app/Examples/modules/Sticky/Types/StickyAdjacentContextExample.js
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react'
+import { Grid, Header, Image, Rail, Segment, Sticky } from 'semantic-ui-react'
+
+const Placeholder = () => (
+  <Image
+    style={{ marginTop: 14, marginBottom: 14 }}
+    className='wireframe paragraph image'
+    src='/assets/images/wireframe/paragraph.png'
```
layershifter: Please sort props in an alphabetical order, eg. `className`, `src`, `style`. `style` object members, too.

docs/app/Examples/modules/Sticky/Types/StickyAdjacentContextExample.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  render() {
+    return (
+      <Grid centered columns={3}>
+        <Grid.Column>
+          <div ref={this.handleContextRef}>
+            <Segment>
+              {[...new Array(10)].map((v, i) => <Placeholder key={i} />)}
```
layershifter: Please use there `_.times` from the lodash:
```js
_.times(i, <Placeholder key={i} />)
```

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+
+  static _meta = {
+    name: 'Sticky',
+    type: META.TYPES.MODULE,
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = { sticky: false, top: null, bottom: null }
```
layershifter: Please use there ES7 property initializers:

```diff
-constructor(props) {
-  super(props)
-  this.state = { sticky: false, top: null, bottom: null }
-}
+state = { 
+ bottom: null,
+  sticky: false,
+  top: null,
+ }
```

layershifter:
@mariolamacchia Also, please solve a merge conflict, tell me if you will have problems with this.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+    this.stick(e)
+    this.setState({ top: this.props.offset, bottom: null })
+  }
+
+  stickToScreenBottom(e) {
+    this.stick(e)
+    this.setState({ top: null, bottom: this.props.bottomOffset })
+  }
+
+  update = (e) => {
```
layershifter: As `update` is an event listener, I think the better name will be `handleUpdate`.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+    })
+    this.setPushing(true)
+  }
+
+  stickToScreenTop(e) {
+    this.stick(e)
+    this.setState({ top: this.props.offset, bottom: null })
+  }
+
+  stickToScreenBottom(e) {
```
layershifter: Can we convert all non-React lifecycle methods to arrow functions?

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+    this.stick(e)
+    this.setState({ top: null, bottom: this.props.bottomOffset })
```
layershifter: Can we use there the object destuction?

```js
const { bottomOffset: bottom } = this.props

this.stick(e)
this.setState({ bottom, top: null })
```

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  stickToContextTop(e) {
+    if (e && this.props.onTop) this.props.onTop(e)
+    this.unstick(e)
+    this.setPushing(false)
+  }
+
+  stickToContextBottom(e) {
+    if (e && this.props.onBottom) this.props.onBottom(e)
```
layershifter: @mariolamacchia Can you clalify in what cases we don't have an `e` object?

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  calcBoundingRects() {
+    this.triggerBoundingRect = this.triggerRef.getBoundingClientRect()
+    this.contextBoundingRect = (this.props.context || document.body).getBoundingClientRect()
+    this.stickyBoundingRect = this.stickyRef.getBoundingClientRect()
+  }
+
+  // Return true when the component reached the bottom of the context
+  hasReachedContextBottom() {
```
layershifter: I don't think that `has` has something meaningful in the method's name, I think `reachedContextBottom` is enough clear.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,230 @@
...
+    return this.triggerBoundingRect.top < this.props.offset
+  }
+
+  // Return true when the bottom of the screen overpasses the Sticky component
+  hasTouchedScreenBottom() {
+    return this.contextBoundingRect.bottom + this.props.bottomOffset > window.innerHeight
+  }
+
+  // Return true if the height of the component is higher than the window
+  isOversized() {
```
layershifter: `isOversized` can be simply `oversized`, when it will be an arrow function, it will very clear:

```js
oversided = () => this.stickyBoundingRect.height > window.innerHeight
```

docs/app/Examples/modules/Sticky/Types/StickyAdjacentContextExample.js
```diff
@@ -0,0 +1,57 @@
+import React, { Component } from 'react'
+import { Grid, Header, Image, Rail, Segment, Sticky } from 'semantic-ui-react'
+import _ from 'lodash'
+
+const Placeholder = () => (
+  <Image
+    className='wireframe paragraph image'
```
layershifter: BTW, these classes are from SUI docs, so I don't sure that they defined in our CSS. @mariolamacchia can you check? Seems they are cruft.

test/specs/modules/Sticky/Sticky-test.js
```diff
@@ -0,0 +1,202 @@
...
+  contextEl.getBoundingClientRect = () => ({ bottom: height + 1 })
+  window.dispatchEvent(new Event('scroll'))
+}
+
+describe('Sticky', () => {
+  common.isConformant(Sticky)
+  common.rendersChildren(Sticky)
+
+  it('should create two divs', () => {
+    mount(<Sticky />).html().should.be.equal('<div><div></div><div></div></div>')
```
layershifter: I think we can use `shallow` there instead of `mount`, also it will be more clear if we will use another enzyme's methods.

```js
const children = shallow(<Sticky />).children()

children.should.have.length(2)
children.everyWhere(child => child.should.have.tagName('div'))

```

src/modules/Sticky/Sticky.js
layershifter: BTW, all our callbacks should pass `data` as second argument, it this case it's `props`.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * Callback when element is fixed to page.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     * */
+    onStick: PropTypes.func,
+
+    /* Callback when element is bound to top of parent container. */
+    onTop: PropTypes.func,
```
layershifter: `onTop` and `onBottom` callbacks should have the documented signature, too. Don't forget also update typings :smile:

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,231 @@
...
+    className: PropTypes.string,
+
+    /** Context which sticky element should stick to. */
+    context: PropTypes.object,
+
+    /** Offset in pixels from the top of the screen when fixing element to viewport. */
+    offset: PropTypes.number,
+
+    /** Offset in pixels from the bottom of the screen when fixing element to viewport. */
+    bottomOffset: PropTypes.number,
```
layershifter: `propTypes` should be sorted in an alphabetical order, too.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,231 @@
+import _ from 'lodash'
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+
+import {
+  customPropTypes,
+  META,
+  getElementType,
+  getUnhandledProps,
```
layershifter: Imports should be sorted in an alphabetical order, too.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,231 @@
...
+} from '../../lib'
+
+/**
+ * Sticky content stays fixed to the browser viewport while
+ * another column of content is visible on the page.
```
layershifter: I sure this comment can be a single line.

src/modules/Sticky/Sticky.js
mariolamacchia: Arrow functions implicitly bind the method to `this` when the instance is created, right? Isn't this kind of bad for performance, since we duplicate the function for each instance?

layershifter:
@mariolamacchia I've pushed some changes, I hope you don't hate me because there can be serious merge conflicts.

test/specs/modules/Sticky/Sticky-test.js
```diff
@@ -0,0 +1,205 @@
...
+    scrollToTop(wrapper, contextEl, { bottomOffset, offset, height })
+    fn.should.have.been.calledOnce()
+  })
+
+  it('should fire onBottom callback', () => {
+    const offset = 5
+    const bottomOffset = 5
+    const height = 5
+    const contextEl = { getBoundingClientRect: () => ({}) }
+    const fn = sandbox.spy()
```
layershifter: Let's rename spies definition, e.g. spy for the `onBottom` should be named `onBottom`:

```js
const onBottom = sandbox.spy()
```

test/specs/modules/Sticky/Sticky-test.js
```diff
@@ -0,0 +1,205 @@
...
+    const wrapper = mount(
+      <Sticky context={contextEl} offset={offset} bottomOffset={bottomOffset} onBottom={fn} />
+    )
+
+    scrollAfterContext(wrapper, contextEl, { bottomOffset, offset, height })
+    fn.should.have.been.calledOnce()
```
layershifter: Let's also add `calledWithMatc` assertion:

```js
onBottom.should.have.been.calledWithMatch({}, { bottomOffset, offset })
```

layershifter (changes requested):

layershifter (approved):
@mariolamacchia Thanks for changes, awesome work :+1: Waiting for @levithomason there

brianjd:
Woot!

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,245 @@
...
+    type: META.TYPES.MODULE,
+  }
+
+  state = {
+    sticky: false,
+  }
+
+  componentDidMount() {
+    this.handleUpdate()
+    window.addEventListener('scroll', this.handleUpdate)
```
levithomason, line 87: We've got a perf leak here that we should handle.  When scrolling the page on my macbook, this method is fired hundreds of times.  This will especially be a concern since `handleUpdate` then calls `getBoundingClientRect`.  This method [forces a reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) and a repaint which is a pretty costly performance hit.

One solution might be to use `requestAnimationFrame` instead.  There is a lot of info on this around the web, but [here](https://www.html5rocks.com/en/tutorials/speed/animations/) is an older article on the topic that gives one explanation.

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,245 @@
...
+    return {
+      bottom,
+      top,
+      position: 'fixed',
+      width: this.triggerRect.width,
+    }
+  }
+
+  // Return true if the height of the component is higher than the window
+  oversized = () => this.stickyRect.height > window.innerHeight
```
levithomason: Let's prefix methods that always return bool:  `isOversized()`

src/modules/Sticky/Sticky.js
```diff
@@ -0,0 +1,245 @@
...
+
+  // Return true when the component reached the bottom of the context
+  reachedContextBottom = () => {
```
levithomason: Let's prefix methods that always return bool:  `didReachContextBottom`

I'll not comment on the rest, but let's do the whole file please.

levithomason:
@mariolamacchia this PR looks really great, thank you so much for the hard and detailed work ❤️ 

I pushed a couple minor fixes and tweaks to the docs.  I also left a couple inline comments.

The only real concern I have before merging is perf related.  We need to handle perf issues with listeners on scroll.  Once that is done, I'd like to do some profiling to be sure we aren't killing resources and causing the browser to do lots of reflow/layout/repaint.  Once we get this done, I think we're good to go here.  Thanks again!

levithomason (changes requested):
@mariolamacchia this PR looks really great, thank you so much for the hard and detailed work ❤️ 

I pushed a couple minor fixes and tweaks to the docs.  I also left a couple inline comments.

The only real concern I have before merging is perf related.  We need to handle perf issues with listeners on scroll.  Once that is done, I'd like to do some profiling to be sure we aren't killing resources and causing the browser to do lots of reflow/layout/repaint.  Once we get this done, I think we're good to go here.  Thanks again!

src/modules/Sticky/Sticky.js
mariolamacchia: Very good point! Thanks for the article, I'll fix it.

src/modules/Sticky/Sticky.js
mariolamacchia: I removed the auxiliary verbs from the names after @layershifter's code review 😅 . Shall I put them back there?

layershifter: @mariolamacchia Seems, that yes :smile:

mariolamacchia:
@levithomason I'm done with the changes

src/modules/Sticky/Sticky.js
fracmak: Is this a solution that should also be done to the Visibility component? It also has a scroll handler that calls getBoundingClientRect() on the dom ref

michaelimperial:
can't wait!

pcompassion:
wating for this!

aktary:
A merge/release of this today would be an AWESOME Friday afternoon gift  ;)

Pleeeeeeease tell me it'll work for tables...

levithomason (approved):

levithomason:

TimNZ:
@mariolamacchia Thanks for this.

I'm trying to use Sticky in a complex nested set of div of varying position (mostly absolute) and with overflowY: auto

The 'scroll' event isn't being triggered when I scroll the div.  

Does it make sense to be allow to provide a scroll element prop to override the 'window' used for addEventListener('scroll') (and other uses of window for calcs)?

I'm using InfiniteScroller (https://github.com/CassetteRocks/react-infinite-scroller) and it has a prop to specify whether to use window, otherwise defaults to parent node.

```
    key: 'attachScrollListener',
    value: function attachScrollListener() {
      if (!this.props.hasMore) {
        return;
      }

      var scrollEl = window;
      if (this.props.useWindow === false) {
        scrollEl = this.scrollComponent.parentNode;
      }

      scrollEl.addEventListener('scroll', this.scrollListener, this.props.useCapture);
      scrollEl.addEventListener('resize', this.scrollListener, this.props.useCapture);

      if (this.props.initialLoad) {
        this.scrollListener();
      }
    }
```

TimNZ:
I'm not a web dev, but I quickly tried adding a 'container' prop, which defaults to 'window', and that seems to work as a starting point.

Now have to play with positioning of the sticky content when it is visible as position=fixed doesn't work right in my layout.

If you can implement a general purpose update for this quickly because you are order of magnitude more experienced than me, please do!  I will be messing around for ages on this as I am weak in CSS.

mariolamacchia:
@TimNZ I've just opened a new PR #1978

laukaichung:
Is there a way to disable the sticky like the prop `isActive` in https://github.com/captivationsoftware/react-sticky ?

mariolamacchia:
@stonecold123 I don't think I understand what you mean... can you show me an example, please?

PR #1752, opened 2017-06-09
fix(Checkbox|Input): fix handling of aria-attributes

### LINKED ISSUE #1750: Checkbox|Input: aria-* attributes incorrectly applied
`aria-label` should be applied at input element level.

```jsx
import React from 'react'
import { Checkbox } from 'semantic-ui-react'

const CheckboxExampleCheckbox = () => (
  <Checkbox label='Make my profile visible' aria-label="test" />
)

export default CheckboxExampleCheckbox
```

#### OR 

```jsx
import React from 'react'
import { Input } from 'semantic-ui-react'

const InputExampleFocus = () => (
  <Input focus placeholder='Search...' aria-label="test" />
)

export default InputExampleFocus
```

### Expected

```html
<div class="column rendered-example modules-checkbox-types-checkbox-example-checkbox">
     <div class="ui checkbox">
                   <input type="checkbox" class="hidden" readonly="" tabindex="0" value="on"  aria-label="test">
                    <label>Make my profile visible</label>
      </div>
</div>
```

#### OR

```html
<div class="column rendered-example elements-input-states-input-example-focus">
    <div class="ui focus input">
        <input type="text" placeholder="Search..." aria-label="test" ></div>
    </div>
</div>

```

### Actual

```html
<div class="column rendered-example modules-checkbox-types-checkbox-example-checkbox">
     <div class="ui checkbox" aria-label="test">
                   <input type="checkbox" class="hidden" readonly="" tabindex="0" value="on">
                    <label>Make my profile visible</label>
      </div>
</div>
```

#### OR

```html
<div class="column rendered-example elements-input-states-input-example-focus">
    <div class="ui focus input" aria-label="test">
        <input type="text" placeholder="Search..." ></div>
    </div>
</div>
```

### Version
0.68.4

layershifter:
Fixes #1750.

### DIFF
--- a/src/lib/htmlInputPropsUtils.js
+++ b/src/lib/htmlInputPropsUtils.js
@@ -1,5 +1,14 @@
 import _ from 'lodash'
 
+export const htmlInputAria = [
+  'aria-activedescendant', 'aria-atomic', 'aria-autocomplete', 'aria-busy', 'aria-checked', 'aria-controls',
+  'aria-describedby', 'aria-disabled', 'aria-dropeffect', 'aria-expanded', 'aria-flowto', 'aria-grabbed',
+  'aria-haspopup', 'aria-hidden', 'aria-invalid', 'aria-label', 'aria-labelledby', 'aria-level', 'aria-live',
+  'aria-multiline', 'aria-multiselectable', 'aria-orientation', 'aria-owns', 'aria-posinset', 'aria-pressed',
+  'aria-readonly', 'aria-relevant', 'aria-required', 'aria-selected', 'aria-setsize', 'aria-sort', 'aria-valuemax',
+  'aria-valuemin', 'aria-valuenow', 'aria-valuetext', 'role',
+]
+
 export const htmlInputAttrs = [
   // REACT
   'selected', 'defaultValue', 'defaultChecked',
@@ -32,7 +41,7 @@ export const htmlInputEvents = [
   'onTouchCancel', 'onTouchEnd', 'onTouchMove', 'onTouchStart',
 ]
 
-export const htmlInputProps = [...htmlInputAttrs, ...htmlInputEvents]
+export const htmlInputProps = [...htmlInputAria, ...htmlInputAttrs, ...htmlInputEvents]
 
 /**
  * Returns an array of objects consisting of: props of html input element and rest.
--- a/src/lib/htmlInputPropsUtils.js
+++ b/src/lib/htmlInputPropsUtils.js
@@ -1,14 +1,5 @@
 import _ from 'lodash'
 
-export const htmlInputAria = [
-  'aria-activedescendant', 'aria-atomic', 'aria-autocomplete', 'aria-busy', 'aria-checked', 'aria-controls',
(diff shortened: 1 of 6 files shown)


src/lib/htmlInputPropsUtils.js
```diff
@@ -1,5 +1,14 @@
...
 
+export const htmlInputAria = [
+  'aria-activedescendant', 'aria-atomic', 'aria-autocomplete', 'aria-busy', 'aria-checked', 'aria-controls',
+  'aria-describedby', 'aria-disabled', 'aria-dropeffect', 'aria-expanded', 'aria-flowto', 'aria-grabbed',
+  'aria-haspopup', 'aria-hidden', 'aria-invalid', 'aria-label', 'aria-labelledby', 'aria-level', 'aria-live',
+  'aria-multiline', 'aria-multiselectable', 'aria-orientation', 'aria-owns', 'aria-posinset', 'aria-pressed',
+  'aria-readonly', 'aria-relevant', 'aria-required', 'aria-selected', 'aria-setsize', 'aria-sort', 'aria-valuemax',
+  'aria-valuemin', 'aria-valuenow', 'aria-valuetext', 'role',
+]
+
```
layershifter: @levithomason I'm not sure that we should pass all `aria` attributes to `input` in these elements, possible you'll want to perform a cleanup in the list.

levithomason: How about we update the existing `partitionHTMLInputProps` loop to check for `aria-` prefixed props instead?  If `/^aria-.*$/.test(prop)` then it is automatically an `inputProp` and not a `rest` prop.

src/lib/htmlInputPropsUtils.js
```diff
@@ -37,15 +37,22 @@ export const htmlInputProps = [...htmlInputAttrs, ...htmlInputEvents]
...
+ * @param {Array} [options.htmlProps] An array of html input props
+ * @param {boolean} [options.includeAria] Includes all input props that starts with "aria-"
  * @returns {[{}, {}]} An array of objects
  */
-export const partitionHTMLInputProps = (props, htmlProps = htmlInputProps) => {
+export const partitionHTMLInputProps = (props, options = {}) => {
+  const {
+    htmlProps = htmlInputProps,
+    includeAria = true,
+  } = options
```
layershifter, line 49: I've applied there our `options` pattern, so we can control `htmlProps` and `ariaProps` separately.

src/lib/htmlInputPropsUtils.js
```diff
@@ -37,15 +37,22 @@ export const htmlInputProps = [...htmlInputAttrs, ...htmlInputEvents]
...
   const inputProps = {}
   const rest = {}
 
   _.forEach(props, (val, prop) => {
-    const target = _.includes(htmlProps, prop) ? inputProps : rest
+    const possibleAria = includeAria && (/^aria-.*$/.test(prop) || prop === 'role')
```
layershifter, line 54: `role` prop belongs to aria attributes, too.

layershifter (commented):
@levithomason I've performed there changes changes that you've requested, looks cleaner :+1:

levithomason:
Looks great 👍

levithomason:

PR #1753, opened 2017-06-09
feat(UniversalInput): add component

### LINKED ISSUE #1744: Input: onChange event triggered when IME character (ex. Chinese) is not completely entered
### Steps

- create an input element with onChange event 

`  <Input 
    placeholder='Search...'
    onChange={() => {console.log("onchange fired!")}}
  />`

- enter some text using Chinese, Japanese, or Korean input method

### Expected Result
onChange event should be triggered only when character is finalized

### Actual Result
onChange event triggered as each sub character is entered 

### Version
0.68.5

### Impact
This severely impacts the usability of Input and Search Components for users using non-latin language
This is probably an inherent issue with React, but maybe Semantic-UI can provide a workaround using compositionstart and compositionend event.

### Related issue
https://github.com/facebook/react/issues/3926

layershifter:
# WIP

Fixes #1744.

### Proposed changes

This PR adds the `Universal Input` (a quite strange name, open for ideas) component, that correctly handles composition events. This component will replace an `input` usages. I grabbed ideas from facebook/react#3926.

#### Adds `finalValue` param to `onChange` handler

`finalValue` contains value after the composition was completed.

peek 2017-06-12 13-49

#### Adds param to `onCompose` handler

It fires only after the composition was completed.

peek 2017-06-12 13-50

### TODO
- [ ] tests

This PR provides an input component that can be used by our other components. It correctly supports the handling of `onChange` with IME.

### DIFF
--- a/docs/app/Examples/addons/UniversalInput/Types/UniversalInputExampleInput.js
+++ b/docs/app/Examples/addons/UniversalInput/Types/UniversalInputExampleInput.js
@@ -1,0 +1,29 @@
+import React, { Component } from 'react'
+import { Form, Grid, Header, UniversalInput } from 'semantic-ui-react'
+
+export default class UniversalInputExampleInput extends Component {
+  state = {
+    finalValue: null,
+    value: null,
+  }
+
+  handleChange = (e, { finalValue, value }) => this.setState({ finalValue, value })
+
+  render() {
+    return (
+      <Grid>
+        <Grid.Column width={8}>
+          <Form>
+            <Form.Field>
+              <UniversalInput onChange={this.handleChange} placeholder='Try to type something...' />
+            </Form.Field>
+          </Form>
+        </Grid.Column>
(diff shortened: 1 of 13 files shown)


layershifter:
@paladinze It will be awesome to have a feedback from you.

### How?

If you could pull this branch `feat/abs-input`, `npm install`, `npm start`, and navigate to http://localhost:8080/addons/universal-input, you'll see the Universal Input component examples. Please have a look at the code, play with the editors, and let me know what you think.

layershifter:
I want to close this for housekeeping as I don't receive any feedback there.

chezhe:
I got a problem, seems related to this.
when I type Chinese in Search Component, it comes out double first character.
Actually, I just type 'fz'，result should be '方庄'.

I tried get value from `data.value` or `e.target.value`, the same.

And，it only comes out in production environment. dev env is okay.
By the way, I use semantic with next.js

PR #1757, opened 2017-06-09
feat(Dropdown): component shorthand in options prop

### LINKED ISSUE #1724: RFC: Dropdown header and divider shorthands
[https://react.semantic-ui.com/modules/dropdown#dropdown-example-header](https://react.semantic-ui.com/modules/dropdown#dropdown-example-header)

I wanna use `Dropdown.Header` and `Dropdown.Divider` in `search` or `multiple` dropdown, but it's currently not supported. So I suppose we should extend shorthand props to support it. But it can be done in many ways so it's better to consider different options.

## Header

I think the header shorthand is pretty straitforward

```
import React from 'react'
import { Dropdown } from 'semantic-ui-react'

const options = [
  { content: 'Important', header: { icon: 'tags', content: 'Filter by tag' } },
  { content: 'Announcement' },
  { content: 'Discussion' }
]

// <Dropdown.Header icon='tags' content='Filter by tag' />
// <Dropdown.Item>Important</Dropdown.Item>
// <Dropdown.Item>Announcement</Dropdown.Item>
// <Dropdown.Item>Discussion</Dropdown.Item>

const DropdownExampleHeader = () => (
  <Dropdown text='Filter' icon='filter' floating labeled button className='icon' options={options} />
)

export default DropdownExampleHeader
```

## Divider

I think it should be implemented by using `divider` shorthand with position as shorthand value. Possible positions will be `'top' | 'bottom' | 'both'`.
```
import React from 'react'
import { Dropdown } from 'semantic-ui-react'

const options = [
  { content: 'Important' },
  { content: 'Announcement', divider: 'top' },
  { content: 'Discussion' }
]

// <Dropdown.Item>Important</Dropdown.Item>
// <Dropdown.Divider />
// <Dropdown.Item>Announcement</Dropdown.Item>
// <Dropdown.Item>Discussion</Dropdown.Item>

const DropdownExampleDivider = () => (
<Dropdown text='Filter' icon='filter' floating labeled button className='icon' options={options} />
)

export default DropdownExampleDivider
```

saitonakamura:
fixes #1724 

WIP

This pull request introduces a possibilty to set `component` shorthand in `options` Dropdown shorthand. Use cases is `Header` and `Divider` use while getting benefits of full state management.

Example

```
const options = [
  { content: 'Important' },
  { component: <Dropdown.Divider /> },
  { content: 'Announcement' },
  { content: 'Discussion' }
]

// <Dropdown.Item>Important</Dropdown.Item>
// <Dropdown.Divider />
// <Dropdown.Item>Announcement</Dropdown.Item>
// <Dropdown.Item>Discussion</Dropdown.Item>

const DropdownExampleDivider = () => (
  <Dropdown options={options} />
)
```

Please note that that API is a subject to change (Discussion is in #1724 ), but I took the risk to start implementing it.

- [ ] Decide how the API should look like
- [ ] Implement it
- [ ] Update examles
- [ ] Update tests


levithomason:
### `component`

Regarding the API: `{ component: <Dropdown.Divider /> } `
Let's go with: `{ component: Dropdown.Divider }`

The primary reason is that `Dropdown.Divider` is a React _component_ (string, class, or function) while `<Dropdown.Divider />` is actually a React _element_.  The JSX is compiled to `React.createElement(Dropdown.Divider)`, which, as the name suggests, creates a new _element_ from the provided _component_.

The secondary reason is that the objects in the array are really props objects.  They are spread on the element when created.  Currently, all objects are `Dropdown.Item` props objects.  By using a component rather than an element, it allows us to still spread the object as props onto the provided component.

The final reason is that it saves a lot of renders.  If an element is passed, it is instantly rendered inline (`<Dropdown.Header />` renders an element) regardless of whether or not the Dropdown menu is ever opened.  However, by passing the component and its props, we lazily render the items only when needed.

### `options`

Since we're now adding much more to the array, these are no longer simply Dropdown (or Select) `options`.  We are, in fact, beginning the shorthand for _any_ the `Dropdown.Menu` item 🎉 

In SUI nomenclature, these are the `items` within in the `menu`.  I'm going back and forth with hardline consistency and ease of use here.

Technically, this should be added as a `menu` prop which takes a props object for the `Dropdown.Menu`. It would contain an `items` array in that case.  I'm leaning this way for consistency and predictability, but it is a little obtuse:

```jsx
const menu = {
  items: [
    { ... }
  ]
}

<Dropdown menu={menu} />
```

I like this as it allows the DropdownMenu to have its own `active` prop and manage its own `className` buildup.  Currently, the Dropdown parent is doing all this.  It should also make animations easier once we get there.

levithomason:
Hm, this looks like it will have direct conflicts also with #1038.  That PR is getting stale, though, I recall changing the options to shorthand introduces a cascade of side effects that have to be dealt with.

This ends up affecting all rendering, filtering, and callbacks.  Let me circle back over there and see where this lands us.

Also, if you have time, go ahead and take a look at it yourself and get up to speed on where it was left off.

layershifter:
### `component`

@levithomason I think that we should go with `component` as string or event `type`:

```jsx
[
{ component: Dropdown.Divider, as: Divider },
{ component: 'divider', as: Divider },
{ type: 'divider', as: Divider },
]
```

It will allow to make `render` logic more simple:

```jsx
const { type } = item
const factories = {
  divider: Dropdown.Divider,
  item: Dropdown.Item,
  menu: Dropdown.Menu,
}
const factory = factories[type] || Dropdown.Item

return factory(item)
```

saitonakamura:
Sorry for my absence, i'm back and eager to continue my work

> Let's go with: { component: Dropdown.Divider }

Perfectly sane, I forgot about renders and stuff

> Technically, this should be added as a menu prop which takes a props object for the Dropdown.Menu

In that case it will be `options` shorthand for ease of use and backwards compatibility and new `menu` shorthand for complex stuff, right?

> I think that we should go with component as string or event type:

I think it can be error-prone, cause linters won't complain in case of typo, but not with `Dropdown.Divider` from import

saitonakamura:
I've updated the code to `{ component: Dropdown.Divider }` and `menu` version
@levithomason I currently don't understand will the `options` prop be appliable for `component` shorthand or not, can you clarify it? Cause at first I was thinking that it won't be, but I can't see the way to separate `getMenuOptions` logic with `options` and with `menu`. The main problem is that we should keep the order of headers, dividers and items. Well, except just copypasting a bunch of code, and I don't fancy that notion.

levithomason:
I've marked this PR for review and will take a look as soon as I can.  Would really love to get this one merged.  Thanks for persisting!

ssageghi:
Can we use this or not?

ssageghi:
Is merge Complete?
can we use it?

saitonakamura:
Hi, guys. Are we agreed on the api? I can start implementing tests and updating documentation if we are and refactor code later.

levithomason:
>@levithomason I think that we should go with component as string or event type:

@layershifter IMHO, the `as` prop should be reserved for rendering the Divider as something else.  I'd like to go with the currently implemented solution where `component` is a real component.

In general, I prefer code over configuration as it is more flexible.  Consider this:

```jsx
{ component: MyHOC(Dropdown.Divider) }
```

This cannot be done with a string based configuration.

docs/app/Examples/modules/Dropdown/Content/DropdownExampleDescription.js
```diff
@@ -1,16 +1,16 @@
...
-  <Dropdown text='Filter Tags' floating labeled button icon='filter' className='icon'>
-    <Dropdown.Menu>
-      <Dropdown.Header icon='tags' content='Filter by tag' />
-      <Dropdown.Divider />
-      <Dropdown.Item description='2 new' text='Important' />
-      <Dropdown.Item description='10 new' text='Hopper' />
-      <Dropdown.Item description='5 new' text='Discussion' />
-    </Dropdown.Menu>
-  </Dropdown>
+  <Dropdown text='Filter Tags' floating labeled button icon='filter' className='icon' options={options} />
```
levithomason, line 13: I think instead of overloading the `options` here, we use a new `items` prop.  There is quite a bit of logic based on the options for keyboard nav and selected/active/disabled states. Options is also meant to reflect the idea of `select` options.

My thinking with `items` is that is reflects any valid `Menu` item, including headers and dividers.  This will also play well with `<Menu items={} />` and `<Dropdown.Menu items={} />`.

Thoughts?

src/modules/Dropdown/Dropdown.js
```diff
@@ -1147,6 +1155,15 @@ export default class Dropdown extends Component {
     })
   }
 
+  renderComponent = (component, props) => {
+    switch (component) {
```
levithomason, line 1204: Rather than a method for this, we should just invoke `React.createElement()` via JSX syntax.  This will allow the user to pass any class or functional component.  They may pass an entirely different component than `Dropdown.Divider`, or one of these wrapped in a HOC which would fail the switch cases.

```diff
-this.renderComponent(component, { ...componentProps, key: i })
+<component {...componentProps} key={i} />
```

I think the `renderComponent` method then goes away completely.

src/modules/Dropdown/Dropdown.js
```diff
@@ -768,6 +773,9 @@ export default class Dropdown extends Component {
 
     // filter by search query
     if (search && searchQuery) {
+      // filter any component elements like Header or Divider
+      filteredOptions = _.filter(filteredOptions, opt => !opt.component)
```
levithomason, line 808: Oh, this is pretty simple.  Perhaps we should just keep `options` as the prop for this case.  For some reason, I hadn't thought of this simple fix.  I believe this will work in all cases.

docs/app/Examples/modules/Dropdown/Content/DropdownExampleHeader.js
```diff
@@ -1,15 +1,16 @@
...
+  items: [
+    { component: Dropdown.Header, icon: 'tags', content: 'Filter by tag' },
+    { text: 'Important', value: 'Important' },
+    { text: 'Hopper', value: 'Hopper' },
+    { text: 'Discussion', value: 'Discussion' },
+  ],
+}
+
+const DropdownExampleHeader = () =>
+  <Dropdown text='Filter' icon='filter' floating labeled button className='icon' menu={menu} />
```
levithomason, line 14: I like this API the most.  This is the most consistent with where we are headed with shorthand and subcomponent alignment.

levithomason:
>will the options prop be applicable for component shorthand or not, can you clarify it?

I'm a little torn on this myself, however, I think we need to abandon the `options` prop.  We have the `<Select />` addon component which takes `options`.  This makes sense because of the parity with an HTML `select` and `option` children.  This is a convenience wrapper around the `Dropdown`.

The Dropdown, however, has a `menu` child.  I think then we should keep parity between subcomponents and shorthand and go with a `menu` prop only for the `Dropdown`:

Final API propsal:

```jsx
<Select options={} />   // as it is today, no change
<Dropdown menu={} />    // Takes Dropdown.Menu shorthand
```

It makes sense for a `Dropdown.Menu` to pass literal shorthand values to its `items` prop:

**DropdownMenu.js**

```jsx
DropdownMenu.create = createShorthandFactory(DropdownMenu, items => ({ items })
```
**Dropdown.js**
```jsx
// When rendering the menu
DropdownMenu.create(this.props.menu)
```

This means arrays will be passed to the `items` prop:

```jsx
// Array of DropdownItem primitive values
<Dropdown menu={['First', 1, 2, 3, 'Last]} />

// Array of DropdownItem props objects 
<Dropdown menu={[{ text: 'One', value: 1 }]} />

// Array of elements
<Dropdown menu={[<Dropdown.Item />, <OtherItem />]} />
```

Whilst props objects and elements can still be used.
```jsx
// DropdownMenu props
<Dropdown menu={{ items: [ ... ], visible: true }} />

// An element
<Dropdown menu={<Dropdown.Menu />} />
<Dropdown menu={<AnotherMenu />} />
```

Note, the `createShorthandFactory()` method handles all this already, I'm just noting the usage we'd get out of it ;)

dennysem:
+1

saitonakamura:
@levithomason `DropdownMenu.create(this.props.menu)` in this case we move a lot of logic from `Dropdown` to `DropdownMenu`, right?
Methods like `renderOptions`, `getMenuOptions`, `search` and keyboard navigation goes into `DropdownMenu`?

levithomason:
Yes, sir.  Right now, the Dropdown is assuming all the functionality of the Menu as it creates and manages both.

saitonakamura:
I started refactoring it and stumbled into a problem, for example: we have method `getSelectedItem` which seem as responsibility of `Menu` cause you unable to get the possible options from `Dropdown`, but it's used in method `renderText`.

I can see a solution, but it seems very hacky: we can get selected item by subscribing by some callback, like `onItemSelected` and save it to `Dropdown` state

levithomason:
>we can get selected item by subscribing by some callback, like onItemSelected and save it to Dropdown state

This sounds like the idiomatic React approach.  Let's go with this for now.

saitonakamura:
@levithomason as you can understand, I can't proceed with this, unfortunately (I left the project in which this feature was needed and I don't have enough willpower). Despite it's obvious, I'm still really sorry for it, guys. Let's close this PR.

levithomason:
Understood, it happens :)

PR #1758, opened 2017-06-10
fix(Accordion): typings inverted to boolean

mcrawshaw:
Tiny fix for typings, inverted type incorrectly set to string.


levithomason:
Thanks much!

levithomason:

PR #1762, opened 2017-06-12
fix(Checkbox): focus should be obtained on mouseDown

### LINKED ISSUE #1335: Checkbox: focus not properly activated on click

layershifter:
Finally fixes #1335.

~~We added focus capture in #1361, but we added it into the `onMouseDown` handler while it should be inside `onClick` handler.~~

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -1,5 +1,5 @@
 import cx from 'classnames'
-import _ from 'lodash/fp'
+import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React from 'react'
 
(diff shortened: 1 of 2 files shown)


src/modules/Checkbox/Checkbox.js
```diff
@@ -1,5 +1,5 @@
 import cx from 'classnames'
-import _ from 'lodash/fp'
+import _ from 'lodash'
```
layershifter, line 2: @levithomason I removed usage of `lodash/fp` because I don't understand how I can make this in FP style:

```js
_.invoke(this.props, 'onMouseDown', e, { ...this.props, checked: !!checked, indeterminate: !!indeterminate })
```

levithomason: No worries, this works 👍

levithomason:
## Order of events

The order of events as fired by the browser are: `mouseDown`, `focus`, `mouseUp`, `click`.  I've added the following handlers to this comment textarea on github and here's the result:

```js
new_comment_field.addEventListener('click', () => console.log('click'))
new_comment_field.addEventListener('mousedown', () => console.log('mousedown'))
new_comment_field.addEventListener('mouseup', () => console.log('mouseup'))
new_comment_field.addEventListener('focus', () => console.log('focus'))
```
A click, long press, and release logs this:



## onBlur

That said, our bug is that the `onBlur` callback is called immediately _after_ the `onFocus` callback.  So, the reported codepen ends up setting state `true` in `onFocus` and setting it back to `false` in `onBlur`.  The `onFocus` event is working correctly, it is just that there is an extra `onBlur` event firing first.  You can see this easily by removing the `onBlur` handler and the state is updated correctly via `onFocus`.

This extra blur event is a side effect of calling `inputRef.focus()` in the `onMouseDown` handler for what is actually the `label` element. We can circumvent this issue by simply preventing the default event in `onMouseDown`.  That way, we call `onMouseDown` then `inputRef.focus()` without causing the extra unnecessary blur:

**Update to this PR**
```diff
// handleClick() {
// ..snip
    _.invoke(this.props, 'onChange', e, { ...this.props, checked: !checked, indeterminate: false })

    this.trySetState({ checked: !checked, indeterminate: false })
-   _.invoke(this.inputRef, 'focus')
  }

  handleMouseDown = e => {
    debug('handleMouseDown()')
    const { checked, indeterminate } = this.state

    _.invoke(this.props, 'onMouseDown', e, { ...this.props, checked: !!checked, indeterminate: !!indeterminate })
+   _.invoke(this.inputRef, 'focus')
+   e.preventDefault()
  }
```

## The test

This results in the correct behavior as well as the correct order of events fired:



***

Give that a try and let me know your thoughts.

layershifter:
@levithomason Thanks for a detailed reply. You're right there, seems that SUI implements focus incorrectly, it triggers `focus` after click. The behaviour that you described is more native. I will perform changes in this PR.

levithomason:
Thanks for the confirmation and second round.

levithomason:

PR #1763, opened 2017-06-12
docs(images): add missing images, update urls

layershifter:
This PR adds some missing images for `Popup` and `Embed` examples.
Also, fixes URLs in `Visibility` examples.


levithomason:

Issue #419, opened 2016-08-23
RFC: Base component


levithomason:
There is basic functionality that is shared by all components:
- merge props className with SUI classes
- spread unhandled props
- get element type (augmentation)
- define `as` and `className` propTypes, most the time also `children`

Updating components in any of these areas requires updating all components.  This was just done to support augmentation, see #414.  There are ~1,800 lines added to accomplish this.  The grand majority of this work was duplicating the logic and propTypes for all components.

It seems we could use a base component factory to generate a "conformant" base component.  It would handle the common functionality for all components.  It could make maintenance much easier.  It could also give us the ability to better standardize how components are generated.

This would either work similar to the reusable component parts factories in #345, or be a base component that we extend.  In either case, the above list of required features would be abstracted into a single place.

jeffcarbs:
I've been doing some cleanup lately and I noticed this as well. Not sure the best way forward here, but just wanted to bump this thread. 

So many of our components, specifically our sub-components, are nearly identical. Here's an example:
<img width="1056" alt="screen shot 2016-10-04 at 3 51 12 am" src="https://cloud.githubusercontent.com/assets/847027/19066168/d6cae3be-89e5-11e6-8481-f9a6126a3767.png">

``` diff
-function MessageContent(props) {
+function MessageHeader(props) {
   const { children, className } = props
-  const classes = cx('content', className)
-  const rest = getUnhandledProps(MessageContent, props)
-  const ElementType = getElementType(MessageContent, props)
+  const classes = cx('header', className)
+  const rest = getUnhandledProps(MessageHeader, props)
+  const ElementType = getElementType(MessageHeader, props)

   return <ElementType {...rest} className={classes}>{children}</ElementType>
 }

-MessageContent._meta = {
-  name: 'MessageContent',
+MessageHeader._meta = {
+  name: 'MessageHeader',
   parent: 'Message',
   type: META.TYPES.COLLECTION,
 }

-MessageContent.propTypes = {
+MessageHeader.propTypes = {
   /** An element type to render as (string or function). */
   as: customPropTypes.as,

   /** Primary content. */
-  children: customPropTypes.children(MessageContent),
+  children: customPropTypes.children(MessageHeader),

   /** Additional classes. */
   className: PropTypes.node,
 }

-export default MessageContent
+export default MessageHeader
```

Literally the only difference (aside from the component name) is the class name applied.

levithomason:
[...]

PR #1770, opened 2017-06-15
fix(AutoControlledComponent) Use this.setState instead of assigning this.state in componentWillMount

### LINKED ISSUE #1761: AutoControlledComponent throws warning in React 16
### Steps
1) Upgrade to React 16
2) Place any of the `AutoControlledComponent`s in your app

### Expected Result
It should render without warning

### Actual Result
It renders with the warning `Warning: Dropdown.componentWillMount(): Assigning directly to this.state is deprecated (except inside a component's constructor). Use setState instead.`
screen shot 2017-06-12 at 10 32 06 am

### Version
0.68.5

### Testcase
It seems to be caused by this line: `src/lib/AutoControlledComponent.js:149`. I'll open a PR shortly so we can continue to fully support React 16.

apiv:
Fixes #1761

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -138,6 +138,11 @@ class Portal extends Component {
 
   state = {}
 
+  constructor(...args) {
+    super(...args)
+    this.setInitialAutoControlledState()
+  }
+
   componentDidMount() {
     debug('componentDidMount()')
     this.renderPortal()
--- a/src/collections/Menu/Menu.js
+++ b/src/collections/Menu/Menu.js
@@ -134,6 +134,11 @@ class Menu extends Component {
   static Item = MenuItem
   static Menu = MenuMenu
 
+  constructor(...args) {
+    super(...args)
+    this.setInitialAutoControlledState()
+  }
+
   handleItemOverrides = predefinedProps => ({
     onClick: (e, itemProps) => {
       const { index } = itemProps
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -68,7 +68,7 @@ export const getAutoControlledStateValue = (propName, props, state, includeDefau
 }
 
 export default class AutoControlledComponent extends Component {
-  componentWillMount() {
+  setInitialAutoControlledState() {
     const { autoControlledProps } = this.constructor
 
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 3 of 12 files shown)


layershifter:
@levithomason We need you there, as I understand that the problem is that `setState()` is async. Can you clarify?

levithomason:
I've retriggered tests here.  I believe you are correct with the async state issue.

In the end, this may end up forcing our hand to a larger refactor to a HOC.  That would be a bit of a bummer, but would also open up the door to complete #419.

apiv:
@levithomason Ah, I see.
- When using `this.setState` in `componentWillMount`, the update is async, which breaks
- Moving this logic to the constructor breaks, because components which extend `AutoControlledComponent` may have their own constructors in which they set state (which this logic should process). Since `super` must be called before the child class's constructor, this flow does not work.
   ```js
   class AutoControlledComponent extends Component {
     state = {}
     constructor () {
       // do stuff with this.state
       console.log(this.state)
     }
   }

   class BrokenExample extends AutoControlledComponent {
     state = {}
     constructor () {
       super();
       this.state = {
         something: 'value'    
       }
       // the console log in AutoControlledComponent will be {}, not { something: 'value' }
     }
   }
   ```

I have found an alternate workaround, which admittedly isn't the most ideal -- but just wanted to throw it out there. If we move the logic to a new method called `setInitialAutoControlledState`, we can do our state setting logic in the constructor, and then call `this.setInitialAutoControlledState` when we are ready.
```js
class AutoControlledComponent extends Component {
  state = {}
  setInitialAutoControlledState () {
    // do stuff with this.state
    console.log(this.state)
  }
}

class BrokenExample extends AutoControlledComponent {
  state = {}
  constructor () {
    super();
    this.state = {
      something: 'value'    
    }
    this.setInitialAutoControlledState()
    // the console log in AutoControlledComponent will be { something: 'value' }
  }
}
```

Again, not ideal, but I have re-pushed my branch with this change in place, and it is running properly. On the negative side, this would be a breaking change, because anyone extending `AutoControlledComponent` in their own codebase would need to change their implementation.

Perhaps an iteration of this could be a good interim solution to the React 16 upgrade, and give us some more time to refactor AutoControlledComponent to an HOC.

But essentially, the logic that I moved to `setInitialAutoControlledState` just needs to happen at any time _after_ the child class's `constructor` call -- which is why it was put into `componentWillMount`. Unfortunately, I don't think there are any other spots in the component lifecycle or ES2015 classes to make this happen without calling it manually...

PR #1780, opened 2017-06-19
docs(props): fix uniqueness

layershifter:
### Why?

`SUI.WIDTHS` contains both numbers and strings (example: `1` and `'1'`), but it does not matter for the docs. Also, as we use `value` as `key` and React converts `key` to string, we have duplicate, but invisible elements.

### DIFF
--- a/gulp/plugins/util/parseType.js
+++ b/gulp/plugins/util/parseType.js
@@ -2,7 +2,9 @@ require('babel-register')
 const _ = require('lodash')
 const SUI = require('../../../src/lib/SUI') // eslint-disable-line no-unused-vars
 
-const evalValue = value => _.uniq(eval(value)) // eslint-disable-line no-eval
+const evalValue = value => eval(value) // eslint-disable-line no-eval
+
+const uniqValues = values => _.uniqWith(values, (val, other) => val == other) // eslint-disable-line eqeqeq
 
 const transformEnumValues = values => _.flatMap(values, ({ value }) => {
   if (_.startsWith(value, '...SUI')) return evalValue(value.substring(3))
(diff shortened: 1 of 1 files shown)


gulp/plugins/util/parseType.js
```diff
@@ -2,7 +2,9 @@ require('babel-register')
 const _ = require('lodash')
 const SUI = require('../../../src/lib/SUI') // eslint-disable-line no-unused-vars
 
-const evalValue = value => _.uniq(eval(value)) // eslint-disable-line no-eval
+const evalValue = value => eval(value) // eslint-disable-line no-eval
+
+const uniqValues = values => _.uniqWith(values, (val, other) => val == other) // eslint-disable-line eqeqeq
```
layershifter: We cannot use there strict equals, because `1` and `'1'` aren't equal.

levithomason: Coerce them both to numbers with the `+` operator:

```jsx
+val === +other
```

Example:
```jsx
val = '5'

typeof val   // => "string"
typeof +val  // => "number"
val === 5    // => false
+val === 5   // => true
```

layershifter: Yes, but not all items of array are numbers:

```js
val = 'five'
+val // => NaN
```

layershifter: I think we can do following:
```js
`${val}` === `${other}`
// or
val.toString() === other.toString()
```

levithomason: Nice catch, that works as well.

layershifter (commented):
I've performed requested change, LGTM

levithomason:
Thank ya!

levithomason:

PR #1781, opened 2017-06-19
feat(Button): add advanced handling of disabled prop

### LINKED ISSUE #1775: Button: the disabled attribute doesn't apply to the button element
### Steps
Create a `Button` component with disabled set to true.

### Expected Result
The button should be disabled.

### Actual Result
The button only has styles to make it appear disabled, but it still functions if you tab to it due to a missing disabled attribute on the button element.

### Version
0.68.3

### Testcase
https://codepen.io/anon/pen/OgbeXr

layershifter:
This PR:
- fixes #1775
- performs a refactoring of  the `render()` method
- improves the test coverage

### DIFF
--- a/src/elements/Button/Button.js
+++ b/src/elements/Button/Button.js
@@ -8,7 +8,6 @@ import {
   createShorthandFactory,
   getElementType,
   getUnhandledProps,
-  makeDebugger,
   META,
   SUI,
   useKeyOnly,
@@ -20,8 +19,6 @@ import Label from '../Label/Label'
 import ButtonContent from './ButtonContent'
 import ButtonGroup from './ButtonGroup'
 import ButtonOr from './ButtonOr'
-
-const debug = makeDebugger('button')
 
 /**
  * A Button indicates a possible user action.
@@ -178,14 +175,14 @@ class Button extends Component {
   focus = () => _.invoke(this.ref, 'focus')
 
   handleClick = (e) => {
-    const { disabled, onClick } = this.props
+    const { disabled } = this.props
 
     if (disabled) {
       e.preventDefault()
       return
     }
 
-    if (onClick) onClick(e, this.props)
+    _.invoke(this.props, 'onClick', e, this.props)
   }
 
   handleRef = c => (this.ref = c)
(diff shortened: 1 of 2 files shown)


test/specs/elements/Button/Button-test.js
```diff
@@ -92,12 +121,19 @@ describe('Button', () => {
       shallow(<Button icon='user' />)
         .should.have.className('icon')
     })
+
+    it('adds className icon when true', () => {
+      shallow(<Button icon />)
+        .should.have.className('icon')
+    })
```
layershifter, line 128: Missing test

src/elements/Button/Button.js
```diff
@@ -256,56 +253,40 @@ class Button extends Component {
...
-    }
-
-    const labelElement = Label.create(label, { defaultProps: {
-      basic: true,
-      pointing: labelPosition === 'left' ? 'right' : 'left',
-    } })
-
-    if (labelElement) {
-      const classes = cx('ui', baseClasses, 'button', className)
+    if (!_.isNil(label)) {
```
layershifter, line 256: If you take attention to these conditions you can see, that in fact there are two conditions: when the `Button` has a container and has not.

src/elements/Button/Button.js
```diff
@@ -256,56 +253,40 @@ class Button extends Component {
...
-        </ElementType>
-      )
-    }
-
-    const labelElement = Label.create(label, { defaultProps: {
-      basic: true,
-      pointing: labelPosition === 'left' ? 'right' : 'left',
-    } })
-
-    if (labelElement) {
```
layershifter, line 275: Also, this condition doesn't makes sence, we should use there check for the `label` prop.

src/elements/Button/Button.js
```diff
@@ -256,56 +253,40 @@ class Button extends Component {
...
+      const labelElement = Label.create(label, { defaultProps: {
+        basic: true,
+        pointing: labelPosition === 'left' ? 'right' : 'left',
+      } })
 
       return (
         <ElementType {...rest} className={containerClasses} onClick={this.handleClick}>
           {labelPosition === 'left' && labelElement}
-          <button className={classes} ref={this.handleRef} tabIndex={tabIndex}>
+          <button className={buttonClasses} disabled={disabled} ref={this.handleRef} tabIndex={tabIndex}>
```
layershifter, line 267: HTML's `button` tag [can have](https://developer.mozilla.org/en/docs/Web/HTML/Element/button) a `disabled` attr.

levithomason:
Perfect, thank you.

levithomason:

PR #1782, opened 2017-06-19
style(typings): add new types

### LINKED ISSUE #1289: RFC: New types for typings

layershifter:
Fixes #1289.

This PR:
- add new types for typings as described in #1289;
- add new tests that check that all types are same as in `propTypes`.


layershifter:
@levithomason we have following validator:

```jsx
export const itemShorthand = (...args) => every([
  disallow(['children']),
  PropTypes.oneOfType([
    PropTypes.array,
    PropTypes.node,
    PropTypes.object,
  ]),
])(...args)
```
However, I don't remember any case when `itemShorthand` can be an array. Can you clarify?

layershifter (commented):
@levithomason Ready for review, also please pay attention to the question above.

levithomason:
>However, I don't remember any case when itemShorthand can be an array. Can you clarify?

I went back and checked some old issues and discussions around this.  I think we can remove the array type from `itemShorthand` now.  Give it a go and see if you run into any test failures or documentation issues and we'll go from there.

layershifter:
@levithomason you have already [removed](https://github.com/Semantic-Org/Semantic-UI-React/commit/afb911028b7aa3075a9ecb5b8bee5349f7fdb191#diff-e69b41f3944b4607ec8c9e9218a6eb4eL285) it in the `Tab`'s PR 😄 So, we're ready for review there.

levithomason:

PR #1786, opened 2017-06-22
feat(Form): safely prevent default on submit

### LINKED ISSUE #1769: RFC: Form submition
Hi! Since semantic-ui force us to use Form component as wrapper to Input, TextArea etc, It would cool to set default hof callback to prevent page reload if `action` prop have not provided
```javascript
(e, callback) => {
  !this.props.action && e.preventDefault()
   callback(e)
}
```

levithomason:
Fixes #1769 

This PR prevents default on submit events when there is no `action`.


levithomason:

PR #1788, opened 2017-06-23
fix(Form): call preventDefault in onSubmit only when it exists

### LINKED ISSUE #1790: Form: using a third-party form library causes "Uncaught TypeError: e.preventDefault is not a function"
Using React Formal (https://github.com/jquense/react-formal) library and Semantic UI React together causes an error on form submission.

(This bug has been fixed in a PR, see bottom of this post)

### Steps

I use the two library together like this:

```javascript
import {default as Formal} from 'react-formal'; //They both have a component called <Form>
import {Form, Button} from 'semantic-ui-react';

class Thing extends Component {
  handleSubmit = (data) => {
  }

  render() {
    return {
      <Form
        as={Formal}
        schema={mySchema}
        onSubmit={this.handleSubmit}
      >
        // fields here...
        <button type="submit">Submit</button>
      </Form>
    }
  }
}
```

When I click on the button to submit the form it gives an error.

### Expected Result

The form should be submitted.

### Actual Result

I receive the following error on the console:

```
Uncaught TypeError: e.preventDefault is not a function
```

### Version
Semantic UI React 0.69.0

A PR (https://github.com/Semantic-Org/Semantic-UI-React/pull/1788) has already been submitted that fixes this problem. A comment there requested an issue for a line comment. This is the issue.

layershifter:
Fixes #1790.

In #1786 we implimented a small, but cool enhancement, but as our user [reports ](https://github.com/Semantic-Org/Semantic-UI-React/issues/1769#issuecomment-310516466) it's not fully compatible with third-party libraries. This PR fixes this.

### DIFF
--- a/src/collections/Form/Form.js
+++ b/src/collections/Form/Form.js
@@ -94,10 +94,13 @@ class Form extends Component {
   static Select = FormSelect
   static TextArea = FormTextArea
 
-  handleSubmit = e => {
-    const { action, onSubmit } = this.props
-    if (!action) e.preventDefault()
-    if (onSubmit) onSubmit(e, { ...this.props })
+  handleSubmit = (e, ...args) => {
+    const { action } = this.props
+
+    // Heads up! Third party libs can pass own data as first argument, we need to check that it has preventDefault()
+    // method.
+    if (!action) _.invoke(e, 'preventDefault')
+    _.invoke(this.props, 'onSubmit', e, this.props, ...args)
   }
 
   render() {
@@ -130,7 +133,7 @@ class Form extends Component {
     const rest = getUnhandledProps(Form, this.props)
     const ElementType = getElementType(Form, this.props)
 
-    return <ElementType {...rest} onSubmit={this.handleSubmit} className={classes}>{children}</ElementType>
+    return <ElementType {...rest} className={classes} onSubmit={this.handleSubmit}>{children}</ElementType>
   }
 }
 
--- a/test/specs/collections/Form/Form-test.js
+++ b/test/specs/collections/Form/Form-test.js
@@ -44,42 +44,46 @@ describe('Form', () => {
 
   describe('onSubmit', () => {
     it('prevents default on the event when there is no action', () => {
-      const preventDefault = sandbox.spy()
+      const event = { preventDefault: sandbox.spy() }
 
-      shallow(<Form onSubmit={_.noop} />)
-        .simulate('submit', { preventDefault })
-      preventDefault.should.have.been.calledOnce()
+      shallow(<Form />)
+        .simulate('submit', event)
+
+      event.preventDefault.should.have.been.calledOnce()
     })
 
     it('does not prevent default on the event when there is an action', () => {
-      const preventDefault = sandbox.spy()
+      const event = { preventDefault: sandbox.spy() }
 
-      shallow(<Form action='do not prevent default!' onSubmit={_.noop} />)
+      shallow(<Form action='do not prevent default!' />)
         .simulate('submit', event)
-      preventDefault.should.not.have.been.called()
+
+      event.preventDefault.should.not.have.been.called()
     })
 
     it('is called with (e, props) on submit', () => {
       const onSubmit = sandbox.spy()
-      const e = { name: 'foo' }
+      const event = { name: 'foo' }
       const props = { 'data-bar': 'baz' }
 
       shallow(<Form {...props} onSubmit={onSubmit} />)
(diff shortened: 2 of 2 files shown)


src/collections/Form/Form.js
```diff
@@ -95,9 +95,10 @@ class Form extends Component {
   static TextArea = FormTextArea
 
   handleSubmit = e => {
-    const { action, onSubmit } = this.props
-    if (!action) e.preventDefault()
-    if (onSubmit) onSubmit(e, { ...this.props })
+    const { action } = this.props
+
+    if (!action) _.invoke(e, 'preventDefault')
```
levithomason: How about a line comment linking to the issue for this?  Looking at this later, I could see it getting reverted as it appears unnecessary.

darkadept: I've added an issue that describes why this fix is needed. (https://github.com/Semantic-Org/Semantic-UI-React/issues/1790)

src/collections/Form/Form.js
```diff
@@ -94,10 +94,14 @@ class Form extends Component {
   static Select = FormSelect
   static TextArea = FormTextArea
 
-  handleSubmit = e => {
-    const { action, onSubmit } = this.props
-    if (!action) e.preventDefault()
-    if (onSubmit) onSubmit(e, { ...this.props })
+  handleSubmit = (...args) => {
+    const { action } = this.props
+    const [e] = args
```
layershifter: @levithomason I've updated this part, too. Main idea to pass all args from third-party lib correctly.

Let me know what you think.

levithomason: Hm, I see the issue.  I don't like that this makes the signature awkward as the position of the `props` argument can never be known for certain anymore.  It depends on how many arguments the underlying component passed back.  I'd like to keep it `(e, data)` for consistency with all our signatures.

That said, perhaps we should extend out signatures to pass any other args _after_ these.  So:

```js
handleSubmit = (e, ...args) => {
  const { action } = this.props

  if (!action) _.invoke(e, 'preventDefault')
  _.invoke(this.props, 'onSubmit', e, this.props, ...args)
}
```

This way our signature is `(e, data, ...args)`.  Thoughts?

layershifter: Makes sence, it also will be more cleaner :+1:

test/specs/collections/Form/Form-test.js
```diff
@@ -44,21 +44,42 @@ describe('Form', () => {
...
     })
 
     it('does not prevent default on the event when there is an action', () => {
-      const event = { preventDefault: sandbox.spy() }
+      const preventDefault = sandbox.spy()
 
-      shallow(<Form action='do not prevent default!' onSubmit={sandbox.spy()} />)
+      shallow(<Form action='do not prevent default!' onSubmit={_.noop} />)
         .simulate('submit', event)
+      preventDefault.should.not.have.been.called()
```
levithomason: I don't understand how this is working, `event` is now undefined.

```js
.simulate('submit', event)
```

Shouldn't this be throwing?

test/specs/collections/Form/Form-test.js
```diff
@@ -44,21 +44,42 @@ describe('Form', () => {
...
-      const event = { preventDefault: sandbox.spy() }
+      const preventDefault = sandbox.spy()
 
-      shallow(<Form onSubmit={sandbox.spy()} />)
-        .simulate('submit', event)
-
-      event.preventDefault.should.have.been.calledOnce()
+      shallow(<Form onSubmit={_.noop} />)
+        .simulate('submit', { preventDefault })
+      preventDefault.should.have.been.calledOnce()
```
levithomason: This is OK for this PR but I'd prefer we don't make arbitrary refactors such as this.  It helps keeps the PRs smaller.

An exception would be refactors for consistency in coding standards, however, we use `const event = {}` quite a bit in tests.

test/specs/collections/Form/Form-test.js
levithomason: Heads up, [`window.event`](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) is a nonstandard proprietary IE global.  This test is not throwing because it is passing the `undefined` `window.event` to `simulate()`.

However, this test is also not testing anything since the event passed to simulate does not include the `preventDefault` spy under test.  I'll update these to assert on the actual event object again.

levithomason:
Rebased and updated.  Will merge on pass.

levithomason:

PR #1789, opened 2017-06-23
fix(children): handle empty array children

brianjd:
This PR properly tests if children are nil.  Fixes compatibility issue with Preact.


src/lib/childrenUtils.js
```diff
@@ -16,3 +16,15 @@ export const someByType = (children, type) => _.some(Children.toArray(children),
...
+ * Tests if any children.
+ * @param {Object} children The children prop of a component.
+ * @returns {Boolean}
+ */
+
+export const isNil = (children) => {
+  if (children === null) return true
+  if (Array.isArray(children) && children.length === 0) return true
+  return false
+}
```
layershifter: Please take attention, that [isNil](https://lodash.com/docs/4.17.4#isNil) checks if value is null or undefined.

brianjd: Understood.  Sorry, I undid the last commit.  I hastily made changes to satisfy the linter.

src/lib/childrenUtils.js
```diff
@@ -16,3 +16,14 @@ export const someByType = (children, type) => _.some(Children.toArray(children),
...
 export const findByType = (children, type) => _.find(Children.toArray(children), { type })
+
+/**
+ * Tests if any children.
+ * @param {Object} children The children prop of a component.
+ * @returns {Boolean}
+ */
+export const isNil = (children) => {
+  if (children == null) return true // eslint-disable-line
+  if (Array.isArray(children) && children.length == 0) return true // eslint-disable-line
```
levithomason: I'd prefer we do not ignore the linter.  Please use `===`.

src/lib/childrenUtils.js
```diff
@@ -16,3 +16,14 @@ export const someByType = (children, type) => _.some(Children.toArray(children),
  * @returns {undefined|Object}
  */
 export const findByType = (children, type) => _.find(Children.toArray(children), { type })
+
+/**
+ * Tests if any children.
```
levithomason: How about something like: `Tests if children are nil in React and Preact.`?

src/lib/childrenUtils.js
brianjd: The Lodash implementation uses `==`, the deeper `===` actually breaks things.  Suggestion?

src/lib/childrenUtils.js
```diff
@@ -16,3 +16,14 @@ export const someByType = (children, type) => _.some(Children.toArray(children),
...
+ * @param {Object} children The children prop of a component.
+ * @returns {Boolean}
+ */
+export const isNil = (children) => {
```
levithomason: Since this is now a very critical util for the lib, let's get some tests on it.  You can see other `lib` tests in `test/specs/lib`.

brianjd: Ok

src/lib/childrenUtils.js
levithomason: Sure, the `==` simply coerces null/undefined comparisons to return true.  We can just explicitly list those. `[].length` always returns a number so a triple equal there should be safe.  Something like:

```js
export const isNil = children => {
  return children === null
    || children === undefined
    || Array.isArray(children) && children.length === 0
}
```

src/lib/childrenUtils.js
brianjd: Done

src/lib/childrenUtils.js
brianjd: Done.

levithomason:
Looks great, thanks for the rapid updates!  We'll get this merged when CI passes.

brianjd:
@levithomason Thank you for reviewing so quickly!

levithomason:
No problem, if CI lint issues are getting in your way note that you can check them locally with:
```sh
npm run lint:fix
```

This will fix many issues and report the remaining errors.  You can safely ignore the `warnings` listed.

brianjd:
Yeah, I tried that before.  It complains about the configuration being invalid.  Which version of eslint are you running?

levithomason:
Hm, as long as you have node >=6 you should be able to do:

```
rm -rf node_modules
npm install

npm run lint:fix
```

You can see all the versions used in `package.json` in `devDependencies`.

brianjd:
Ah, my eyes were looking for "eslint*" entries, not "babel-eslint".  My mistake.

brianjd:
The global installation of eslint is being called, I had 4.0.  Reverting to 3.x resolved.  Should probably be pegged explicitly in the project.

levithomason:
Oh yucky, we're apparently using `eslint` by reaching through `eslint-config-ta` which installs `eslint@2.x` as a dependency.  Yep, this should be fixed.

That said, we're good here now.  Thanks much!

levithomason:
As a heads up, `babel-eslint` is just a parser that enables `eslint` to understand ES2015+.  The `eslint` CLI still does the linting.  I've opened #1792 to fix the missing dep issue, thanks for the report.

brianjd:
Awesome news! Sorry about the linting debacle 😅

levithomason:
No worries, totally on me there.

levithomason:

PR #1792, opened 2017-06-23
chore(package): add missing eslint dev dep

levithomason:
As noted in https://github.com/Semantic-Org/Semantic-UI-React/pull/1789#issuecomment-310732307 we didn't actually have `eslint` installed locally.  We were reaching through `eslint-config-ta` to get the cli.

This PR simply adds it and fixes that.


levithomason:

PR #1793, opened 2017-06-23
fix(TextArea): compute height when hidden

### LINKED ISSUE #1405: TextArea incorrect height with autoHeight prop when element is not visible
**Update**
Managed to find out exactly why this issue happens. Will come back and edit this with and jsfiddle that replicates exactly this issue. Spoiler, it has to do with setting tabular items.

I am getting this weird issue when using `Form.TextArea` with `autoHeight` and `disabled` props. The height of the textarea is not set correctly as you can see in the SS below:

screen shot 2017-03-01 at 12 13 47 pm

Here is the JSX:
```
<Grid>
  <Grid.Row columns={2}>
    <Grid.Column>
      <Header size="small">General Conditions</Header>
      <Form.TextArea label="Textarea 1" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
    <Grid.Column>
      <Header size="small">Special Conditions</Header>
      <Form.TextArea label="&nbsp;" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
  </Grid.Row>
  <Grid.Row columns={2}>
    <Grid.Column>
      <Form.TextArea label="Textarea 2" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
    <Grid.Column>
      <Form.TextArea label="&nbsp;" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
  </Grid.Row>
  <Grid.Row columns={2}>
    <Grid.Column>
      <Form.TextArea label="Textarea 3" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
    <Grid.Column>
      <Form.TextArea label="&nbsp;" defaultValue="Lorem Ipsum" disabled readOnly autoHeight />
    </Grid.Column>
  </Grid.Row>
</Grid>
```

Under inspect element, the style properties for the TextArea are being set as:
```
min-height: 0px;
resize: none;
height: 2px;
```

Where the `height: 2px` is obviously wrong.

### Version
0.67.0

levithomason:
Fixes #1405

This PR changes the height calculation do not depend on a visible DOM node.  We simply total up the box model height and text height rather than using the `scrollHeight`.

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -14,6 +14,8 @@ import {
  * @see Form
  */
 class TextArea extends Component {
+  state = {}
+
   static _meta = {
     name: 'TextArea',
     type: META.TYPES.ADDON,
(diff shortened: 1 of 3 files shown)


src/addons/TextArea/TextArea.js
```diff
@@ -83,29 +85,46 @@ class TextArea extends Component {
...
+    const resize = autoHeight ? 'none' : ''
+
     return (
       <ElementType
         {...rest}
         onChange={this.handleChange}
         ref={this.handleRef}
         rows={rows}
-        style={{ ...style, minHeight }}
+        style={{ resize, height, ...style }}
```
layershifter: Can we sort `resize` and `height` in alphabetical order? 😄

levithomason: NO!  😛

levithomason: Sorted all keys in objects and force pushed 👍

levithomason:

PR #1794, opened 2017-06-23
chore(package): use satisfied react version

levithomason:
image


levithomason:

PR #1795, opened 2017-06-24
fix(Dropdown): fix multiple problems

### LINKED ISSUE #1787: Dropdown: problems with minCharacters
### 1. Single select - Not allowing for typing
**Steps**
- Click the placeholder

**Expected Result**
- Able to start typing

**Actual Result**
- Not able to start typing

### 2. Single select - Active selection not greyed out
**Steps**
- Click the dropdown (behind the placeholder)
- Start typing (e.g. 'CDE') and select an option
- Click again

**Expected Result**
- Active selection is greyed out, to indicate that typing is allowed again

**Actual Result**
- Font color remains black

### 3. Multi select - Not allowing for typing at all
**Steps**
- Click the dropdown (doesn't matter where)

**Expected Result**
- Able to start typing

**Actual Result**
- Not able to start typing

### 4. User experience
These two items just don't feel intuitive:

1. There's no indication that there's a minimum number of characters required. You could put instructions outside the dropdown, but I should expect them inside while using the dropdown. Initially you could use the placeholder, but after selecting a value, the instructions are obviously gone.
2. From user perspective the arrow down is frozen most of the time.

### Version
v0.69.0

### Testcase
https://codepen.io/klerkendev/pen/PjKWyK

### LINKED ISSUE #1858: Dropdown: Filtered list resets when using arrow keys
### Steps
I'm using a `Dropdown` component. When entering a `searchQuery` and then using the `arrow keys` on my keyboard, the filter resets while cycling. This is only happening when `multiple=false`. In `multiple=true`, everything works properly.

### Expected Result
I would expect the same behavior for both modes.

### Actual Result
The filtered list resets

### Version
0.71.0

### Testcase
[Inside the documentation, scroll to Remote](https://react.semantic-ui.com/modules/dropdown)

[Link to demo video](http://webm.land/media/dLAK.webm)

layershifter:
Fixes #1787.
Fixes #1858.
[Codepen](https://codepen.io/klerkendev/pen/PjKWyK)

# Single select, placeholder

Click on placeholder doesn't capture the search input focus.

### Expected behaviour

e2ccbe57b3b7cb9527222b1be5b67634

# Single select, arrows

Click on `up` or `down` arrows resets current search value.

### Current behaviour

ec2e512c71c23889349a691a9051554c

### Expected behaviour

335f2250d304c78a79e0e88003d3122a

# Single select

Click on `dropdown icon` should always open the dropdown.

### Current behaviour

d9561f84b192b3556bda9120caf92cf4

### Expected behaviour

d84572e673d80f22ce2d665812a8b33d

# Multi select, click

Click the dropdown (doesn't matter where) doesn't make anything.

### Expected behaviour

38461b54e548206f3304e657c6e66f37


src/modules/Dropdown/Dropdown.js
```diff
@@ -650,7 +650,20 @@ export default class Dropdown extends Component {
 
     if (!search) return this.toggle(e)
     if (open) return
-    if (searchQuery.length >= minCharacters || minCharacters === 1) this.open(e)
+    if (searchQuery.length >= minCharacters || minCharacters === 1) {
+      this.open(e)
+      return
+    }
+    if (this.searchRef) this.searchRef.focus()
```
layershifter, line 673: @levithomason this method does not make me happy, do you have ideas how we can refactor it?

src/modules/Dropdown/Dropdown.js
```diff
@@ -650,7 +650,20 @@ export default class Dropdown extends Component {
...
+  }
+
+  handleIconClick = e => {
+    debug('handleIconClick()', e)
+
+    _.invoke(this.props, 'onClick', e, this.props)
+    // prevent handleClick()
+    e.stopPropagation()
+    this.toggle(e)
```
layershifter, line 682: I don't want to split this to separate method, however I don't see any other way. The initial idea was to compare `e.target` in `handleClick` this the `Icon`'s ref, but the `Icon` component is stateless and doesn't have refs.

src/modules/Dropdown/Dropdown.js
```diff
@@ -848,12 +861,12 @@ export default class Dropdown extends Component {
   // Setters
   // ----------------------------------------
 
-  setValue = (value) => {
+  setValue = (value, resetQuery = true) => {
```
layershifter: I'm not sure that the idea with `resetQuery` is the best possible, hovewer I don't see another path, as I see in all other cases `searchQuery` should be reset.

layershifter:
I performed required changes, in most cases the `Dropdown` is working as expected, but there are there some behaviours whose destiny should be solved before I will fix tests and add new.

@levithomason your review is very necessary, also I will be happy to hear @klerkendev there.

layershifter:
I extracted RFCs to separate issues: #1806 and #1807.

src/modules/Dropdown/Dropdown.js
```diff
@@ -949,6 +962,17 @@ export default class Dropdown extends Component {
   }
 
   // ----------------------------------------
+  // Overrides
+  // ----------------------------------------
+
+  handleIconOverrides = predefinedProps => ({
+    onClick: (e) => {
+      _.invoke(predefinedProps, 'onClick', e, predefinedProps)
```
layershifter, line 986: Please take an attention to this line, `Icon` component doesn't have `onClick` handler, so it doesn't have second argument. There are two possible solutions: remove the second argument or add `onClick` hanlder to the `Icon` component.

test/specs/commonTests/implementsShorthandProp.js
```diff
@@ -32,9 +33,11 @@ export default (Component, options = {}) => {
     ShorthandComponent,
     shorthandDefaultProps = {},
     shorthandOverrideProps = {},
+    strictAssert = true,
     requiredProps = {},
   } = options
   const { assertRequired } = helpers('implementsShorthandProp', Component)
+  const assertMethod = strictAssert ? 'contain' : 'containMatchingElement'
```
layershifter: Why? [`contains`](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/contains.md) is too strict while I added `onClick` handler to the `Icon` it was real pain. I think it's a quite elegant solution, as I remember we had same problem with the `Input` component.

layershifter (commented):
@levithomason we're ready for review there.

layershifter:
Requires #1845.

levithomason:
@layershifter #1845 has been merged, should be unblocked here.

src/modules/Dropdown/Dropdown.js
levithomason: Yep, I'd much rather have a separate explicit click handler here.  This is OK.  We are doing something entirely different when the icon is clicked.

src/modules/Dropdown/Dropdown.js
levithomason: I think we should split this into two methods.  `setValue` only sets the value and `clearSearchQuery` only clears the query.  It is easier to test and reason about complex interactions when the methods are small and explicit.

There are also enough use cases to warrant seprating them I think:

### componentWillMount

This calls `setValue`, however, it calls it with `this.state.value`.  This is a redundant operation.  It seems the only reason we call it is for the other logic in `setValue`, clearing the search query and setting the selected index.

We don't need to clear the query on mount, but we do need to set the selected index.  I think this is duplicating work, but at the least, we don't need to clear the query on mount.

### componentWillReceiveProps

This calls `setValue` when a new value is received.  Though this is correct, I also don't think we want to clear the user's query when this happens.  If a new value comes down while a query is entered, I think it should only change the value, not the query.

### makeSelectedItemActive

This method currently takes a `resetQuery` argument and passes it down to `setValue`.  I think this method should only make the selected item active and not do anything else.  The call sites for this method should be responsible for also deciding to clear the query or not:

- `moveSelectionOnKeyDown` - should never clear the query
- `selectItemOnEnter` - should only clear query when !multiple
- `handleBlur` - should clear the query (it is already only called when !multiple)

### removeItemOnBackspace

Currently resets the query but is not necessary, there is no query at this point.  Also, if the user enters a query and moves the cursor to the begining of the query and then presses backspace, we wouldn't want to clear the query either:
```
[ LabelA x ] [ LabelB x ] | My query
                          ^ cursor, backspace should not clear the query
```

I know we don't support this yet, but we technically could/should in the future.

### handleItemClick

This currently resets the query when an menu item is made active, however, I have always found this to be a usability issue with `multiple` dropdowns.  When there are lots of items, it is often more helpful to keep the query while clicking items.  Otherwise, the user has to enter the same query over and over for every item.  Example, selecting all states that start with `ma`:



If we separate `setValue` and `clearSearchQuery`, then we can call only `setValue` in the multiple condition in this method.  The other branch (not multiple) we can safely clear the query becuase we know the user is done making selections.

### handleLabelRemove

This will currently clear the query also but probably never should have done so.  Here I'm again selecting states with `ma`.  I "accidentally" select `Alabama` and continue making selections.  I then remove `Alabama` while I have a query entered and my query gets removed :/

src/modules/Dropdown/Dropdown.js
levithomason: I see the issue here, however, I think this is going to be rare enough of an edge case to leave as-is for now. We shall see :)

src/modules/Dropdown/Dropdown.js
levithomason: Hm, I don't think so.  React has no focus mechanism so we'll have to use a ref or querySelector here.  Since this section of logic is only reachable if this is a closed search dropdown, I think it makes sense.

What specifically are you not happy with and perhaps I can recommend better?

levithomason (changes requested):
I've left some notes regarding splitting `setValue`, `clearSearchQuery`, and `makeSelectedItemActive`.

src/modules/Dropdown/Dropdown.js
layershifter: Too many conditions for one method as I think.

src/modules/Dropdown/Dropdown.js
layershifter: @levithomason Thanks for a detailed feedback, I'm fully agree with your suggestions. Will make a changes soon

layershifter: I think that we should extract behaviour of `removeItemOnBackspace` to separate issue, I fully agree that we need to impliment it.

layershifter (commented):
@levithomason I've performed requested changes

levithomason:
This looks great, thanks for pressing through on this!

levithomason (approved):

levithomason:

PR #1796, opened 2017-06-24
docs(ComponentProps): fix compatibility with IE

layershifter:
IE doesn't support [`includes`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).


levithomason:

PR #1799, opened 2017-06-25
fix(AutoControlledComponent): use setState instead of an assign

### LINKED ISSUE #1682: Dropdown value not populated on initial render w/ react 16
### Steps
<Dropdown selection options={states} value='CA' placeholder='Select State' />

### Expected Result
California would be selected/active

### Actual Result
Placeholder is shown

### Version
semantic-ui-react@0.68.3
react@16.0.0-alpha.6

### Testcase
http://codepen.io/anon/pen/jmpavw

---

Upon further investigation it looks like semantic-ui-react is following some conventions not recommended by the react docs. Specifically I tracked the issue down to:

https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.68.3/src/lib/AutoControlledComponent.js#L149

> Never mutate this.state directly, as calling setState() afterwards may replace the mutation you made. Treat this.state as if it were immutable.

This turns out to be a messy problem. The dropdown is dependent on `this.state` being set sync (it uses `this.state.value` in order to find the `selectedIndex`), but the state needs to be set with the async `setState` method otherwise it gets blown out on the event loop that `setState` finally fires.

There may need to be a larger discussion around how controlled components behave, but I was able to side step the problem by replacing the above line with:

```
      this.state = _extends({}, this.state, initialAutoControlledState)
      this.setState(this.state)
```

The overall idea being it can take advantage of `this.state` being set sync but it doesn't get blown out because I also add the update to the `setState` queue.

### LINKED ISSUE #1761: AutoControlledComponent throws warning in React 16

layershifter:
Fixes #1761.
Fixes #1682.

In #1770, I suggested that the problem is in that `setState` is an async, I was right. `AutoControlledComponent` component's children such as `Dropdown` and  `Search` had own logic inside the `componentWillMount` method that relied on the current state.

The `componentWillMount` method was renamed to `constructor` as it's a single place where `state` is assignable. However, we need to have ability to assign an initial state, so I added the `initialState` method that accepts `props` as an argument.

@apiv Thanks for you suggestions in #1770.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -136,8 +136,6 @@ class Portal extends Component {
     type: META.TYPES.ADDON,
   }
 
-  state = {}
-
   componentDidMount() {
     debug('componentDidMount()')
     this.renderPortal()
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -68,8 +68,11 @@ export const getAutoControlledStateValue = (propName, props, state, includeDefau
 }
 
 export default class AutoControlledComponent extends Component {
-  componentWillMount() {
+  constructor(...args) {
+    super(...args)
+
     const { autoControlledProps } = this.constructor
+    const state = _.invoke(this, 'initialState', this.props) || {}
 
     if (process.env.NODE_ENV !== 'production') {
       const { defaultProps, name, propTypes } = this.constructor
@@ -130,7 +133,7 @@ export default class AutoControlledComponent extends Component {
     // Also look for the default prop for any auto controlled props (foo => defaultFoo)
     // so we can set initial values from defaults.
     const initialAutoControlledState = autoControlledProps.reduce((acc, prop) => {
-      acc[prop] = getAutoControlledStateValue(prop, this.props, this.state, true)
+      acc[prop] = getAutoControlledStateValue(prop, this.props, state, true)
 
       if (process.env.NODE_ENV !== 'production') {
         const defaultPropName = getDefaultPropName(prop)
@@ -146,7 +149,7 @@ export default class AutoControlledComponent extends Component {
       return acc
     }, {})
 
-    this.state = { ...this.state, ...initialAutoControlledState }
+    this.state = { ...state, ...initialAutoControlledState }
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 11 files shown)


src/addons/Portal/Portal.js
```diff
@@ -136,8 +136,6 @@ class Portal extends Component {
     type: META.TYPES.ADDON,
   }
 
-  state = {}
-
```
layershifter, line 140: `state` is assigned inside the constructor now, cruft.

src/modules/Dropdown/Dropdown.js
```diff
@@ -358,7 +358,6 @@ export default class Dropdown extends Component {
   static Menu = DropdownMenu
 
   componentWillMount() {
-    if (super.componentWillMount) super.componentWillMount()
```
layershifter, line 361: Cruft.

src/lib/AutoControlledComponent.js
```diff
@@ -68,8 +68,11 @@ export const getAutoControlledStateValue = (propName, props, state, includeDefau
 }
 
 export default class AutoControlledComponent extends Component {
-  componentWillMount() {
+  constructor(...args) {
+    super(...args)
+
     const { autoControlledProps } = this.constructor
+    const state = _.invoke(this, 'initialState', this.props) || {}
```
levithomason: How about we go with `createReactClass`'s method name `getInitialState()` here?

layershifter: I like, saw that you made changes :+1:

apiv:
Ah, assumed my solution wasn't ideal and closed my PR, guess I should have left it open 😂.

Really excited to finish our migration to React 16 now that this is fixed, thanks!

levithomason:

PR #1802, opened 2017-06-26
fix(Input): Allow "right" to be passed as `actionPosition`/`iconPosition` on Input element

eldarshamukhamedov:
`actionPosition` and `iconPosition` both default to "right", but prop-types for both warn when "right" is passed in. Came up in the following case:

```jsx
<Form.Input label="Some label" actionPosition="right">
  <input />
  {flag && <Button content="Sometimes Visible" />}
  <Button content="Always Visible" />
</Form.Input>
```

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -39,7 +39,7 @@ class Input extends Component {
     ]),
 
     /** An action can appear along side an Input on the left or right. */
-    actionPosition: PropTypes.oneOf(['left']),
+    actionPosition: PropTypes.oneOf(['left', 'right']),
 
     /** Primary content. */
     children: PropTypes.node,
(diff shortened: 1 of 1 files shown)


layershifter (commented):
I don't think that this PR will bring anything useful. The current idea of SUIR is to form a valid markup for  the usage with SUI.

SUI [has](https://github.com/Semantic-Org/UI-Input/blob/master/input.css#L419) definition for `left action`, but hasn't for `right`. So, `actionPosition` can't be `right`, too, as I think.

If we'll merge it, typings should be updated, too.

levithomason:
An `action` input defaults to right position.  Instead of `actionPosition` use `action`.

```diff
-<Form.Input label="Some label" actionPosition="right">
+<Form.Input label="Some label" action>
```

The `actionPosition` prop is only required when you want to change the position of the action.

PR #1808, opened 2017-06-29
style(typings): export Tab definitions

pierre-H:


index.d.ts
```diff
@@ -156,6 +156,8 @@ export { default as Sidebar, SidebarProps } from './dist/commonjs/modules/Sideba
 export { default as SidebarPushable, SidebarPushableProps } from './dist/commonjs/modules/Sidebar/SidebarPushable';
 export { default as SidebarPusher, SidebarPusherProps } from './dist/commonjs/modules/Sidebar/SidebarPusher';
 
+export { default as Tab, TabProps} from './dist/commonjs/modules/Tab';
+
```
layershifter: The `Tab` has also the `TabPane` subcomponent, let's export it, too

layershifter (changes requested):

index.d.ts
pierre-H: @layershifter You're right ! It's done.

layershifter (approved):
@pierre-H Thanks for update, LGTM

levithomason:

PR #1810, opened 2017-06-29
feat(Embed): add iframe shorthand

### LINKED ISSUE #1804: Embed: add shorthand for iframe
When embedding a youtube video the fullscreen option is disabled. This should be a property for the Embed component.

layershifter:
Fixes #1804.


layershifter:
@levithomason There is single failing test:

> iframe shorthand prop (common)
> ✖ renders a iframe from elements

I can't fix it, looks like a problem with default value `src`, but nothing changes if I define it in `shorthandDefaultProps`. BTW, we can simply pick the `strictAssert` feature from #1795.

layershifter:
Requires #1845.

layershifter (commented):
Unblocked, ready for review 👍

levithomason:
👍

levithomason:

PR #1811, opened 2017-06-29
fix(Tab): fix children propType

### LINKED ISSUE #1798: Tab.Pane's children data type needs to be updated
### Steps
Tab.Pane's children is supposed to be String according to PropType, but it accepts React component as well. So i think propType for children of Tab.Pane should be oneOfType([PropTypes.string, PropTypes.object]) instead of PropTypes.string

1. <Tab.Pane>String</Tab.Pane>
2. This should render ok.
3. 
```
<Tab.Pane>
     <Grid>
       <Grid.Column>Test</Grid.Column>
     </Grid>
   </Tab.Pane>
```
4. Warning : Invalid prop 'children' of type 'object' supplied to 'TabPane'

### Expected Result
Please check my comment above.

### Actual Result
Please check my comment above.

### Version
0.70.0

### Testcase
Sorry I don't have the test case.

layershifter:
Fixes #1798.

### DIFF
--- a/src/modules/Tab/TabPane.js
+++ b/src/modules/Tab/TabPane.js
@@ -53,7 +53,7 @@ TabPane.propTypes = {
   as: customPropTypes.as,
 
   /** Primary content. */
-  children: PropTypes.string,
+  children: PropTypes.node,
 
   /** Additional classes. */
   className: PropTypes.string,


levithomason:

PR #1812, opened 2017-06-29
style(typings): Add tab panes prop definition

pierre-H:


src/modules/Tab/Tab.d.ts
```diff
@@ -27,7 +27,7 @@ export interface TabProps {
   onTabChange?: (event: React.MouseEvent<HTMLDivElement>, data: TabProps) => void;
 
   /** Shorthand props for the Menu. */
-  panes?: any;
+  panes?: Array<{menuItem: string; render: () => React.ReactNode;}>;
```
layershifter: Can we make it multiline?

```ts
panes?: Array<{
  menuItem: string;
  render: () => React.ReactNode;
}>;
```

layershifter (changes requested):

src/modules/Tab/Tab.d.ts
pierre-H: @layershifter done

layershifter (commented):
@pierre-H Thanks again, LGTM

layershifter:
I've relaxed type for changes from #1823.

levithomason:

PR #1820, opened 2017-06-30
fix(Input): add typings update

layershifter:
`Input` is a class component.


levithomason:

PR #1823, opened 2017-06-30
fix(Tab): relax menuItem prop type

### LINKED ISSUE #1822: RFC: Tab, allow custom menuItem
The new `Tab` component is awesome but sometimes instead of the plain text menu item you might want to add for example a `Label` component to it.

It does currently works if you do the following:

```javascript
import React from 'react'
import { Label, MenuItem, Tab } from 'semantic-ui-react'

const panes = [
  {
    menuItem: <MenuItem>Tab1<Label>2</Label></MenuItem>,
    render: () => <Tab.Pane>Tab 1 Content</Tab.Pane>
  }
]

const TabExampleBasic = () => (
  <Tab panes={panes} />
)

export default TabExampleBasic


```
You do however get a prop type warning. I'm not sure if it's just a matter of changing `menuItem: PropTypes.string.isRequired` to `menuItem: PropTypes.node.isRequired` or if any other changes are required.

levithomason:
Fixes #1822 

The Tab was designed to allow any shorthand value to be passed as the `menuItem`.  The propType was limited to string.

This PR relaxes the propType and adds an example with shorthand `menuItem`s.

### DIFF
--- a/docs/app/Examples/modules/Tab/Usage/TabExampleCustomMenuItem.js
+++ b/docs/app/Examples/modules/Tab/Usage/TabExampleCustomMenuItem.js
@@ -1,0 +1,19 @@
+import React from 'react'
+import { Label, Menu, Tab } from 'semantic-ui-react'
+
+const panes = [
+  {
+    menuItem: { key: 'users', icon: 'users', content: 'Users' },
+    render: () => <Tab.Pane>Tab 1 Content</Tab.Pane>,
+  },
+  {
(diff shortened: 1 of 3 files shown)


layershifter (changes requested):
@levithomason ~~`typings` should be also updated, use `any` there.~~

Whoops, it will be done only in #1812.

layershifter (approved):

levithomason:

PR #1825, opened 2017-07-02
feat(Search): prop to return the active result

treyhuffine:
Return the active value in `<Search>` when navigating using the arrow keys

### DIFF
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -105,7 +105,7 @@ export interface SearchProps {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onResultSelect?: (event: React.MouseEvent<HTMLDivElement>, data: SearchProps) => void;
+  onResultSelect?: (event: React.MouseEvent<HTMLDivElement>, data: SearchResultData) => void;
 
   /**
    * Called on search input change.
@@ -114,6 +114,14 @@ export interface SearchProps {
    * @param {object} data - All props, includes current value of search input.
    */
   onSearchChange?: (event: React.MouseEvent<HTMLElement>, data: SearchProps) => void;
+
+  /**
+   * Called when the active selection index is changed.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onSelectionChange?: (event: React.MouseEvent<HTMLElement>, data: SearchResultData) => void;
 
   // ------------------------------------
   // Style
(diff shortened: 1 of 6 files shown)


src/modules/Search/Search.js
```diff
@@ -138,6 +138,14 @@ export default class Search extends Component {
     onResultSelect: PropTypes.func,
 
     /**
+     * Called when the active selection index is changed.
+     *
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
+    onActiveSelectionChange: PropTypes.func,
```
layershifter: I think that `onActiveSelectionChange` can be renamed to `onSelectionChange`.

src/modules/Search/Search.js
```diff
@@ -289,6 +297,13 @@ export default class Search extends Component {
     if (onResultSelect) onResultSelect(e, result)
   }
 
+  handleActiveSelectionChange = () => {
+    debug('handleActiveSelectionChange()')
+    const { onActiveSelectionChange } = this.props
+    const result = this.getSelectedResult()
+    if (onActiveSelectionChange) onActiveSelectionChange(result)
```
layershifter: We can exclude there if-condition if we will use lodash's `invoke`. Also, our callbacks should return `props` as the second argument.

```js
_.invoke(this.props, `onActiveSelectionChange`, e, { ...this.props, result})
```

layershifter (changes requested):
@treyhuffine thanks for PR. 

I left some feedback, it will awesome if we will also add new test for the prososed callback.

treyhuffine:
@layershifter thanks, I'll fix and resubmit.

layershifter:
@treyhuffine Thanks 👍 You can simply push your commits into the current branch instead of the new PR

src/modules/Search/Search.js
```diff
@@ -301,11 +316,11 @@ export default class Search extends Component {
     switch (keyboardKey.getCode(e)) {
       case keyboardKey.ArrowDown:
         e.preventDefault()
-        this.moveSelectionBy(1)
+        this.moveSelectionBy(e, 1)
```
layershifter, line 319: We will need `e` in `handleSelectionChange`.

src/modules/Search/Search.d.ts
```diff
@@ -105,7 +105,7 @@ export interface SearchProps {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onResultSelect?: (event: React.MouseEvent<HTMLDivElement>, data: SearchProps) => void;
+  onResultSelect?: (event: React.MouseEvent<HTMLDivElement>, data: SearchResultData) => void;
```
layershifter, line 108: `data` there also contains `result`.

layershifter (approved):
@levithomason we're for review there.

levithomason:

PR #1828, opened 2017-07-03
breaking(Search): update onResultSelect and onSearchChange signature

layershifter:
# BREAKING

This wasn't done in #951 and there were arguments for that reason, however I think that all our callbacks should be consistent.

onResultSelect | (e, result) | (e, { ...props, result })
onSearchChange | (e, value) | (e, { ...props, value })

### Before

```jsx
<Search onResultSelect={(e, result) => console.log(e, result)} />
<Search onSearchChange={(e, value) => console.log(e, value)} />
```

### After

```jsx
<Search onResultSelect={(e, data) => console.log(e, data.result)} />
<Search onSearchChange={(e, data) => console.log(e, data.value)} />
```

### DIFF
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -105,15 +105,15 @@ export interface SearchProps {
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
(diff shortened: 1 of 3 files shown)


levithomason:

PR #1833, opened 2017-07-05
fix(Dropdown): fix touch event issue on mobile devices

### LINKED ISSUE #1832: Dropdown: Touch event is not closing/collapsing the dropdown
### Description
When the dropdown component is used with mobile touch events the dropdown items list is not closed when touching the outer layer of the component (e.g.: "touchend" event on document or body). This is due to the fact that "touchend" event is not added/listed on componentDidUpdate lifecycle.

### Steps
1. Open the dropdown
2. Touch on outer layer using a mobile device
 
### Expected Result
The dropdown should close on touching the outer layer of the component (document or body) as in web devices when using the mouse & click on the outside layer.

### Actual Result
The dropdown remains open until a value is selected or the dropdown is touched again.

### Version
0.70.0

### Testcase
https://codepen.io/hetmann/pen/dRKJwQ

hetmann:
Fixes #1832.

When the dropdown component is used with mobile touch events the dropdown items list is not closed when touching the outer layer of the component (e.g.: "touchend" event on document or body). This is due to the fact that "touchend" event is not added/listed on componentDidUpdate lifecycle.

The fix was tested on iOS device & Chrome.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -453,6 +453,7 @@ export default class Dropdown extends Component {
       document.addEventListener('keydown', this.selectItemOnEnter)
       document.addEventListener('keydown', this.removeItemOnBackspace)
       document.addEventListener('click', this.closeOnDocumentClick)
+      document.addEventListener('touchend', this.closeOnDocumentClick)
       document.removeEventListener('keydown', this.openOnArrow)
       document.removeEventListener('keydown', this.openOnSpace)
       this.scrollSelectedItemIntoView()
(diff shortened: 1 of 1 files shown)


layershifter (commented):
I want to test this on my iOS device before merge.

hetmann:
@layershifter Thanks for reviewing it. What is the average time for merging to master & updating the npm package?

layershifter:
I'm not ready to give this PR a green light before I test it. Some time ago @levithomason said that we try to stick to the weekly schedule, but now he's very busy, so I can't promise anything.

layershifter:
I need some clarification there. We speaking about closing `Dropdown` when user performs a tap event without moving, i.e a click. I've tested this on my mobile with Android & Chrome and there is no a such problem. However, it is on the iPhone & Safafi, there is [react-tap-event-plugin](https://github.com/zilverline/react-tap-event-plugin) which solves it.

-------

As I see, this PR will break this behaviour and it's not correct.

ezgif com-resize

hetmann:
Thanks, I'll try the plugin & I'll let you know.

levithomason:
This is actually a bad design decision by Apple, IMO.  iOS devices only fire `onClick` events on elements that have a `cursor: pointer` style.  Without this style, they only fire touch events.

An easy fix is to apply `cursor: pointer` to the body on touch enabled devices.  This way, all taps result in click events:

```js
if ('ontouchstart' in document.documentElement) {
  document.body.style.cursor = 'pointer';
}
```

PR #1836, opened 2017-07-05
fix(Dropdown): accessibility problem (screen readers)

### LINKED ISSUE #1834: Dropdown: accessibility problem (screen readers)
### Description
If you use Dropdown passing the items using options, it is keyboard accessible but with a screen reader the options are not read aloud as you move through them.

### Steps
Reach a dropdowns with keyboard, and uses arrows keys to move through options
### Expected Result
When you move  through options, options are read aloud 
### Actual Result
Options are not read aloud 
### Version
0.70.0

### Testcase
e.g. https://react.semantic-ui.com/modules/dropdown#dropdown-example-inline

### 
I can contribute patching it modifying the Semantic-UI-React/src/modules/Dropdown/Dropdown.js, in function renderText = () => (line 1028),
adding the following aria labels to the redered div (when className='text'):
role='alert' aria-live='polite'

Sole-Valero:
Fixes #1834
Screen readers read aloud the different options.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1049,7 +1049,7 @@ export default class Dropdown extends Component {
     }
     if (classes === 'text') {
       return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
-
+    }
     return <div className={classes}>{_text}</div>
   }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1047,7 +1047,7 @@ export default class Dropdown extends Component {
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
     }
-    
+
     return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
   }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1047,8 +1047,13 @@ export default class Dropdown extends Component {
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
     }
-
-    return <div className={classes}>{_text}</div>
+    let finalRenderedText = null
+    if (classes === 'text') {
+      finalRenderedText = <div className={classes} role='alert' aria-live='polite'>{_text}</div>
+    } else {
+      finalRenderedText = <div className={classes}>{_text}</div>
+    }
+    return finalRenderedText
   }
 
   renderSearchInput = () => {
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1047,10 +1047,8 @@ export default class Dropdown extends Component {
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
     }
-    if (classes === 'text') {
-      return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
-    }
-    return <div className={classes}>{_text}</div>
+    
+    return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
   }
 
   renderSearchInput = () => {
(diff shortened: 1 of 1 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -1047,8 +1047,13 @@ export default class Dropdown extends Component {
...
     }
-
-    return <div className={classes}>{_text}</div>
+    let finalRenderedText = null
+    if (classes === 'text') {
+      finalRenderedText = <div className={classes} role='alert' aria-live='polite'>{_text}</div>
+    } else {
+      finalRenderedText = <div className={classes}>{_text}</div>
+    }
+    return finalRenderedText
```
layershifter: I think that we can make it simplier:

```jsx
if (classes === 'text') return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
return <div className={classes}>{_text}</div
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -1047,8 +1047,13 @@ export default class Dropdown extends Component {
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
     }
-
-    return <div className={classes}>{_text}</div>
+    let finalRenderedText = null
+    if (classes === 'text') {
```
layershifter: Can you clarify why we don't add `aria` attributes in other cases?

layershifter (changes requested):

src/modules/Dropdown/Dropdown.js
Sole-Valero: I don't add them because I don't know when are used the other cases. If you can explain me them, I can notice if they are needed or not, and also, which ones.  If I know in which examples from the documentation they are rendered...I can try also to improve them.

src/modules/Dropdown/Dropdown.js
Sole-Valero: Done

Sole-Valero: Ups...I had a mistake...I'll patch it

src/modules/Dropdown/Dropdown.js
```diff
@@ -1047,7 +1047,9 @@ export default class Dropdown extends Component {
     } else if (hasValue) {
       _text = _.get(this.getItemByValue(value), 'text')
     }
-
+    if (classes === 'text') {
+      return <div className={classes} role='alert' aria-live='polite'>{_text}</div>
+    }
```
levithomason: I believe `classes === 'text'` is the same as the final return case here.  When none of the above conditions match, the text is rendered as:

```html
<div class="text">File</div>
```



Is there a reason not to add `role='alert' aria-live='polite'` to the final return value instead?

Sole-Valero: @levithomason  we only need to mark as 'alert' a DIV which text changes and the user should know about the update. In the dropdowns case, the DIV which displays the value selected by the user uses the class 'Text'.  
To the best of my knowledge, renderText  returns a div which can be marked with classes: text , default and filtered. Looking at the examples of the documentation, I noticed than 'text' is always used to the DIV which contains the value selected, but I don't know when the other two options are used. Can you give me examples of them to decide if we need to add the alert tag in  the three cases?

src/modules/Dropdown/Dropdown.js
Sole-Valero: I updated the code. Now returns always the rendered text with the aria-labels suggested. I tested the documentation page, and all dropdown added work as in the main documentation page (http://react.semantic-ui.com)

src/modules/Dropdown/Dropdown.js
Sole-Valero: I updated the code. Now returns always the rendered text with the aria-labels suggested. I tested the documentation page, and all dropdown added work as in the main documentation page (http://react.semantic-ui.com)

Sole-Valero (commented):
Now, it always renders the text with the aria labels: role=alert aria-live='polite'

Sole-Valero (commented):
thanks, my editor doen't mark that lint error..sorry

levithomason:
@Sole-Valero 

`text` Shows the active value, though, this may change in the future to the selected (highlighted value in the menu) _or_ active value.

`filtered` Used to hide the text when a search is in progress.

`default` Used for placeholder text.

Let's merge this as is for now and update should we hit any future issues.  Thanks for the work and thoughtfulness here!

levithomason:

PR #1837, opened 2017-07-05
fix(Input): list prop should be passed to the input

### LINKED ISSUE #1835: Input: list prop should be passed to input
Input doesn't support 'list' attribute so that it can be used with a datalist, eg.   

```
<input list="languages">
<datalist id="languages">
 <option value="English"></option>
 <option value="Chinese"></option>
 <option value="Dutch"></option>
</datalist>
```

layershifter:
Fixes #1835.


levithomason:

PR #1845, opened 2017-07-09
tests(shorthand): allow non-strict assert

layershifter:
# Why? 

Extracted from #1795.

[`contains`](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/contains.md) is too strict, we should allow non-strict assert in these cases, useless conditions should gone.


test/specs/elements/Input/Input-test.js
```diff
@@ -63,6 +63,7 @@ describe('Input', () => {
   common.implementsHTMLInputProp(Input, {
     alwaysPresent: true,
     shorthandDefaultProps: { type: 'text' },
+    strictAssert: false,
```
levithomason: How about `assertExactMatch` or similar?  It is not clear _what_ is being strictly asserted.

layershifter: Yep, I like your suggestion, will change :+1:

layershifter:
@levithomason I've changed an option name :+1:

levithomason:

Issue #1851, opened 2017-07-11
SUI: release 2.2.11


layershifter:
# What is going there?

SUI 2.2.11 was released today and it's awesome, there are quite big release notes. Release notes were transformed to several PRs.

# PRs

- **package**: update SUI version (#1855)
- **Menu**: add `left` value to the position prop (#1852)
- **Modal**: add `mini` and `tiny` sizes (#1853)
- **Modal**: add `scrolling` prop to Modal.Content (#1857)
- **Step**: add `unstackable` prop (#1856)

PR #1852, opened 2017-07-11
feat(Menu): add `left` value to the position prop

layershifter:
Rel #1851.

A `MenuMenu` and a `MenuItem` [should](https://github.com/Semantic-Org/Semantic-UI/blob/2.2.10/src/definitions/collections/menu.less#L463) also accept `left` value on the `position` prop.


levithomason:

PR #1853, opened 2017-07-11
feat(Modal): add `mini` and `tiny` sizes

layershifter:
Rel #1851.
Rel [#5123](https://github.com/Semantic-Org/Semantic-UI/issues/5123).

A `Modal` should also accept `mini` and `tiny` values on the `size` prop.


levithomason:

PR #1855, opened 2017-07-12
chore(package): update SUI version

layershifter:
Ref #1851.

Updates SUI version in `package.json` and our examples.


levithomason:

PR #1856, opened 2017-07-12
feat(Step): add `unstackable` prop to Step.Group

layershifter:
Ref #1851.
Ref [#3714](https://github.com/Semantic-Org/Semantic-UI/issues/3714).

This PR adds `unstackable` prop to the `Step.Group` component.


levithomason:

PR #1860, opened 2017-07-12
docs(Portal): add Example under Addons Section of Docs

alexUXUI:
Documentation Update
Adds Portal Example

Hello, I have added a simple yet functional example of the SUI Portal component. It mainly demonstrates the callback functions that are available through the Portal component API. It is not exhaustive but would provide a good starting point for future iterations. 

There is only one problem with the way that this example fits (or does not) fit into the documentation website. Because of the positioning that is baked into of the Portal component, it is rendering underneath the left navigation of the website. I am unsure the best way to resolve as I do not want to add css. 

Many thanks to the maintainers and authors of this library. it is a true gem and I would love to be a part of the awesome community around it.


alexUXUI:
Hmm, I do not see a place to add test coverage to the example I have provided in docs/app/examples. I do see an example test in test/specs/docs/examples-test. I am not sure what I am missing in terms of testing but would like to provide - just need a little guidance.

layershifter (changes requested):
@alexUXUI Please restore files' permissions.

alexUXUI:
@layershifter oh, silly me! Will do now.

layershifter (approved):
@alexUXUI Thanks for PR. I've restored permissions and pushed some style changes @levithomason we're ready for review there

levithomason:
Great, thanks!

levithomason:

PR #1865, opened 2017-07-14
fix(AutoControlledComponent): rename state method

### LINKED ISSUE #1863: Tab: browser throwing a warning - getInitialState
**Steps to Reproduce**
1. Using the Tab component and importing it to my page

**Expected**
The Tab should do this: there shouldn't be any errors or warnings

**Result**
The Tab does not do this: the browser gives a warning:

> Warning: getInitialState was defined on Tab, a plain JavaScript class. This is only supported for classes created using React.createClass. Did you mean to define a state property instead?

The Tab.js (in the semantic-ui-react node_module) is using getInitialState which is not used in ES6
```
getInitialState() {
    return { activeIndex: 0 }
}
```
Could you please help me with this if there is a way to fix this or is changing the syntax to ES6 by using constructor and assigning the state is the only way to go?
Thanks!

**Testcase**
If the docs show the issue, use: https://react.semantic-ui.com/modules/tab#tab-example-secondary-pointing
Otherwise, fork this to get started: http://codepen.io/levithomason/pen/ZpBaJX

layershifter:
Fixes #1863.

In #1799 we improved compatibility with React 16, that PR introcuced `getInitialState` method as was [suggested](https://github.com/Semantic-Org/Semantic-UI-React/pull/1799#discussion_r124040515). However, this name is reserved and [produces](https://github.com/facebook/react/blob/b840229286ac2a82fa49553ce793cf7b953d1845/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L256) warnings even in [Fiber](https://github.com/facebook/react/blob/f9af9aacd3c14288cd5869ff23cc71cdfb54d447/src/renderers/shared/fiber/ReactFiberClassComponent.js#L171).

### DIFF
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -72,7 +72,7 @@ export default class AutoControlledComponent extends Component {
     super(...args)
 
     const { autoControlledProps } = this.constructor
-    const state = _.invoke(this, 'getInitialState', this.props) || {}
+    const state = _.invoke(this, 'getInitialAutoControlledState', this.props) || {}
 
     if (process.env.NODE_ENV !== 'production') {
       const { defaultProps, name, propTypes } = this.constructor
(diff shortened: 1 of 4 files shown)


levithomason:
For parity with our `autoControlledProps` and React's `getInitialState` I went with `getInitialAutoControlledState `.

levithomason:

PR #1872, opened 2017-07-18
feat(Responsive): add component

layershifter:
# Why?

This component is the idea picked from my production project. `Breakpoint` uses syntax from `Grid`'s device visibility, but has two important differences:
- it doesn't require `Grid` markup, so two close items can have a different visibility
- it doesn't render children when they doesn't visible

### Warning

`Breakpoint` adds a listener to `window` that can cause a performance issue, however it will be solved by #1733.

---

P.S. Better ideas about the component's name are welcome.


package.json
```diff
@@ -130,7 +130,7 @@
     "satisfied": "^1.1.0",
     "semantic-ui-css": "^2.2.11",
     "simulant": "^0.2.2",
-    "sinon": "^2.1.0",
+    "sinon": "^2.3.8",
```
layershifter: Sinon should be updated to support `value`:

```js
sandbox.stub().value()
```

src/addons/Breakpoint/Breakpoint.js
```diff
@@ -0,0 +1,149 @@
...
+
+  // ----------------------------------------
+  // Helpers
+  // ----------------------------------------
+
+  visible = () => {
+    const { only = '' } = this.props
+    const points = only.replace('large screen', 'largeScreen').split(' ')
+
+    return _.some(points, point => _.invoke(this, point))
```
levithomason: Where does `this.point` get defined?

```
_.invoke(this, point))
```

src/addons/Breakpoint/Breakpoint.js
```diff
@@ -0,0 +1,149 @@
...
+    const { width } = this.state
+
+    return width >= widescreen
+  }
+
+  // ----------------------------------------
+  // Helpers
+  // ----------------------------------------
+
+  visible = () => {
```
levithomason: Let's prefix bools `isVisible`

src/addons/Breakpoint/Breakpoint.js
layershifter: It equals to `this[point]()`, while `point` is an element of array that is splitted from `only` prop. The idea is to call breakpoint matchers (methods) until match.

layershifter:
Also need merge with `master`

src/addons/Breakpoint/Breakpoint.js
layershifter: I will prefix breakpoints with `is`, it will be easier to read

layershifter (commented):
@levithomason I've prefixed all methods that returns `boolean` with `is`. It will be easier to understand what it makes.

levithomason:
### Naming
Per the request for naming, I've renamed this component:

```jsx
<Responsive breakpoints={} />
```

I think this makes sense as many devs should be familiar with Bootstrap's ["responsive utilities"](https://getbootstrap.com/docs/3.3/css/#responsive-utilities), which are classes for the same purpose.

### breakpoints

I'm not sure about this prop.  It seems it would be most helpful to set this at the LESS/CSS level instead.  If we did allow overriding a specific `<Responsive />` element with different values, what about something like this instead?

```jsx
<Responsive minWidth={400} />
<Responsive maxWidth={800} />
```

What is the use case for overriding the value of `mobile`, etc?

layershifter:
## Naming

Agree about `Responsive`.

## breakpoints

I've add it because widths that SUI uses are [variables](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/globals/site.variables#L207) and it can be possible, that someone override them in own theme. Thats why I think we allow to change used values.

levithomason:
Oh, I see.  I was mistaken in thinking the `only='mobile'` here was relying on a `mobile` `className` and a CSS media query for visibility.  I see now that this is only mimicking the CSS visibility styles.

The only issue I see here is that users with custom CSS breakpoints in their theme would expect this component to respect them, but it won't.  Imagine I set my theme to have a mobile breakpoint of 400px:

```jsx
<Column only='mobile' />      // only <400px (uses my LESS/CSS value)
<Responsive only='mobile' />  // only <300px (uses defaultProps value)
```

I'm worried about adding complexity and configuration to the theming process.  It is quite cumbersome already.

My next suggestion was going to be that we should add CSS support for these utilties, however, it looks like SUI core isn't fond of [responsive visibility](https://github.com/Semantic-Org/Semantic-UI/issues?utf8=%E2%9C%93&q=is%3Aissue%20responsive%20visibility%20is%3Aclosed%20) or [responsive utilities](https://github.com/Semantic-Org/Semantic-UI/issues?utf8=%E2%9C%93&q=is%3Aissue%20responsive%20utilities%20is%3Aclosed%20).  There is a design decision to not permit them.  While I don't agree, chances of getting into CSS core are minimal.

I'm still not comfortable merging this while we have two components with `only='device'` which would render differently.  We need some kind of way to ensure all components use the same values.

layershifter:
You justly paid attention to these things, but unfortunately I see no other way to solve this problem.

I want to solve the problem with useless rendering, `Grid` renders all elements in contrast to this component. It's very usefull for navbars and other responsive things.

levithomason:
What if we remove the `only` prop and go with min/max width props instead?  Users can create the same values locally:

```jsx
const mediaQueries = { tablet: { minWidth: 768, maxWidth: 992 }  }

<Responsive {...mediaQueries.tablet} />
```

I would be OK with this as it doesn't suggest to the user that there is any relation between Responsive's `only` prop and Grid's `only` prop.

layershifter:
I will think about this, but `<Responsive only='mobile' />` looks more intuitive.

levithomason:
You're right, it is very intuitive.  If it used the same CSS as the Grid it would be a no brainer.  Since they will produce 2 different behaviors, I think it becomes unintuitive.

One hack I didn't want to mention was somehow creating a `ui grid` element and checking the styles.  I would rather not go down that road, but it may get us to where we need to be.

layershifter (commented):
@levithomason I refactored it to `minWidth` and `maxWidth` and added breakpoints as static properties. Also, it now uses `eventStack` and RAF for `resize` handler.

levithomason:
Cool, seems like a good middle ground 👍 Thanks much for the cooperative spirit.

levithomason:

Issue #1849, opened 2017-07-11
onKeyDown on <Modal/>


mkaradeniz:
Not sure if this was intended behavior, but I added an `onKeyDown` attribute on the `<Modal />`-Component to navigate through pictures shown inside Modal. But since Version 0.65.0 this doesn't work anymore.

### Steps
Adding an `onKeyDown` attribute to `<Modal />`
### Expected Result
Should fire the `onKeyDown` event.
### Actual Result
Does not fire the `onKeyDown` event.
### Version
0.65.0+

### Testcase
0.64.8: https://codepen.io/anon/pen/bRQQwV?editors=1111 (Working, outputs the pressed key to the console.)
0.65.0: https://codepen.io/anon/pen/OgaBYe?editors=1111 (Not working, does not output the pressed key to the console.)

layershifter:
### Why?

The breaking change was done in #1341, also you can read there why it was done.

### keyDown

`onKeyDown` doesn't work because `Modal` doesn't have a `tabIndex` attribute in `0.65.0+`, while it had in `0.68.8`. You can simply fix this, if you `tabIndex={-1}`, see [codepen](https://codepen.io/layershifter/pen/jwQjvw?editors=1111).

P.S. Thanks for a detailed report 👍

mkaradeniz:
Thank you!

Sorry for posting again, but I can't recreate the way it was working on `0.68.8` as it this still doesn't focus the `<Modal />` upon mount to `onKeyDown` right away without clicking on the Modal first. (Like mentioned in your referenced PR.) So I should tell react to focus it, but I'm not sure how exactly I go about this. If I set a ref to `<Modal/>` and let `ReactDOM` focus it on mounting, it doesn't work. I could wrap the `<Modal />` inside a wrapper div (which has `onKeyDown` and focus on that, but then Chrome thinks it is selected and draws a border around it. So how could I foucs the `<Modal />` manually?

layershifter:
It's a hateful workaround, but you can use there following:

```jsx
const id = 'my-modal'
const captureFocus = () => {
  const el = document.querySelector(`#${id}`)
  el.focus()
}

<Modal id={id} onOpen={captureFocus} trigger={<Button />} />
```

Also, check #1732 for more details, there is also a [codepen](https://codepen.io/anon/pen/JJdbjm) that will be helpful.

mkaradeniz:
Thank you, I'll try this!

Issue #1740, opened 2017-06-07
Button: add focus() method


Sole-Valero:
### Steps
It is not possible to call focus() method when you hav a ref to an Button component
<Button   ref={(button) => this.buttonRef=button} >Select All</Button>

This is a similar issue also solved for Input component:
https://github.com/Semantic-Org/Semantic-UI-React/issues/1475

### Expected Result
this.buttonRef.focus() --> send the focus to the button
### Actual Result
 Uncaught TypeError:this.buttonRef.focus is not a function

levithomason:
I think focusing a Button is common enough that this makes sense to implement this.  Thanks for the idea.

In the meantime, here's a workaround:

```jsx
const $button = document.querySelector('#my-button')
if ($button) $button.focus()

<Button id='my-button' />
```

Issue #1732, opened 2017-06-02
Modal should capture all key presses while being open


codeaid:
### Steps
I have a login form with username and password fields and a submit button. When the form is submitted, an API call is made and if the response is not successful a `Modal` is displayed notifying the user.

The problem is - while the modal is being displayed, the cursor is still in the username/password field and if I press Enter hoping to close the modal it resubmits the form even though the modal is covering the whole page. I've not found an easy way to automatically focus on the button in the modal.

### Expected Result
Modal/dimmer to capture key presses and act as a layer covering the whole page. At least give us an option which modal action to activate (default to) on open.

### Actual Result
Dimmed modal is acting as if it's "transparent" even though it's covering the whole page and not capturing focus.

### Version
0.68.5

### Testcase
https://codepen.io/anon/pen/wewbjX

Open console, focus on the username field and press Enter multiple times. Watch form submission messages being logged on every Enter press.

### Note
The same isssue is happening with the `Form` element if `loading` is set to `true`. One can keep submitting the form even though it's covered by the loading overlay.

Probably related to #1341.

layershifter:
@levithomason commented on this before [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/1341#issuecomment-281464149) and [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/1341#issuecomment-281527212). I agree with him, `Modal` shouldn't capture focus by default.

You can simply remove focus from `Form` in the `handleFormSubmit`, it will solve your issue. I will keep this issue opened before we'll get a response from @levithomason.

levithomason:
https://codepen.io/anon/pen/JJdbjm

**Add a button `id`**
```jsx
<Button id='my-button' color='green' inverted onClick={this.closeModal.bind(this)}>
  <Icon name='checkmark' /> Yes
</Button>
```

**After the state change to `open`, focus the button**
```jsx
this.setState({
  open: true,
}, () => {
  // after modal opens, focus the modal button
  const $button = document.querySelector('#my-button')
  if ($button) $button.focus()
})
```

codeaid:
@levithomason Isn't that a rather dirty solution having to use `document.querySelector`? I suppose I could use refs but it would be nice if there was a built-in functionality to specify which button/action to default to when the modal opens.

[...]

PR #1879, opened 2017-07-20
feat(Ref): implement `innerRef` on all components

### LINKED ISSUE #1321: Ensure DOM nodes can be found for `trigger` elements and `as` components
# See https://github.com/Semantic-Org/Semantic-UI-React/pull/1190 for context
**TL;DR** 
You cannot locate a DOM node for a functional component.  We allow users to pass components and elements into our components.  We then sometimes try to locate the DOM node of those passed in components.  This results in an error if the user passed a stateless functional component as there is no DOM node.

The solution, create a wrapper class and HOC that locates its own DOM node but renders the child or wrapped component in place.

## Progress to date

OK ugly code here, but, I've dug up the experiment I did for this here: http://www.react.run/HJa1BVYQl/24

This would need to be adapted to also work as a HOC rather than only on a child element:

```jsx
const RefsNotSupported = () = <p>I'm a functional component</p>

<RefsNotSupported ref={node => console.log(node)} />
// => null

const RefsSupported = Ref(RefsNotSupported)

<RefsSupported ref={node => console.log(node)} />
// => <p>I'm a functional component</p>
```

Finally, this component and HOC would need to be implemented everywhere we use a `trigger` and or `ref`.  It would also need to be applied in `src/lib/getElementType`, which powers the `as` prop, so that all `as` components can use `ref`s.

If anyone would like to pick this up, that would beyond amazing 👍

layershifter:
# WIP

Fixes #1321.

# Why?

`ref` isn't a prop, it has a same behaviour as `key` - it never passed to a component. However, there were numerous issues about refs (#1740, #1602, #1849, #1732 and more) and it's time to make something in this way. `document.querySelector` is a workaround,  but it's not a solution.

Problem is a well known in React's world, there is a long-live issue facebook/react#4213, there was also the RFC facebook/react#6974 from Dan.

# Variants

| Project | Prop | Example |
| ------- | ------ | ----- |
| [formsy-react-components](https://github.com/twisty/formsy-react-components) | `componentRef` & `innerRef` | [`refs.md`](https://github.com/twisty/formsy-react-components/blob/master/docs/refs.md) |
| [office-ui-fabric-react](https://github.com/OfficeDev/office-ui-fabric-react) | `componentRef` | [office-ui-fabric-react#1356](https://github.com/OfficeDev/office-ui-fabric-react/issues/1356) |
| [react-redux](https://github.com/reactjs/react-redux)  | `connect` HOC, `withRef` | reactjs/react-redux@2d3d0be |
| [react-router](https://github.com/ReactTraining/react-router) | `innerRef` | [`Link.md`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md) |

# TODO

- [ ] fix broken tests
- [ ] update `ref` and `trigger` usages
- [ ] update `context` and `scrollContext` usages


levithomason:
Very cool, thanks for getting this going.  It has been a long time running and will solve many issues.  ❤️ 

I assume this PR will also apply this to all the problem areas noted in the original issue?

layershifter:
Yep, however there is an unresolvable problem with a HOC.

```jsx
const RefsSupported = Ref(RefsNotSupported)
<RefsSupported ref={node => console.log(node)} />
```

There is no way to make this working because `ref` is not a prop, so component doesn't know anything about it. I've read many issues about this in React's repository and there is no any progress. I think that we should add a `componentRef` or `innerRef` prop to each component like React Bootstrap team done. But I need to do more progress before I can make final proposals.

levithomason:
Agreed, in my example on react.run I think I used `withDOMNode={this.handleDOMNode}` or something similar.  My thought was that it will not return the `ref`, but the DOM node.  Their site no longer works, so the example is not accessible.

test/utils/shallow.js
```diff
@@ -0,0 +1,18 @@
...
+  return wrapper
+}
+
+class Wrapper extends ShallowWrapper {
+  shallow(options) {
+    return this.single('shallow', n => dive(new Wrapper(n, null, options)))
+  }
+}
+
+export default (...args) => dive(new Wrapper(...args))
```
layershifter: @levithomason I need you there. 

After I injected `withRef` HOC to `getElementType` I received many broken tests. Half of then fixed with the code above, its idea to `dive` automatically when `shallow` is called.

Another half shoud be fixed with overriding of the `shallow` method on `ShallowWrapper`, but the inheritance **doesn't** work. WTF? Can you explain what I'm doing wrong? I think that I'm missing something basic, but at now it destroys my mind. `phantomjs` doesn't support neither `Proxy`, nor `__noSuchMethod__` :cry: 

Example test:

```jsx
 shallow(<Confirm cancelButton='foo' />)
   .find('Button')
   .first()
    .shallow()
    .should.have.text('foo')
```

```js
expected <Button /> to have text 'foo', but it has '<refHOC />'
```

layershifter (commented):
I returned to this PR and want to finish it before this weekend because I want to get initial review on it. I fixed all `elements` tests, will back to all other tomorrow.

test/utils/shallow.js
layershifter:
```js
const shallowMethod = ShallowWrapper.prototype.shallow

ShallowWrapper.prototype.shallow = function (...args) {
  return dive(shallowMethod.apply(this, ...args))
}
```

I found way to solve this, looks ugly, but works. Open for better solution

layershifter:
@levithomason I didn't finished this PR, but I want to get your review. There are many things that don't make me happy.

### getElementType

Now it returns an `withRef` hoc instead of a `tag`. It's not fatal, but requires the rewrite of code where we relied on `tagName` like [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-ce18ab6f0764450733723283e944fe22R73) and [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-94eeec164a2a14ec928ff1be8eeb8421R113).

### withRef

This hoc contains [magic conditions](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-d8a09b532f7f6537cea60a6fe7750ff9R19). I'm not sure that it's the best idea, but calls of `findDOMNode` aren't performance-free and we should omit them.

### Tests

Most of test are broken because now we get `withRef` hoc instead of component when we do `shallow` or `mount`. I've made [`shallow`](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-1c59c6a0ab24222629f28a45ebb410fbR1) wrapper, but it also adds problems:
- we can't use [`state()`](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-257bc851392ef05d97a06ce4a5958e23R56) on it
- [strange differences](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-3c4362b47157a758971c6d910292d36dR79) after it
- broken [`find()`](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-73f146001fc85c0e62bca4ce81f4e0daR35)

----

@levithomason I can fix all left tests, but I need comments and signal from you, that I've chosen the right way :smile:

src/hoc/withRef.js
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+
+import Ref from '../addons/Ref'
+
+const withRef = Child => class refHOC extends Component {
```
levithomason: Capitalized classes, please.

nit: Could also just be `WithRef`

src/lib/getElementType.js
```diff
@@ -19,7 +21,7 @@ function getElementType(Component, props, getDefault) {
   // computed default element type
 
   if (getDefault) {
-    const computedDefault = getDefault()
+    const computedDefault = typeof getDefault === 'function' ? getDefault() : getDefault
```
levithomason: If this change is made, the signature needs to be updated as well.

```jsx
 * @param {function} [getDefault] A function that returns a default element type.
```

That said, I'm not sure there is too much advantage to allowing a conditional here.

levithomason:
>Now it returns an withRef hoc instead of a tag. It's not fatal, but requires the rewrite of code where we relied on tagName like there and there.

As for the [image example](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-ce18ab6f0764450733723283e944fe22R73), this is actually a bug :)  The 3rd argument is a function:

```diff
-const ElementType = getElementType(Image, props, wrappedImage && 'div')
+const ElementType = getElementType(Image, props, () => wrappedImage && 'div')
```
**getElementType.js**
```jsx
* @param {function} [getDefault] A function that returns a default element type.
```

As for the [list item example](https://github.com/Semantic-Org/Semantic-UI-React/pull/1879/files#diff-94eeec164a2a14ec928ff1be8eeb8421R113), I'm not entirely sure right now.  One idea, but feels hacky :( is to have `withRef` add the `Child` component to the `refHOC` as a static, like `originalComponent`.  Then, we could check it something like:

```jsx
<Button as='div' />

const Button (props) => {
  const ElementType = getElementType(Button, props)

  // given the above invocation:
  // ElementType.originalComponent === 'div'

  return <ElementType />
}
```

I'm not super fond of this, but it seems like one of the only possibilities for "unwrapping" the HOC and accessing the original component.

levithomason:
I will have to come back and address the others another time!

levithomason:
Note, `context` props should also use the `Ref` component. This way, other React elements can be passed as `context` nodes.  See https://github.com/Semantic-Org/Semantic-UI-React/pull/1978#issuecomment-325505964.

src/hoc/withRef.js
layershifter: Nice catch :+1:

src/lib/getElementType.js
layershifter: Yes, I used `original` as you wrote below, so this change was reverted.

levithomason:
> withRef

Regarding the conditions, I'm not fond of them either but I think for the sake of progress we'll have to accept them for now.  `_meta` may go away in the future, but we can address it then.

> Tests

It seems this is a popular issue over at the Enzyme.  Seems some sort of `.until` method may be added to deal with the stack of HOCs issue.  There is this great https://github.com/airbnb/enzyme/issues/539#issuecomment-259843469 which offers [this gist](https://gist.github.com/matthieuprat/5fd37abbd4a4002e6cfe0c73ae54cda8).  It includes code for the method and lots of usage examples.

The basic idea is to shallow render the components all the way down to the one you want to test.  This is great as I believe it should also work in the cases you've noted having issues with your wrapper.

layershifter:
Requires #2142.

levithomason:
Unblocked, has some conflicts.

layershifter:
Needs #2144.

levithomason:
#2144 is now unblocked.

levithomason:
Note, #2144 was merged and this PR should be clear to move forward.

PR #1890, opened 2017-07-21
fix(Tab): use correct onTabChange activeIndex

### LINKED ISSUE #1888: Tab: data.activeIndex is overwritten in onTabChange for controlled Tab
### Steps
Make a controlled Tab with both activeIndex and onTabChange props.
Click an item in the tab menu.
Read the data.activeIndex value in onTabChange.

### Expected Result
data.activeIndex should be the index of the tab that was clicked.

### Actual Result
data.activeIndex is the value of the activeIndex prop passed to the Tab component.

### Version
0.71.1

### Testcase
https://codepen.io/anon/pen/QMLydJ

levithomason:
Fixes #1888

The proposed `activeIndex` was being overridden by spreading props _after_ it was set.

Tests were passing due to forgiving `calledWithMatch` logic.  I've updated tests to be more strict and added an explicit test.

I've also updated the controlled example to handle `onTabChange` in addition to the range input.  This way we have test and doc coverage on this.

### DIFF
--- a/docs/app/Examples/modules/Tab/Usage/TabExampleActiveIndex.js
+++ b/docs/app/Examples/modules/Tab/Usage/TabExampleActiveIndex.js
@@ -10,7 +10,8 @@ const panes = [
 class TabExampleActiveIndex extends Component {
   state = { activeIndex: 1 }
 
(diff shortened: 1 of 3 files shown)


levithomason:

PR #1891, opened 2017-07-21
fix(Button): fix animated prop type definition

ecozoic:
The type definitions for the Button component were slightly off. The `animated` prop can be a boolean in addition to the `'fade'` and `'vertical'` values. This matches the propTypes definition as well.


levithomason:
Perfect, thank you.

levithomason:

PR #1894, opened 2017-07-23
feat(Tab): Add vertical tabbing

mclarentgp:
In the current implementation it's impossible to set tabs to be in a vertical format.
This pr changes this.

### DIFF
--- a/src/collections/Grid/GridColumn.js
+++ b/src/collections/Grid/GridColumn.js
@@ -115,8 +115,6 @@ GridColumn.propTypes = {
 
 GridColumn.create = createShorthandFactory(
   GridColumn,
-  stretched => ({ stretched }),
-  width => ({ width }),
   children => ({ children })
 )
 
--- a/src/collections/Grid/GridColumn.js
+++ b/src/collections/Grid/GridColumn.js
@@ -4,6 +4,7 @@ import React from 'react'
 
 import {
   customPropTypes,
+  createShorthandFactory,
   getElementType,
   getUnhandledProps,
   META,
@@ -111,5 +112,12 @@ GridColumn.propTypes = {
   /** Represents width of column. */
   width: PropTypes.oneOf(SUI.WIDTHS),
 }
+
+GridColumn.create = createShorthandFactory(
+  GridColumn,
+  stretched => ({ stretched }),
+  width => ({ width }),
+  children => ({ children })
+)
 
 export default GridColumn
--- a/src/collections/Grid/GridColumn.js
+++ b/src/collections/Grid/GridColumn.js
@@ -113,9 +113,6 @@ GridColumn.propTypes = {
   width: PropTypes.oneOf(SUI.WIDTHS),
 }
 
-GridColumn.create = createShorthandFactory(
-  GridColumn,
-  children => ({ children })
-)
+GridColumn.create = createShorthandFactory(GridColumn, children => ({ children }))
 
 export default GridColumn
--- a/src/collections/Menu/Menu.d.ts
+++ b/src/collections/Menu/Menu.d.ts
@@ -81,7 +81,7 @@ export interface MenuProps {
   stackable?: boolean;
 
   /** A menu can be formatted to show tabs of information. */
-  tabular?: boolean | 'right' | 'left';
+  tabular?: boolean | 'right';
 
   /** A menu can be formatted for text content. */
   text?: boolean;
--- a/src/collections/Menu/Menu.d.ts
+++ b/src/collections/Menu/Menu.d.ts
@@ -81,7 +81,7 @@ export interface MenuProps {
   stackable?: boolean;
 
   /** A menu can be formatted to show tabs of information. */
-  tabular?: boolean | 'right';
+  tabular?: boolean | 'right' | 'left';
 
   /** A menu can be formatted for text content. */
   text?: boolean;
--- a/src/modules/Tab/Tab.js
+++ b/src/modules/Tab/Tab.js
@@ -9,6 +9,8 @@ import {
   getUnhandledProps,
   META,
 } from '../../lib'
+import Grid from '../../collections/Grid/Grid'
+import GridColumn from '../../collections/Grid/GridColumn'
 import Menu from '../../collections/Menu/Menu'
 import TabPane from './TabPane'
 
@@ -37,6 +39,9 @@ class Tab extends Component {
     /** Shorthand props for the Menu. */
     menu: PropTypes.object,
 
+    /** Shorthand props for the Grid */
+    grid: PropTypes.object,
+
     /**
      * Called on tab change.
      *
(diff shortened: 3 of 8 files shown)


src/collections/Menu/Menu.d.ts
```diff
@@ -81,7 +81,7 @@ export interface MenuProps {
   stackable?: boolean;
 
   /** A menu can be formatted to show tabs of information. */
-  tabular?: boolean | 'right';
+  tabular?: boolean | 'right' | 'left';
```
layershifter: Can you clarify necessity of `left` value? There [is no special](https://github.com/Semantic-Org/UI-Menu/blob/master/menu.css#L683) CSS for it, so it's `left` by default.

src/collections/Grid/GridColumn.js
```diff
@@ -112,4 +113,11 @@ GridColumn.propTypes = {
   width: PropTypes.oneOf(SUI.WIDTHS),
 }
 
+GridColumn.create = createShorthandFactory(
```
layershifter: `createShorthandFactory` accepts only two agruments, in this case correct usage is: 
```js
createShorthandFactory(GridColumn,  children => ({ children }))
```

layershifter (dismissed):
@mclarentgp Thanks for PR, I've left some comments, but I need time to pay more attention to this PR.

DENSENDEW:
impossible to set tabs to be in a vertical format.  project collaborators,

DENSENDEW:
? = missing data
Powered by Codecov. Last update 02daf0f...1f29cb3. Read the comment docs.?

src/collections/Menu/Menu.d.ts
mclarentgp: Ahh, sorry you are correct this doesn't need to be changed.  Will undo changes.

src/collections/Grid/GridColumn.js
mclarentgp: I will update this to the correct usage.

mclarentgp:
@DENSENDEW Sorry I don't quite understand?  I'm still fairly new to some of this and it's my first PR.

src/modules/Tab/Tab.js
```diff
@@ -96,19 +102,42 @@ class Tab extends Component {
     })
   }
 
+  renderColumn(columnProps) {
+    return GridColumn.create(columnProps)
+  }
```
levithomason: Since this is a pass-through method, let's just call `GridColumn.create` instead of `this.renderColumn` and drop this method.

src/modules/Tab/Tab.js
```diff
@@ -64,7 +69,8 @@ class Tab extends Component {
   ]
 
   static defaultProps = {
-    menu: { attached: true, tabular: true },
+    menu: { attached: true, tabular: true, vertical: false, fluid: false },
+    grid: { segmentWidth: 12, tabWidth: 4 },
```
levithomason: In keeping with the Tab terminology, let's go with `paneWidth`  instead of `segmentWidth`.

src/modules/Tab/Tab.js
```diff
@@ -96,19 +102,42 @@ class Tab extends Component {
...
+  renderVertical(menu) {
+    const { grid, panes } = this.props
+    const { activeIndex } = this.state
+
+    return (
+      <Grid>
+        {menu.props.tabular !== 'right' && this.renderColumn({ width: grid.tabWidth, children: menu })}
+        {this.renderColumn({
+          width: grid.segmentWidth,
+          children: _.invoke(_.get(panes, `[${activeIndex}]`), 'render', this.props),
```
levithomason: This line:
```js
_.invoke(_.get(panes, `[${activeIndex}]`), 'render', this.props)
```
is duplicated in the main `render()` method.  Let's pull this line into a method like `renderActivePane()` to contain it.

src/modules/Tab/Tab.js
```diff
@@ -96,19 +102,42 @@ class Tab extends Component {
...
+          stretched: true }
```
levithomason: Closing curly on the same line as the closing paren and wrapping curly please.
```diff
-  stretched: true }
-})
+  stretched: true,
+})}
```

src/modules/Tab/Tab.js
```diff
@@ -64,7 +69,8 @@ class Tab extends Component {
   ]
 
   static defaultProps = {
-    menu: { attached: true, tabular: true },
+    menu: { attached: true, tabular: true, vertical: false, fluid: false },
```
levithomason: No need to add `vertical: false, fluid: false` as these are not set by default.

src/modules/Tab/Tab.js
```diff
@@ -37,6 +39,9 @@ class Tab extends Component {
     /** Shorthand props for the Menu. */
     menu: PropTypes.object,
 
+    /** Shorthand props for the Grid. */
+    grid: PropTypes.object,
```
levithomason, line 43: Shorthand can be a string, number, props object, or an element.  It is then passed to a shorthand factory, such as `Grid.create(grid)`.

In this case, the user cannot pass normal Grid props such as `celled` or `padded` because this object is not used in a shorthand factory for the Grid. This is some kind of configuration rather than shorthand.

I'd like to reserve the `grid` prop for actual Grid props so users can control that if needed.  That said, I'm not so sure of the best API for these values.  Here are some ideas, would love feedback:

### Top level props
Make `paneWidth` and `tabWidth` top level props rather than nested in `grid`.  This makes room for `grid` to be true shorthand, however, those props are only applicable if `menu.vertical` is set so it might be a bit odd as well.

### Omit and use rest for shorthand
The current `grid` object needs to be spread on the `Grid` in order to be true Grid props.  We could leave it as is and just omit `paneWidth` and `tabWidth` from the `grid` value before spreading it:

```jsx
const { paneWidth, tabWidth, ...gridProps } = grid

<Grid {...gridProps>
  {GridColumn.create({ width: tabWidth })}
  {GridColumn.create({ width: paneWidth })}
</Grid>
```

***

I think I'm leaning toward the omit/rest approach, what are your ideas?

src/modules/Tab/Tab.js
```diff
@@ -96,19 +102,42 @@ class Tab extends Component {
...
+    const grid = menu.props.vertical === true ? this.renderVertical(menu) : null
     const rest = getUnhandledProps(Tab, this.props)
     const ElementType = getElementType(Tab, this.props)
 
     return (
       <ElementType {...rest}>
-        {menu.props.attached !== 'bottom' && menu}
-        {_.invoke(_.get(panes, `[${activeIndex}]`), 'render', this.props)}
-        {menu.props.attached === 'bottom' && menu}
+        {menu.props.vertical === true && grid}
```
levithomason: The `vertical` menu prop makes the case for two clean return values here.  We can also avoid creating a `null` grid in the case of non-vertical menus:

```js
if (menu.props.vertical) {
  return <ElementType {...rest}>{this.renderVertical(menu)}<ElementType>
}

// ...the old return value as-is
```

src/modules/Tab/Tab.js
mclarentgp: @levithomason I went ahead and made the omit/rest change.  Let me know if you decide you want to use a different approach.

levithomason:
I've pushed a minor update and will merge on pass.

levithomason:
Whoops, sorry for the delay!

levithomason:

PR #1896, opened 2017-07-24
FormFields marked as error should mark control as aria-invalid

fracmak:
Simple accessibility fix for FormFields


fracmak:
actually, on second thought, this is probably better handled outside the library, since aria-invalid can have multiple values, and this would force the user to always set true

PR #1898, opened 2017-07-25
fix(TextArea): update autoHeight on text wrap

### LINKED ISSUE #1843: TextArea: wrong autoHeight counting
### Steps

1. create `TextArea` instance with `autoHeight` property
`<TextArea autoHeight />`
2. put a content in there (just keep typing or copy & paste some text); while doing this do not press `<enter>` key

### Expected Result
TextArea grows with the content 

### Actual Result
TextArea keeps the original size; It only grows when pressing `<enter>` key. Actually it only counts newLines * lineHeight (simplified) which is IMO wrong

### Version
0.71.0

### Testcase
https://react.semantic-ui.com/addons/text-area#text-area-example-auto-height

### Contribution
I have a fix ready locally if interested

slunecniborecek:
Fixes #1843

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -79,6 +79,14 @@ class TextArea extends Component {
     _.invoke(this.props, 'onChange', e, { ...this.props, value })
   }
 
+  handleInput = e => {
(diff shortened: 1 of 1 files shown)


levithomason:

PR #1904, opened 2017-07-28
fix(Header): remove absent `mini` size

### LINKED ISSUE #1903: fix(Header): 'mini' size does not work
Document shows 'mini' as a size option. However it does not show the correct text size.

layershifter:
Fixes #1903.


levithomason:

PR #1918, opened 2017-08-01
feat(Dropdown): lazyLoad prop for menu items

dsirnk:
Lazy-loading options will be very useful and efficient when there are many options.
We could eventually bind it to a prop like 'lazyLoad={true}'.


dsirnk:
Sorry this is my first PR ever.

Anyone know what the issue is here?
I made the change so that the _dropdown_ only populates with options when it's open (lazy loading).
I made respective changes to the unit tests to account for the change. And ensuring that all the tests pass.

However, I don't know what **codecov** is and why the **_codecov/patch_** and **_codecov/project_** are failing.

levithomason:
Codecov seems to be broken.  It measures code coverage in our project.  They changed how they calculate that recently and it hasn't recovered.  You can safely ignore it for now.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -372,7 +372,9 @@ describe('Dropdown', () => {
         .find('DropdownItem')
         .first()
         .simulate('click')
-        .should.have.prop('active', true)
+        // `first().simulate('click')` closes the Dropdown so can't verify `prop('active', true)`
+        // TODO: Find a way to verify `first()` has `prop('active', true)`
+        // .should.have.prop('active', true)
```
levithomason: I would think re-opening the dropdown would retain the active item.  You could simply try something like (untested):

```jsx
wrapper
  .find('DropdownItem')
  .first()
  .simulate('click')

// re-open the dropdown
wrapper.setProps({ open: true })

wrapper
  .find('DropdownItem')
  .first()
  .should.have.prop('active', true)
```

levithomason (changes requested):
Would like to resolve https://github.com/Semantic-Org/Semantic-UI-React/pull/1918#discussion_r132563071 before merge.

levithomason:
I've rebased this to master, moved the logic to a `lazyLoad` prop, and cleaned up tests.  This can be merged after tests pass and a review is done.

levithomason (approved):

levithomason:

hubertguillemain:
I just created a bug (#2975) , as I don't see this working as I would expect.
But maybe I'm doing it wrong?
If so, can you please explain and add a sample usage to the doc site ?

PR #1934, opened 2017-08-06
HomepageLayout.js is not compatible in webpack: "^2.1.0-beta.21"

david30907d:
I update react syntax to make it compatible to new webpack version
My version is "^2.1.0-beta.21"
i get this error

```
ERROR in ./assets/js/semantic_layout.jsx
Module build failed: SyntaxError: Unexpected token (36:8)

  34 | 
  35 | export default class HomepageLayout extends Component {
> 36 |   state = {}
     |         ^
  37 | 
  38 |   hideFixedMenu = () => this.setState({ visible: false })
  39 |   showFixedMenu = () => this.setState({ visible: true })

 @ ./assets/js/index.jsx 3:0-51
```

hope my code work on others webpack too haha.

### DIFF
--- a/docs/app/Layouts/HomepageLayout.js
+++ b/docs/app/Layouts/HomepageLayout.js
@@ -34,18 +34,17 @@ const FixedMenu = () => (
 
 export default class HomepageLayout extends Component {
   constructor(props) {
-    super(props);
+    super(props)
     this.state = {}
-    
     this.hideFixedMenu = this.hideFixedMenu.bind(this)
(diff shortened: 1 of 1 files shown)


layershifter (changes requested):
Thanks for the contribution.

This example doesn't work because of the missing preset for `babel`. Our docs and code use the syntax and features from experimental presets. Example that you try to update uses [`class properties transform`](https://babeljs.io/docs/plugins/transform-class-properties/).

You can write code as you want, but we will not accept such changes. BTW, latest version of webpack is `3.4.1`.

PR #1895, opened 2017-07-24
chore(package): update all dependencies

layershifter:
# WIP

This PR:
- updates all deps to their latest version
- removes usage of `eslint-config-ta`, vanilla `airbnb` config will be used
- removes `del`, we have already `rimraf` for such cases
- removes `empty`, cruft

# TODO

- [x] fix lint errors
- [x] ES-ify configs
- [x] clean up `eslint` configs
- [x] merge #1938
- [x] merge #1939
- [x] merge #1949
- [x] replace [`babel-plugin-react-transform`](https://github.com/gaearon/babel-plugin-react-transform), it's deprecated (#1946)
- [x] replace [`babel-plugin-__coverage__`](https://github.com/dtinth/babel-plugin-__coverage__), it's deprecated (#1940)

PR #1939, opened 2017-08-07
style(tests): mixed lint updates

layershifter:
This PR contains only styling changes, it was splitted from #1895 to simplify review.

### DIFF
--- a/test/specs/commonTests/isConformant.js
+++ b/test/specs/commonTests/isConformant.js
@@ -5,10 +5,10 @@ import ReactDOMServer from 'react-dom/server'
 import * as semanticUIReact from 'semantic-ui-react'
 
 import { META } from 'src/lib'
+import { consoleUtil, sandbox, syntheticEvent } from 'test/utils'
 import helpers from './commonHelpers'
 import componentInfo from './componentInfo'
 import hasValidTypings from './hasValidTypings'
-import { consoleUtil, sandbox, syntheticEvent } from 'test/utils'
 
 /**
  * Assert Component conforms to guidelines that are applicable to all components.
@@ -76,13 +76,13 @@ export default (Component, options = {}) => {
       expect(isTopLevelAPIProp).to.equal(
         false,
         `"${constructorName}" is private (starts with  "_").` +
-        ' It cannot be exposed on the top level API'
+        ' It cannot be exposed on the top level API',
       )
 
       expect(isSubComponent).to.equal(
         false,
         `"${constructorName}" is private (starts with "_").` +
-        ' It cannot be a static prop of another component (sub-component)'
+        ' It cannot be a static prop of another component (sub-component)',
       )
     })
   } else {
@@ -100,7 +100,7 @@ export default (Component, options = {}) => {
       expect(isSubComponent).to.equal(
         true,
         `\`${constructorName}\` is a child component (has a _meta.parent).` +
-        ` It must be a static prop of its parent \`${_meta.parent}\``
+        ` It must be a static prop of its parent \`${_meta.parent}\``,
       )
     })
   }
@@ -158,7 +158,7 @@ export default (Component, options = {}) => {
     })
 
     it('renders as a ReactClass or passes "as" to the next component', () => {
-      class MyComponent extends React.Component {
+      class MyComponent extends React.Component { // eslint-disable-line react/prefer-stateless-function
         render() {
           return <div data-my-react-class />
         }
@@ -199,7 +199,7 @@ export default (Component, options = {}) => {
 
       Component.handledProps.should.to.deep.equal(expectedProps,
         'It seems that not all props were defined in Component.handledProps, you need to check that they are equal ' +
-        'to the union of Component.autoControlledProps and keys of Component.defaultProps and Component.propTypes'
+        'to the union of Component.autoControlledProps and keys of Component.defaultProps and Component.propTypes',
       )
     })
   })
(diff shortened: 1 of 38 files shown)


test/specs/commonTests/isConformant.js
```diff
@@ -264,10 +264,10 @@ export default (Component, options = {}) => {
 
         // Components should return the event first, then any data
         handlerSpy.calledWithMatch(...expectedArgs).should.equal(true,
-          `<${constructorName} ${listenerName}={${handlerName}} />\n` +
+          `${`<${constructorName} ${listenerName}={${handlerName}} />\n` +
           `${leftPad} ^ ${errorMessage}\n` +
-          'It was called with args:\n' +
-          JSON.stringify(handlerSpy.args, null, 2)
+          'It was called with args:\n'}${
+            JSON.stringify(handlerSpy.args, null, 2)}`,
```
levithomason: Hm, looks like the update here went awry.

test/specs/lib/numberToWord-test.js
```diff
@@ -26,7 +26,7 @@ describe('numberToWord', () => {
   })
 
   it('returns word if input is word', () => {
-    words.forEach((word) => numberToWord(word).should.equal(word))
+    words.forEach(word => numberToWord(word).should.equal(word))
```
levithomason, line 29: Why is it that in some (most?) cases the parens are added, but in others they are removed?

https://github.com/Semantic-Org/Semantic-UI-React/pull/1939/files#diff-ca99bb99c3b7ee63716bc6db755e5413R107

https://github.com/Semantic-Org/Semantic-UI-React/pull/1939/files#diff-ca99bb99c3b7ee63716bc6db755e5413R107

..etc

test/tests.bundle.js
```diff
@@ -9,7 +9,7 @@ const testsContext = require.context('./', true, /-test\.js$/)
 // only re-run changed tests, or all if none changed
 // https://www.npmjs.com/package/karma-webpack-with-fast-source-maps
 const __karmaWebpackManifest__ = []
-let runnable = testsContext.keys().filter((path) => __karmaWebpackManifest__.indexOf(path) >= 0)
+let runnable = testsContext.keys().filter(path => __karmaWebpackManifest__.indexOf(path) >= 0)
```
levithomason, line 12: Here's another example where parens are removed while most others are added.

levithomason (changes requested):
I believe there is syntax error in `isConformant.js`, see inline comment.

Other than this, it seems the style is also inconsistent between files (arrow func arg parens).  That said, I'd rather implement `eslint` rules that fixable for these so we don't have to manage them ourselves.

test/specs/lib/numberToWord-test.js
layershifter: https://github.com/airbnb/javascript#arrows--implicit-return

test/specs/commonTests/isConformant.js
layershifter: Yep, should be fixed. Will fix tomorrow

layershifter:
@levithomason  this come from airbnb rules, I gave link in a comment:

> If the function body consists of a single statement returning an expression without side effects, omit the braces

levithomason:
That would refer to the curly braces of the function body`{}`, I'm referring to the parentheses of the arguments `()`.

When there is a single argument, the changes here arbitrarily add parens in some cases and remove them in other cases.   I can't seem to figure out what the reason is.

**tests.bundle.js** (Removes parens)
```diff
-.filter((path) =>
+.filter(path =>
```

**test/specs/views/Feed/FeedEvent-test.js** (Adds parens)
```diff
-contentProps.forEach(propKey => {
+contentProps.forEach((propKey) => {
```

**test/specs/lib/numberToWord-test.js** (Removes parens)
```diff
-words.forEach((word) =>
+words.forEach(word =>
```

**test/specs/behaviors/Visibility/Visibility-test.js** (Adds parens)
```diff
-expectations.forEach(expectation => {
+expectations.forEach((expectation) => {
```

...there are many of these in the PR.

layershifter:
@levithomason This code was automatically fortmatted with `eslint` and `airbnb` preset. I split #1895 to smaller PRs only for review purposes.

It looks confusing, but it's correct. Take a look [8.2](https://github.com/airbnb/javascript#arrows--implicit-return), this paragraph affects `arrow-parens` rule, too.

```js
// bad
[1, 2, 3].map(number => {
  const nextNumber = number + 1;
  `A string containing the ${nextNumber}.`;
});

// good
[1, 2, 3].map(number => `A string containing the ${number}.`);

// good
[1, 2, 3].map((number) => {
  const nextNumber = number + 1;
  return `A string containing the ${nextNumber}.`;
});

// good
[1, 2, 3].map((number, index) => ({
  [index]: number,
}));
```

Paragraph [8.4](https://github.com/airbnb/javascript#arrows--one-arg-parens) also describes behaviour of parentheses.

```js
// bad
[1, 2, 3].map((x) => x * x);

// good
[1, 2, 3].map(x => x * x);
```

test/specs/commonTests/isConformant.js
layershifter: I've fixed this.

layershifter (commented):
@levithomason I've fixed formatting of that message and answered about parens.

levithomason:
Got it, thank you again.  I'm OK with this so long as it is fully automated.  👍

levithomason:

PR #1940, opened 2017-08-07
chore(package): replace deprecated babel-plugin-__coverage__

layershifter:
Rel #1895.
[Details](https://github.com/dtinth/babel-plugin-__coverage__).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -18,7 +18,7 @@
   "env": {
     "test": {
       "plugins": [
-        ["__coverage__", { "only": "src/" }]
+        ["istanbul", { "include": ["src"] }]
       ]
     }
   }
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -46,6 +46,7 @@ module.exports = (karmaConfig) => {
     singleRun: true,
     preprocessors: {
       // do not include 'coverage' preprocessor for karma-coverage
+      // TODO
```
levithomason: I believe this will also be true for istanbul since the bundle is instrumented before it ever gets to Karma (the browser).

That said, an easy check would be to run tests and open `coverage/lcov-report/index.html` in browser an see if it looks right.

levithomason (changes requested):
Let's resolve the TODO before merge.

karma.conf.js
layershifter: Yep, I doubt that I left it. However, I will check

layershifter: If's a cruft TODO, coverage is collected normally without coverage preprocessor

levithomason (approved):

levithomason:

PR #1946, opened 2017-08-08
chore(package): switch to react-hot-loader

layershifter:
Rel #1895.

[Details](https://github.com/gaearon/babel-plugin-react-transform)


webpack.config.js
```diff
@@ -165,18 +166,21 @@ if (!__TEST__) {
...
   //
   test: /\.js$/,
   exclude: /node_modules/,
-  use: {
-    loader: 'babel-loader',
-    options: {
-      cacheDirectory: true,
-    },
-  },
+  loaders: jsLoaders,
```
levithomason: I think this should be a `use: []` and not `loaders: []`, correct?

https://webpack.js.org/concepts/loaders/#configuration

layershifter: @levithomason Thanks for your eyes, [docs](https://github.com/gaearon/react-hot-loader/tree/master/docs#migration-to-30) sent me the wrong way

webpack.config.js
```diff
@@ -165,18 +166,21 @@ if (!__TEST__) {
 // ------------------------------------
 // Rules
 // ------------------------------------
+const jsLoaders = [{
+  loader: 'babel-loader',
+  options: {
+    cacheDirectory: true,
+  },
+}]
+if (__DEV__) jsLoaders.unshift('react-hot-loader/webpack')
```
levithomason, line 175: Does this work as is?  I'm not as solid with webpack 2 yet, but the [docs](https://webpack.js.org/concepts/loaders/#configuration) show only this syntax:

```diff
-if (__DEV__) jsLoaders.unshift('react-hot-loader/webpack')
+if (__DEV__) jsLoaders.unshift({ loader: 'react-hot-loader/webpack' })
```

layershifter: I can find this in docs, but it [works](http://andrewhfarmer.com/webpack-2-loader-variations/#3-rule-with-multiple-loaders-no-options).

levithomason:

PR #1949, opened 2017-08-09
style(mixed): update components styling

layershifter:
Stripped from #1895. Doesn't have conflicts with current code-style.


levithomason:
Wow, a ton of work.  Thanks for that!

I would like to consider using `prettier` in the future to enforce code style and relieve us having to do this again.  I've been using it in some other projects with moderately good results so far.  There are some things we'd have to give up since there is 0 configuration, however, the tradeoff and standardization is likely worth it in the long run.

levithomason:

PR #1950, opened 2017-08-10
docs(Icon): add FontAwesome attribution

### LINKED ISSUE #1901: Attribute the icons to FontAwesome
It took me a while, and a little source code digging to find where the icon set was coming from (FontAwesome). It's [mentioned on the other doc site](https://semantic-ui.com/elements/icon.html), but not on the react one. I think it should be added there, and the readme of this repo and the main repo.

layershifter:
Fixes #1901.


levithomason:
Thanks

levithomason:

PR #1951, opened 2017-08-10
docs(Button): add example with replicating button behaviour

### LINKED ISSUE #1797: Attached <Button /> does not respond on keyboard events
### Steps
Create a Button with an `attached` property.

### Expected Result
The button is accessible for keyboard users (hit Space or Enter and the button pressed).

### Actual Result
The button does not receive keyboard input.

### Version
0.70.0

### Testcase
https://codepen.io/anon/pen/jwGEYJ

layershifter:
Fixes #1797.

### DIFF
--- a/docs/app/Examples/elements/Button/Usage/ButtonExampleAttachedEvents.js
+++ b/docs/app/Examples/elements/Button/Usage/ButtonExampleAttachedEvents.js
@@ -1,0 +1,45 @@
+import React, { Component } from 'react'
+import { Button, Grid, Segment } from 'semantic-ui-react'
+
+export default class ButtonExampleAttachedEvents extends Component {
+  state = { log: [] }
+
+  handleClick = () => this.updateLog('Button received click with mouse')
+
+  handleKeyPress = (e) => {
+    if (e.charCode === 32 || e.charCode === 13) {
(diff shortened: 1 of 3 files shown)


levithomason:
How about we just bake this into the button itself?

layershifter:
Yes, it was an initial idea, but how we will handle it? In case when `Button` will be `button` it will receive `click` event when key pressed, too. So we will have double click there.

levithomason:
Could we not check `ElementType !== 'button'` then add `onKeyPress`, `tabIndex`, and `role`?  I'm imagining using the `overrideProps` in the factory to wrap the user's props for these.

I haven't tested any of this so I could be wrong, but it seems like we could bake in the same logic the user would manually add.

layershifter:
#1879 will disable ability to this check with `ElementType !== 'button'` because `ElementType` will return `withRef` HOC.

Possible we can handle this in event handler with `e.target.tagName`, if it will not be `button`, we call `onClick`. What you think?

levithomason:
Hmm, you are right.  Actually, anything other than a DOM component will pose this problem.  I think given the complication around trying to detect the type, we go with your current solution.

Thanks for the careful thought.

levithomason:

PR #1953, opened 2017-08-11
fix(TabPane): remove extra loading prop

### LINKED ISSUE #1941: Tab.Pane unknown prop warning when using as='div'
### Steps
1. Use Tab.Pane with ```as='div'```:
```js
<Tab.Pane as='div'>Tab1</Tab.Pane>
```
2. Receive react warning:
```
Warning: Unknown prop `loading` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
```

### Expected Result
Not to receive the warning.

### Actual Result
Warning displayed.

### Version
0.71.3

### Testcase
https://codepen.io/dfilipidisz/pen/yoMBRe

levithomason:
Fixes #1941

The `loading` prop was an artifact from a previous design which relied strictly on the Segment as the Tab Pane component.  The Segment does accept a loading prop.

We no longer need this as the `tab` class can accept an associated `loading` class, which is slightly different from the Segment's `loading` prop.  We already have this in place as well.

### DIFF
--- a/src/modules/Tab/TabPane.js
+++ b/src/modules/Tab/TabPane.js
@@ -32,7 +32,7 @@ function TabPane(props) {
   }
 
   return (
-    <ElementType {...calculatedDefaultProps} {...rest} className={classes} loading={loading}>
+    <ElementType {...calculatedDefaultProps} {...rest} className={classes}>
       {children}
     </ElementType>
   )


levithomason:

Issue #1954, opened 2017-08-11
RFC: Multiple Search Dropdown -> not clearing search query on select


creedmangrum:
This isn't a bug as the code is intentional, but I'm curious as to why in the search dropdown if multiple items are allowed, why the search query is not cleared on enter of item click? I would be happy to help out and make this a prop option, unless you would say otherwise. It seems like something useful as I can't really think of a use case for leaving the searchQuery after select, though I'm sure there are some. If this would be something that would help, happy to contribute, just want to make sure that I wouldn't be totally off base.

Thanks for all your work on this!

### Steps
search for an item on multi select, then select an item that matches

### Expected Result
i would expect the search query to be cleared

### Actual Result
search query remains. it is due to

```
    if (!multiple) this.clearSearchQuery();
```
that is in both on enter and item click.

### Version
0.71.4

levithomason:
**tl;dr** When selecting many options with the same query required, you have to enter after every selection. Since it poses no usability issues keeping the query, we've defaulted to this option.

For more, see https://github.com/Semantic-Org/Semantic-UI-React/pull/1795#discussion_r128802050

>### handleItemClick
>
>This currently resets the query when an menu item is made active, however, I have always found this to be a usability issue with `multiple` dropdowns.  When there are lots of items, it is often more helpful to keep the query while clicking items.  Otherwise, the user has to enter the same query over and over for every item.  Example, selecting all states that start with `ma`:
>
>![http://g.recordit.co/OqqnZDHR5u.gif](http://g.recordit.co/OqqnZDHR5u.gif)

Will move the discussion to the PR.

AndrewCi:
I believe there was a regression somehow and this issue should be re-opened #4063 @levithomason @creedmangrum

PR #1955, opened 2017-08-11
feat(Dropdown): add prop for clearing searchQuery data on item select or on enter

creedmangrum:
I wanted to add a prop to the dropdown where the search query could be cleared after selection on the multiple select, if desired. 

This maintains the current behavior of how everything works today, but if the prop is added then the search query will clear after item selection. 

I asked this question on issue #1954 but just figured it was easy enough to PR before waiting for an answer.


src/modules/Dropdown/Dropdown.d.ts
```diff
@@ -216,6 +216,9 @@ export interface DropdownProps {
   /** A shorthand for a search input. */
   searchInput?: any;
 
+  /** Whether the searchQuery is cleared or not when selecting an item from search. */
+  searchQueryClearOnSelect?: boolean;
```
levithomason: Nit: Could we go with `clearSearchQueryOnSelect`?

levithomason (changes requested):
I've requested a prop name change above.  We just need some tests for this prop:

```js
describe('clearSearchQueryOnSelect', () => {
  it('does not clear the query by default on item click', () => {})
  
  it('does not clear the query by default on enter', () => {})
  
  it('clears the query on item click', () => {})
  
  it('clears the query on enter', () => {})
})
```

src/modules/Dropdown/Dropdown.d.ts
creedmangrum: not a nit at all, was hoping for some specific feedback on the naming anyway. i just put it near the search ones as it felt grouped, but really was unsure about naming convention. will switch to your suggestion.

creedmangrum:
thanks for the feedback. will add the tests and prop name change.

creedmangrum:
okay should be good to go!

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1659,6 +1659,80 @@ describe('Dropdown', () => {
...
+        .simulate('click')
+        .setState({ searchQuery })
+
+      wrapper
+        .find('DropdownItem')
+        .first()
+        .simulate('click', nativeEvent)
+
+      // cleared search query
+      wrapper.should.have.state('searchQuery', searchQuery)
```
levithomason: Comment needs updated, or removed :)

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1659,6 +1659,80 @@ describe('Dropdown', () => {
...
+
+      wrapper
+        .simulate('click')
+        .setState({ searchQuery })
+
+      domEvent.keyDown(document, { key: 'ArrowDown' })
+      domEvent.keyDown(document, { key: 'Enter' })
+
+      // cleared search query
+      wrapper.should.have.state('searchQuery', searchQuery)
```
levithomason: Comment out of date

levithomason (changes requested):
There are a couple out of date comments that should be updated or removed.

We merge another PR which handles clearing the search query when there is only one option.  This PR has some conflicts that need to be resolved.

creedmangrum:
Great taken care of. Sorry about the bad comments in the tests. Whoops.

creedmangrum:
I think that the other PR that was merged probably solves for the use case, where the search query should be cleared if it is the available options === 1. I am likely going to close this, unless you feel that it still makes sense to do.

levithomason:
Fewer options in the Dropdown, the happy I am 😄!  This component is quite complex and getting hairy to work on.

Assuming we're all good here, I'll take your suggestion and close this.  Let me know if we need to reopen it.

PR #1963, opened 2017-08-14
fix(Search): only call onBlur & onFocus event handler once

### LINKED ISSUE #1962: SearchField triggers blur & focus event twice
### Steps
Click in the SearchField to set focus

### Expected Result
The focus event handler is called once.

### Actual Result
The focus event handler is called twice. the same applies to blur event.

### Version
0.71.4

### Testcase
https://codepen.io/anon/pen/rzGzPp?editors=1111

chopstikk:
Fixes #1962 

Search component onBlur and onFocus event handers should not be called twice.

### DIFF
--- a/test/specs/modules/Search/Search-test.js
+++ b/test/specs/modules/Search/Search-test.js
@@ -525,6 +525,28 @@ describe('Search', () => {
     })
   })
 
+  describe('onBlur', () => {
+    it('is called with (event, data) on search input blur', () => {
+      const spy = sandbox.spy()
+      wrapperMount(<Search onBlur={spy} />)
+        .simulate('blur', nativeEvent)
+
+      spy.should.have.been.calledOnce()
+      spy.should.have.been.calledWithMatch(nativeEvent)
+    })
+  })
+
+  describe('onFocus', () => {
+    it('is called with (event, data) on search input focus', () => {
+      const spy = sandbox.spy()
+      wrapperMount(<Search onFocus={spy} />)
+        .simulate('focus', nativeEvent)
+
+      spy.should.have.been.calledOnce()
+      spy.should.have.been.calledWithMatch(nativeEvent)
+    })
+  })
+
   describe('onResultSelect', () => {
     let spy
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


layershifter (changes requested):
@chopstikk Thanks for PR. However, I think that we should add tests there to omit such cases in future.

Something like:
```js
describe('onFocus', () => {
  it('is called with (e, data)', () => {
    const onFocus = sandbox.spy()
    wrapperMount(<Search onFocus={onFocus} />)
    
    wrapper.simulate('focus')

    onFocus.should.have.been.calledOnce()
    onFocus.should.have.been.calledWithMatch({ }, { options })
  })
})
```

test/specs/modules/Search/Search-test.js
```diff
@@ -525,6 +525,28 @@ describe('Search', () => {
     })
   })
 
+  describe('onBlur', () => {
+    it('is called with (event) on search input blur', () => {
+      const spy = sandbox.spy()
```
layershifter: Please name `spy` as event handler, i.e. `onBlur`.

test/specs/modules/Search/Search-test.js
```diff
@@ -525,6 +525,28 @@ describe('Search', () => {
...
+      wrapperMount(<Search onBlur={spy} />)
+        .simulate('blur', nativeEvent)
+
+      spy.should.have.been.calledOnce()
+      spy.should.have.been.calledWithMatch(nativeEvent)
```
layershifter: Let's also check that second argument provides a valid props:

```diff
-wrapperMount(<Search onBlur={spy} />)
+wrapperMount(<Search onBlur={spy} results={options} />)

-spy.should.have.been.calledWithMatch(nativeEvent)
+spy.should.have.been.calledWithMatch(nativeEvent, { onBlur, results: options })
```

test/specs/modules/Search/Search-test.js
```diff
@@ -525,6 +525,28 @@ describe('Search', () => {
     })
   })
 
+  describe('onBlur', () => {
+    it('is called with (event) on search input blur', () => {
```
layershifter: Let's also correct a title `is called with (e, data) on...`

layershifter (changes requested):

layershifter (approved):

levithomason:
I love you two, thanks! 😄

levithomason:

PR #1972, opened 2017-08-16
fix(Input|Button|TextArea): add `focus` method to typings

### LINKED ISSUE #1970: Input: 'focus' method does not exist in typings
### Steps

This example was tested using [create-react-app-typescript](https://github.com/wmonk/create-react-app-typescript)

```
import * as React from 'react';
import * as ReactDOM from 'react-dom';
import { Input } from 'semantic-ui-react';

ReactDOM.render(
    <Input ref={(i) => { if (i) { i.focus(); } }} />,
    document.getElementById('root') as HTMLElement
);
```

### Expected Result
It compiles and work.
The same example works with Babel using ES6 only (no TypeScript, as with [create-react-app](https://github.com/facebookincubator/create-react-app)).

### Actual Result
It fails to compile with error:

```
(6,37): error TS2339: Property 'focus' does not exist on type 'Component<InputProps, ComponentState>'.
```
### Version
0.71.4
(React 15.6.1)

layershifter:
Fixes #1970.

### DIFF
--- a/src/addons/TextArea/TextArea.d.ts
+++ b/src/addons/TextArea/TextArea.d.ts
@@ -31,6 +31,8 @@ export interface TextAreaOnChangeData extends TextAreaProps {
   value?: string;
 }
 
-declare const TextArea: React.ComponentClass<TextAreaProps>;
+declare class TextArea extends React.Component<TextArea, {}> {
+  focus: () => void;
+}
 
 export default TextArea;
--- a/src/elements/Button/Button.d.ts
+++ b/src/elements/Button/Button.d.ts
@@ -101,12 +101,12 @@ export interface ButtonProps {
   toggle?: boolean;
 }
 
-interface ButtonComponent extends React.ComponentClass<ButtonProps> {
-  Content: typeof ButtonContent;
-  Group: typeof ButtonGroup;
-  Or: typeof ButtonOr;
-}
+declare class Button extends React.Component<ButtonProps, {}> {
+  static Content: typeof ButtonContent;
+  static Group: typeof ButtonGroup;
+  static Or: typeof ButtonOr;
 
-declare const Button: ButtonComponent;
+  focus: () => void;
+}
 
 export default Button;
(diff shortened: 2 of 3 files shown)


levithomason:

codeaid:
This change completely breaks the component. `focus` attribute alongside others is now always required. `<TextArea />` results in:

```
error TS2322: Type '{}' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<TextArea> & Readonly<{ children?: ReactNode; }> & ...'.
  Type '{}' is not assignable to type 'Readonly<TextArea>'.
    Property 'focus' is missing in type '{}'.
```
If I add `focus={null}` it then says:

```
error TS2322: Type '{ autoHeight: true; disabled: boolean; ref: (textarea: any) => void; value: string; onChange: (ev...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<TextArea> & Readonly<{ children?: ReactNode; }> & ...'.
  Type '{ autoHeight: true; disabled: boolean; ref: (textarea: any) => void; value: string; onChange: (ev...' is not assignable to type 'Readonly<TextArea>'.
    Property 'setState' is missing in type '{ autoHeight: true; disabled: boolean; ref: (textarea: any) => void; value: string; onChange: (ev...'.
```

If you look at the `React.Component` declaration (which is the class `TextArea` now extends, you can see that it expects quite a few methods to be present:

```typescript
class Component<P, S> {
    constructor(props?: P, context?: any);
    setState<K extends keyof S>(f: (prevState: S, props: P) => Pick<S, K>, callback?: () => any): void;
    setState<K extends keyof S>(state: Pick<S, K>, callback?: () => any): void;
    forceUpdate(callBack?: () => any): void;
    render(): JSX.Element | null | false;

   // etc.
```
I've managed to fix it by changing the declaration to the following:

```typescript
declare class TextArea extends React.Component<TextAreaProps, {}> {
  focus: () => void;
}
```
It looks like `TextArea` was copied instead of `TextAreaProps`.

codeaid:
I've created a pull request to fix the issue. Please find a link to it above this comment.

PR #1973, opened 2017-08-16
fix(Popup): change position(right/left) for small width components

vipul-21:
The position of the Popup on elements with smaller width(<25px) not aligned.

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -140,6 +140,20 @@ export default class Popup extends Component {
 
   state = {}
 
+  computePopupOffset(width) { // computing the left/right offset based on width
+    let offset = 0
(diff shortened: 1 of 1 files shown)


src/modules/Popup/Popup.js
```diff
@@ -140,6 +140,20 @@ export default class Popup extends Component {
 
   state = {}
 
+  computePopupOffset(width) { // computing the left/right offset based on width
+    let offset = 0
+    if (width > 17 && width <= 25) {
+      offset = 5.8
```
layershifter: Please use `return` there:
```js
if (width > 17 && width <= 25) return 5.8
```

layershifter (changes requested):

PR #1976, opened 2017-08-17
feat(Tab): add `renderActiveOnly` prop

### LINKED ISSUE #1877: <Tab /> re-renders child pane on switching.
### Steps
The default behavior of the `<Tab />` component seems to be to remove it's active `<Tab.Pane>` from the DOM every time it is made inactive, and to push it back in the DOM when it is selected and made active. This re-renders the child components of `<Tab.Pane>` and is a problem if you have children like Leaflet or CesiumJS which internally manage their display state, as this state is reset on every re-render.

### Expected Result
One should expect that the default behavior would be to keep all Tab.Pane children rendered and in the DOM, and only display the one that is active. The inactive tabs should be hidden from the display while keeping them in the DOM.

### Actual Result
Inactive Tabs get removed from the DOM and pushed in again when made active. This behavior breaks the flow of many stateful child components, by resetting their internal state.

### Version
v0.71.0

Any quick fixes would be appreciated, while this gets fixed in main.

layershifter:
Fixes #1877.


layershifter:
Has merge conflicts

layershifter (commented):
Solved merge conflict and merged with `master`, ready for review

levithomason:

drheinheim:
Fyi, this change does not work. Setting `renderActiveOnly={false}` results in no panes being rendered.

layershifter:
It's not enough to set this prop to `false`, take a look on [this](https://react.semantic-ui.com/modules/tab#tab-example-basic-all) and [this](https://react.semantic-ui.com/modules/tab#tab-example-pane-shorthand) examples.

levithomason:
>It's not enough to set this prop to false,

We should make it so.

drheinheim:
Thanks for pointing out those examples. They worked. I'm still having trouble with the example that the OP in #1877 provides of Leaflet, but perhaps this is a Leaflet issue.

PR #1977, opened 2017-08-17
feat(Popup): Multiple event triggering a Popup

vipul-21:
Popup triggering by multiple events(hover, focus or click)

### DIFF
--- a/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
+++ b/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
@@ -5,7 +5,7 @@ const PopupExampleClick = () => (
   <Popup
     trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
     content={<Button color='green' content='Confirm the launch' />}
-    on={'click'}
+    on='click'
     position='top right'
   />
 )
--- a/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
+++ b/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
@@ -5,7 +5,7 @@ const PopupExampleClick = () => (
   <Popup
     trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
     content={<Button color='green' content='Confirm the launch' />}
-    on='click'
+    on={['click']}
     position='top right'
   />
 )
--- a/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
+++ b/docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
@@ -5,7 +5,7 @@ const PopupExampleClick = () => (
   <Popup
     trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
     content={<Button color='green' content='Confirm the launch' />}
-    on={['click']}
+    on={'click'}
     position='top right'
   />
 )
--- a/docs/app/Examples/modules/Popup/Usage/PopupExampleControlled.js
+++ b/docs/app/Examples/modules/Popup/Usage/PopupExampleControlled.js
@@ -26,7 +26,7 @@ class PopupExampleControlled extends React.Component {
           <Popup
             trigger={<Button content='Open controlled popup' />}
             content={`This message will self-destruct in ${timeoutLength / 1000} seconds!`}
-            on={['click']}
+            on={'click'}
             open={this.state.isOpen}
             onClose={this.handleClose}
             onOpen={this.handleOpen}
--- a/docs/app/Examples/modules/Popup/Usage/PopupExampleControlled.js
+++ b/docs/app/Examples/modules/Popup/Usage/PopupExampleControlled.js
@@ -26,7 +26,7 @@ class PopupExampleControlled extends React.Component {
           <Popup
             trigger={<Button content='Open controlled popup' />}
             content={`This message will self-destruct in ${timeoutLength / 1000} seconds!`}
-            on='click'
+            on={['click']}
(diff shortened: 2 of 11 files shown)


src/modules/Popup/Popup.js
```diff
@@ -250,15 +250,16 @@ export default class Popup extends Component {
       portalProps.closeOnPortalMouseLeave = true
       portalProps.mouseLeaveDelay = 300
     }
-
-    if (on === 'click') {
+    if (_.indexOf(on, 'click') > -1) {
```
layershifter: `_.includes` can be used there

src/modules/Popup/Popup.js
```diff
@@ -127,7 +127,7 @@ export default class Popup extends Component {
 
   static defaultProps = {
     position: 'top left',
-    on: 'hover',
+    on: ['hover'],
```
layershifter: We can left this

src/modules/Popup/Popup.js
```diff
@@ -72,7 +72,7 @@ export default class Popup extends Component {
     offset: PropTypes.number,
 
     /** Event triggering the popup. */
-    on: PropTypes.oneOf(['hover', 'click', 'focus']),
+    on: PropTypes.arrayOf(PropTypes.oneOf(['hover', 'click', 'focus'])),
```
layershifter: Valid propType will be:
```jsx
propTypes.oneOfType([
  PropTypes.oneOf(['hover', 'click', 'focus']),
  PropTypes.arrayOf(['hover', 'click', 'focus']), 
])
```

src/modules/Popup/Popup.d.ts
```diff
@@ -39,7 +39,7 @@ export interface PopupProps extends PortalProps {
   offset?: number;
 
   /** Event triggering the popup. */
-  on?: 'hover' | 'click' | 'focus';
+  on?: Array<string>;
```
layershifter: `'hover' | 'click' | 'focus' | Array<'hover' | 'click' | 'focus'>`
This should be correct as I remember

docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
```diff
@@ -5,7 +5,7 @@ const PopupExampleClick = () => (
   <Popup
     trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
     content={<Button color='green' content='Confirm the launch' />}
-    on='click'
+    on={['click']}
```
layershifter: Let's revert this

src/modules/Popup/Popup.js
layershifter: We can normalize value there and then use it:

```js
const normalizedOn = _.isArray(on) ? on : [on]
```

This will allow to pass `on='hover'` and `on={['hover']}`

layershifter (changes requested):
@vipul-21 Thanks for PR! But we can't merge it as is because it will introduce a breaking change. However, we can easily omit it, let's use normalization of `on`, this will allow to left compability with current API.

vipul-21:
@layershifter I have updated the changes as recommended. But the array being assigned to `on` is giving a type error. I am not able to figure out why. Can you help me out ?

docs/app/Examples/modules/Popup/Usage/PopupExampleClick.js
```diff
@@ -5,7 +5,7 @@ const PopupExampleClick = () => (
   <Popup
     trigger={<Button color='red' icon='flask' content='Activate doomsday device' />}
     content={<Button color='green' content='Confirm the launch' />}
-    on='click'
+    on={'click'}
```
layershifter: Just `on='click'`, please

docs/app/Examples/modules/Popup/Usage/PopupExampleControlled.js
```diff
@@ -26,7 +26,7 @@ class PopupExampleControlled extends React.Component {
           <Popup
             trigger={<Button content='Open controlled popup' />}
             content={`This message will self-destruct in ${timeoutLength / 1000} seconds!`}
-            on='click'
+            on={'click'}
```
layershifter: ^

docs/app/Examples/modules/Popup/Usage/PopupExampleFocus.js
```diff
@@ -6,7 +6,7 @@ const PopupExampleFocus = () => (
     trigger={<Input icon='search' placeholder='Search...' />}
     header='Movie Search'
     content='You may search by genre, header, year and actors'
-    on='focus'
+    on={'focus'}
```
layershifter: ^

layershifter:
You're right, `PropTypes.arrayOf` accepts type, not value, so correct propType will be `PropTypes.arrayOf(PropTypes.oneOf(['hover', 'click', 'focus'])),`

layershifter:
@vipul-21 Thanks for changes, we need only to add test

```jsx
describe('trigger', () => {
    //  .....


    it('it appears on multiple', (done) => {
      const trigger = <button>foo</button>
      const button = wrapperMount(<Popup on={['click', 'hover']} content='foo' header='bar' trigger={trigger} />)
        .find('button')

      button.simulate('click')
      assertInBody('.ui.popup.visible')

      domEvent.click('body')

      button.simulate('mouseenter')
      setTimeout(() => {
        assertInBody('.ui.popup.visible')
        done()
      }, 51)
    })
  })
```

layershifter (approved):

levithomason:
Very cool, thanks

levithomason:

PR #1978, opened 2017-08-17
feat(Sticky|Visibility): add scroll context

### LINKED ISSUE #1959: Feature request: context prop to Visibility component
Apologies if this is not the right forum to make this request, but it would be great to have a `context` prop for the Visibility component, much like you can change the `context` in Semantic UI: https://semantic-ui.com/behaviors/visibility.html#/settings. 

For example. I want to change the `scroll` event listener to listen on the scroll of a surrounding <List> component (i.e. a component which has a height less than the window height, but has an overflow so is scrollable). 

Currently, when this component is scrolled, none of the Visibility component's props fire because it does not think that the window is being scrolled. 

Thanks!

mariolamacchia:
Fixes #1959.

### DIFF
--- a/src/behaviors/Visibility/Visibility.d.ts
+++ b/src/behaviors/Visibility/Visibility.d.ts
@@ -9,6 +9,9 @@ export interface VisibilityProps {
   /** Primary content. */
   children?: React.ReactNode;
 
+  /** Context which sticky element should stick to. */
+  context?: object;
+
   /**
    * When set to true a callback will occur anytime an element passes a condition not just immediately after the
    * threshold is met.
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -20,6 +20,9 @@ export default class Visibility extends Component {
     /** Primary content. */
     children: PropTypes.node,
 
+    /** The scroll context visibility should use. */
+    context: PropTypes.object,
+
     /**
      * When set to true a callback will occur anytime an element passes a condition not just immediately after the
      * threshold is met.
@@ -138,6 +141,7 @@ export default class Visibility extends Component {
   }
 
   static defaultProps = {
+    context: window,
     continuous: false,
     once: true,
   }
@@ -166,11 +170,13 @@ export default class Visibility extends Component {
   }
 
   componentDidMount() {
-    window.addEventListener('scroll', this.handleScroll)
+    const { context } = this.props
+    context.addEventListener('scroll', this.handleScroll)
   }
 
   componentWillUnmount() {
-    window.removeEventListener('scroll', this.handleScroll)
+    const { context } = this.props
+    context.removeEventListener('scroll', this.handleScroll)
   }
 
   execute = (callback, name) => {
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -20,7 +20,7 @@ export default class Visibility extends Component {
     /** Primary content. */
     children: PropTypes.node,
 
-    /** The scroll context visibility should use. */
+    /** Context which visibility should attach onscroll events. */
     context: PropTypes.object,
 
     /**
--- a/src/modules/Sticky/Sticky.js
+++ b/src/modules/Sticky/Sticky.js
@@ -67,7 +67,7 @@ export default class Sticky extends Component {
     /** Whether element should be "pushed" by the viewport, attaching to the bottom of the screen when scrolling up. */
     pushing: PropTypes.bool,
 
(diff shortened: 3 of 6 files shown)


TimNZ:
@mariolamacchia Thanks.
If these (https://help.github.com/articles/checking-out-pull-requests-locally/) instructions work I'll pull down PR today and provide feedback.

TimNZ:
@mariolamacchia Seems to work as expected.

My positioning requirements can be likely be handled using the on* events and css.

Thanks for the rapid PR.

TimNZ:
@mariolamacchia I spoke too soon.

You didn't change the hardcoded window.* to use the scrollContext instead.
e.g.
```
window.innerHeight
```
->
```
this.props.scrollContext.innerHeight
```

Sticky sections are staying visible and stacking up as I scroll.

screen shot 2017-08-18 at 9 09 30 am

src/behaviors/Visibility/Visibility.d.ts
```diff
@@ -9,6 +9,9 @@ export interface VisibilityProps {
   /** Primary content. */
   children?: React.ReactNode;
 
+  /** Context which sticky element should stick to. */
+  context?: object;
```
levithomason, line 13: I'd assume the description here would match the Visibility component:
```js
/** The scroll context visibility should use. */
```

src/modules/Sticky/Sticky.js
```diff
@@ -66,11 +66,15 @@ export default class Sticky extends Component {
 
     /** Whether element should be "pushed" by the viewport, attaching to the bottom of the screen when scrolling up. */
     pushing: PropTypes.bool,
+
+    /** Context which sticky should attach `onscroll` events. */
```
levithomason: nit: Let's match the style of the typings comment, no backticks.

src/modules/Sticky/Sticky.js
```diff
@@ -66,11 +66,15 @@ export default class Sticky extends Component {
...
+    scrollContext: PropTypes.object,
```
levithomason: Based on the usage, looks like these should be `.isRequired`.

src/behaviors/Visibility/Visibility.js
```diff
@@ -20,6 +20,9 @@ export default class Visibility extends Component {
     /** Primary content. */
     children: PropTypes.node,
 
+    /** The scroll context visibility should use. */
+    context: PropTypes.object,
```
levithomason: Needs to be `.isRequired` here given usage in code.

levithomason:
I'm curious as to the reasoning behind `context` for Visibility and `scrollContext` for Sticky.  Why different names for these?

```jsx
<Visibility context={div} />

<Sticky scrollContext={div} />
```

mariolamacchia:
@TimNZ `scrollContext` is only responsible for attaching the scroll event to a different element. If you want it to stick to something else, you should use `context`.

mariolamacchia:
@levithomason I'm using the names from the SemanticUI docs, although it isn't very consistent:
- `context` in Sticky = which element the Sticky should stick to
- `context` in Visibility = `scrollContext` in Sticky = context of the scroll event

src/behaviors/Visibility/Visibility.js
mariolamacchia: There is a default value of `window`. `isRequired` means that the user must specify the value, right?

src/modules/Sticky/Sticky.js
mariolamacchia: same as below

src/modules/Sticky/Sticky.js
mariolamacchia: I'm using the same descriptions from the SemanticUI docs. I'll use this one, without backticks, for both

levithomason:
Thanks for the explanation @mariolamacchia, let's keep the parity for now then 👍

levithomason:

theCuriousOne:
Visibility BREAKS server side rendering!!!

frogbandit:
Hi @mariolamacchia, thank you for adding the context prop for the Visibility component! :+1: :+1: for moving so fast on this.

However, I've tried using it and `context` doesn't seem to accept either a React ref or an HTML element. I took a look at the source code and it's just supposed to be a `PropTypes.object`. Can you give some more info on how to use it? i.e. what kind of element should be passed in.

levithomason:
@frogbandit we have a new component being added to the library to solve this issue, #1879.  It is a `Ref` component.

Stateless functional components do not support refs, so we cannot be sure that a component passed in on `context` will support `ref`s.  Once we land #1879 it will include support for all props that accept DOM nodes and/or components, like `context` and `as`.

PR #1981, opened 2017-08-18
perf(docs): optimize performance

layershifter:
This PR is first round of docs performance optimizations. I've profiled our docs page and found that there are two components that aren't optimized: `ComponentExample` and `ComponentProps`.

## `ComponentProps`

Renders many elements, has a non-optimized render of enum values. I've refactored it to several small components that have controlled render cycle. I've also refactored usage of double curly braces:
- [Extra](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.71.4/docs/app/Components/ComponentDoc/ComponentProps.js#L17)
- [values](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.71.4/docs/app/Components/ComponentDoc/ComponentProps.js#L129)

## `ComponentExample`

My profile showed that the next bottleneck is control items at each example. All of them use `Popup` which uses `Portal`, page that has 30 examples produces 120 `Portal`s, which rendered immediately. It's not fast.

I saw that in fact, it's not necessary because controls should be visible only when example is hovered. I've factored CSS animation with a `Transition` component, so controls will mount only when they will be needed. Each control was optimized, too.


layershifter (commented):
@levithomason I've finished this PR, it will be interesting to hear your opinion

levithomason:
Oh man, this has been on my list for so long ❤️ !

The update all deps PR has made some conflicts here, could you take a look?

layershifter:
@levithomason I'll solve them tomorrow, 22PM there :sleeping:

layershifter (commented):
@levithomason I've resolved conficts and merged with `master` :+1:

levithomason:

PR #1985, opened 2017-08-20
docs(Transition): enhancements

keksipurkki:
Here are my 2 cents. I had some issues with the `Transition.Group` component resulting from poorly picked `key` props on the child components. As I suspect this is a common blunder, I thought I'd make a note about it in the docs. Cheers!

### DIFF
--- a/docs/app/Examples/modules/Transition/Types/index.js
+++ b/docs/app/Examples/modules/Transition/Types/index.js
@@ -6,23 +6,30 @@ import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
 
 const TransitionTypesExamples = () => (
   <ExampleSection title='Types'>
+
     <ComponentExample
       title='Transition'
       description='A Transition animates a single child by toggling the the visible prop.'
       examplePath='modules/Transition/Types/TransitionExampleTransition'
     >
       <Message warning>
-        <p>Do not unmount a Transition child or else it cannot be animated.</p>
+        <p>An animated component must not unmount unexpectedly or the transition will not run as intended.</p>
         <Message.List>
           <Message.Item>
-            Use the <code>unmountOnHide</code> prop to unmount the child after the animation exits.
+            If you need to animate children as they mount and unmount, use a <code>Transition.Group</code>. 
           </Message.Item>
           <Message.Item>
-            Use a <code>Transition.Group</code> to animate children as they mount and unmount.
+            It is paramount that each child of a <code>Transition.Group</code>
+            has an unique <code>key</code> prop that stays constant through rerenders.
+            In particular, it is likely that <code>keys</code> based on pseudo-random identifiers or simple loop increment variables will lead to unexpected behavior.
+          </Message.Item>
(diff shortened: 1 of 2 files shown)


levithomason:
I'm not sure this belongs in our repo.  The `key` here is a vanilla React `key`.  They have great docs on what they are for, how to use them, and how not to use them.

Perhaps we provide a simple note like:

>`Transition.Group` children must have proper [React keys](https://facebook.github.io/react/docs/lists-and-keys.html#keys).

Then, leave React's detailed docs to explain the details and show samples.  Feedback?

docs/app/Examples/modules/Transition/Types/index.js
```diff
@@ -6,23 +6,30 @@ import ExampleSection from 'docs/app/Components/ComponentDoc/ExampleSection'
...
+        <p>An animated component must not unmount unexpectedly or the transition will not run as intended.</p>
         <Message.List>
           <Message.Item>
-            Use the <code>unmountOnHide</code> prop to unmount the child after the animation exits.
+            If you need to animate children as they mount and unmount, use a <code>Transition.Group</code>. 
+          </Message.Item>
+          <Message.Item>
+            It is paramount that each child of a <code>Transition.Group</code>
+            has an unique <code>key</code> prop that stays constant through rerenders.
+            In particular, it is likely that <code>keys</code> based on pseudo-random identifiers or simple loop increment variables will lead to unexpected behavior.
```
levithomason, line 24: Per my comment, unique, consistent, and non-random keys are covered here:

https://facebook.github.io/react/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

keksipurkki: I think a link to the React docs is good but I would still be a bit more verbose about the keys, especially since some of the devs adopting Semantic UI may not be ReactJS experts. 

In my case I was using proper React keys in the sense described in the docs: a simple loop variable that was unique among siblings. It was just that when my list was modified and the keys were reset, Transition.Group lost track of which components to animate.

brianespinosa: @keksipurkki This is definitely a React issue and is most definitely not unique for this project. In fact, [there are lint rules specifically to make sure you are not using something like an index to set your keys](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) because when the DOM updates your keys will not be unique. If they are not always going to be unique then you're not following the documentation. The keys need to be unique in ALL states of your application for the siblings you are referring to.

levithomason: I want to make sure I don't miss a nuance difference here.  @keksipurkki could you share what your keys were?  I currently don't see a difference between the Transition.Group keys and valid React keys.  Were they indexes by chance?  I could see this failing and possibly needing clarification.

layershifter:
I close this for housekeeping. `Transtion.Group` invents nothing and vanilla React keys.

PR #1989, opened 2017-08-21
fix(TextArea): fix typings definition

pierre-H:
Fix false definition of TextArea


layershifter (approved):
@pierre-H Thanks!

thchia:
Should focus be an optional property?

layershifter:
`focus` isn't React's prop there, it's a class method. So, definition is correct and this will be work:

```tsx
 <TextArea ref={c => c.focus()} />
```

thchia:
Oops you are absolutely right 🙃

levithomason:

PR #1990, opened 2017-08-21
fix(Sticky|Visibility): add null check on window

### LINKED ISSUE #1987: ReferenceError: window is not defined
From the latest version: 0.71.5, server-side rendering produces following issue

```
/var/app/current/node_modules/semantic-ui-react/dist/commonjs/behaviors/Visibility/Visibility.js:260
  context: window,
           ^
ReferenceError: window is not defined
    at Object.<anonymous> (/var/app/current/node_modules/semantic-ui-react/dist/commonjs/behaviors/Visibility/Visibility.js:260:12)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/var/app/current/node_modules/semantic-ui-react/dist/commonjs/behaviors/Visibility/index.js:8:19)
    at Module._compile (module.js:569:30)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! snowy-console@1.0.0 start: `NODE_ENV=production node build/server`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the snowy-console@1.0.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

lott-ai:
fix(Sticky|Visibility): add null check on window for server side rendering

fixes https://github.com/Semantic-Org/Semantic-UI-React/issues/1987

### DIFF
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -142,7 +142,7 @@ export default class Visibility extends Component {
   }
 
   static defaultProps = {
-    context: isBrowser ? window : {},
+    context: isBrowser ? window : null,
     continuous: false,
     once: true,
   }
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -171,11 +171,15 @@ export default class Visibility extends Component {
   }
 
   componentDidMount() {
+    if (!isBrowser) return
+
     const { context } = this.props
     context.addEventListener('scroll', this.handleScroll)
   }
 
   componentWillUnmount() {
+    if (!isBrowser) return
+
     const { context } = this.props
     context.removeEventListener('scroll', this.handleScroll)
   }
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -7,6 +7,7 @@ import {
   getElementType,
   getUnhandledProps,
   META,
+  isBrowser,
 } from '../../lib'
 
 /**
@@ -141,7 +142,7 @@ export default class Visibility extends Component {
   }
 
   static defaultProps = {
-    context: window,
+    context: isBrowser ? window : {},
     continuous: false,
     once: true,
   }
--- a/src/modules/Sticky/Sticky.js
+++ b/src/modules/Sticky/Sticky.js
@@ -75,7 +75,7 @@ export default class Sticky extends Component {
   static defaultProps = {
     bottomOffset: 0,
     offset: 0,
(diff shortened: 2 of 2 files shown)


src/behaviors/Visibility/Visibility.js
```diff
@@ -141,7 +142,7 @@ export default class Visibility extends Component {
   }
 
   static defaultProps = {
-    context: window,
+    context: isBrowser ? window : {},
```
layershifter: Let's use `null` there:
```js
isBrowser ? window : null,
```

layershifter (changes requested):

src/behaviors/Visibility/Visibility.js
lott-ai: @layershifter if we use null, should we also add null checks around https://github.com/ChrisNLott/Semantic-UI-React/blob/d3bd18cd7fca2be2ea5a32381b70798aa5de2c57/src/behaviors/Visibility/Visibility.js#L173? basically anywhere context is used? But I suppose the empty object would error in that case too

layershifter (approved):

lulalachen:
Looking forward to this PR! Thanks

tim-soft:
I can't wait for this to merge, thank you!

levithomason:
Has this been tested in SSR?  I don't see how the `componentDidMount` would work when `context` is `null`:

```jsx
  componentDidMount() {
    const { context } = this.props
    context.addEventListener('scroll', this.handleScroll)
  }
```

In other components, we also exit methods early if they are accessing the `document` or `window`:

```jsx
  componentDidMount() {
    if (!isBrowser) return

    const { context } = this.props
    context.addEventListener('scroll', this.handleScroll)
  }
```

Would love to get some validation from SSR users before shipping.

tim-soft:
I can confirm that these changes fix my SSR apps, although I also need to make the same changes in `dist/commonjs/behaviors/Visibility/Visibility.js` and `dist/commonjs/modules/Sticky/Stick.js`

lott-ai:
@levithomason I tested it against my SSR app, but I will add null checks around the lifecycle methods too

levithomason:
@tim-soft and @ChrisNLott, thanks for investigating and reporting back.  Also, to @ChrisNLott for the updates.

Merging!  Will be released today :)

levithomason:

Issue #837, opened 2016-11-11
RFC: `transition` and `popup` common shorthand


jeffcarbs:
### Background

I've been using `<Popup>` in my app for tooltips and it's a little cumbersome to use. Here's an example usage for showing a popup on a disabled button (note: this doesn't actually work out of the box since pointer-events are disabled on buttons 😐 )

```js
const buttonElement = (
  <Button
    content='Some action!'
    disabled={disabled}
    loading={loading}
    onClick={onClick}
    primary
  />
)

if (!disabled) {
  return buttonElement
}

return (
  <Popup
    content='This button is disabled!'
    trigger={buttonElement}
  />
)
```

### Proposal

What if we were to add a `popup` common shorthand prop to all components. The same example from above could be written as:

```js
<Button
  content='Some action!'
  disabled={disabled}
  loading={loading}
  onClick={onClick}
  popup={disabled && 'This button is disabled!'}
  primary
/>
```

Under the hood, it would use a `Popup.create` shorthand factory:
```js
Popup.create(props.popup, { trigger: /*regularly rendered button*/ })

// In this case would render
<Popup content='This button is disabled!' trigger={<Button>Some action</Button>} />
```

### Transition

The `Transition` component (currently underway in #813) strikes me as something that will have the same cumbersome usage.
```js
<Transition animation='slide'><Button /></Transition>
```
could be written as
```js
<Button transition='slide' />
```

To me the former reads like "render a transition that results in a button", whereas the latter reads more naturally: "render a button with the slide transition".

Both the `transition` and `popup` props would be shorthand so they could be customized even further. For example:

```
<Button popup={{ inverted: true, on: 'click', content: 'Clicked!' }} />
```

### Implementation

I think it would be crazy to try to do this with our components as-is given that we'd have to touch every single file. I really think we need to solve #419 first and get all of our components rendering via some common method. Whether it's a render method utility, a base component that every component eventually renders down to, or some factory.

cc @layershifter @levithomason - would love your thoughts!

levithomason:
Man, I love your proposals :)  Yes, a `popup` prop has been on mind since day one.  I also think we should solve the base component idea before doing this.

I hadn't considered transition, but it also makes perfect sense to me.  This gets me thinking that the Dropdown menu could also take advantage of this pattern, using `trigger`.  Really, any component wraps or appears beside another component could / should be added as a prop IMO.  

Imagine:

``` jsx
[...]

PR #2001, opened 2017-08-24
feat(Dimmer): add transition

layershifter:
# WIP


levithomason:
Very cool, can't wait to get Transitions in the entire framework :)  This is largely what is stopping a v1 release.  The other blocking work is minor in comparison.

tim-soft:
I have been scratching my head over the dimmer transitions used in the docs site, can't wait for this!

levithomason:
It would be great if we tackled #837 and made `transition` a prop on every component possible.  We'd just need a HOC that did something like (warning, untested!):

```jsx
import React from 'react'
import Transition from './modules/Transition/Transition'

const withTransition = Component => {
  const WithTransition = props => {
    const { transition } = props

    const element = <Component {...props} />

    if (!transition) return element

    return Transition.create(transition, {
      defaultProps: { 
        ...props,
        children: element,
      },
    })
  }

  return WithTransition
}
```

This requires adding a shorthand factory to the `Transition`.

Then, in our component files:
```jsx
// Button.js
export default withTransition(Button)
```

Now, we can do this:

```jsx
<Button transition='fade' />
<Button transition={{ animation: 'fade', duration: 100 }} />
```

I actually did this for `popup` and I cannot find the branch 😭 The only issue I had was getting the docs to understand extended propTypes.  Example, the Button above would need to spread the `withTransition` propTypes.  I guess we could just hard code them, but it would be nice to be able to inherit.  Especially as we'll have propTypes from `withPopup` and any other HOCs we add.

layershifter:
In fact, the key problem that I see there is with `page` `Dimmer`. Instead of regular in renders inside `Portal` which unmounts its content when it closes. I'm still have no idea how to resolve this without tons of callbacks.

layershifter:
Needs #2155.

levithomason:
TransitionablePortal has been merged, unblocking this.

padrefuture:
New to this site, and Im hoping this is the right place to post this. Im looking for a a menu with an animated (transition?) dropdown, that has a delay on mouseout. ANy ideas where i can find or is this too much to ask in the SemanticUI/React world?

levithomason:
@padrefuture animations are a work in progress here in the React port of Semantic UI.  That feature is not yet available.  We do have a [Transition component](https://react.semantic-ui.com/modules/transition).  That component needs to be integrated into our components so we have the transitions you are looking for.  PRs welcome!

ghost:
Am I wrong or is just a linting problem the latest ci/circleci problem?

Can we force this and then fix the lint problems?

I'm newbie, sorry for maybe wrong questions.

levithomason:
I've merged master and fixed the lint issues.  This PR still appears to be in progress, as the description states.

We'll need the feature completed, some tests, and some docs.

jermsam:
Hey guys. Any success with @levithomason 's sugestion? I don't see myself moving away from semantic ui react just because of a simple animation. This css framework is awesome for react. someone do something please

brianespinosa:
Good morning @jermsam 

This is an open source project and anyone is welcome to contribute. If there is an area of the framework where you have a need for something specific to your project, pull requests are always welcome.

ghost:
@stale, please, shut up.

PR #2004, opened 2017-08-25
fix(Modal): fix typo in className

### LINKED ISSUE #1979: Modal: scroll breaks if two are adjacent and of different lengths
This is probably related to #1157 but it's not just if you have a modal within a modal that this bug exists, having two in the same component causes the same issue. 

### Steps

1. Create a component which contains two Modals where one has content long enough to require a scroll and the other doesn't.

2. Open the longer modal and scroll. Close this Modal. 

3. Open the shorter modal. Close this Modal.

4. Open the longer modal. Scroll no longer works. 

### Expected Result

The longer modal should always be scrollable.

### Actual Result

The longer modal gets stuck if opened after the shorter modal is opened and closed. 

### Version
0.71.3

### Testcase

https://codepen.io/anon/pen/Ljerpm

layershifter:
We have the awesome [codepen](https://codepen.io/anon/pen/Ljerpm) in #1979. It show that `scrolling` className isn't removed from `body` after `Portal`'s unmount. In fact, there was a typo and this PR fixes it.

Fixes #1979.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -219,7 +219,7 @@ class Modal extends Component {
     mountNode.classList.remove('blurring')
     mountNode.classList.remove('dimmable')
     mountNode.classList.remove('dimmed')
-    mountNode.classList.remove('scrollable')
+    mountNode.classList.remove('scrolling')
 
     cancelAnimationFrame(this.animationRequestId)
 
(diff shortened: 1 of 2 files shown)


levithomason:
Looking into why our tests for this were passing :)

levithomason:
We were missing quite a bit of coverage on adding/removing classes.  I've updated tests and added missing coverage.

Oddly enough, fixing these tests somehow fails the Sticky tests?  I may not have time today to figure out why today.

layershifter:
I'm almost sure that problem is with `window.innerHeight` usage in tests.

layershifter:
Yep, working on fix.

levithomason:
Heh, oh boy... Thanks!

levithomason:

PR #2006, opened 2017-08-25
fix(Button|Menu|List|Step): fix behaviour of `onClick` when is disabled

### LINKED ISSUE #1964: Menu.Item - disabled=true - onClick still firing...intentional?
### Steps
MenuItem with prop `disabled=true` - the onClick event is being triggered.

### Expected Result
When a menu item is disabled, and onClick is fired - internally the MenuItem component should override the onClick prop with an empty anonymous function - effectively `disabling` the menu item.

I have done the following as a work around - however I believe it should be handled internal to the Menu.Item component.
```
<Menu.Item disabled={activePageIndex === pages.length - 1}
    onClick={
      activePageIndex === pages.length - 1
      ? () => {}
      : this.handleNextPage}  
/>
```
### Actual Result
onClick is being fired on `disabled` menu items. Requiring the above pattern to 'disable' the onClick handler.

Is this intentional to handle some kind of edge cases?
Thoughts?

### Version
0.71.4

### Testcase
https://codepen.io/harmon25/pen/GvMdow

layershifter:
Fixes #1964.

Affects:
- `Button`
- `ListItem`
- `MenuItem`
- `Step`

We have `disabled` prop also in:
- `FormField` - I don't touch because it doesn't have `onClick` handler and I'm not sure that it will be correct behaviour
- `TableCell`/`TableRow` - I think we can update the behaviour there, but these components don't have `onClick` handler
- `Icon`, `Image`, `Loader`, `Reveal`, `Segment`, `Progress` - These components don't have `onClick` handler and I'm not sure that we need to change there something
`Checkbox`, `Dropdown`, `Rating` - already handles `disabled`
`Dimmer` - `disabled` has there another context

### DIFF
--- a/src/collections/Menu/MenuItem.js
+++ b/src/collections/Menu/MenuItem.js
@@ -86,9 +86,9 @@ export default class MenuItem extends Component {
   }
 
   handleClick = (e) => {
-    const { onClick } = this.props
+    const { disabled } = this.props
 
-    if (onClick) onClick(e, this.props)
+    if (!disabled) _.invoke(this.props, 'onClick', e, this.props)
   }
 
   render() {
--- a/src/elements/List/ListItem.js
+++ b/src/elements/List/ListItem.js
@@ -87,9 +87,9 @@ class ListItem extends Component {
   }
 
   handleClick = (e) => {
-    const { onClick } = this.props
+    const { disabled } = this.props
 
-    if (onClick) onClick(e, this.props)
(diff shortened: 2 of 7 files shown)


levithomason:
Let's merge this as it is a great step forward.

We may need to double back and add `onClick` handlers for the others you've mentioned, but let's see what the community response is before investing the time.  Also, the CSS for many disabled classes includes `pointer-events: none` which effectively does the same thing.

levithomason:

PR #2009, opened 2017-08-27
feat(Dropdown): implement `selectOnNavigation` prop

rijk:


levithomason:
Great, next steps here will be adding an example to the docs in `/docs/app/Examples/modules/Dropdown/Usage`.  Be sure to add the example to the `index.js` in that directory as well.

Then, we also need a test for this in `Dropdown-test.js`.  We should `describe('selectOnKeydown')` and add assertions for what `it()` does when `true` and `false`.

levithomason:
Cool, thanks much!  We'll get a review on this as soon as we can.  I just made a release so we'll target next weekend for releasing this one.

rijk:
Just to let you know: I was not able to test the Doc example.

levithomason:
If you mean automated tests, that's OK.  Doc example tests are a little looser than they should be right now.  We just render them and assert no errors occurred.  This is done automatically for you.

rijk:
No, I meant I was unable to open the docs locally after building, to see if/how the example looked and worked. The tests should be passing.

levithomason:
Fixed lint issues, testing manually.  Hope to ship this one today as well!

levithomason:
Ok, finished updating docs.  I renamed this prop to `selectOnNavigation` since key down still applies for the <kbd>Enter</kbd> case.

I added test assertions and doc descriptions to clarify the prop controls the change in value.  The fact that `onChange` is called is a side-effect of changing the value only.

levithomason:

Issue #1020, opened 2016-12-12
RFC: Use PureComponent


layershifter:
Hello, guys!

### PureComponent

I want to speak about next optimization topic - `PureComponent`. It's previously known as [pure render mixin](https://facebook.github.io/react/docs/pure-render-mixin.html), it makes shallow compare of props inside `shouldComponentUpdate()`. Inheriting from `React.PureComponent` indicates to React that your component doesn't need to rerender when the props are unchanged ([#7195](https://github.com/facebook/react/pull/7195)). 

I'm using it in my app and it's perfect 😸  [There](http://moduscreate.com/react_component_rendering_performance/) are some benchmarks and conclusions.

### Benefits

* Faster rerendering.
* Refs.
* Unification.

We will loose stateless components, according to [Dan Abramov](https://twitter.com/dan_abramov/status/755343749983657986) they are same as statefull inside, so the same slow as `Component` child. Also, now we have stateless, statefull and statefull component without real state (#607), after refactoring we will have `Component` and `PureComponent` classes 👍 

### Limitations

* React [^0.15.3](https://github.com/facebook/react/releases/tag/v15.3.0).
* Stateless components looks really nice 😿 

---

@jcarbo @levithomason thoughts?

levithomason:
I don't think we should implement this, here's why:

Quoting `@jimfb` from https://github.com/facebook/react/pull/7195#issuecomment-230793226:
>In addition to event handlers bound in render, it is probably also worth mentioning that any component which takes jsx children will also always fail shallowEqual, since the children elements will be re-created for every render. I think most people don't realize this.

Almost all of our components can render JSX children, meaning, someone's app can potentially be shallow comparing all our components for no reason.

Quoting Abramov from https://github.com/facebook/react/pull/7195#issuecomment-236361372:
>However just making every one of them work like PureComponent can slow down your app. Please don't think shallow equality checks are extremely cheap. They can help your app when placed strategically but just making every single component pure can actually make your app slower. Tradeoffs.

Quoting Abramov again, from [`react-pure-render`](https://github.com/gaearon/react-pure-render#known-issues):
[...]

PR #2011, opened 2017-08-28
perf(Flag|Icon): add shouldComponentUpdate with shallowEqual

layershifter:
Rel #1020.

I've imported `shallowEqual` in #1981 and remembered about #1020. Converts `Flag` and `Icon` to class components and adds `shouldComponentUpdate` with `shallowEqual`.


levithomason:

PR #2012, opened 2017-08-28
perf(docs): optimize ComponentProps

layershifter:
## ComponentProps

Components was splitted to multiple sub-components. Props info transforms was moved to our gulp doc plugin.

#### Before
| Action | Time |
| --------------- | ---- |
| Show/hide enums _(4 times)_ | 40.88 |
| Switch props tab _(4 times)_ | 37.03 |
#### After

| Action | Time |
| --------------- | ---- |
| Show/hide enums _(4 times)_ | 11.16 |
| Switch props tab _(4 times)_ | 36.67


layershifter (commented):
@levithomason I wanted to add there changes that will affect `ComponentExample`, but later I understood that it has already enough changes. So, it's ready for review.

docs/app/HOC/pure.js
```diff
@@ -0,0 +1,15 @@
...
+  shouldComponentUpdate(nextProps) {
+    return !shallowEqual(this.props, nextProps)
+  }
+
+  render() {
+    return <ChildComponent {...this.props} />
+  }
+}
+
+export default pure
```
levithomason: I believe this can be replaced by extending [`React.PureComponent`](https://facebook.github.io/react/docs/react-api.html#react.purecomponent), correct?

docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
```diff
@@ -38,123 +36,22 @@ export default class ComponentProps extends Component {
...
-        </Table.Header>
+        <ComponentPropsHeader />
         <Table.Body>
-          {_.map(content, this.renderRow)}
+          {_.map(propsDefinition, item => (
+            <ComponentPropsRow
+              {...item}
+              key={item.name}
+              showEnums={showEnumsFor[item.name]}
+              onToggleEnum={this.toggleEnumsFor(item.name)}
```
levithomason: Since the `showEnums` state is only needed within `ComponentPropsRow`, I think we can push it down another level.  The parent doesn't need to re-render when the child shows/hides enums.

docs/app/Components/ComponentDoc/ComponentProps/ComponentPropsEnum.js
```diff
@@ -7,7 +7,9 @@ import ComponentPropsExtra from './ComponentPropsExtra'
 import ComponentPropsToggle from './ComponentPropsEnumToggle'
 import ComponentPropsValue from './ComponentPropsEnumValue'
 
-const ComponentPropsEnum = ({ limit, showAll, toggle, values }) => {
+const ComponentPropsEnum = ({ limit, showAll, toggle, type, values }) => {
```
levithomason, line 10: The `limit` was added because of the `Icon` names and `Flag` names.  However, those props do not show enums as they are `{custom}`.  I almost want to just show all enums and do away with the state.

At the least, let's increase the limit to `50`.  Most examples have much less than 50 enums, and 50 will still fit OK.  Here are a few examples of our larger lists that show we can easily work with a limit of 50:

**Grid.Row**
`color` is especially funny as it a very common prop and we hide only 3 of them 😂 


**Transition**

layershifter: Thanks for reminding, `customPropTypes.suggest` should be also parsed as `enum`, I will update it separate PR.

layershifter (commented):
@levithomason I've applied requested changes :+1:

levithomason:
💥

levithomason:

PR #2013, opened 2017-08-28
fix(Checkbox): fix `checked` value returned by `onClick`

### LINKED ISSUE #1936: Checkbox onClick provides wrong data [onChange works fine]

### Steps
Check or uncheck a checkbox

### Expected Result
Checked empty checkbox returns true for 'checked' value in data params and vice versa

### Actual Result
onClick registers too early and the 'checked' value is what is was prior (inverse of truth) 

### Version
0.71.3

### Testcase
Toggle a checkbox and view the data params for checked. It's the inverse of what it's supposed to be. I'm assuming the onClick event registers before it actually changes.

layershifter:
Fixes #1936.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -159,7 +159,7 @@ export default class Checkbox extends Component {
 
     if (!this.canToggle()) return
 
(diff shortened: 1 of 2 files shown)


layershifter:
Closing in favor of #2014.

PR #2016, opened 2017-08-29
feat(Visibility): add offset

layershifter:
Rel #1146.

- adds the `offset` prop
- cleanups `Visibility` docs, make them more usable


levithomason:
Fixed a couple issues with the `ComponentExample` in this branch.  It didn't know how to `import` the `Wireframe` component.  The examples would not re-render on change, but throw. 
 Need a better way of resolving modules in the future.  Right now, we have a very hacky hard-coded way of resolving modules.

Also, I made some updates to the event log experience and improved performance.  It is now truncated to 20 items and shows the total number of callbacks called.

layershifter:
I solved merge conflicts :+1:

levithomason:

PR #2018, opened 2017-08-29
feat(Visibility): add fireOnMount

### LINKED ISSUE #1911: Visibility: onOnScreen doesn't fire if element is onScreen on first render
### Steps
The goal here is to create an element that has an image that only shows when the image is visible.
1. Create an element with an image.
2. Wrap the image with a visibility element that fires onOnScreen.
3. When the image first draws, set its src to ''
4. When the image is on screen, set its src to the desired image so it downloads.

### Expected Result
1. Images that are rendered on screen (above the fold) on the first render have onOnScreen to set their source to the desired image.
2. Images that are rendered off screen are only loaded once the page is scrolled to a point where the image is on the screen.

### Actual Result
Images that are rendered on screen (above the fold) on the first render do not fire onOnScreen.

### Version
0.69.0

### Testcase
https://codepen.io/anon/pen/MvKPJX?editors=0010

Note, I just used icons in this example.  The example says that the github icon should be 'users' when it is off screen and 'github' when it is onscreen.  Notice that on initial load it is 'users'.  If you scroll the window ever so slightly, it'll change to 'github'.

If there is another way I should be accomplishing this first render case, please let me know.

layershifter:
Fixes #1911.

### DIFF
--- a/docs/app/Examples/behaviors/Visibility/Settings/VisibilityExampleFireOnMount.js
+++ b/docs/app/Examples/behaviors/Visibility/Settings/VisibilityExampleFireOnMount.js
@@ -1,0 +1,112 @@
+import React, { Component } from 'react'
+import { Divider, Grid, Image, Table, Segment, Visibility } from 'semantic-ui-react'
+
+export default class VisibilityExampleFireOnMount extends Component {
+  state = {
+    calculations: {
+      height: 0,
+      width: 0,
+      topPassed: false,
+      bottomPassed: false,
+      pixelsPassed: 0,
+      percentagePassed: 0,
+      topVisible: false,
+      bottomVisible: false,
+      fits: false,
+      passing: false,
+      onScreen: false,
+      offScreen: false,
+    },
+  }
+
+  handleOnScreen = (e, { calculations }) => this.setState({ calculations })
+
+  handleOffScreen = (e, { calculations }) => this.setState({ calculations })
+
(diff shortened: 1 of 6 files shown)


levithomason:
Made a small verbiage tweak, will merge on pass.

levithomason:

boostup:
Hello Guys!  

I'm basically trying to show a react element that is onScreen on first render.  So, I found [Visibility: onOnScreen doesn't fire if element is onScreen on first render](https://github.com/Semantic-Org/Semantic-UI-React/issues/1911), which lead me to this page here.

1)According to the comment on this page on Sep 11, a fix was released at version 0.73.0  
I have version 0.76.0 in my project.  However, when wrapping it with a Visibility react tag, an element, say an image, does not show it until I scroll even though the image is already partially onScreen.  

2)Also, if I include the fireOnMount prop on the Visibility tag, then calculations.onScreen is always equal to true.  Is that normal?  I can't understand the reason for this...What is fireOnMount for?  I don't get it from the docs https://semantic-ui.com/behaviors/visibility or https://react.semantic-ui.com/behaviors/visibility

Here is a codesandbox:  https://codesandbox.io/s/nr5k9k6zwp

Thanks for any assistance on these 2 points

layershifter:
If you want to get help, you should create a new issue and fully fill issue template.

PR #2021, opened 2017-08-29
feat(Dropdown): remove diacritics on filter

AgentChris:
I really need this feature, can you guys help me?

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -795,7 +795,14 @@ export default class Dropdown extends Component {
         filteredOptions = search(filteredOptions, searchQuery)
       } else {
         const re = new RegExp(_.escapeRegExp(searchQuery), 'i')
-        filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text))
+        filteredOptions = _.filter(filteredOptions, function (opt) {
+          let str = opt.text
+          if(str) {
+            // remove diacritics on search
+            str = str.normalize('NFD').replace(/[\u0300-\u036f]/g, "");
+          }
+          return re.test(str);
+        });
       }
     }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -795,14 +795,8 @@ export default class Dropdown extends Component {
         filteredOptions = search(filteredOptions, searchQuery)
       } else {
         const re = new RegExp(_.escapeRegExp(searchQuery), 'i')
-        filteredOptions = _.filter(filteredOptions, function (opt) {
-          let str = opt.text
-          if(str) {
-            // remove diacritics on search
-            str = str.normalize('NFD').replace(/[\u0300-\u036f]/g, "");
-          }
-          return re.test(str);
-        });
+        // remove diacritics on search
+        filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text ? _.deburr(opt.text): opt.text))
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -795,7 +795,14 @@ export default class Dropdown extends Component {
         filteredOptions = search(filteredOptions, searchQuery)
       } else {
         const re = new RegExp(_.escapeRegExp(searchQuery), 'i')
-        filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text))
+        filteredOptions = _.filter(filteredOptions, function (opt) {
```
levithomason: This should just be:

```diff
-filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text))
+filteredOptions = _.filter(filteredOptions, opt => re.test(_.deburr(opt.text)))
```

https://lodash.com/docs/4.17.4#deburr

levithomason:
See inline comment on `_.deburr`.  We just need a test case for this now.  See `Dropdown-test.js`.

AgentChris:
for example when i have this options:
when i search after: floresti
i will get no response, but when i click outside i can view the options

rsz_1screenshot_from_2017-08-29_19-53-59

rsz_1screenshot_from_2017-08-29_19-54-07

levithomason:
Yep, I'm noting we should use lodash's `_.deburr` method rather than writing our own filter:

image

AgentChris:
ty for your quick response, i will try to add a test, first time I am doing a pull request so please have patience :)

src/modules/Dropdown/Dropdown.js
```diff
@@ -795,7 +795,8 @@ export default class Dropdown extends Component {
         filteredOptions = search(filteredOptions, searchQuery)
       } else {
         const re = new RegExp(_.escapeRegExp(searchQuery), 'i')
-        filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text))
+        // remove diacritics on search
+        filteredOptions = _.filter(filteredOptions, opt => re.test(opt.text ? _.deburr(opt.text): opt.text))
```
levithomason: Lodash gracefully handles `null` and `undefined` values, so you can safely remove the ternary check here.  The method will always return a string:

image

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -610,6 +610,37 @@ describe('Dropdown', () => {
...
+
+      // search for 'floresti'
+      wrapperMount(<Dropdown options={opts} search selection />)
+        .simulate('click')
+        .find('input.search')
+        .simulate('change', { target: { value: 'floresti' } })
+
+      wrapper
+        .find('.selected')
+        .should.contain.text('FLOREŞTI')
```
levithomason, line 628: This looks great.  We can end the test here as one assertion is enough to prove the feature.  

Skipping the additional arrow down behaviors will make this test stronger.  This way, the `filter after diacritics` test doesn't fail if the arrow down behavior changes, but only if the search filter behavior changes. 👍

levithomason:
Awesome, congrats on your first PR :)  I left a couple more comments, then I think we're ready to go here.

levithomason:
Be sure to run `npm run lint:fix` as well.  This will run lint checks and also fix some issues for you.  You can ignore the `warnings`, but the `errors` will need fixed.  This one was failing CI:

```
/home/ubuntu/Semantic-UI-React/src/modules/Dropdown/Dropdown.js
  799:97  error  Infix operators must be spaced  space-infix-ops
```

AgentChris:
hey i've made the changes, i hope they are good, thank you soooo much for your help, i am going to brag all day to my work colleagues :))) good job on this library :+1:

levithomason:
Thanks! Nice work, brag away my friend :)

levithomason:

PR #2024, opened 2017-08-30
chore(tests): replace phantomjs with chromium

layershifter:
## Why?

Chrome teams released an awesome [puppeteer](https://github.com/GoogleChrome/puppeteer), the main advantage is that it installs headless Chromium. So, we can use it in CI without any problems. This PR replaces `phantomjs`, there are maintaining problems while we need more features.

## Advantages

- performance, Chromium (35s) vs PhantomJS (44s) to run tests bundle
- drop `es-shim` and `karma-phantomjs-shim`, minus two dependencies
- #1879 will be unblocked, I need `Proxy` there for tests while `phantomjs` doesn't support it


levithomason:
Awesome!  LGTM.

levithomason:

PR #2038, opened 2017-09-01
feat(Card): add textAlign prop to Card subcomponents

### LINKED ISSUE #2027: Card subcomponents are missing `textAlign` props
Per SUI core, a Card's children can be aligned:

https://github.com/Semantic-Org/Semantic-UI/blob/master/dist/components/card.css#L203

```jsx
<Card>
  <Card.Content textAlign="center">
    This should be centered.
  </Card.Content>
</Card>
```

This is true for all Card subcomponents.

itamar244:
Fixes #2027


itamar244:
Adding only changes for CardContent to check if they are good.
Is it good?

itamar244:
Why the test fail?

itamar244:
finished

itamar244:
ready for a merge

test/specs/views/Card/CardContent-test.js
```diff
@@ -24,5 +24,6 @@ describe('CardContent', () => {
     mapValueToProps: val => ({ content: val }),
   })
 
+  common.implementsTextAlignProp(CardContent, ['left', 'center', 'right'])
```
layershifter: Lets use `_.without(SUI.TEXT_ALIGNMENTS, 'justified')` in tests.

src/views/Card/CardGroup.js
```diff
@@ -26,11 +27,13 @@ function CardGroup(props) {
     items,
     itemsPerRow,
     stackable,
+    textAlign,
   } = props
 
   const classes = cx('ui',
```
layershifter: Lets add newline there:

```diff
-const classes = cx('ui',
+const classes = cx(
+  'ui',

File: src/views/Card/CardMeta.js [hunk 3]
```diff
@@ -8,14 +9,16 @@ import {
...
 } from '../../lib'
 
 /**
  * A card can contain content metadata.
  */
 function CardMeta(props) {
-  const { children, className, content } = props
-  const classes = cx(className, 'meta')
+  const { children, className, content, textAlign } = props
+  const classes = cx(className, useTextAlignProp(textAlign), 'meta')
```
layershifter: Lets make this multiline:

```jsx
const classes = cx(
  className, 
  useTextAlignProp(textAlign),
   'meta',
)
```

layershifter (changes requested):
@itamar244 Thanks for PR :+1: I've add some comments

test/specs/views/Card/CardContent-test.js
itamar244: Maybe a new constant equals to `['left', 'center', 'right']` is needed?
It is smaller when bundling and faster.

layershifter: I don't think that there is real necessity. This code doesn't work in production builds.

src/views/Card/CardContent.js
```diff
@@ -26,11 +29,13 @@ function CardContent(props) {
...
     meta,
+    textAlign,
   } = props
 
   const classes = cx(
     className,
     useKeyOnly(extra, 'extra'),
+    useTextAlignProp(textAlign),
     'content',
   )
```
layershifter: As we update almost all `Card` components, lets also move `className` to end, it's one of our style rules. I overlooked this on initial review.

```jsx
cx(
  useKeyOnly(extra, 'extra'),
  useTextAlignProp(textAlign),
   'content',
  className,
)	
```

layershifter (changes requested):

test/specs/views/Card/CardContent-test.js
```diff
@@ -1,3 +1,6 @@
+import _ from 'lodash';
```
layershifter: Heads up, tests are failing due to lint:

```
/home/ubuntu/Semantic-UI-React/test/specs/views/Card/CardContent-test.js
  1:23  error  Extra semicolon  semi
```

You can run `npm run lint:fix` to fix most errors and then push.  Note that you can ignore warnings.

src/views/Card/CardContent.js
itamar244: Should i first write `'ui'`, then functions, then other literals and then `className`?

layershifter (approved):
@itamar244 Thanks, awesome work 👍

levithomason:
You guys are great!  Thanks again @layershifter for always filling in for me ❤️

levithomason:

PR #2043, opened 2017-09-03
fix(Image): relax `wrapped` propType

layershifter:
This PR relaxes `wrapped` propType. It makes a sence, but it will throw warning when it will be used with `Item.Image` because it always passes `wrapped={true}`:

```jsx
<Item.Image
  as='a'
  href='http://example.com'
  src={src}
  size='tiny'
/>
```


levithomason:

PR #2047, opened 2017-09-04
fix(childMapping): correct argument order for getValue()

### LINKED ISSUE #2040: Value of input not updating correctly when inside Transition.Group
### Steps
* Created new React app using `create-react-app` in WebStorm
* Installed `semantic-ui-react` and `semantic-ui-css` npm packages
* Changed the `App` class in `App.js` to the following:

      class App extends Component {
        constructor(props) {
          super(props);
          this.state = {
            x: ""
          };
        }

        render() {
          return (
            <div>
              <Transition.Group>
                <Input value={this.state.x} onChange={e => {
                  this.setState({x: e.target.value});
                  console.log(e.target.value);
                }}/>
              </Transition.Group>
            </div>
          );
        }
      }

### Expected Result
With the above code, `value` should contain the inputted text, so the inputted text is displayed in the browser and outputted to the console.

### Actual Result
Value of input field seems to not be updating properly.

Input field is not updated in the browser (so it appears read-only), and `value` appears to only contain the last character entered when outputted to the console.

Same behaviour also occurs with `Form.Input` and normal `input`.

### Version
0.72.0

### Testcase
https://codepen.io/anon/pen/Ljqwdb

layershifter:
Fixes #2040. Corrects param order for `getValue()`.

```js
const getValue = (key, prev, next) => (_.has(next, key) ? next[key] : prev[key])
```

### DIFF
--- a/src/lib/childMapping.js
+++ b/src/lib/childMapping.js
@@ -46,15 +46,15 @@ export const mergeChildMappings = (prev = {}, next = {}) => {
   _.forEach(_.keys(next), (nextKey) => {
     if (_.has(nextKeysPending, nextKey)) {
       _.forEach(nextKeysPending[nextKey], (pendingKey) => {
-        childMapping[pendingKey] = getValue(pendingKey, next, prev)
+        childMapping[pendingKey] = getValue(pendingKey, prev, next)
       })
     }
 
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2058, opened 2017-09-08
chore(debug): remove debug from builds

### LINKED ISSUE #1238: Using rollup.js for web builds instead of webpack has unexpected results.

layershifter:
Fixes #1238.

## Why?

This issue is one of  the most longstanding problems. We tried to solve it in #1237 and #1663, but solution wasn't elegant. Levi [suggested](https://github.com/Semantic-Org/Semantic-UI-React/pull/1663#issuecomment-309489919) to use a babel plugin there.

I've tried multiple plugins and selected [`babel-plugin-filter-imports`](https://github.com/ember-cli/babel-plugin-filter-imports). I've done there big work and now we can see it there :smile: 

This PR:
- removes `debug` from our builds
- moves `debug` to dev deps


levithomason:
AHHH!! Thanks so much for this.  This has been a nagging nightmare for me.  Very happy to see this.

levithomason:

PR #2059, opened 2017-09-08
docs(readme): add according.to.localsourc.es

martin-ericsson:
Added another example of web app using Semantic UI React


levithomason:
Awesome, love it!

levithomason:

PR #2063, opened 2017-09-09
fix(Popup): Remove hideOnScroll from window scroll

### LINKED ISSUE #2062: Popup component does not clean event handlers attached to window when it unmounts
### Steps
* Have a Popup component with `hideOnScroll` enabled, inside a view that can get unmounted at any time (for example in the contents of a view which changes its contents based on the selected tab)
* Hover over the Popup
* Without scrolling, change the view which the Popup lived (unmount it)
* Scroll

### Expected Result
I would expect the `hanlePortalUnmount` method of the Popup to remove the `hideOnScroll` method from the `window`'s `scroll` event, when the Popup is unmounted. [Code](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Popup/Popup.js#L307)

### Actual Result
The `hideOnScroll` is not removed from the `window`'s `scroll` event, when the Popup is unmounted. As a result when I am in the new view and scroll, the `hideOnScroll` of the currently unmounted Popup will be executed and try to `setState` in a currently unmounted component, resulting in this error

```
warning.js:36 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Popup component.
```

### Version
Latest

mkarajohn:
Fixes #2062

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -306,6 +306,7 @@ export default class Popup extends Component {
 
   handlePortalUnmount = (e) => {
     debug('handlePortalUnmount()')
+    window.removeEventListener('scroll', this.hideOnScroll)
     const { onUnmount } = this.props
     if (onUnmount) onUnmount(e, this.props)
   }


mkarajohn:
~~Hmm, tests are passing on my machine. What should I look to correct?~~ Probably the commit name

layershifter (approved):
Thanks, you're right. I'm performing refactoring under #1146 and saw same thing.

levithomason:
Thank you!

levithomason:
Released in `semantic-ui-react@0.73.1`.

PR #2067, opened 2017-09-11
feat(Icon|Flag): PureComponent for Icon/Flag

brianespinosa:
Per [comments here](https://github.com/Semantic-Org/Semantic-UI-React/issues/1020#issuecomment-304082926), I have added in support for `PureComponent` in `Icon` and `Flag`. It looks like since that discussion happened last, someone has already added support for SCU and `shallowEqual` so adding `PureComponent` was the last piece for this.

### DIFF
--- a/src/elements/Flag/Flag.d.ts
+++ b/src/elements/Flag/Flag.d.ts
@@ -59,7 +59,7 @@ export interface FlagProps {
     'zambia' | 'zw' | 'zimbabwe';
 }
 
-declare class Flag extends React.Component<FlagProps, {}> {
+declare class Flag extends React.PureComponent<FlagProps, {}> {
 }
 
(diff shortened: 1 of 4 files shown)


layershifter:
@brianespinosa Thanks for the contribution, but we can't merge it. I've done these changes in #2011, in fact this equal to the `PureComponent`:
```js
  shouldComponentUpdate(nextProps) {
    return !shallowEqual(this.props, nextProps)
  }
```

We also support [range](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/package.json#L145) of react vesions, while `PureComponent` was introduced in `15.3`.

brianespinosa:
ha thanks @layershifter it took me so long to get to the updates that you beat me to it.

PR #2070, opened 2017-09-12
docs(ComponentSidebar): add sidebar navigation

layershifter:
This PR is part of work (#1981, #2012) that improves docs usability and performance.

### sidebar

This PR will also add the navigation sidebar like on SUI website.

----

## TODO

- [x] grab title from section, not path
- [x] fix order of sections
- [x] cleanup new parsers
- [x] review components
- [x] refactor doc utils
- [ ] fix rounding in `scrollToAnchor()`
- [x] finish sidebar component


levithomason:
This is great, been on my list for a long :)

I think this can be done by using the already existent `exampleContext` in `docs/app/utils.js`.  We may not need `menuInfo.json`.

Note, in the future, I've been considering the use of dynamic imports for each component route.  This way, we only load the docs people need.  It will make the docs much lighter and faster to load.  Along with this change, I'd like to split up `docInfo.json` and output a json file to each component directory.

docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -70,6 +75,10 @@ class ComponentExample extends Component {
     })
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
```
layershifter: Small optimization. Without it, all examples will be rerendered

layershifter:
### Naming

Moved to #2123.

### `menuInfo.json`

I really want to avoid it, but I do not know another way to collect information for the sidebar. @levithomason do you have any better ideas?

levithomason:
I will try to take a look and comment.  I've marked this PR for review to remind me.

layershifter:
@levithomason Don't pay attention to the refactoring, I moved it into #2123. However, question about `menuInfo.json` is still open.

layershifter (commented):
@levithomason I merged this PR with and made the cleanup there. it's ready for review :v: 

However, I can't fix the problem with `scrollToAnchor()`. It's easy to reproduce, just pull this branch and try to click on different sidebar items on the `Header` page. The scroll sometime will not reach the matching position and the wrong sidebar item will be active. There is a rounding problem, but I don't have any idea how to fix it :thinking:

levithomason:
I've pushed some usability fixes and would like to do more.

The issue you stated still persists.  It seems to me the cause is that Visibility expects the element to be _passed_, while `scrollToAnchor` does not pass the element but scrolls directly to the element.  This could be due to Visibility's use of `getBoundingClientRect`, which returns fractional values for top/bottom/left/right.

I will come back to this again.  I am out of time for today. 👋

layershifter:
Solved merge conflict there 👍

levithomason:
I'm going to merge this and we can iterate on it.  Maybe someone will feel motivated and submit a PR as well.

levithomason:

Issue #2041, opened 2017-09-03
Dropdown: ignores closeOnChange if it has children and isn't a search


9teen90nine:
**Steps to Reproduce**
1. Create a dropdown with children and closeOnChange={false}
2. Select an option

**Expected**
The Dropdown should not close

**Result**
The Dropdown closes

**Testcase**

https://codepen.io/9teen90nine/pen/PKLzQz

9teen90nine:
A little digging got the bug back.

layershifter:
@9teen90nine Thanks for detailed report :+1:

levithomason:
Per the docs, using children is not supported and requires you to manage the state yourself:

![image](https://user-images.githubusercontent.com/5067638/30524955-534a9ad6-9bb2-11e7-9405-05c3973a9918.png)

Note that we'll likely completely remove support for using children and replace it with shorthand props only.  See #1757 for that work.

If the JSX children in the codepen are changed to the supported `options` prop, each case works correctly: https://codepen.io/levithomason/pen/wrKxjo?editors=0010

I will close this for now.  We'll still merge the PR if it is inline with the future Dropdown API.

PR #2074, opened 2017-09-14
fix(Dropdown): keep dropdown open when closeOnChange is set to false

ericmak624:
Root cause: When `options` prop is used, clicking on the dropdown item will trigger `handleItemClick` which will utilize `closeOnChange` properly.
When `options` prop is undefined(using `children`), clicking on the dropdown item will trigger `handleClick` instead which will use `toggle`(if `search` prop is falsy) instead of `closeOnChange`.

Fix: Replaced `this.close(e)` in `toggle` with `this.closeOnChange(e)` to keep dropdown open when `closeOnChange` is set to `false`.

issue: Dropdown: ignores closeOnChange if it has children and isn't a search #2041

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1098,7 +1098,7 @@ export default class Dropdown extends Component {
     this.setState({ focus: hasFocus })
   }
 
-  toggle = e => (this.state.open ? this.close(e) : this.open(e))
+  toggle = e => (this.state.open ? this.closeOnChange(e) : this.open(e))
 
   // ----------------------------------------
   // Render


levithomason:
I'm hesitant on this.  We don't support state management when there are children in the Dropdown, the user needs to manage state in that case.

image

The fact that the click event is bubbling to the top level `handleClick` is due the user not handling the click themselves.  This subcomponent Dropdown API will likely be removed in favor of shorthand props only, see #1757.

The main issue I have is that the `toggle()` method should simply toggle the Dropdown, without regard to any other state or conditions.  Currently, changing `toggle` to use `closeOnChange` works only by happenchance.  However, if `toggle` is called in a case where it should absolutely toggle irrespective of a change event, it will fail.

I'm leaning toward not moving forward on this change, but it would be great to get some feedback.

ericmak624:
it's good to know the API changes for `Dropdown`. 

You're spot on regarding the issue and `toggle()`. Currently, using `closeOnChange()` in `toggle()` simply checks if either `multiple` is `true` or `closeOnChange` is `false` in addition to calling `this.close()`. However, I can understand this may be confusing when `toggle()` is being used more often in the future than just `handleClick()` and `handleIconClick()` rightnow. 

My opinion on the shorthand only is I'm afraid it may lose some flexibilities, but I might be wrong.

If not moving forward with the PR, my suggestion is to let users know that they need to use the `onClick` prop with `e.stopPropagation()` on `children` in order for `closeOnChange` to work properly.

levithomason:
OK, we'll close this for now then.

>My opinion on the shorthand only is I'm afraid it may lose some flexibilities, but I might be wrong.

I would be interested to see the use cases you are concerned with over at #1757.  My thinking is that JSX components have the ability to map 1:1 to JS objects, making all scenarios possible.

>If not moving forward with the PR, my suggestion is to let users know that they need to use the onClick prop with e.stopPropagation() on children in order for closeOnChange to work properly.

I don't want to begin telling users what they'd need to do to get subcomponents working as in the future we will not have support for them.  Also, to get them working properly, they'd actually need to rewrite the all dropdown logic surrounding options :)  This is largely why we're not supporting them, it isn't really practical for users to try to rewrite the state management for options.

PR #2081, opened 2017-09-18
chore(package): replace npm with yarn

levithomason:
NPM is broken.  `package-lock.json` still doesn't work, and even if it did, it only works for users using the latest `npm` cli.  Users on old versions still get a worse experience a no lock file safety.

This PR:

- replaces `npm` with `yarn`
- upgrades CircleCI to v2


levithomason:
Well, I cannot seem to figure out why ChromeHeadless is not included with pupeteer when installing with yarn.

I will have to come back next time and SSH into the box for more debugging.  This is the only thing blocking this PR!

layershifter:
# max callStack

We regularly get max call stack errors on CI, I'm think that we increase this value:

```
   customLaunchers: {
      puppeteer: {
        base: 'ChromeHeadless',
        flags: ['--stack-trace-limit 50000']
      }
    }
```

I'm not sure, but it should work.

### missing libs

Seems Docker image of Circle missing required libs. I think that the most correct solution is use own Docker image there.

```
FROM circleci/node:8

RUN apt-get update && \
apt-get install -yq gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 \
libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 \
libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 \
libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 \
ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget
```

GoogleChrome/puppeteer#290

levithomason:
Thanks for the lead, the list of missing deps was indeed the issue.  Instead of managing our own image, I just updated those deps as the first step on CI.

We did not, however, have the issues related to sandboxing so I have not added any flags for that.

I also added the increased stack trace flag, we'll see if that resolves the maximum call stack exceeded issue!

levithomason:

dvdzkwsk:
Praise be.

PR #2083, opened 2017-09-18
fix(Sticky): disable styles when `active` is false

### LINKED ISSUE #2082: Sticky: `active` prop doesn't reset sticky state
### Steps

Create two sticky components , one on the left, and one on the right. 
The right one should be inactive in default. 
The left one has an `onStick` event to update the state to make the right sticky active. 

### Expected Result

The right sticky should become `active` after the left sticky fires `onStick` event.

### Actual Result

The `active` prop doesn't seem to update on state change. The right sticky remains inactive. 

### Version
0.73.1

### Testcase
https://codepen.io/anon/pen/xXwBxj (Invalid example, please read post 2)

layershifter:
Fixes #2082.


levithomason:

PR #2086, opened 2017-09-19
docs(parser): parse custom values and mixed enums

### LINKED ISSUE #2078: Docs: parse {custom} and display correct values
should list some acceptable values here

```
only |   | {custom} | A row can appear only for a specific device, or screen sizes.
```
i get error
```
ReferenceError: computer is not defined
```
on `<Grid.Column only={computer}>`

layershifter:
Fixes #2078.

#### after

image


levithomason:
The `type` in the `{bool|enum}` case should remain `{bool|enum}`.  The list of enums should show `very`.  The value `bool` is not a valid enum value, it is a type.

layershifter:
Thanks for feedback, you're right there. I will fix it tomorrow

layershifter:
@levithomason I've updated PR as you requested, screenshot in PR header was also updated :+1:

levithomason:
Great, thanks again!

levithomason:

PR #2088, opened 2017-09-19
fix(Visibility): fix behaviour of reverse calls

layershifter:
I'm working on #2070, but it can't be finished without changes there. 

We have wrong behaviour for `*reverse` callbacks and it's easy to show, see GIFs. We fire them always, but in fact they should be called in the oppositite condition.

```js
// topPassed
if(calculations.topPassed) module.execute(callback, callbackName);

// topPassedReverse
if(!calculations.topPassed) module.execute(callback, callbackName);
```

#### SUI

<details>
  <summary>Click to expand</summary>
  
</details>

#### SUIR

<details>
  <summary>Click to expand</summary>
  
</details>

### DIFF
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -199,7 +199,7 @@ export default class Visibility extends Component {
 
     const { context, fireOnMount } = this.props
 
-    context.addEventListener('scroll', this.handleUpdate)
+    context.addEventListener('scroll', this.handleScroll)
     if (fireOnMount) this.update()
   }
 
@@ -207,7 +207,7 @@ export default class Visibility extends Component {
     if (!isBrowser) return
 
     const { context } = this.props
-    context.removeEventListener('scroll', this.handleUpdate)
+    context.removeEventListener('scroll', this.handleScroll)
   }
 
   // ----------------------------------------
@@ -260,7 +260,7 @@ export default class Visibility extends Component {
     })
   }
 
-  handleUpdate = () => {
+  handleScroll = () => {
     if (this.ticking) return
 
     this.ticking = true
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -181,7 +181,7 @@ export default class Visibility extends Component {
     topVisible: false,
   }
 
-  occurredCallbacks = []
+  firedCallbacks = []
 
   // ----------------------------------------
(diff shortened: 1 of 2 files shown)


layershifter (commented):
Ready for review. Now the component's behaviour matches to SUI. I've also shipped some cleanups for tests there.

src/behaviors/Visibility/Visibility.js
```diff
@@ -171,108 +171,76 @@ export default class Visibility extends Component {
...
     passing: false,
-    onScreen: false,
     offScreen: false,
+    onScreen: false,
+    topPassed: false,
+    topVisible: false,
   }
 
-  firedCallbacks = []
+  occurredCallbacks = []
```
levithomason: I think the original name here is more appropriate, `firedCallbacks`.

src/behaviors/Visibility/Visibility.js
```diff
@@ -171,108 +171,76 @@ export default class Visibility extends Component {
...
   }
 
   componentDidMount() {
     if (!isBrowser) return
 
     const { context, fireOnMount } = this.props
 
-    context.addEventListener('scroll', this.handleScroll)
-    if (fireOnMount) this.handleUpdate()
+    context.addEventListener('scroll', this.handleUpdate)
```
levithomason: nit: suggest keeping `handleScroll` for the scroll handler.  It makes it clear when/why it is called.

src/behaviors/Visibility/Visibility.js
layershifter: Restored :+1:

test/specs/behaviors/Visibility/Visibility-test.js
```diff
@@ -26,67 +26,72 @@ const mockScroll = (top, bottom) => {
     }
   }
 
-  window.dispatchEvent(new Event('scroll'))
+  domEvent.scroll(window)
```
levithomason, line 29: ❤️

src/behaviors/Visibility/Visibility.js
layershifter: This comes from #2091, this handler will also handle `resize` event

levithomason (approved):
Good to merge after the small naming updates.

layershifter:
@levithomason I've restored naming, as you approved changes in this PR, I merged it :boom:

levithomason:

PR #2090, opened 2017-09-20
feat(Visibility): add `direction` for calculations

layershifter:
Cut from #2088.

# What?

SUI has [`direction`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/visibility.js#L869) value in `calculations`, this PR adds it into the component. I don't added `static` value because I don't know when it really can be.


README.md
```diff
@@ -111,7 +111,7 @@ Any other issue labeled [`help wanted`][4] is ready for a PR.
 |-----------------|-----------------|-----------------|-----------------|--------------------|
 | ✓ Button        | ✓ Breadcrumb    | ✓ Advertisement | ✓ Accordion     |   Form Validation  |
 | ✓ Container     | ✓ Form          | ✓ Card          | ✓ Checkbox      | *API (NA)*         |
-| ✓ Divider       | ✓ Grid          | ✓ Comment       | ✓ Dimmer        | ✓ Visibility (NA)  |
```
layershifter, line 114: Cruft :smile:

levithomason:

PR #2091, opened 2017-09-20
feat(Sticky|Visibility): listen for `resize` events

layershifter:
Cut from #2088.

# What?

`Sticky` and `Visibility` should also listen for `resize` events as SUI component does.

- [`Sticky`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/sticky.js#L209)
- [`Visibility`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/visibility.js#L201)


levithomason:
Minor conflict to address here.

levithomason:
Fixed conflict, good to merge after tests pass and you confirm the conflict fix is correct.

layershifter:
@levithomason I will fix it now

layershifter:
Fixed, waiting for CI pass

layershifter:
@levithomason As you approved these changes, I merged them

levithomason:

PR #2094, opened 2017-09-20
feat(eventStack): support for different targets

layershifter:
Cut from #2088.

# What?

This PR adds ability to use `eventStack` on custom targets, except `document` and `window`. This means that we can use it in `Sticky` and `Visibility` and gains its benefits :+1: 

In fact, I moved old `eventStack` to `EventTarget` and fully kept its functionality. New `eventStack` now only manages instances of `EventTarget`s.


levithomason:
Fixed conflict, good to merge once test pass and you confirm.

layershifter:
@levithomason thanks, looks correct :+1: Let's merge this before #2091, because there will be conflict again :smile:

levithomason:

PR #2099, opened 2017-09-21
fix(Modal|Popup|Portal): fix usage of eventStack sub/unsub

### LINKED ISSUE #2075: eventStack: Multiple Popups do not behave correctly
I have some popups like this in my project (this is a simplified version with the important bits):

``` jsx
class SomePopup extends React.Component {
  constructor(props) {
    super(props);

    this.state = {
      open: false
    };

    this.close = this.close.bind(this);
    this.open = this.open.bind(this);
  }

  close() {
    this.setState({ open: false });
  }

  open() {
    this.setState({ open: true });
  }

  render() {
    const { /*Destructured props*/, ...rest } = this.props;
    return (
      <Popup
        {...rest}
        onClose={e => {
          this.close();
          // do something with a prop method
        }}
        onOpen={e => {
          this.open();
          // do something with a prop method
        }}
        trigger={
          <SomeButton
            onClick={e => {
              // do something  with a prop method
            }}
            active={this.state.open}
          />
        }
        on="click"
      >
        <div>Hello</div>
      </Popup>
    );
  }
}
```

Where `SomeButton` is a stateless button component.

As you can see the Popup is not controlled by its parent, it's left to its default behaviour (i.e. click outside to close) and when it opens or closes, the `SomePopup`'s state will update accordingly in order to reflect the fact to the button that triggered the Popup, and have it stay in "active" state as long as its popup is open.

Now, I have noticed that Popups will close when you click outside, **however they will not close when you click a button that opens another popup**.

Am I doing something wrong with the Popups or is this some kind of bug?

### Steps
Assume we have Popup A and Popup B that are implemented as shown above.

Do the following in this order:

* Click on Popup A's trigger button
* Click on Popup B's trigger button

### Expected Result

I would expect that once I click on the trigger of Popup B then Popup A would close and Popup B would open

### Actual Result

When the Popup A is open, and you click on the trigger component of Popup B the following will happen in this order: 

* The `onClick` handler of the Popup B's  trigger component will execute first
* Then the `onOpen` of  Popup B will execute 
* Then immediately after, the `onClose` handler of  Popup B will execute .

As a result it looks like as if  Popup B did not open at all, when in reality, it opened and closed immediately. It makes no sense for  Popup B's `onClose` to fire immediately after it was opened.

In contrast, when Popup A is open, and you click outside of it (not on the trigger of Popup B) the following will happen in this order: 

* First the `onClick` handler on Popup A's trigger component will execute 
* Then Popup A's `onClose` handler will execute .

I have the feeling that the Portal component used, does not keep proper track of the Popups it has open, and while it's supposed to close the previous Popup it closes the new one

### Version
0.72.0

### Testcase
https://codepen.io/mkarajohn/pen/xXbJYK?editors=0010

austinfox:
fixes #2075

Without this change, any event handlers added to EventStack under the 'Portal' namespace will be appended to an array. If another Portal is then mounted, new event handlers will be added to the EventStack and take precedence over the original portal's event handlers.

This change allows each portal to store its event handlers under its own namespace in EventStack, so when you click from one portal to another, the first one is appropriately closed.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -62,6 +62,9 @@ class Portal extends Component {
     /** Initial value of open. */
     defaultOpen: PropTypes.bool,
 
+    /** Event pool namespace that is used to handle component events */
+    eventPool: PropTypes.string,
+
     /** The node where the portal should mount. */
     mountNode: PropTypes.any,
 
@@ -115,9 +118,6 @@ class Portal extends Component {
     /** Controls whether or not the portal should open when mousing over the trigger. */
     openOnTriggerMouseEnter: PropTypes.bool,
 
-    /** Event pool namespace that is used to handle component events */
-    eventPool: PropTypes.string,
-
     /** Controls whether the portal should be prepended to the mountNode instead of appended. */
     prepend: PropTypes.bool,
 
@@ -128,8 +128,8 @@ class Portal extends Component {
   static defaultProps = {
     closeOnDocumentClick: true,
     closeOnEscape: true,
-    openOnTriggerClick: true,
     eventPool: 'default',
+    openOnTriggerClick: true,
   }
 
   static autoControlledProps = [
@@ -376,9 +376,9 @@ class Portal extends Component {
     debug('mountPortal()')
 
     const {
+      eventPool,
       mountNode = isBrowser ? document.body : null,
       prepend,
-      eventPool,
     } = this.props
 
     this.rootNode = document.createElement('div')
@@ -398,7 +398,6 @@ class Portal extends Component {
     if (!isBrowser || !this.rootNode) return
 
     debug('unmountPortal()')
-
     const { eventPool } = this.props
 
     ReactDOM.unmountComponentAtNode(this.rootNode)
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -139,6 +139,7 @@ class Portal extends Component {
 
   componentDidMount() {
     debug('componentDidMount()')
+    this.uniqueId = _.uniqueId()
     this.renderPortal()
   }
 
@@ -384,8 +385,8 @@ class Portal extends Component {
       mountNode.appendChild(this.rootNode)
     }
 
-    eventStack.sub('click', this.handleDocumentClick, 'Portal')
-    eventStack.sub('keydown', this.handleEscape, 'Portal')
+    eventStack.sub('click', this.handleDocumentClick, `Portal${this.uniqueId}`)
+    eventStack.sub('keydown', this.handleEscape, `Portal${this.uniqueId}`)
     _.invoke(this.props, 'onMount', null, this.props)
   }
 
@@ -403,8 +404,8 @@ class Portal extends Component {
     this.rootNode = null
     this.portalNode = null
 
-    eventStack.unsub('click', this.handleDocumentClick, 'Portal')
-    eventStack.unsub('keydown', this.handleEscape, 'Portal')
+    eventStack.unsub('click', this.handleDocumentClick, `Portal${this.uniqueId}`)
+    eventStack.unsub('keydown', this.handleEscape, `Portal${this.uniqueId}`)
     _.invoke(this.props, 'onUnmount', null, this.props)
   }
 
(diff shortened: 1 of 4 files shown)


layershifter (changes requested):
@austinfox thanks for picking up this :+1: 

Proposed changes useless, because in fact it's equals to `default` pool of the `eventStack`. My initial proposal about the `exclusive` prop is also not ideal.

### Problem

We have a couple of bugs with multiple `Modal`s, `eventStack` solves them and improves performance.

What `eventStack` actually does? It has an entity called `pool`. Event handlers that included in `default` pool will be fired usual. But, only the last added event handler to the named pool will be fired when corresponding event will be occured.

You can open [this](https://react.semantic-ui.com/modules/modal#modal-example-multiple) `Modal` example. Open both modals and then press `Esc` key. Only top modal will be closed and it's correct. But, as we can see it causes problems for `Popup`s.

### Solution

So, my idea is to add the `eventPool` prop to the `Portal` which defaults to `default`:

```js
const { eventPool } = this.props

eventStack.sub('keydown', this.handleEscape, eventPool)
```

This prop should be also added to `Modal` and `Popup`, where it will be defaults to `Modal` and `default` accordingly. It will keep `Modal` working correctly and solve the problem with event propagation in `Popup`.

Let me know what you think.

austinfox:
@layershifter Yea I was thinking about adding a prop to Modal, Popup, and Portal. I'm not convinced it will be clear to others how to prevent this bug though even if we add eventPool as a prop. If they have an in-depth knowledge about the inner workings of Portal and eventStack, then yes they will be able to fix the problem. But by default, clicking on a popup and then clicking on another popup will be broken. Why not default to having each Popup use its own eventPool? I think that is standard behavior, whenever you click away from a popup is should almost always close.

austinfox:
@layershifter ignore my last comment. Taking a look at the EventStack code, I now understand what you mean. When the 'default' poolName is used, all handlers are called for that event. This will work just fine, I'll modify the PR in a few hours. Is there any way we can get this change into a new version of SUI react asap? My team site is currently breaking for IE users because we cannot use this feature. I had to rely on='focus' and do some other hacks that work well for chrome, Firefox, safari, and edge, but... sigh, IE...

src/addons/Portal/Portal.js
```diff
@@ -115,6 +115,9 @@ class Portal extends Component {
     /** Controls whether or not the portal should open when mousing over the trigger. */
     openOnTriggerMouseEnter: PropTypes.bool,
 
+    /** Event pool namespace that is used to handle component events */
+    eventPool: PropTypes.string,
```
layershifter: Please add prop in alphabetical order.

src/addons/Portal/Portal.js
```diff
@@ -126,6 +129,7 @@ class Portal extends Component {
     closeOnDocumentClick: true,
     closeOnEscape: true,
     openOnTriggerClick: true,
+    eventPool: 'default',
```
layershifter: ^

src/addons/Portal/Portal.js
```diff
@@ -374,6 +378,7 @@ class Portal extends Component {
     const {
       mountNode = isBrowser ? document.body : null,
       prepend,
+      eventPool,
```
layershifter: ^

src/addons/Portal/Portal.js
```diff
@@ -394,6 +399,8 @@ class Portal extends Component {
 
     debug('unmountPortal()')
 
+    const { eventPool } = this.props
```
layershifter: New line before this declaration is useless.

layershifter (changes requested):
Looks good to me. I left some comments about styling.

layershifter:
> When the 'default' poolName is used, all handlers are called for that event.

Exactly :+1: 

> Is there any way we can get this change into a new version of SUI react asap?

I will perform style changes tomorrow, if you will not do them before me :smile: I want to hear there @levithomason, too. I think that we can include it to next release on this weekend.

layershifter (approved):

levithomason:
I've been following this issue/PR and I'm happy with this result 👍.  Thanks for hashing it out guys.

levithomason:

drewsbrew:
Looking at the multiple modals example in the docs for v0.76, a single press of the Escape key closes both modals. I am seeing the same behavior in my project even with unique `eventPool` props values. Has this regressed or am I missing something?

brianespinosa:
@drewsbrew If that is what you are experiencing, please open up a separate bug report form with a codesandbox example.

layershifter:
@drewsbrew Bug actually exists, however @brianespinosa please create new issues instead of necroposting.

Will be fixed in #2329.

PR #2102, opened 2017-09-22
breaking(Dropdown): correct onSearchChange signature

### LINKED ISSUE #2052: Dropdown: onSearchChange signature should match that of Search
### Steps
onSearchChange previously provided event, value however the event.target no longer provides target data like name.

We have been using a generic search handler on a series of Dropdown components that populates state using field names and stripping the -search

I've noted several issues in recent months changing all event handler signatures to match a (event, data) pattern. Should this also be the case for Dropdown onSearchChange?

```
const name = event.target.name.match(/(\w+?)-search/)[1]
```

### Expected Result

name is a valid fieldname on the form

### Actual Result

event.target.name evaluates to null

Inspection of target in the console shows it no longer includes a name attribute

### Version
0.72.0

layershifter:
# BREAKING

Fixes #2052.
Rel #1828.

onSearchChange | (e, value) | (e, { ...props, searchQuery })

### Before

```jsx
<Dropdown onSearchChange={(e, value) => console.log(e, value)} />
```

### After

```jsx
<Dropdown onSearchChange={(e, data) => console.log(e, data.searchQuery)} />
```

### DIFF
--- a/src/modules/Dropdown/Dropdown.d.ts
+++ b/src/modules/Dropdown/Dropdown.d.ts
@@ -174,9 +174,9 @@ export interface DropdownProps {
    * Called on search input change.
    *
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
-   * @param {string} value - Current value of search input.
+   * @param {object} data - All props, includes current value of searchQuery.
    */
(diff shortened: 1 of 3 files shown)


levithomason:

Issue #2048, opened 2017-09-04
fix(Dropdown): input doesn't get cleared when adding item


skulptur:
**Steps to Reproduce**
1. Using the example specified on the Testcase bellow, type 'hi' and hit enter
2. Hit delete to remove the added selection
3. Type 'hi' again and hit enter
4. Repeat steps 1 to 3 and use the word 'hello'

**Result**
The second time the user hits enter, the word 'hi' remains in the input and the dropdown still suggests the addition of 'hi' to the options. When adding 'hello', the input is cleared, but the dropdown still suggests a new addition of 'hello'.

**Expected**
It is expected that the dropdown clears the input after an addition and recognizes that the option has been added instead of suggesting a new addition.

**Testcase**
If the docs show the issue, use: https://react.semantic-ui.com/modules/dropdown#dropdown-example-multiple-allow-additions
Otherwise, fork this to get started: http://codepen.io/levithomason/pen/ZpBaJX

layershifter:
Duplicate #1954. This behaviour was introduced in #1795, there was also PR for this #1955, but it wasn't finished. I like idea suggested [there](https://github.com/Semantic-Org/Semantic-UI-React/issues/1829#issuecomment-326625654).

spredemann:
I see, but if I understand correctly it's only partially duplicate. They don't mention the dropdown still suggesting a new addition despite the new option having been added already.

skulptur:
I didn't find mention on the problem where the dropdown suggests the addition of an item that has already been added. I'm talking about the allowAdditions flag, as opposed to when selecting a new item.

dardub:
The new behavior introduced in #1795 also introduced a bug when the allowAdditions is used. Can this issue be re-opened?

This bug makes it too easy to create duplicate entries by the user.

layershifter:
You can control your options with `onAddItem` handlers and skip duplicates. After we will make `searchQuery` contollable you will can to clear its value after `onChange`. I also think that we should transform the `additionalLabel` prop to shorthand, this will allow to pass `null` when option has already added.

PR #2109, opened 2017-09-23
feat(Dropdown): add `searchQuery` prop

### LINKED ISSUE #1829: Dropdown: how to stop it from clearing the input field? [feature request]
I'm using the dropdown like so,

 ```
      <Dropdown 
                    placeholder='Enter basket name...' 
                    search 
                    selection
                    options={this.getDropDownOptions()}
                    onBlur={this.onDropDownBlur}
                    onChange={this.onChange}
                    onKeyPress={this.onKeyPress}
                    maxLength='64'
                    style={{ width: this.props.width ? this.props.width : '100px'}}
                    />

```

is there a way I can prevent the default behavior of clearing the input field onBlur? I'd like whatever the user typed to still remain in the input field. It would be nice if there was a prop so I could turn off this default behavior.

layershifter:
Fixes #1829.
Rel #2048.

This PR allows to control the`searchQuery` via autocontrolled prop, now it has same behaviour as the `value` prop.
Adds new tests and example.


layershifter:
We should merge #2102 before and update the docs example there.

layershifter:
I will fix merge conflict now

levithomason:

PR #2110, opened 2017-09-23
docs(Dropdown): fix searchQuery example

layershifter:
It's strange, but the change of example wasn't included in #2109. This PR fixes it.


levithomason:

patrikmolsson:
Hi, 

This also means that https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleRemote.js#L24

needs updating.

PR #2111, opened 2017-09-23
perf(lodash): remove _.omit and _.isEqual

### LINKED ISSUE #864: perf(lodash): remove use of _.omit (and optimize lodash)

levithomason:
Fixes #864 

Most tasks there were complete.  This PR removes the final two methods with perf issues, omit and isEqual.  We've already optimized lodash in the bundle by picking only the methods we use.

### DIFF
--- a/src/lib/objectDiff.js
+++ b/src/lib/objectDiff.js
@@ -19,5 +19,6 @@ export default (source, target) => _.transform(source, (res, val, key) => {
   // deleted keys
   if (!_.has(target, key)) res[key] = '[DELETED]'
   // new keys / changed values
+  // Note, we tolerate isEqual here as this is a dev only utility and not included in production code
   else if (!_.isEqual(val, target[key])) res[key] = target[key]
 }, {})
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -14,6 +14,7 @@ import {
   makeDebugger,
   META,
   objectDiff,
+  shallowEqual,
   useKeyOnly,
   useKeyOrValueAndKey,
 } from '../../lib'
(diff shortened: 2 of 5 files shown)


levithomason:

PR #2112, opened 2017-09-24
breaking(Dropdown): clean searchQuery after selection in multiple search Dropdown

### LINKED ISSUE #1829: Dropdown: how to stop it from clearing the input field? [feature request]

layershifter:
Fixes #1829.
Rel #1795.
Rel #2048.
Rel #2109.

```jsx
<Dropdown search multiple />
```

| Type  | Behaviour |
| ------------- | ------------- |
| **Before** | Search query was kept after item selection |
| **After** | Search query will be cleared after selection, if you want to keep previous behaviour, use [controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components) and handle `searchQuery` manually |

In #1795 we changed the behaviour of `searchQuery`, we don't clear it after selection in `multiple search` `Dropdown`. This PR reverts this change because:
- this behaviour isn't intuitive for users and we got claims about it
- #2109 added ability to control `searchQuery`

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleSearchQuery.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleSearchQuery.js
@@ -7,10 +7,7 @@ import { stateOptions } from '../common'
 export default class DropdownExampleSearchQuery extends Component {
   state = { searchQuery: '' }
 
-  handleChange = (e, { value }) => this.setState({
-    value,
-    searchQuery: '',
-  })
+  handleChange = (e, { searchQuery, value }) => this.setState({ searchQuery, value })
 
   handleSearchChange = (e, { searchQuery }) => this.setState({ searchQuery })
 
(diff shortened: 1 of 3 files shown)


levithomason:
Thank you sir!

levithomason:

spredemann:
@layershifter you mentioned 'This PR reverts this change' (keeping the searchQuery after selection). I just upgraded to 0.75 and the behavior seems to still default to keeping the query. Did I misinterpret your description or am I doing something wrong?

PR #2113, opened 2017-09-24
fix(Dropdown): fix call order of `onAddItem` and `onChange` handlers

### LINKED ISSUE #2106: Dropdown: call onAddItem after onChange
### Steps
(Using the testcase)
1. Use options with distinct `text` and `value` properties;
2. In a Dropdown with `allowAdditions` set, type a new entry and confirm addition;

### Expected Result
A way to set a custom value to the newly added item is provided before it is auto-selected.

### Actual Result
The newly added item has both its `text` and `value` properties set to the input text and is then selected based on that string.

This is caused by the item insertion logic in [getMenuOptions](https://github.com/Semantic-Org/Semantic-UI-React/blob/e2fc8f5c6056521166fa7269ef0788af65e9b334/src/modules/Dropdown/Dropdown.js#L779) that sets both the `text` and `value` properties of new items to the user-inserted text when the user confirms the entry. `onAddItem` is then called with the new value (which is fine), immediately followed by `setValue`, `setSelectedIndex` and `handleChange`, which attempt to select the item using the same value. 

This behavior gives no opportunity to set an appropriate value property (a unique ID, for instance) to the newly inserted item before it is auto-selected. In the testcase, this is attempted in the `onAddItem` handler, but setting the value property there means the `setValue`, `setSelectedIndex` and `handleChange` calls fail as they use the old value (the user-inserted text). There is then no elegant way to select the newly added item.

### Version
0.73.1

### Testcase
https://codepen.io/anon/pen/xXEVqK

layershifter:
Fixes #2106.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -579,7 +579,7 @@ export default class Dropdown extends Component {
 
   makeSelectedItemActive = (e) => {
     const { open } = this.state
-    const { multiple, onAddItem } = this.props
+    const { multiple } = this.props
 
     const item = this.getSelectedItem()
     const value = _.get(item, 'value')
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2117, opened 2017-09-25
fix(eventStack): correct handling of removed targets

layershifter:
Rel [#2075#comment](https://github.com/Semantic-Org/Semantic-UI-React/issues/2075#issuecomment-331749246).

Changes:
- `eventStack` now uses `Map`
- corrected behaviour of `unsub`

### DIFF
--- a/src/lib/eventStack/eventStack.js
+++ b/src/lib/eventStack/eventStack.js
@@ -1,35 +1,32 @@
-import _ from 'lodash'
-
 import isBrowser from '../isBrowser'
 import EventTarget from './EventTarget'
 import normalizeTarget from './normalizeTarget'
 
 class EventStack {
-  _eventTargets = {}
-  _targets = []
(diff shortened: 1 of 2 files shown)


src/lib/eventStack/eventStack.js
```diff
@@ -49,10 +46,12 @@ class EventStack {
     if (!isBrowser) return
 
     const { target = document, pool = 'default' } = options
-    const eventTarget = this._find(target)
+    const eventTarget = this._find(target, false)
```
layershifter, line 49: Before this change, `_find` created always new instance of `EventTarget` when it was removed

layershifter:
@levithomason This problem looks quite critical to me. Can you review this and release `0.74.2`?

levithomason:
Released in `semantic-ui-react@0.74.2`

Issue #1604, opened 2017-04-21
Thank you, maintainers and contributors


jayphelps:
I'm an open source maintainer myself, so I realize people making issue tickets for things that aren't issues can be annoying. That said, I think maintainers and contributors don't get enough props, so I wanted to take a moment.

At Netflix my particular team (Edge Developer Experience) aims to make the lives of other engineers here easier, part of which is the creation of many fairly complex _internal_ tools. So we can focus on features instead of design our team started using Semantic-UI almost two years ago and while no project is perfect, it has made our lives much much easier. We maintained internal React version of some of the components as things progressed, but when Semantic-UI-React came along and started to mature, we started to switch over to it. Needless to say, Semantic-UI-React has quickly become impressive.

Thank you maintainers and all contributors for your valuable and impactful library. We receive positive feedback on the look, functionality, and performance of our internal apps and we owe a large part of that to Semantic-UI and Semantic-UI-React ❤️ 

(This comment is my personal views, and does not represent the views of Netflix. Netflix is a big company and teams are free to use any technology. Other teams may and do choose other things.)

Here's an example of one internal app, Raven. If you're interested, I discuss the goals of it in the context of Reactive Programming [in my talk](https://www.youtube.com/watch?v=uODxUJ5Jwis) (but I didn't mention Semantic-UI 😞)

![image](https://cloud.githubusercontent.com/assets/762949/25294580/b4ec1780-2694-11e7-9dd3-e5163170befb.png)

levithomason:
This such great news!  Thank you for taking the time to express this.  I don't think most users understand the commitment and time sacrifice it takes.  We have full-time jobs, kids, and other obligations as well.  Doing something like this certainly takes a lot out of someone.

Even though this is internal tooling and your personal views, is it possible to get this listed in our README.md under [Built With Semantic UI React](https://github.com/Semantic-Org/Semantic-UI-React#built-with-semantic-ui-react)?

Again, thanks for the issue and your very kind words.

jayphelps:
@levithomason absolutely. 👍

Anderzon00:
well

PR #2122, opened 2017-09-25
docs(readme): add amazon publishing to built with

austinfox:


levithomason:
This is superb!  I hoist well-known companies to the top of the list.  I've taken a snippet from the apub.com landing page for the description.  It should also generate a little more SEO hits.

Once you verify the changes are OK to publicize, I'll merge.

austinfox:
Thanks for the changes, LGTM :)
I'll post in this issue (https://github.com/Semantic-Org/Semantic-UI-React/issues/1604) in a bit with an internal screen shot of our app. We have been migrating from stock semantic-ui with custom react wrappers to semantic-ui-react, and it has been an absolute pleasure to use. I'm so glad to be rid of jquery, thanks so much for all your work. I'll try to be an active contributor and make this package even better.

levithomason:
Very cool, we're looking for a few hands to join!  Will wait for the screenshot.

levithomason:
Merging for now.  Would be great to get some screenshots on this issue and update the README to link here as well.

PR #2123, opened 2017-09-26
docs(ComponentDoc): refactor and optimize

layershifter:
_Splitted from #2070 for the easier review of proposed changes._

This PR is part of work (#1981, #2012) that improves docs usability and performance. Contains the deep refactor of `ComponentDoc`.

### Naming

This PR includes refactor of `ComponentDoc`. As before, I've refactored it to smaller components and move down `showPropsFor` to `ComponentProps`. But I met naming problem and I hope for your help there.

- `ComponentProps` => `ComponentTable` - after performed refactoring, this component has been renamed, since by context it is a **table** with of props
- `ComponentProps/*` => `ComponentProp/*` - there are numerous small components that represent props, I moved them there and it's what makes me doubt


gulp/plugins/gulp-react-docgen.js
```diff
@@ -52,9 +52,13 @@ export default (filename) => {
...
         }
       })
+
+      parsed.path = file.path
+        .replace(`${process.cwd()}/`, '')
+        .replace(new RegExp(_.escapeRegExp(path.sep), 'g'), '/')
       parsed.props = _.sortBy(parsed.props, 'name')
 
-      result[relativePath] = parsed
+      result[componentName] = parsed
```
layershifter: `posixPath` is now calculated by plugin. For `key` in `docgenInfo.json` will be used a component name, it's unique, so we don't to regexps to find the component info :v:

layershifter:
@levithomason It's another try, it's different from that you see in #2070.

### `withDocInfo`

Now we have the HOC that generates all needed info about component, later we can easily refactor it to support codesplitting :+1:

gulp/plugins/gulp-react-docgen.js
levithomason: We need to make it windows compatible, so it needs to use Noed's `path` module for path modification.  We cannot hardcode forward or backward slashes.

layershifter: I've shipped a small change, looks correct. I don't have a windows PC near and I'm not familar with `path`. If you have better solution, feel free to push updates.

levithomason:
There are still some windows build breaking forward slashes in here, but we'll address those if we hear from windows users.

levithomason:

PR #2126, opened 2017-09-26
docs(Dropdown): fix DropdownRemoteExample

### LINKED ISSUE #2125: docs(Dropdown): Recent update breaks DropdownExampleRemote
**Steps to Reproduce**
1. Write something in the dropdown

**Expected**
state.searchQuery should equal that string

**Result**
state.searchQuery instead equals the entire Dropdown's state

This can be fixed the same way like https://github.com/Semantic-Org/Semantic-UI-React/pull/2110,

but instead on https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleRemote.js#L24

patrikmolsson:
Fixes #2125


levithomason:
Thanks!

levithomason:

PR #2130, opened 2017-09-27
fix(Transition): fix component wrapping inside Transition.Group

### LINKED ISSUE #2127: Transition: Transition.Group doesn't perform transitions correctly
**Steps to Reproduce**
1. Adding at least 2 new elements with the '+' sign, then trying to remove elements with the '-' results errorneous behaviour, since the list items will not disappear properly according to the clicks on '-'.

**Expected**
The Transition should do remove item 1 by 1 when i click on the '-' sign.

**Result**
The Transition does not do this

**Testcase**
If the docs show the issue, use: https://react.semantic-ui.com/modules/transition#transition-example-group
Otherwise, fork this to get started: http://codepen.io/levithomason/pen/ZpBaJX

layershifter:
Fixes #2127.

### What?

I've fixed the bug with `mergeChildMapping` in #2047, but this fix gave birth to another bug. Before it, we [actually picked](https://github.com/Semantic-Org/Semantic-UI-React/compare/fix/transition-group?expand=1#diff-5c5e8b4da7557815875ed532103beb9eR95) an old wrapped child, but in fact we need to wrap it again. I also made changes in tests, now they assert that items are wrapped.

I also picked changes from #1542 with `setSafeState`, because I need them in #1879.

### DIFF
--- a/src/modules/Transition/Transition.js
+++ b/src/modules/Transition/Transition.js
@@ -125,6 +125,7 @@ export default class Transition extends Component {
   componentDidMount() {
     debug('componentDidMount()')
 
+    this.mounted = true
     this.updateStatus()
   }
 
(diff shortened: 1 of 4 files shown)


levithomason:

PR #2134, opened 2017-09-27
docs(readme): url update in built with

GautierT:
Hey, we switch the domain name of our name website. But we are still using React Semantic UI ! :)
lifebot.fr is a 301 to seeuletter.com but i think it's better to have the correct url.
Thanks.


layershifter:
@GautierT Thanks for update :+1:

PR #2137, opened 2017-09-28
feat(Responsive): add `fireOnMount` prop

### LINKED ISSUE #2133: Responsive: add `fireOnMount` prop and utils
This may be an issue for the `Grid` component `API` but it may be something that should probably be better solved in the context of the `Responsive` add on. Hopefully I am just misguided and there is a simple way to accomplish what I am trying to do that requires no changes to the library.

Currently for me to get responsive column text alignment within a `Grid.Column` component I need to write my own bespoke function copying the functionality of the `Responsive` add on in order to select the `textAlign` value of the column.

Shouldn't there be some kind of utility function for this sort of thing? 

The one I made I happened to call `Responsive.value`. My code looks a little like this:

```js
class ResponsiveGridExample extends Component {
  this.onResizeHandler = () => this.forceUpdate();
  render() {
    return (
      <Responsive as={Grid} onUpdate={this.onResizeHandler}>
        <Grid.Column 
          mobile={16} 
          computer={4} 
          textAlign={Responsive.value({mobile: 'center', tablet: 'left'})} >
          {/* content */}
        </Grid.Column>
        {/* other columns */}
      </Responsive>
    );
  }
}

```

I am not necessarily suggesting this as a syntax this is just my initial idea of how to solve this. 

The code for the function is a little hairy at around 40 lines without using lodash or ramda and is the kind of thing I would expect the library to handle for me instead of having to implement it myself especially as the logic is already used within the `Responsive` component. 

In this particular instance perhaps extending the Grid API to accomodate responsive text alignment makes sense but I am sure there are other instances where having some kind of standard utility would be appreciated.

I also appreciate this brings up questions about how to handle responsive props other than size as so far the convention has been to use the breakpoints as prop names and trying to add responsive other properties might lead to inconsistency.

layershifter:
Rel #2133.

### DIFF
--- a/docs/app/Examples/addons/Responsive/Usage/ResponsiveExampleFireOnMount.js
+++ b/docs/app/Examples/addons/Responsive/Usage/ResponsiveExampleFireOnMount.js
@@ -1,0 +1,29 @@
+import React, { Component } from 'react'
+import { Grid, Responsive, Segment } from 'semantic-ui-react'
+
+export default class ResponsiveExampleFireOnMount extends Component {
+  state = {}
+
+  handleOnUpdate = (e, { width }) => this.setState({ width })
+
+  render() {
+    const { width } = this.state
+    const textAlign = width >= Responsive.onlyComputer.minWidth ? 'right' : 'left'
+
+    return (
+      <Responsive
+        as={Grid}
+        columns={1}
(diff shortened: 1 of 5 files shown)


levithomason:

PR #2138, opened 2017-09-28
fix(Search): simplify propTypes on SearchResult

### LINKED ISSUE #2135: Search: the `id` prop on SearchResult should accept `string` as value
From looking at the type definition for SearchResult.d.ts, the [key:strig]: any should allow me to add any additional props.  However, when in typescript it will give warnings about them.

Warning: Failed prop type: Invalid prop results supplied to Search

It seems like it doesn't effect the functionality, but console gets flooded with these warnings  Is this a typescript issue only and is there a workaround.

Example:

When you define the results:

```
  {
    "title": "Gusikowski LLC",
    "description": "User-friendly reciprocal parallelism",
    "image": "https://s3.amazonaws.com/uifaces/faces/twitter/alecarpentier/128.jpg",
    "price": "$34.11"
  }
```

In my case, I added 

```
  {
    "tester": "a"
    "title": "Gusikowski LLC",
    "description": "User-friendly reciprocal parallelism",
    "image": "https://s3.amazonaws.com/uifaces/faces/twitter/alecarpentier/128.jpg",
    "price": "$34.11"
  }

```

"tester" will be treated as an unknown property and warnings will come.

Also another question, is it possible to update the id to be an any.  In my case, i have a string value of a GUID but because it's a number I can't set it.  

```
export interface SearchResultProps {
  [key: string]: any;

  as?: any;

  /** The item currently selected by keyboard shortcut. */
  active?: boolean;

  /** Additional classes. */
  className?: string;

  description?: string;

  /** A unique identifier. */
  id?: number;
```
Thanks,
Derek

layershifter:
Fixes #2135.

### DIFF
--- a/src/modules/Search/SearchResult.d.ts
+++ b/src/modules/Search/SearchResult.d.ts
@@ -16,7 +16,7 @@ export interface SearchResultProps {
   description?: string;
 
   /** A unique identifier. */
-  id?: number;
+  id?: number | string;
 
   /** Add an image to the item. */
   image?: string;
--- a/src/modules/Search/SearchResult.js
+++ b/src/modules/Search/SearchResult.js
@@ -42,7 +42,10 @@ export default class SearchResult extends Component {
     description: PropTypes.string,
 
     /** A unique identifier. */
-    id: PropTypes.number,
+    id: PropTypes.oneOfType([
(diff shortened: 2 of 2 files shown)


levithomason:

Issue #1533, opened 2017-03-30
Replace manually constructed keys with factory keys


levithomason:
We currently have several components that build up a `key`, `finalKey`, etc.  All factories now support generating keys from string and number shorthand.  If a user passes an object or an element, they must define their own `key`.

We need to remove all manual build up of keys and leave this to factories and users only.

levithomason:
See https://github.com/Semantic-Org/Semantic-UI-React/pull/2424

PR #2143, opened 2017-09-29
breaking(Statistic): refactor Statistic to use factories, update content props

layershifter:
# Why?

- use our factories
- removes manual `key` generation, #1533
- use `content` instead of custom props
- cleanup docs

-----

### `Statistic.Label`, `Statistic.Value`

| Before | After |
| ------------- | ------------- |
| `<Statistic.Label label='foo' />`  | `<Statistic.Label content='foo' />`  |
| `<Statistic.Label label={<div>foo</div>} />`  | `<Statistic.Label content={<div>foo</div>} />`  |
| `<Statistic.Value value='foo' />`  | `<Statistic.Value content='foo' />`  |
| `<Statistic.Value value={<div>foo</div>} />`  | `<Statistic.Value content={<div>foo</div>} />`  |

#### Before

```jsx
<Statistic
  label={<div>label</div>}
  value={<div>value</div>}
/>
```

#### After

```jsx
<Statistic
  label={{ content: <div>label</div> }}
  value={{ content: <div>value</div> }}
/>
// or
<Statistic
  label={<Statistic.Label>label</Statistic.Label>}
  value={<Statistic.Value>value</Statistic.Value>}
/>
```

#### Before

```jsx
const items = [
  { label: 'foo', value: 'bar' },
  { label: 'baz', value: 'qux' },
]
<Statistic.Group items={items} />
```

#### After

`key` is **required** now.

```jsx
const items = [
  { key: 'foo', label: 'foo', value: 'bar' },
  { key: 'baz', label: 'baz', value: 'qux' },
]
<Statistic.Group items={items} />
```


levithomason:
One question, why is this syntax not supported anymore?

```jsx
<Statistic
  label={<div>label</div>}
  value={<div>value</div>}
/>
```

Factories support passing elements.

layershifter:
`Statistic` now uses factories to create `Statistic.Label` and `Statistic.Value`, so this example will produce following markup:
```html
<div class="ui statistic">
  <div>value</div>
  <div>label</div>
</div>
```

It will don't contain valid `className`s.

levithomason:
Heads up, in the future, I'd like to make factories ensure required classNames are present 👍 Thanks for the work here!

levithomason:

Issue #1364, opened 2017-02-21
RFC: Standartize content prop


layershifter:
I'm think that we need to remove `content` prop from all our components.
This will be breaking change.

### Confusing

It's represents different types in our components:
- [contentShorthand](https://github.com/Semantic-Org/Semantic-UI-React/blob/060e81db9c18e8cb8d20b804abafa75e1e8579e8/src/views/Feed/FeedContent.js#L66)
- [itemShorthand](https://github.com/Semantic-Org/Semantic-UI-React/blob/060e81db9c18e8cb8d20b804abafa75e1e8579e8/src/views/Feed/FeedEvent.js#L65)

### Poor coverage

There are many components that doesn't have `content` prop. [Sometime](https://github.com/Semantic-Org/Semantic-UI-React/blob/8c619da2dc2e94e0e1c4f09746e6c42ed1adeab4/src/elements/Step/StepDescription.js#L42) they have another shorthand instead of `content`.

### Useless

It does same job as `children` prop:

```jsx
<Component>Children</Component>
<Component children='Children' />
<Component content='Children' />
```

### Poor tests

We doesn't have special tests for `contentShorthand`, our [existing](https://github.com/Semantic-Org/Semantic-UI-React/blob/060e81db9c18e8cb8d20b804abafa75e1e8579e8/test/specs/views/Feed/FeedContent-test.js#L40) is not so powerfull as `rendersChildren`.

### Simplify render function

Our `render` funtions will be more simple, while we have less usages of `lodash` (`_.isNil`).

```diff
- {_.isNil(children) ? content : children}
+ {children}
```

-------------

@levithomason @jcarbo I'm glad to hear your throughts :smile

levithomason:
I agree on confusing, poor coverage, and simplifying the render, however, it is not only useful but required to facilitate the shorthand factories.

The purpose of the `content` prop is that it supports shorthand creation.  This is different than `children` in two very important ways:

1. Children does not support props objects
1. Children does not create an element, but renders the value as-is

Example, a Popup takes `content` which ultimately creates a `content` div within the Popup:

```jsx
<Popup content='Some content' />

<div class="ui popup transition visible">
  <div class="content">Some content</div>
</div>
```

Whereas using `children` would, of course, render the text directly as a child.  We also cannot pass `children` to the factory in every case because there would be no way for the user to pass certain values, such as an array of strings or a props object.

[...]

PR #2144, opened 2017-09-29
 test(mixed): add test for `content` prop, remove useless tests

layershifter:
Required by #1879.
Requires #2145.
Requires #2152.

### What?

In #1364 we decided that each component that has `children` should also have `content` prop. 

This PR:
- removes all non-common tests for `content` (fewer changes in #1879)
- adds common tests `content` prop (same as for `children`)
- adds `content` prop to all components that missing it


layershifter (commented):
It's ready, requires cleanup and rebase after #2145 and #2152 will be merged.

levithomason:
Unblocked, needs conflicts resolved.

layershifter (commented):
@levithomason I solved merge conflicts, ready for review :+1:

levithomason:

PR #1240, opened 2017-01-27
feat(datetime): Add component

harel:
This is a work in progress mainly focused on exploring the possiblity of adding a datetime component which supports date ranges, time and localisation.

#### Bugs and Features
- [ ] allow typing/pasting dates into the input field
- [x] do not show completely disabled weeks in month view
- [ ] Menu needs some work, buttons don't make sense in all states
- [ ] Expose ways to customize the menu and table as well
- [ ] Add some more props for configuring the view and doc examples
  - [ ] showWeekends
  - [ ] showDaysOfWeek
- [ ] if at all possible, remove inline styles.  required styles should be configurable.
- [ ] move to 5 button CalendarMenu: `<<    <      Mo, Yr      >    >>`.  double chevrons change year, single change month, mo
- [ ] don't allow selecting invalid ranges: start > end
- [ ] use 12hr time grid with AM/PM option
- [ ] provide onChange callbacks with `(e, { value })` where `value` is the proposed new value.
- [ ] Add dateSeparator prop to DateRange to control the separator string between the two dates

#### Docs
- [ ] document function prop signatures
- [x] change `Week starts Sunday` example to `First Day Of Week` show two examples, start on Sun, another on Monday.
- [x] fix disabled days example, do not import date utils
#### Cleanup
- [x] move component to `/addons`
- [x] remove cruft propTypes that do not exist
- [ ] sort propTypes alphabetically
- [ ] update Month.js to use DateTimeGrid, https://github.com/Semantic-Org/Semantic-UI-React/pull/1240#issuecomment-289264758
- [ ] complete all code `TODO`s in `/Datetime`
- [x] reverse `onDateSelect` args to match other callbacks, with `event` first.
- [ ] remove unused methods in classes

PR #2146, opened 2017-09-30
test(common): support isConformant for portals

levithomason:
These commits were cherry picked from Datetime #1240.  We need to be able to test conformance for components that render portals.

- Add a `rendersPortal` isConformant option.
- Skip `as` prop conformance tests, not idea but better than skipping all conformance tests as we do now.  Portals render other components, which handle the `as` prop anyway.
- Look for the user's className in the `document.body` if rendering a portal.


levithomason:

PR #2149, opened 2017-10-02
perf(customPropTypes): Memoize `suggest`

### LINKED ISSUE #1184: Optimize Icon validation in development (and others?)
This should be a fairly easy fix, but it is quite important as I noticed non-negligible slowdown in one of our applications today until implementing local workarounds. Steps below, posting this as a reminder so we don't forget about it.

### Steps
1. Render the `Icon` component.
2. Note the render time, using something like the React perf tools
3. For extra fun, run a bunch of repeated renders fairly quickly, as in a normal app environment

### Expected Result
It should be super fast, the icon is a tiny component!

### Actual Result
It can cripple applications in development when `propTypes` are not ignored due to expensive validation on supported icon names.

### Version
v0.64.2

### Testcase
Not yet provided

ayastreb:
First time contributing here 🤓

cache finding suggestions for prop words.
Fixes #1184

### DIFF
--- a/src/lib/customPropTypes.js
+++ b/src/lib/customPropTypes.js
@@ -12,6 +12,30 @@ export const as = (...args) => PropTypes.oneOfType([
   PropTypes.func,
 ])(...args)
 
+/* eslint-disable max-nested-callbacks */
+const findBestSuggestions = _.memoize((propValueWords, suggestions) => _.flow(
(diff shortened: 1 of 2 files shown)


levithomason:
Awesome will give it a try!

levithomason:

PR #2155, opened 2017-10-02
feat(TransitionablePortal): add component

layershifter:
### Why?

- `Modal` requires the page `Dimmer`
- Page `Dimmer` requires transitions, but it uses `Portal`
- `Portal` unmounts components when closes

### Solution

We need to simply wrap `Portal` into the new components that respects `Transition`'s callbacks.

### But, we can use `Portal`?

`Portal` is already complicated, this will be overhang for it as callback handling is quite complex.


layershifter:
@levithomason It's ready for initial review.

### Naming

I'm not sure that I chose the best name for the component, too many characters as I think.

### Behaviour

I have trouble with `onClose` handling, the try to click multiple on `Button` in docs and you will undertand troubles :)

---

I also need to add more tests after initial review.

levithomason:
OK, thanks.  Skimming through PRs/Issues right now trying to get a release out.  I will come back to this one as it seems like it needs more attention than I have at the moment.

levithomason:
Reviewing and testing...

src/addons/TransitionablePortal/TransitionablePortal.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // ----------------------------------------
+  // Lifecycle
+  // ----------------------------------------
+
+  componentWillReceiveProps({ open }) {
+    debug('componentWillReceiveProps()', { open })
+
+    // Heads up! We apply `open` prop only when it's defined, otherwise it will break autocontrolled Portal
```
levithomason: This component does not extend ACC, is this still valid?

src/addons/TransitionablePortal/TransitionablePortal.js
```diff
@@ -0,0 +1,173 @@
...
+  // ----------------------------------------
+
+  handlePortalClose = (e, data) => {
+    debug('handlePortalClose()')
+    const { open } = this.props
+    const { portalOpen, transitionVisible } = this.state
+
+    // Heads up! We simply call `onClose` when component is controlled with `open` prop.
+    // But, when it's autocontrolled we should change the state to opposite to keep the transition
+    // queue
```
levithomason: This component does not extend ACC, is this still valid?

levithomason:
My first thought was just to add an `unmountOnClose` prop to the Portal, however, leaving lots of Portal components mounted to the DOM doesn't sound like a good idea.  Also, we do want to unmount them, just not until the Transition is done.

Your solution seems like a sane approach to this problem for now.

### Name

It's great.  I'd rather have it verbose and clear than short.  Perhaps since we have `Transition` and `Portal`, we just call it `PortalTransition`?

### Behavior

Ah, this is tricky.  Clicking the trigger immediately calls `onClose`/`onOpen`, because those callbacks are triggered from the Portal props.  However, the `PortalTransition` is not truly closed until the transition has completed.  Therefore, the authoritative callback for "opened" is `onOpen` from the Portal and the authoritative callback for "closed" is `onHide` from the Transition.

Changing the example from `onClose` to `onHide` resolves the issue.  The button state also waits for the Transition before changing.  This is correct and prevents entering an invalid mix of states.

#### Suggest Fix

We probably shouldn't expose the `onOpen` and `onClose` callbacks at all.  Just use `onStart` and `onHide`.  These are called at the actual time the state changes.

src/addons/TransitionablePortal/TransitionablePortal.js
levithomason: Oh, I see.  This is referring to the underlying Portal which extends ACC.

src/addons/TransitionablePortal/TransitionablePortal.js
levithomason: Resolved, see my next...

layershifter:
@levithomason I've made changes, thanks for your suggestions. Ready for review

levithomason:

PR #2158, opened 2017-10-03
test(typings): restore assertion for typings and propTypes order

layershifter:
Rel #2146.

@levithomason You updated typings tests and now they have more understandable errors. However, we lost the order assertion there, this PR restores it.


levithomason:
[Key order isn't guaranteed in objects](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order), but we can address that if it becomes an issue.  This is pretty useful as is 👍

levithomason:

PR #1956, opened 2017-08-11
fix(Dropdown): more logic for clearing search

lott-ai:
Fixes #1952.

* fix(Dropdown): clear search if adding item or if selection is only option #1952
* test(Dropdown): add test for clear search if selection is only option

PR #2162, opened 2017-10-04
fix(Dropdown): update text when item selected with keyboard

### LINKED ISSUE #2161: Dropdown: Does not display value on enter when using search selection
✖ CSS ISSUES → Post on https://github.com/Semantic-Org/Semantic-UI

✖ USAGE QUESTIONS → Use these dedicated resources:
      Docs - http://react.semantic-ui.com
      Chat - https://gitter.im/Semantic-Org/Semantic-UI-React
      SO - http://stackoverflow.com/search?q=semantic-ui-react

### Steps
```
<Form.Dropdown
      label={label}
      search
      selection
      selectOnBlur={false}
      selectOnNavigation={false}
      options={options}
      value={input.value}
      onChange={(param,data) => input.onChange(data.value)}
    />
```

Search a couple of letters in dropdown search selection
Arrow down to selection
Hit enter

### Expected Result
Call onChange and displays new value in input box

### Actual Result
Calls onChange but displays the searched letters instead of new value
OnBlur sets display of input box to value

### Version
0.74.2

### Testcase

ayastreb:
Fixes #2161 
the text on dropdown wasn't updated when item is selected with keyboard after search because `searchQuery` wasn't cleared if there are multiple search results (it is cleared if there's only one search result). And if `searchQuery` is not cleared, we do not update text in [renderText](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L1136)

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -603,7 +603,7 @@ export default class Dropdown extends Component {
 
   selectItemOnEnter = (e) => {
     debug('selectItemOnEnter()', keyboardKey.getName(e))
-    const { search } = this.props
+    const { multiple, search } = this.props
 
     if (keyboardKey.getCode(e) !== keyboardKey.Enter) return
     e.preventDefault()
@@ -617,7 +617,7 @@ export default class Dropdown extends Component {
     this.makeSelectedItemActive(e)
     this.closeOnChange(e)
 
-    if (isAdditionItem || optionSize === 1) this.clearSearchQuery()
+    if (!multiple || isAdditionItem || optionSize === 1) this.clearSearchQuery()
     if (search && this.searchRef) this.searchRef.focus()
   }
 
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -603,7 +603,7 @@ export default class Dropdown extends Component {
 
   selectItemOnEnter = (e) => {
     debug('selectItemOnEnter()', keyboardKey.getName(e))
-    const { multiple, search } = this.props
+    const { search } = this.props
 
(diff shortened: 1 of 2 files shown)


layershifter:
@ayastreb Thanks for the PR, this needs more investigation as it's opposite change to #2112

layershifter:
Our behaviour should match to SUI:
- https://semantic-ui.com/modules/dropdown.html#search-selection
- https://semantic-ui.com/modules/dropdown.html#multiple-search-selection

layershifter:
In fact, problem comes from condition with `optionSize`, which was added in #1956. As I think, we should clear `searchQuery` **always**, all other cases can be manually handled by user with [`searchQuery` prop](https://react.semantic-ui.com/modules/dropdown#dropdown-example-search-query).

As I see the following change will not break any test:
```diff
-if (isAdditionItem || optionSize === 1) this.clearSearchQuery()
+this.clearSearchQuery()
```

@levithomason can you clarify?

levithomason:
Always clearing seems like the correct solution to me.  @ayastreb are there any issues with this feature if we always clear the search query?

ayastreb:
@levithomason seems to be no issues, all tests are green and it works as expected.

levithomason:
Thanks guys!

levithomason:

scottrblock:
Hi @ayastreb and @levithomason, I'm curious as to why this is only the expected behavior if the item is selected with a keyboard. It currently seems that if there are multiple options and one is clicked with the mouse, the searchQuery is not cleared.

This can be handled by locally managing the `searchQuery` state and clearing it `onClose`, but I would expect the same default behavior for selecting an item from the keyboard or mouse.

levithomason:
Agreed.  It should be consistent.  This was an oversight.  A new issue and PR would be much appreciated.

PR #2165, opened 2017-10-04
fix(Form): don't prevent form submission when action=''

eug48:
When the action attribute is an empty string it means to submit to the current URL which is quite useful.

### DIFF
--- a/src/collections/Form/Form.js
+++ b/src/collections/Form/Form.js
@@ -102,7 +102,7 @@ class Form extends Component {
 
     // Heads up! Third party libs can pass own data as first argument, we need to check that it has preventDefault()
     // method.
-    if (!action) _.invoke(e, 'preventDefault')
+    if (!action && action !== '') _.invoke(e, 'preventDefault')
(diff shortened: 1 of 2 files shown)


src/collections/Form/Form.js
```diff
@@ -102,7 +102,7 @@ class Form extends Component {
 
     // Heads up! Third party libs can pass own data as first argument, we need to check that it has preventDefault()
     // method.
-    if (!action) _.invoke(e, 'preventDefault')
+    if (action === undefined) _.invoke(e, 'preventDefault')
```
layershifter: I think that `false` and `null` are also blocking values. Let's update condition and test assertions

layershifter (changes requested):

src/collections/Form/Form.js
eug48: Ok sure, I was thinking `false` and `null` would be incorrectly typed here.

layershifter (approved):

levithomason:
Excellent, thanks!

levithomason:

PR #2167, opened 2017-10-04
feat(typings): add icon names for Icon

ghost:


src/elements/Icon/Icon.d.ts
```diff
@@ -45,7 +45,7 @@ export interface IconProps {
   loading?: boolean;
 
   /** Name of the icon. */
-  name?: string;
+  name?: SemanticICONNAMES;
```
layershifter: Can we rename it to `SemanticICONS`? It will have more parity with `SemanticCOLORS`, `SemanticWIDTHS`, etc.

layershifter (changes requested):
@vincealdrin Thanks for PR, I left only one comment about naming.

In fact, this will not cover all possible values. Typings allow `mail outline` while `outline mail` is also valid. But, I'm sure that we don't need to cover these cases.

src/elements/Icon/Icon.d.ts
ghost: yeah sure

layershifter (approved):

levithomason:

Issue #1897, opened 2017-07-25
Dropdown: item clipped if positioned right of a Menu


hubertguillemain:
### Steps
Create a top menu bar with the last item being a dropdown and with position ="right".
Some of the downdown's items have a long content
### Expected Result
The dropdown is displayed completely and not clipped when expanded
### Actual Result
The dropdown items are clipped and not fully visible
### Version
0.71.1

### Testcase
https://codepen.io/youbi/pen/RZNWjz

levithomason:
The Dropdown currently does not support [menu direction](http://react.semantic-ui.com/modules/dropdown#dropdown-example-menu-direction).  PRs welcome.

### The Fix
We need to add a `direction` prop to the `Dropdown.Menu` which simply adds the `left` className when `direction='left'`.

We then need to also add a `direction` prop to the `Dropdown`.  The `Dropdown` would pass the direction value to the `Dropdown.Menu` when rendering it.  The CSS will then open it on the correct side.

### A Workaround
Adding an inline style to the `Dropdown.Menu` will get the results you need for now:
https://codepen.io/levithomason/pen/XaXjma
```jsx
<Dropdown.Menu style={{ left: 'auto', right: 0 }}>
```

Note that in future versions we are considering deprecating the Dropdown sub component API.  We would then have only props for defining the menu and items.  This will only happen if we can support all features via props, but I wanted to make you aware.

mprambadi:
is the ticket available can i contribute to this issue ?

levithomason:
@mprambadi I've not seen a PR for this yet.  I think you're in the clear.

[Dropdown PRs](https://github.com/Semantic-Org/Semantic-UI-React/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20dropdown)

PR #2174, opened 2017-10-06
feat(Dropdown): add `direction` prop

luisbd:
This PR adds left directioning support to DropdownMenu.

Closes issue #1897

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -115,6 +115,9 @@ export default class Dropdown extends Component {
       ])),
     ]),
 
+    /** A dropdown can be positioned to the left */
+    direction: PropTypes.oneOf(['left']),
+
     /** A disabled dropdown menu or item does not allow user interaction. */
     disabled: PropTypes.bool,
 
@@ -1249,6 +1252,7 @@ export default class Dropdown extends Component {
       className,
       compact,
       disabled,
+      direction,
       error,
       fluid,
       floating,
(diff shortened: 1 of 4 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -115,6 +115,9 @@ export default class Dropdown extends Component {
       ])),
     ]),
 
+    /** A dropdown can be positioned to the left */
+    directionLeft: PropTypes.bool,
```
layershifter: Please rename the prop to `direction` that will accept `left` as value.

src/modules/Dropdown/Dropdown.js
```diff
@@ -1298,6 +1302,7 @@ export default class Dropdown extends Component {
       useKeyOnly(simple, 'simple'),
       useKeyOnly(scrolling, 'scrolling'),
       useKeyOnly(upward, 'upward'),
+      useKeyOnly(directionLeft, 'left'),
```
layershifter: Please sort it in alphabetical order, i.e.:

```js
useKeyOrValueAndKey(direction, 'direction'),
useKeyOrValueAndKey(pointing, 'pointing'),
```

layershifter (changes requested):
@luisbd Thanks for PR :+1: I've left some comments about naming

levithomason:
This branch has a conflict that needs resolved.  Also, let's update docs with the new prop as well, there's `DropdownExampleMenuDirection` and `DropdownExampleMenuDirectionLeft`.  Not sure if we can do both, but let's see.

krjw:
What is the status here?

levithomason:
See https://github.com/Semantic-Org/Semantic-UI-React/issues/720, please refrain from status update requests.  It makes it difficult to keep up with notifications and responses to bugs/feature requests.

The best way to check on the status is to check the docs, issues, and PRs.  We have no work or status available outside of this.

PR #2180, opened 2017-10-09
fix(typings): fix definition for Accordion.Accordion

swASO:
Hey,
is my first PR, so if i forgot something... sorry :)! 

this commit fix the failure
> TS2604:JSX element type 'Accordion.Accordion' does not have any construct or call signatures.


layershifter (approved):
@swASO Thanks :+1:

levithomason:

PR #2181, opened 2017-10-09
breaking(Tab): remove render(), standartize API

### LINKED ISSUE #2168: Tab: renderActiveOnly={false} does not render
### Steps
I am trying render two components in two Tab Panes with renderActiveOnly={false}

### Expected Result
Show two tabs which contain my two components.

### Actual Result
Shows two tabs with proper labels and empty content.

### Version
0.74.2

### Testcase
https://codepen.io/deilv/pen/qPVKXb

### Hint
If I change renderActiveOnly to {true} or in fact anything other than {false}, the panes render, though I guess the renderActiveOnly functionality is disabled.

### LINKED ISSUE #2184: Accordion open/close state is persisted when it's not suppose to be.
Accordion doesn't reinitialize the state of whether it's open or not.  Whenever there is an update to the props that are passed into the accordion, whether it's open or not is not reset. As a result, it's persisted.

I used Tabs as an example of a way to switch different contents in the accordion.   Imagine you have a component that uses accordion to display content.  If the component updates it's data and want to display a new accordion, this problem will occur

### Steps
https://codepen.io/darewreckk/pen/VMxJoP?editors=1111
1) In Tab 1, close all the accordion
2) Switch to Tab2, notice that the accordion is close.  Open the accordion.  
3) Switch to Tab1, notice that the accordion is open.  Before it was closed.  

### Expected Result
I would expect that the accordion state (whether its open or not) is not persisted between the tabs.

### Actual Result
Whether the accordion is open or not is persisted to the other tabs.

### Version
2.2.4

### Testcase
https://codepen.io/darewreckk/pen/VMxJoP?editors=1111

layershifter:
# BREAKING

Fixes #2168.
Fixes #2184.

# What?

This PR:
- fully replaces the `render()` with `pane` prop, it wasn't the best design from the beginning
- sets `renderActiveOnly` to `false` by default, this behaviour borned made issues

## `renderActiveOnly`

This prop is set to `false` by default.

| **Before** | **After** |
| -------------- | --------------- |
| `<Tab renderActiveOnly />` | `<Tab />` |
| `<Tab />` | `<Tab renderActiveOnly={false} />` |

#### Before

```jsx
const panes = [
  { menuItem: 'Tab 1', render: () => <Tab.Pane>Tab 1 Content</Tab.Pane> },
]

<Tab panes={panes} />
```

#### After

```jsx
const panes = [
  { menuItem: 'Tab 1', pane: 'Tab 1 Content' },]

<Tab panes={panes} />
```

#### Before

```jsx
const panes = [
  { menuItem: 'Tab 1', render: () => <Tab.Pane attached={false}>Tab 1 Content</Tab.Pane> },
]

<Tab panes={panes} />
```

#### After

```jsx
const panes = [
  { menuItem: 'Tab 1', pane: { key: 'pane1', attached: false, content: 'Tab 1 Content' } },
]

<Tab menu={{ secondary: true }} panes={panes} />
```

#### Before

```jsx
const panes = [
  { menuItem: 'Tab 1', render: () => <Tab.Pane><div>Tab 1 Content</div></Tab.Pane> },
]

<Tab panes={panes} />
```

#### After

```jsx
const panes = [
  { menuItem: 'Tab 1', pane: { key: 'pane1', content: <div>Tab 1 Content</div> },
]

<Tab panes={panes} />
```


layershifter (commented):
Finally finished :checkered_flag:

levithomason:
The issue with removing `render` is that the content of all tab panes will be rendered on creation.  React.createElement is going to be called for all tabs since they exist statically in the object.

Using a `render` method ensures the tab pane content is not created until called.  I see no other way of lazily creating tabs.

levithomason:
@layershifter any feedback on my comment ^ ?

FMCorz:
I have just discovered this issue by _accident_, but having used `Tab` I am not sure whether I really liked the `panes` property. And if everything will be initialised within that property, isn't there a way to use children instead?

Example:
```jsx
<Tabs>
  <Tab>
    <Tab.Item>Menu 1</Tab.Item>
    <Tab.Content>Hello, there!</Tab.Content>
  </Tab>
  <Tab content={_ => getContent()} menuItem="Menu 2" />
</Tabs>
```

I'll admit that this is probably more verbose than it needs to be, so I wouldn't argue that this is the way to go, it's just an idea. I'll just end with the fact that I find the `panes` prop to be different from common React components.

Edit: I would also think that the `render` method should still be an option in `panes`.

levithomason:
It is difficult and unreliable to parse children in React.  Especially if those children are wrapped in HOCs.  Child syntax also invites users to add other components that shouldn't be there, which will break the component.

The advantage to using a prop with config is that we can construct the elements ourselves, guarantee parsing, and handle HOCs.  Also, it opens the door for more powerful shorthand props.

layershifter:
@levithomason I understood what you mean, okay we will leave `render`. But this mean that `panes` API will be removed, I will update PR soon.

levithomason:
Curious to see the proposal for removing the panes API.  I've had much good use from the `render()` API in Tab.  As I see, we can accomplish lazy loading, loading all tabs up front, and conditionally rendering the active tab only.  All of this is only possible if we defer the creation of tab panes, at least, that is the only option I can see.

layershifter:
Yep, I will close this PR because it's confusing. Will rework it soon.

PR #2182, opened 2017-10-10
fix(Popup): execute `onClose` when Popup closes on scroll

mkarajohn:
I believe when a popup closes on scroll the `onClose` handler should be called, since the popup *closed*.

This PR makes it so.


mkarajohn:
I am not sure what is breaking the tests. It passed locally and the commit name follows the guidelines. Help?

src/modules/Popup/Popup.js
```diff
@@ -278,11 +278,13 @@ export default class Popup extends Component {
...
   }
 
-  hideOnScroll = () => {
+  hideOnScroll = (e) => {
     this.setState({ closed: true })
 
     eventStack.unsub('scroll', this.hideOnScroll, { target: window })
     setTimeout(() => this.setState({ closed: false }), 50)
+
+    this.handleClose(e)
```
layershifter, line 287: @mkarajohn Did you investigate what `closed` value in state actually does? Why it becomes `false` after 50ms? This looks confusing to me.

layershifter:
I've retrigged the build, now it passes

src/modules/Popup/Popup.js
mkarajohn: Well, I I saw it reopens the modal, even though I do not understand the reason as to why it does so. 

From my understanding of the docs, and from the visual feedback, when the modal closes on scroll, it closes for good. 

Could you please explain to me?

levithomason: Seems it is only used in one place, render.  If true, the `trigger` is rendered directly, else, the Portal is rendered with the `trigger` set:

```jsx
if (closed) return trigger

// ..snip

return (
  <Portal...
```

I can't see why a timeout is added here?

Also, I've been racking my brain as to why we have an error logged regarding setState being called on an unmounted component, and I believe this is the issue :)  There is no cancelling of this timeout when the Popup unmounts, therefore, setState can still be called 50ms after unmount.

This looks like it might be a hack to "reset" the portal.  By rendering the `trigger` only, there is no Portal so it appears to have closed.  Then, by nearly immediately (50ms?) rending the Portal again this ensures the Portal behavior is added to the trigger again.  Future click/hover/etc will re-open the Portal.

I would have to investigate a solution to have a solid answer, but my suggestion would be to explore moving Popup to an AutoControlledComponent.  Then, in `hideOnScroll`, we likely need to `this.trySetState({ open: false })` _without_ a timeout.  This way, we are attempting to immediately close the Popup on scroll.

mkarajohn: @levithomason 
>Also, I've been racking my brain as to why we have an error logged regarding setState being called on an unmounted component, and I believe this is the issue :) There is no cancelling of this timeout when the Popup unmounts, therefore, setState can still be called 50ms after unmount.

This was supposed to be closed [here](https://github.com/Semantic-Org/Semantic-UI-React/pull/2063/files). It seems like it's been undone or something.

So, what about the initial concern that this PR addresses? The fact that the `onClose` is not called when the Popup closes because of scrolling?

levithomason: >So, what about the initial concern that this PR addresses? The fact that the onClose is not called when the Popup closes because of scrolling?

Yes, any state change due to user interaction should fire callbacks.  This would include scroll.  We only skip callbacks when state is changed through props, such as toggling the `open` prop.

layershifter:
@mkarajohn Let's add a test that asserts that `onClose` will be called on scroll and we will ready for review

layershifter (approved):
@mkarajohn Thanks for PR. I've added a missing test and simplified existing. I will investigate the case with `closed` during the refactor in future

levithomason:

PR #2196, opened 2017-10-14
style(mixed): fix spelling and grammar issues in comments and tests

direvus:
This PR corrects some spelling and grammar errors in docs, code comments and example content.

### DIFF
--- a/src/addons/Ref/Ref.js
+++ b/src/addons/Ref/Ref.js
@@ -29,7 +29,7 @@ export default class Ref extends Component {
   componentDidMount() {
     const { innerRef } = this.props
 
-    // Heads up! Don't move this condition, it's a short circle that avoids run of `findDOMNode`
+    // Heads up! Don't move this condition, it's a short circuit that avoids run of `findDOMNode`
     // if `innerRef` isn't passed
     // eslint-disable-next-line react/no-find-dom-node
     if (innerRef) innerRef(findDOMNode(this))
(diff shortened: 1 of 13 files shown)


layershifter (approved):
@direvus Thanks for your eyes :+1:

src/addons/Ref/Ref.js
```diff
@@ -29,7 +29,7 @@ export default class Ref extends Component {
   componentDidMount() {
     const { innerRef } = this.props
 
-    // Heads up! Don't move this condition, it's a short circle that avoids run of `findDOMNode`
+    // Heads up! Don't move this condition, it's a short circuit that avoids run of `findDOMNode`
```
levithomason, line 32: This one made me chuckle :)

levithomason:
Very cool, thanks much.  There are quite a few of these in the code base and I'd love to clean them all up.... priorities!

direvus:
@levithomason You're very welcome!  I spotted a couple of typos in the documentation and then went on a bit of a rampage.  I've checked every instance of it's/its in the codebase, so I'm pretty confident that those are all correct (for now anyway).

There were a fair number of code comments I came across that seemed non-idiomatic or strangely worded, but I was reluctant to fix those without a solid understanding of what each comment intends to say.

If there are any other particular spelling/grammar errors you think might be widespread in the repo, feel free to let me know and I'll do a sweep.

levithomason:
Awesome, I think the props doc blocks could use a sweep.   As well as the `index.js` files in `docs/app/Examples`.  These contain `<ComponentExample />`s which have `title` and `description` props.  That is what powers the text above every example in the docs.

Many of our contributors are non-native English speakers and awesome developers so I've consciously chosen not to slow them down for grammar's sake.  However, a clean sweep over any of this would be super appreciated!

levithomason:

direvus:
@levithomason No worries, I'll have a look soon.

With the props doc blocks, does it make sense to change both the *.js and the *.d.ts files, or is one of these generated from the other?

levithomason:
Unfortunately, we manage both manually.  Please update both to match.

PR #2199, opened 2017-10-16
fix(TextArea): fix autoHeight calculate wrong value on long text

### LINKED ISSUE #2173: TextArea's autoHeight sometimes makes it taller than an actual text
### Steps
1. visit https://react.semantic-ui.com/addons/text-area#text-area-example-auto-height
2. start typing or copy-pasting some very random long text.
3. The `textarea` height will eventually grow higher than the content itself
### Expected Result
the `textarea` height match the content height
### Actual Result
the `textarea` height is more than the content height
### Version
0.74.2
### Browsers tested
Chrome  61.0.3163.100
FireFox 55.0.3
### Explanation
I think this happen because the way we calculate height. By setting `height` to auto the overflow content will sit in the textarea with a scroll bar. So `scrollHeight` will *sometimes* has a value larger than the actual content height *without* a scroll bar. Because there could be a scenario where the appearance of the scroll bar will push some words down to a new line.

I think this will happen in an environment where the scroll bar actually take spaces in the viewport. 

### Testcase
https://codepen.io/WiNloSt/pen/qPxEaY

WiNloSt:
Calculate content height without taking a vertical scroll bar into account.

fixes #2173

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -115,7 +115,9 @@ class TextArea extends Component {
 
     // Measure the scrollHeight and update the height to match.
     this.ref.style.height = 'auto'
+    this.ref.style.overflowY = 'hidden'
     this.ref.style.height = `${Math.max(parseFloat(minHeight), Math.ceil(this.ref.scrollHeight + borderHeight))}px`
+    this.ref.style.overflowY = ''
   }
 
   render() {


WiNloSt:
Tests failed at
```
adds/removes body classes "dimmable dimmed" on mount/unmount
```

Not sure if it has anything to do with my code, but tests passes on my machine.

layershifter:
Retrigger build, now CI passing

levithomason:
Thanks folks!

levithomason:

PR #2205, opened 2017-10-17
feat(typings): export DropdownOnSearchChangeData

pierre-H:
Useful when we need to define the function : 
```ts
private onSearchChange = (e: React.SyntheticEvent<HTMLElement>, data: DropdownOnSearchChangeData) => {
    // my function
};
```


layershifter (approved):
@pierre-H thanks :+1: I've retrigged built and it passed as expected

levithomason:

PR #2212, opened 2017-10-18
fix(FormField): pass content prop through

### LINKED ISSUE #2211: Form.Button's content attribute
### Steps
Create a `Form.Button` with a text given in a `content` attribute:

    <Form.Button content="text"/>

### Expected Result
The button should have a text on.

### Actual Result
It doesn't.

### Version
0.75.0

### Testcase
https://codesandbox.io/s/8yvjnow9y2
https://react.semantic-ui.com/collections/form#form-example-capture-values

levithomason:
Fixes #2211 

`Form.Button` was updated to handle the `content` prop, however, only unhandled props are passed to underlying components.  This resulted in the `content` prop no longer being passed through to the underlying `component`.

This PR explicitly passes it through as it is explicitly handled.  Tests have also been added.  Can be merged once tests pass.

### DIFF
--- a/src/collections/Form/FormField.js
+++ b/src/collections/Form/FormField.js
@@ -74,7 +74,7 @@ function FormField(props) {
   // ----------------------------------------
   // Checkbox/Radio Control
   // ----------------------------------------
-  const controlProps = { ...rest, children, disabled, required, type }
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2215, opened 2017-10-19
docs(Props): make large code blocks more legible

redbmk:
Components with large blocks of Default code
cause the table to extend beyond the page, making
it difficult to read. Make the code wrap properly
and allow the container to scroll.

As a specific example, this can be seen in the props
for Search.Result.

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
+++ b/docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
@@ -7,6 +7,8 @@ import ComponentPropsComponents from './ComponentPropsComponents'
 import ComponentPropsDescription from './ComponentPropsDescription'
 import ComponentPropsHeader from './ComponentPropsHeader'
 
+const propsTableContainerStyle = { overflowX: 'auto' }
+
 export default class ComponentProps extends Component {
   static propTypes = {
     componentGroup: PropTypes.arrayOf(PropTypes.object),
@@ -50,7 +52,7 @@ export default class ComponentProps extends Component {
         />
 
         {activeName && (
-          <div>
+          <div style={propsTableContainerStyle}>
             <ComponentPropsDescription description={description} />
             <ComponentTable name={activeName} props={props} />
           </div>
--- a/docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
+++ b/docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
@@ -7,7 +7,7 @@ import ComponentPropsComponents from './ComponentPropsComponents'
 import ComponentPropsDescription from './ComponentPropsDescription'
 import ComponentPropsHeader from './ComponentPropsHeader'
 
-const propsTableContainerStyle = { overflowX: 'auto' }
+const propsContainerStyle = { overflowX: 'auto' }
 
 export default class ComponentProps extends Component {
   static propTypes = {
@@ -52,7 +52,7 @@ export default class ComponentProps extends Component {
         />
 
         {activeName && (
-          <div style={propsTableContainerStyle}>
+          <div style={propsContainerStyle}>
             <ComponentPropsDescription description={description} />
             <ComponentTable name={activeName} props={props} />
           </div>
--- a/docs/app/index.ejs
+++ b/docs/app/index.ejs
@@ -117,13 +117,11 @@
     }
 
     code:not(.hljs) {
-      padding: 0;
-      padding-top: 0.1em;
-      padding-bottom: 0.2em;
-      margin: 0;
       font-size: 87.5%;
       background-color: rgba(0, 0, 0, 0.04);
       border-radius: 3px;
+      white-space: pre;
+      display: inline-block;
     }
 
     code:not(.hljs)::before {


docs/app/index.ejs
```diff
@@ -124,6 +124,8 @@
       font-size: 87.5%;
       background-color: rgba(0, 0, 0, 0.04);
       border-radius: 3px;
+      white-space: pre;
+      display: inline-block;
```
levithomason, line 124: With proper whitespace, now the `padding-top` and `padding-bottom` need to be adjusted to match.  The block element now has too much bottom padding.

Let's go with `padding: 0.1em 0` instead of the 3 rules currently in place (see line 120-123).

levithomason:
Nice fix, thanks!  I've left one comment.  Also, please run `yarn test` and `yarn lint` locally to see why tests are failing.

redbmk:
It looks like all tests are passing. I'm getting a few linting warnings, but not errors, and not in anything I touched.

```
SUMMARY:
✔ 9319 tests completed

=============================== Coverage summary ===============================
Statements   : 97.08% ( 3362/3463 )
Branches     : 93.93% ( 1610/1714 )
Functions    : 97.27% ( 642/660 )
Lines        : 98.23% ( 3048/3103 )
================================================================================
Done in 159.52s.
yarn run v1.2.1
$ cross-env NODE_ENV=production eslint .

/home/bmk/code/semantic-ui-react/docs/app/utils.js
  19:31  warning  Unexpected unnamed function  func-names

/home/bmk/code/semantic-ui-react/src/collections/Form/FormField.js
  83:9  warning  Form label must have associated control  jsx-a11y/label-has-for

/home/bmk/code/semantic-ui-react/src/elements/Image/Image.js
  94:7  warning  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text

/home/bmk/code/semantic-ui-react/src/lib/factories.js
  20:8  warning  Function 'createShorthand' has a complexity of 25  complexity

/home/bmk/code/semantic-ui-react/src/modules/Checkbox/Checkbox.js
  244:36  warning  Form label must have associated control  jsx-a11y/label-has-for

/home/bmk/code/semantic-ui-react/src/modules/Embed/Embed.js
  182:25  warning  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text

/home/bmk/code/semantic-ui-react/src/modules/Rating/RatingIcon.js
  117:9  warning  Elements with the ARIA role "radio" must have the following attributes defined: aria-checked  jsx-a11y/role-has-required-aria-props

✖ 7 problems (0 errors, 7 warnings)
```

docs/app/index.ejs
redbmk: No problem. After looking at it some more though, it looks like with `padding: 0.1em 0`, `display: inline-block` pushes the text up. Try setting both those rules in e.g. the Icon docs, then toggle the `display` setting on and off in dev tools - you'll see the text bounce up and down. If we use `padding: 0` though, the `code` padding will still look larger but without pushing the text around.

It's your call though - just let me know.

redbmk: I went ahead and added the `padding: 0.1em 0` as a separate commit. Let me know if you want me to change it or squash them.

levithomason: I see what you're saying.  I think `padding: 0;` looks great.  In fact, removing all padding rules may be the same thing in this case.  Try it with no padding rules and see what you think.

I'll let you make the call and we'll ship it.

levithomason:
Once we get final word from @redbmk that he is happy with the `code` padding, we can ship this one.

docs/app/index.ejs
redbmk: A good point. Let's go with that

levithomason:
Oh boy, appears the doc perf refactor just merged gives quite the conflict here.  I believe the component which had the inline style has moved.  Apologies, you may need to run the docs and see which component that is now.

redbmk:
Ah, yeah I wasn't expecting that. No worries though, I found where it needs to go.

I took out all the padding rules, which do have the same effect as `padding: 0`, then I figured I might as well take out the margin rules as well for the same reason. 

I merged in with the latest master and squashed it back down to a single commit.

docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
```diff
@@ -50,7 +50,7 @@ export default class ComponentProps extends Component {
         />
 
         {activeName && (
-          <div>
+          <div style={{ overflowX: 'auto' }}>
```
layershifter: Please extract this style to variable like [there](https://github.com/redbmk/Semantic-UI-React/blob/ea316aee9dc35c6a1086459323555ace9ad76701/docs/app/Components/ComponentDoc/ComponentProps/ComponentPropsHeader.js#L8). Double curly braces are performance killer as they always return the new object on each render. I'm not sute about this case, possible @levithomason can correct me

layershifter (changes requested):

docs/app/Components/ComponentDoc/ComponentProps/ComponentProps.js
redbmk: No problem. I went ahead and moved that into a `const`. Even if it doesn't make a difference, it couldn't hurt.

layershifter (approved):

levithomason:

PR #2216, opened 2017-10-19
docs(index): don't use prod libs in dev

levithomason:
This explains why prop type warnings were being swallowed in local development, we were using production libs in local development.

This PR adds the prod flag for all CDN libs, not just React.


levithomason:

PR #2219, opened 2017-10-19
feat(Portal): use Ref for handling trigger refs

layershifter:
Split from #1879. Also fixes problem with warnings about refs on stateless components and React 16.

`Portal` is the single component where we have handle refs on `trigger` prop as I see.


levithomason:
>Portal is the single component where we have handle refs on trigger prop as I see.

Though this is true, I think we need to implement `Ref` anywhere we use `this.refs` on `<ElementType />` as the user may pass a SFC in `as` also.

I will merge this one and we can do the others separately.

levithomason:

PR #2220, opened 2017-10-20
docs(Icon): fix similar icon search

### LINKED ISSUE #2193: docs(Icon): "Show similar" does not filter at all on the deployed docs site
### Steps
Go to https://react.semantic-ui.com/elements/icon, have the "Show similar" checked (it is by default), and search for "home".

#### Environment
Have tested in Firefox 56.0.1 and Chrome 61.0.3163.100.

### Expected Result
The icon "home" will show up (and maybe some related icons)

### Actual Result
No filtering happens at all, for no search strings. It does however filter when **not** having the "Show similar" checked. 

It does work on the local build, however. I.e. when running ```yarn start``` in the source directory and accessing it from localhost, the "Show similar" functionality works.

patrikmolsson:
Fixes #2193.

Addresses the issue described in https://github.com/Semantic-Org/Semantic-UI-React/issues/2193

Merely tested on fork's deployed site: https://patrikmolsson.github.io/Semantic-UI-React/.

### DIFF
--- a/docs/app/Components/IconSearch/IconSearch.js
+++ b/docs/app/Components/IconSearch/IconSearch.js
@@ -1,7 +1,8 @@
 import copyToClipboard from 'copy-to-clipboard'
 import React, { Component } from 'react'
 import _ from 'lodash/fp'
-import { SUI, leven } from 'src/lib'
+import { SUI } from 'src/lib'
+import leven from 'leven'
 import {
   Form,
   Grid,
--- a/docs/app/Components/IconSearch/IconSearch.js
+++ b/docs/app/Components/IconSearch/IconSearch.js
@@ -1,8 +1,9 @@
 import copyToClipboard from 'copy-to-clipboard'
-import React, { Component } from 'react'
 import _ from 'lodash/fp'
-import { SUI } from 'src/lib'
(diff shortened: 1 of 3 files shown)


layershifter (approved):
@patrickgaskill Awesome work :+1: Thanks for finishing it

patrikmolsson:
@layershifter Of course :) Just a question, how come the tests were failing?

layershifter:
There is a strange bug with one of `Modal` tests, I can't reproduce it locally. Sometime it fails, sometime - no :thinking:

patrikmolsson:
@layershifter Is this the new Modal test? From 0.75?

layershifter:
Nope, it fails periodically from `0.74.x` as I remember:

```
FAILED TESTS:
  Modal
    dimmer
      false
        ✖ does not add any dimmer classes to the body
              at test/utils/assertBodyClasses.js:22:0
              at Array.forEach (<anonymous>)
              at assertBodyClasses (test/utils/assertBodyClasses.js:17:0
              at Context.<anonymous> (test/specs/modules/Modal/Modal-test.js:280:0
```

levithomason:

PR #2223, opened 2017-10-20
feat(Dropdown): make deburr opt-in, and deburr input

### LINKED ISSUE #2163: Dropdown: remove diacritics on filter also converts national letters
✖ CSS ISSUES → Post on https://github.com/Semantic-Org/Semantic-UI

✖ USAGE QUESTIONS → Use these dedicated resources:
      Docs - http://react.semantic-ui.com
      Chat - https://gitter.im/Semantic-Org/Semantic-UI-React
      SO - http://stackoverflow.com/search?q=semantic-ui-react

https://github.com/Semantic-Org/Semantic-UI-React/pull/2021
introduced a filter that removes diacritics on the options in dropdowns, this has an unfortunate side effect as it also converts national characters (ex. ø = o). Scandianvian users would never expect to try searching using o instead of ø - so it breaks the feature.

I suggest that deburr is made optional - and is false by default.

### Steps
Create a dropdown that contains the letter ø in one of it's options
Search for 'ø'
(same goes for æåäö etc)

### Expected Result
List the option that contains ø

### Actual Result
Nothing is listed

### Version
0.73.1

### Testcase

patrikmolsson:
WIP pull request. Would gladly appreciate feedback on both tests and source code!

Fixes #2163.

- [x] Make deburr opt-in
- [x] When having deburr, the input string should also be deburred.
- [x] Add proper documentation
- [x] Add example of using deburr
- [x] Add example of custom search function
- [x] Add proper "should not filter diacritics when deburr is not set" test

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import _ from 'lodash'
+import { Dropdown } from 'semantic-ui-react'
+
+const caseSensitiveSearch = (options, query) => {
+  const re = new RegExp(_.escapeRegExp(query))
+  return options.filter(opt => re.test(opt.text))
+}
+
+const options = [
+  { key: 'a', value: 'a', text: 'UPPERCASE' },
+  { key: 'b', value: 'b', text: 'lowercase' },
+]
+
+const DropdownExampleCustomSearchFunction = () => (
+  <Dropdown
+    options={options}
+    search={caseSensitiveSearch}
+    selection
+    fluid
+    placeholder={'Try to search for case or CASE'}
+  />
+)
+
+export default DropdownExampleCustomSearchFunction
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
@@ -1,5 +1,5 @@
-import React from 'react'
 import _ from 'lodash'
+import React from 'react'
 import { Dropdown } from 'semantic-ui-react'
 
 const caseSensitiveSearch = (options, query) => {
@@ -14,11 +14,11 @@ const options = [
 
 const DropdownExampleCustomSearchFunction = () => (
   <Dropdown
+    fluid
     options={options}
+    placeholder={'Try to search for case or CASE'}
     search={caseSensitiveSearch}
     selection
-    fluid
-    placeholder={'Try to search for case or CASE'}
   />
 )
 
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleDeburrSearch.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleDeburrSearch.js
@@ -1,0 +1,24 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+
+const options = [
+  { key: 'a', value: 'a', text: 'Café with accent' },
+  { key: 'b', value: 'b', text: 'Cafe without accent' },
+  { key: 'c', value: 'c', text: 'Déjà vu' },
+  { key: 'd', value: 'd', text: 'Deja vu' },
+  { key: 'e', value: 'e', text: 'Scandinavian å ä æ ø ö' },
+  { key: 'f', value: 'f', text: 'Scandinavian a a ae o o' },
+]
+
+const DropdownExampleSearchSelection = () => (
+  <Dropdown
+    placeholder='Try to search for "Deja vu"'
+    fluid
+    search
+    selection
+    deburr
+    options={options}
+  />
+)
+
+export default DropdownExampleSearchSelection
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleDeburrSearch.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleDeburrSearch.js
@@ -12,12 +12,12 @@ const options = [
 
 const DropdownExampleSearchSelection = () => (
   <Dropdown
-    placeholder='Try to search for "Deja vu"'
+    deburr
     fluid
+    options={options}
+    placeholder='Try to search for "Deja vu"'
     search
     selection
-    deburr
-    options={options}
   />
 )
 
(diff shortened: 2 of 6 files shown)


patrikmolsson:
@levithomason I think I'm done now, but I would love some input on the prop naming before we merge!

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import _ from 'lodash'
+import { Dropdown } from 'semantic-ui-react'
```
layershifter: Let's sort imports in alphabetical order

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
```diff
@@ -0,0 +1,25 @@
...
+  { key: 'b', value: 'b', text: 'lowercase' },
+]
+
+const DropdownExampleCustomSearchFunction = () => (
+  <Dropdown
+    options={options}
+    search={caseSensitiveSearch}
+    selection
+    fluid
+    placeholder={'Try to search for case or CASE'}
```
layershifter: And props in alphatical order, too :smile:

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleDeburrSearch.js
```diff
@@ -0,0 +1,24 @@
...
+]
+
+const DropdownExampleSearchSelection = () => (
+  <Dropdown
+    placeholder='Try to search for "Deja vu"'
+    fluid
+    search
+    selection
+    deburr
+    options={options}
```
layershifter: ^

src/modules/Dropdown/Dropdown.js
```diff
@@ -787,7 +791,7 @@ export default class Dropdown extends Component {
   // There are times when we need to calculate the options based on a value
   // that hasn't yet been persisted to state.
   getMenuOptions = (value = this.state.value, options = this.props.options) => {
-    const { multiple, search, allowAdditions, additionPosition, additionLabel } = this.props
+    const { multiple, search, allowAdditions, additionPosition, additionLabel, deburr } = this.props
```
layershifter: Let's sort props there too

layershifter (changes requested):
@patrikmolsson Thanks, awesome work :+1: I've left styling comments, let's fix them

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
patrikmolsson: Of course! :smile: On it!

Is there a linting rule for this that I'm missing?

layershifter: Nope, we don't have or don't use it :thinking:

patrikmolsson: Alright. Btw, how do you sort it?  Alphabetically by module name, i.e. ```react```, ```lodash```, and ```semantic-ui-react```?

patrikmolsson: --> ```lodash```, ```react```, ```semantic-ui-react```

layershifter: Yes, by module name.

patrikmolsson:
@layershifter There it is! 

By the way, what are your thoughts on the prop naming? Is it clear what ```deburr``` means?

layershifter (approved):

layershifter:
@patrickgaskill `fuzzy` looks better to me, however lodash uses `deburr`. It's a complicated question

patrikmolsson:
@layershifter ```fuzzy``` might also imply that searching for ```ac``` will match ```abc```...

patrikmolsson:
I've thought about this, but cannot really find an appropriate third option.

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
levithomason: Thanks for updating.  FWIW, there is a fixable lint rule for this that we should turn on [`sort-imports`](https://eslint.org/docs/rules/sort-imports) so this is done automatically for people.

levithomason:
Beautiful PR, I can see the care and attention to detail you put in.  ❤️ 

### Prop name

Regarding the prop name, lodash is the most installed NPM package and it calls this process `deburr`ing.  It is likely many people will recognize that term.  However, there is a formal specification and process for this.

A Unicode character with diacritics is called a "[precomposed character](https://en.wikipedia.org/wiki/Precomposed_character)".  Example, `é` is composed of the letter `e` and `´` (acute accent).  When these characters are separated like this they are "decomposed" (see previous link).

In terms of Unicode equivalence, this is called "[normalization](https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization)".  Java has a class called "[Normalizer](https://docs.oracle.com/javase/7/docs/api/java/text/Normalizer.html)" which performs this operation.  The Unicode Standard includes a [Normalization Chart](http://www.unicode.org/charts/normalization/) which maps these characters to one another.  The technical process by which equivalent Unicode values are determined is called [_decomposition_](http://www.unicode.org/reports/tr15/tr15-23.html#Decomposition).

So, I think we could go with `deburr`, `normalize`, or `decompose`.  I have my opinion, but I'll let you two choose.

***

@layershifter this is good to merge once you guys are settled on the most fitting prop name.

layershifter:
I'm think that we can go with `deburr`.

patrikmolsson:
@levithomason Thanks for the input, and for the kind words! As @layershifter says, let's go with deburr :relaxed:

patrickgaskill:
@layershifter You tagged the wrong Patrick ;)

layershifter:
@levithomason Can you merge this? As `codecov` failed, I don't have enough rights :smile:

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleCustomSearchFunction.js
patrikmolsson: @levithomason It seems that lint does not sort on the module name, but rather on the imported member, or am I missing something?

levithomason:
Sorry folks, Codecov is having issues finding a base to compare some PRs against and incorrectly failing statuses.

This will go out in the next release, today or tomorrow.

levithomason:

PR #2145, opened 2017-09-29
breaking(Step): refactor Step to use factories, update content props, add missing props

layershifter:
# Why?

- use our factories
- removes manual `key` generation, #1533
- use `content` instead of custom props
- cleanup docs
- add missing props (`attached`, `widths`) and examples

-----

### `Step.Description`, `Step.Title`

| Before | After |
| ------------- | ------------- |
| `<Step.Description description='foo' />`  | `<Step.Description content='foo' />`  |
| `<Step.Description description={<div>foo</div>} />`  | `<Step.Description content={<div>foo</div>} />`  |
| `<Step.Title title='foo' />`  | `<Step.Title content='foo' />`  |
| `<Step.Title title={<div>foo</div>} />`  | `<Step.Title content={<div>foo</div>} />`  |

#### Before

```jsx
<Step
  description={<div>description</div>}
  title={<div>title</div>}
/>
```

#### After

```jsx
<Step
  description={{ content: <div>description</div> }}
  title={{ content: <div>title</div> }}
/>
// or
<Step
  description={<Step.Description>description</Step.Description>}
  title={<Step.Title>title</Step.Title>}
/>
```

#### Before

```jsx
const items = [
  { description: 'foo', title: 'bar' },
  { description: 'baz', title: 'qux' },
]
<Step.Group items={items} />
```

#### After

`key` is **required** now.

```jsx
const items = [
  { key: 'foo', description: 'foo', title: 'bar' },
  { key: 'baz', description: 'baz', title: 'qux' },
]
<Step.Group items={items} />
```

PR #2225, opened 2017-10-20
breaking(Image): Inconsistent prop usage between Image and Icon for circular shape

### LINKED ISSUE #2217: Image: Inconsistent prop usage between Image and Icon for circular shape
### Steps
Currently to create a circular image and a circular icon, we would do the following:
```html
<Image shape='circular' src='url' />

<Icon circular name='icon' />
```
but that's really inconsistent, especially where `Image` and `Icon` can be used interchangeably such as in `Header`
### Expected Result
Ideally, both `Image` and `Icon` would use bool props for adjective based attributes as it is the case with many other attributes like `disabled`, `bordered`, etc...
```html
<Image circular src='url' />

<Icon circular name='icon' />
```
### Version
0.75.1

Please let me know if there is a reason to keep it the way it is.

RetroCraft:
Fixes #2217.

# Why?
Make properties more concistent with other elements. See #2217 for more details.

---

## `Image`
Before|After
:---:|:---:
`<Image shape="rounded" />`|`<Image rounded />`
`<Image shape="circular" />`|`<Image circular />`

### DIFF
--- a/docs/app/Examples/collections/Table/Types/TableExampleCollapsing.js
+++ b/docs/app/Examples/collections/Table/Types/TableExampleCollapsing.js
@@ -14,7 +14,7 @@ const TableExampleCollapsing = () => (
       <Table.Row>
         <Table.Cell>
           <Header as='h4' image>
-            <Image src='/assets/images/avatar/small/lena.png' shape='rounded' size='mini' />
+            <Image src='/assets/images/avatar/small/lena.png' rounded size='mini' />
             <Header.Content>
                 Lena
               <Header.Subheader>Human Resources</Header.Subheader>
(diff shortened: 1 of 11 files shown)


layershifter (changes requested):
@RetroCraft Please fix lint errors, `yarn run lint:fix`

```
  126:1  error    Trailing spaces not allowed                                                                                no-trailing-spaces
  138:1  error    Trailing spaces not allowed                                                                                no-trailing-spaces
  171:1  error    Trailing spaces not allowed                                                                                no-trailing-spaces
```

layershifter:
@RetroCraft Also this PR should be marked as **breaking** in title and initial comment. We also need to provide before/after usage example in the initial comment. See #2145 for example.

RetroCraft:
I think everything is good now.

layershifter (approved):

levithomason:

PR #2232, opened 2017-10-21
docs(Usage): switch from NPM to yarn

lielfr:
The README file states that the project has switched from NPM to yarn, so I changed the Usage page in the docs to use yarn instead of NPM.


levithomason:
Thanks for the careful eyes and support!

levithomason:

PR #2244, opened 2017-10-23
fix(docs): fix propTypes warnings

layershifter:
Rel #2216.

This PR fixes all known propType's warning in docs.


levithomason:

Issue #2120, opened 2017-09-25
Bug on Dropdown.componentWillUnmount()


MeXaaR:
Simple bug on every component dependants on Dropdown (FormSelect, Select, etc...)
it happens since the last update

```
Error: Cannot read property '_bound' of undefined
TypeError: Cannot read property '_bound' of undefined
    at EventStack._unlisten (modules.js:47055)
    at EventStack.unsub (modules.js:47083)
    at Dropdown.componentWillUnmount (modules.js:48677)
    at modules.js:20397
    at measureLifeCyclePerf (modules.js:20064)
    at ReactCompositeComponentWrapper.unmountComponent (modules.js:20396)
    at Object.unmountComponent (modules.js:13040)
    at ReactCompositeComponentWrapper.unmountComponent (modules.js:20406)
    at Object.unmountComponent (modules.js:13040)
```
anybody has the same issue ?

levithomason:
I have not, but this may be related to #2117.  It is being released now as `0.74.2`.  Let's see if it fixes it.

levithomason:
@MeXaaR please try the latest release, `0.74.2`, and let us know if you're still seeing this bug.

MeXaaR:
Thanks guys, it work fine now

Great fix

ewiechowicz:
@levithomason  I'm still seeing this on `0.74.2` and `0.75.1`. Please see the trace below:

```
Error: Cannot read property '_bound' of undefined
TypeError: Cannot read property '_bound' of undefined
    at EventTarget._unlisten (http://localhost:3000/packages/modules.js?hash=0051c69b4aad0f180713cf5acfcf13ef81b7be36:91254:18)
    at EventTarget.unsub (http://localhost:3000/packages/modules.js?hash=0051c69b4aad0f180713cf5acfcf13ef81b7be36:91282:11)
    at EventStack.unsub (http://localhost:3000/packages/modules.js?hash=0051c69b4aad0f180713cf5acfcf13ef81b7be36:91125:19)
    at Dropdown.componentWillUnmount (http://localhost:3000/packages/modules.js?hash=0051c69b4aad0f180713cf5acfcf13ef81b7be36:99245:23)
    at http://localhost:3000/packages/modules.js?hash=0051c69b4aad0f180713cf5acfcf13ef81b7be36:39978:25
[...]

PR #2246, opened 2017-10-24
fix(eventPool): check existance of handler in EventTarget._unlisten

mhagmajer:
Fixes https://github.com/Semantic-Org/Semantic-UI-React/issues/2120

### DIFF
--- a/src/lib/eventStack/EventTarget.js
+++ b/src/lib/eventStack/EventTarget.js
@@ -40,8 +40,9 @@ export default class EventTarget {
   }
 
   _unlisten = (name) => {
-    if (_.some(this._pools, name)) return
     const handler = this._handlers[name]
+
+    if (_.some(this._pools, name)) return
     if (!handler) return
 
     this.target.removeEventListener(name, handler)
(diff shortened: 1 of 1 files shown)


mhagmajer:
Failing check seems like false negative

layershifter:
@levithomason I've marked this as ready for review, however I'm not sure that this is fixes something. We don't have reproduction and I don't see how this can be covered by unit tests.

levithomason:
As I've shown here https://github.com/Semantic-Org/Semantic-UI-React/issues/2120#issuecomment-340239605, both our code and the browser's native code already gracefully handle removing `undefined` handlers.  I want to get a broken test case that this fixes before merging.

levithomason:
I should note, I spent 20-30 minutes researching and also trying to reproduce this.  I could not do so in Chrome nor PhantomJS no matter what abuse I threw at our EventTarget and eventStack. I was also unabled to reproduce this in Chrome with `removeEventListener`.

The closest I could find was this https://github.com/angular/zone.js/issues/21 issue.  They had this error, but they also seemed unable to reproduce it.  It seemed to resolve itself, unfortunately.

Lastly, neither `_bound` nor `bound` are properties that occur in our codebase.  This leads me to believe the error is something native to the browser, or some other conflicting lib.

levithomason:
Any more progress on this one?

mhagmajer:
Not yet, I'll get back to you when I have something

layershifter:
Closing for housekeeping. Will be reopened if we will have a repro case.

PR #2251, opened 2017-10-25
docs(Dropdown): add dedicated example for controlled Dropdown

### LINKED ISSUE #2249: Dropdown  event.target.value  is undefined
I'm trying to do controlled component as it shown in official Rect documentation (https://reactjs.org/docs/forms.html):

`
<Dropdown value={this.state.value} onChange={this.handleChange} />
`

````
handleChange = (event) => {
    this.setState({value: event.target.value});
  }
````

But **event.target.value is undefined ** so it's impossible to do controlled component as usual.
May be you've made your own way to do this. But then where is documentation for this? I've search a lot... Couldn't find how to do controlled component with Semantic UI Dropdown.

layershifter:
Fixes #2249.


levithomason:

PR #2252, opened 2017-10-25
fix(Dropdown): isEqual over shallowEqual for props

### LINKED ISSUE #2243: Dropdown: setState in onChange leads to mismatch of 'active' and 'selected' option
### Steps

I want to setState() by onChange of the Dropdown

### Expected Result

The state should be set and the dropdown should change its value.
The selected option should be active and selected

### Actual Result
The state is set as expected
The selected option is set to active, the previously selected option is kept as selected.
=> **Selected and active do not match the same option anymore**

### Version
0.75.1

### Testcase
https://codesandbox.io/s/lrvqlpy7oz

chrisjbrown:
fixes #2243

Component on `componentWillReceiveProps` will always run `setSelectedIndex` because shallowEqual on two objects internally runs `object.is(a, b)` to compare objects which will always be false because `nextProps.options` and `this.props.options` are not the same object

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -427,7 +427,7 @@ export default class Dropdown extends Component {
       this.setSelectedIndex(nextProps.value)
     }
 
-    if (!shallowEqual(nextProps.options, this.props.options)) {
+    if (!_.isEqual(nextProps.options, this.props.options)) {
       this.setSelectedIndex(undefined, nextProps.options)
     }
   }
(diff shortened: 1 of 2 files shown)


levithomason:
Let's get a test here that fails with the workflow reported in the issue.  This way we don't hit the regression again.

`isEqual` was removed as part of the lodash optimizations as it is a very slow and heavy method.  We can use it here if we must, but I'd like to ensure there is a test that catches the failing behavior.

chrisjbrown:
Added UT for when props are updated with options that are the same and when props are updated with new options.

levithomason:
Awesome, thank you very much @chrisjbrown!

levithomason:

Issue #2150, opened 2017-10-02
Modal: Scrolling doesn't work properly after reopen


kien8995:
### Steps
1. Open scroll modal
2. Scroll modal
3. Close scroll modal
4. Open scroll modal again
5. Scroll modal 
### Expected Result
Modal must be scrollable
### Actual Result
Modal no longer scrollable
### Version
0.74.2

### Testcase
https://react.semantic-ui.com/modules/modal

layershifter:
There are no need to scroll this modal. In fact, error comes from this [condition](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.74.2/src/modules/Modal/Modal.js#L258). After modal will be closed it should reset scrolling state as I think:

```diff
  handleClose = (e) => {
    debug('close()')

    const { onClose } = this.props
    if (onClose) onClose(e, this.props)

-    this.trySetState({ open: false })
+    this.trySetState({ open: false }, { scrolling: undefined })
  }
```

PR #2254, opened 2017-10-25
fix(Modal): fix state update after close

### LINKED ISSUE #2236: Modal: on close warning in 0.75
### Steps

[0.75.1 example](https://codesandbox.io/s/0xj174kwpl )
[0.74 example](https://codesandbox.io/s/4r55j6n7l7 )

Sometimes I have to pragmatically open or close a modal without the help of `trigger` prop. This is just an example:

````
class Example extends React.Component {
  state = { open: false }
  constructor(props){
      super(props);
      this._toggle = this._toggle.bind(this);
  }
  render() {
    const { open } = this.state
    return (
      open?
      <Modal
        open={open}
        onOpen={this.open}
        onClose={this._toggle}
        size='small'>
        dgfdgfdgfdgf
      </Modal>
       : <Button onClick={this._toggle} primary>Open</Button>

    )
  }

  _toggle(){
    this.setState({open:!this.state.open})
  }
}
````
Before 0.75, I can do that without a warning, as shown in the 0.74 example. 

#### Actual Result

However, since 0.75, I'm getting this warning when `onClose` is fired:

````
Warning: Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.
````
Is it a bug or do I really need to use a `trigger` element from now on?

#### Version

0.75.1

### PR DESCRIPTION
**Pull request by layershifter**
Fixes #2236.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -175,10 +175,8 @@ class Modal extends Component {
   handleClose = (e) => {
     debug('close()')
 
-    const { onClose } = this.props
-    if (onClose) onClose(e, this.props)
-
-    this.trySetState({ open: false }, { scrolling: false })
+    _.invoke(this.props, 'onClose', e, this.props)
+    this.trySetState({ open: false })
   }
 
   handleIconOverrides = predefinedProps => ({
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/modules/Modal/Modal.js [hunk 1]
```diff
@@ -175,10 +175,8 @@ class Modal extends Component {
   handleClose = (e) => {
     debug('close()')
 
-    const { onClose } = this.props
-    if (onClose) onClose(e, this.props)
-
-    this.trySetState({ open: false }, { scrolling: false })
+    _.invoke(this.props, 'onClose', e, this.props)
+    this.trySetState({ open: false })
```
layershifter, line 179: My initial suggestion [there](https://github.com/Semantic-Org/Semantic-UI-React/issues/2150#issuecomment-333460199) was falsy. It causes state after `onClose` was done.

The same effect can be archived when `scrolling` will be reseted in `handleMount`.

levithomason:

PR #2255, opened 2017-10-26
fix(Modal): scrolling disabled with multiple modals bug

### LINKED ISSUE #1157: Multiple modals scrolling bug

### LINKED ISSUE #2248: Modal: scrolling issue when using sub-modals

I have a modal that invokes a modal, and this blocs scrolling in the first modal.

Here is how to reproduce:

### Steps

1. Open: https://codesandbox.io/s/pw5on5o7m7
2. In the running app, click on 'Show modal'.
3. Reduce window height until the button row of the modal is no longer visible.
4. You will notice you can access the button row by scrolling the screen. Scroll so you can see the buttons and click on 'Open sub modal'.
5. Close the modal with the 'x' at the top-right corner.
6. You will be back at the first modal. Try to scroll to access the buttons.

### Expected Result

You can access the button row by scrolling.

### Actual Result

Scrolling does not work.

### Version
0.75.1

### Testcase
* https://codesandbox.io/s/pw5on5o7m7
* Also: https://github.com/guillaumep/semantic-ui-react-modal-issue/blob/master/src/App.js

RetroCraft:
Fixes #1157, fixes #2248

See the discussion at #1157 for more details, bug was reopened as #2248. The gist of it is that closing a modal removes the scrolling class from body, so if multiple scrollable modals are open the top one closing disables the bottom one's scroll.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -259,7 +259,12 @@ class Modal extends Component {
         if (scrolling) {
           mountNode.classList.add('scrolling')
         } else {
-          mountNode.classList.remove('scrolling')
+          // Check if another modal exists before removing the scolling
+          const anotherModal = document.getElementsByClassName('ui page modals').length
+          // Only remove scrolling class if there is no other modals active
+          if (anotherModal === 0) {
+            mountNode.classList.remove('scrolling')
+          }
         }
       }
 


guillaumep:
Wow, that's great, I'm amazed by the responsiveness of the Semantic-UI-React community!

I would have a suggestion for an alternative approach: check if the scrolling class is already present when the modal component wants to add it, and when closing the modal, only remove the class if it was added. This could lead to a more generic fix.

RetroCraft:
I'm really just a high school kid who's got too much time on his hands 😉

I'll see if I can figure something like that out, that sounds like a lot more sense than the workaround I ripped off of that other issue.

guillaumep:
It's nice to see you using this free time this way!

levithomason:
Thanks for the contribution!  Looks like we need to update tests.  You can run tests with `yarn test:watch`:

```
FAILED TESTS:
  Modal
    dimmer
      false
        ✖ does not render a dimmer
              at test/utils/assertBodyClasses.js:22:0
              at Array.forEach (<anonymous>)
              at assertBodyClasses (test/utils/assertBodyClasses.js:17:0
              at Context.<anonymous> (test/specs/modules/Modal/Modal-test.js:275:0

    scrolling
      ✖ adds/removes the scrolling class to the body when the window grows/shrinks
        Error: Uncaught AssertionError: document.body should not have class "scrolling", but it has class="dimmable dimmed scrolling": expected true to equal false (node_modules/chai/lib/chai/assertion.js:141:0
```

While we're at it, let's add a test with a second scrolling Modal to ensure this works like we think it does.  You can see many test examples in `Modal-test.js`.  Specifically, we have tests for the scrolling behavior right [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/test/specs/modules/Modal/Modal-test.js#L463).

guillaumep:
I have proposed an alternative approach to this fix in #2394.

layershifter:
Closing in favor of #2407 as it promotes more proper solution.

PR #2256, opened 2017-10-26
fix: actionPosition should require action

redbmk:
if there is no action, then actionPosition should not add the 'action' class to Input

I created a codesandbox to show off the issue along with a workaround

[Edit r0pq6xy7zp](https://codesandbox.io/s/r0pq6xy7zp)

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -198,7 +198,7 @@ class Input extends Component {
       useKeyOnly(inverted, 'inverted'),
       useKeyOnly(loading, 'loading'),
       useKeyOnly(transparent, 'transparent'),
-      useValueAndKey(actionPosition, 'action') || useKeyOnly(action, 'action'),
+      useKeyOnly(action, useValueAndKey(actionPosition, 'action') || 'action'),
       useValueAndKey(iconPosition, 'icon') || useKeyOnly(icon || loading, 'icon'),
       useValueAndKey(labelPosition, 'labeled') || useKeyOnly(label, 'labeled'),
       'input',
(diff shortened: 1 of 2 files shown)


levithomason:
Thanks for the contribution!  I think this would fall to the responsibility of the developer.  If there is no `action` available, don't add the `actionPosition` prop:

```jsx
const action = undefined

<Input actionPosition={action && 'left'} action={action} />
```
Renders:
```html
<div class="ui input">
  <input type="text" />
</div>
```

We could introduce a prop warning for this:

```diff
 /** An action can appear along side an Input on the left or right. */
-actionPosition: PropTypes.oneOf(['left'])
+actionPosition: customPropTypes.every([
+  customPropTypes.demand(['action']),
+  PropTypes.oneOf(['left']),
+]),
```

However, I don't think we should silently remove the className for `actionPosition` when the developer has explicitly applied it.

redbmk:
Hmm, that's weird. When I ran into this issue before that was the first thing I tried and I remember seeing it not responding when the conditional changed. I tried added tests for that but it didn't seem to be an issue so I took those out, and now I can't seem to recreate that bug. Maybe that was just before my daily dose of caffeine :coffee:.

At first I liked the warning idea, but on second thought I feel like it might be annoying. If you're putting `left` there you are probably either:

A) not expecting it to apply the class unless you have an `action`, in which case the warning would be a clue to change it, but you would also likely notice the left border missing when there's no action; or
B) you're intending to have the class applied even when there's no action, in which case you would get console warnings even when your code is working as intended, with no way to turn it off.

It's your call, but I'm leaning towards just closing the ticket, as I feel like if I were purposely trying to add `actionPosition="left"` when there was no `action`, I would just be irritated by the warning.

levithomason:
Sounds good.  In this case, I think explicit and exact behavior wins out over automatic help.  Thanks for the careful consideration and dialog.

Issue #2259, opened 2017-10-26
Modal: bug with SSR rendering


adam-26:
### Steps
See #1531 for steps

### Expected Result
Modal should not cause SSR to result in invalid checksum on client

### Actual Result
Client checksum is invalid

### Version
0.72.0

### Testcase
From #1531
https://github.com/kolpav/semantic-ui-ssr-bug

As per the closed issue #1531, rendering a `Modal` on the server results in an invalid checksum on the client.

The [bug is caused in the render() method at this line](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Modal/Modal.js#L324). The problem is that on the server it does not render the `trigger` content. 

A decent partial solution is to simply change from this (current code @ line 324):
```js
// Short circuit when server side rendering
if (!isBrowser) return null
```

To this, rendering the trigger component on the server:

```js
// Short circuit when server side rendering
if (!isBrowser) {
  return this.props.trigger;
};

```

It's not a perfect solution, it **will work on the condition that the Modal has a state of _closed_ when rendering on the server**. I'm guessing this will be acceptable in most use-cases, as `Modal` components are generally displayed in response to user interaction and therefore don't need to be rendered on the server.

If this solution would be accepted as a PR, i'll submit one. Please let me know!

Adam.

layershifter:
I looked a little, you seem right. Feel free to open PR :+1: 

The only thing I think is that we should wrapper `trigger` with `isValidElement()` and returns `null` when it's false.

layershifter:
Fixed in #2260.

PR #2260, opened 2017-10-26
fix(Modal): fix errors with SSR

adam-26:
Resolve Issue #2259

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -158,16 +158,13 @@ class Modal extends Component {
   static Description = ModalDescription
   static Actions = ModalActions
 
-  // expose for testing
-  static isBrowser = isBrowser
-
   componentWillUnmount() {
     debug('componentWillUnmount()')
     this.handlePortalUnmount()
   }
 
   // Do not access document when server side rendering
-  getMountNode = () => (Modal.isBrowser ? this.props.mountNode || document.body : null)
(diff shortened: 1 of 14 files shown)


adam-26:
@layershifter PR submitted for #2259

src/modules/Modal/Modal.js
```diff
@@ -155,13 +158,16 @@ class Modal extends Component {
   static Description = ModalDescription
   static Actions = ModalActions
 
+  // expose for testing
+  static isBrowser = isBrowser
```
levithomason: Is there a reason to not import `isBrowser` in tests instead?

levithomason: Oh, I see.  I'd like to have a way to spoof `isBrowser` that is not dependant on the Modal.  One idea is to convert isBrowser to a function.  This way, we can set a flag to override it:

```js
const isBrowser = () => !_.isNil(isBrowser, 'override')
  ? isBrowser.override
  : hasDocument && hasWindow

export default isBrowser
```

Then we can do something similar to what is being done now, except with isBrowser rather than Modal:

```jsx
before(() => {
  isBrowser.override = false
  // isBrowser() === false
})

after(() => {
  isBrowser.override = null
  // isBrowser() === <appropriate value>
})
```

levithomason:
Thanks!

levithomason:

PR #2265, opened 2017-10-27
BREAKING(Tab): allow right/left without `tabular`

### LINKED ISSUE #2264: Tab menu option tabular is being overused for both style and position.
### Steps
Given the following menu options for Tab:
```
const menu = {
  attached: false,
  fluid: true,
  vertical: true,
  tabular: 'right',
  secondary: true,
  borderless: true,
}
```
### Expected Result
The tabs should be on the right, and the style should be secondary (non-tabular).

### Actual Result
The tabs are on the right, but the style is tabular, ignoring the secondary configuration.

### Version
0.75.1

### Testcase
[Example](https://codesandbox.io/s/womxq04zp5)
[Proposed solution](https://codesandbox.io/s/0mm43m33xl)
In my PR #2265 - I've added an "aligned" option, which is used to determine right or left alignment, regardless of which style configuration you've set.

ryanmcgarvey:
# BREAKING CHANGE

It is currently not possible to align vertical tabs to the left or right unless you use the `tabular` prop.  This PR allows left/right vertical tab alignment without requiring the `tabular` menu variation.

### Before

Right aligned tabs required the `tabular` prop and therefore the tabular menu variation.

```jsx
const panes = [ ...{} ]
const menu = {
  fluid: true,
  vertical: true,
  tabular: 'right',
 }

<Tab menu={menu} panes={panes} />
```
image

### After

Right aligned tabs no longer require the `tabular` prop and can be used with any menu variation.

```jsx
const panes = [ ...{} ]
const menu = {
  fluid: true,
  vertical: true,
  secondary: true,
  aligned: 'right'
}

<Tab menu={menu} panes={panes} />
```
image

***

This allows tab menus to be right aligned without having the tabular styles set.

Resolves #2264

### DIFF
--- a/src/modules/Tab/Tab.js
+++ b/src/modules/Tab/Tab.js
@@ -73,7 +73,7 @@ class Tab extends Component {
 
   static defaultProps = {
     grid: { paneWidth: 12, tabWidth: 4 },
-    menu: { attached: true, tabular: true },
+    menu: { attached: true, tabular: true, alignment: 'left' },
     renderActiveOnly: true,
   }
 
@@ -124,13 +124,13 @@ class Tab extends Component {
 
     return (
       <Grid {...gridProps}>
-        {menu.props.tabular !== 'right' && GridColumn.create({ width: tabWidth, children: menu })}
+        {menu.props.alignment !== 'right' && GridColumn.create({ width: tabWidth, children: menu })}
         {GridColumn.create({
           width: paneWidth,
           children: this.renderItems(),
           stretched: true,
         })}
-        {menu.props.tabular === 'right' && GridColumn.create({ width: tabWidth, children: menu })}
+        {menu.props.alignment === 'right' && GridColumn.create({ width: tabWidth, children: menu })}
       </Grid>
     )
   }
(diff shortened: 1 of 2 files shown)


levithomason:
Thanks for the contribution!  Could you state the problem that this is solving?  I'd like to be sure I understand before merging.

ryanmcgarvey:
@levithomason this PR Resolves #2264. I'm still working on creating a sandbox testcase, but I'm not sure how to do it with my fork.

brianespinosa:
@ryanmcgarvey please take a moment to review the pen I forked for you on #2264 as I think it resolves what you were hoping to achieve.

levithomason:
I'm awaiting a completed issue template as well.  See original issue.

ryanmcgarvey:
@levithomason @brianespinosa 

I've updated the original issue as per the template. Is the problem I'm trying to solve more clean now?

Resolves #2264

levithomason:
@ryanmcgarvey very much appreciated.  I get it now and agree. We should not couple the left/right alignment to the `tabular` prop.  They can be used separately.

This is a breaking change, so I've updated the description to reflect that.  Thanks for the update.

levithomason:

PR #2269, opened 2017-10-29
chore(test): fix browser config

### LINKED ISSUE #2227: Test: Cannot run local tests without modifying test config
When running ```yarn test``` on local environment (Arch Linux) I ran into the issues described in: https://github.com/GoogleChrome/puppeteer/issues/290. Although, just adding the flags as described in the referred issue wasn't enough. 

I also had to rename the browser: https://github.com/Semantic-Org/Semantic-UI-React/blob/4f29636738fffbaf3b8bcd179f85b0f2c2248b7b/karma.conf.babel.js#L32 in the test config file to reflect the key in the customLaunchers: https://github.com/Semantic-Org/Semantic-UI-React/blob/4f29636738fffbaf3b8bcd179f85b0f2c2248b7b/karma.conf.babel.js#L47

I.e. I renamed ```ChromeHeadless``` browser to ```puppeteer``` and added the flags ```--no-sandbox``` and ```--disable-setuid-sandbox``` on the customLauncher, since the flags wasn't respected when trying to run the browser. This leads me to believe the flag present today (```--stack-trace-limit 50000```) isn't respected as well.

Is this something we need to address? We should maybe communicate how to do the work-around to add the ```--no-sandbox``` and ```--disable-setuid-sandbox``` flags in case more users run into this issue as well.

layershifter:
Fixes #2227.

### DIFF
--- a/karma.conf.babel.js
+++ b/karma.conf.babel.js
@@ -29,7 +29,7 @@ const formatError = (msg) => {
 export default (karmaConfig) => {
   karmaConfig.set({
     basePath: process.cwd(),
-    browsers: ['ChromeHeadless'],
+    browsers: ['puppeteer'],
     client: {
       mocha: {
         reporter: 'html', // change Karma's debug.html to mocha web reporter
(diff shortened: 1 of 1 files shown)


layershifter:
Thanks to @patrikmolsson for suggestions.

patrikmolsson:
@layershifter Thank you! :smile:

levithomason:

PR #2270, opened 2017-10-29
docs(HOC): remove reexport to avoid warning in test

layershifter:


levithomason:

PR #2293, opened 2017-11-05
feat(Checkbox): add support of htmlFor

### LINKED ISSUE #2274: Checkbox: Labels don't have htmlFor
### Steps
Using the [wave toolbar](http://wave.webaim.org/extension/)

1. Navigate to https://react.semantic-ui.com/maximize/checkbox-example-checkbox 
2. Run the wave toolbar, notice the "missing form label error"

### Expected Result
All labels are connected to inputs via the `for` attribute

### Actual Result
No `for` attribute

### Version
0.71.3

I'm working on a fix for this that is similar to [these lines of code](https://github.com/Semantic-Org/Semantic-UI-React/blob/91a0710990bba2d50f768cf1bbd545650d948159/src/collections/Form/FormField.js#L96) but wanted to ask some questions before I submit a PR.

1. The `for` attribute value should be the `id` of the input that it relates to, but `id` is not a required proptype. I'd like to make that required but I'm not sure what the ramifications will be since `as` prop allows for a checkbox to be some other type of html element. Does anyone see a scenario where a required `id` will cause issues?
2. In the same thought, the `id`, if it is passed in, is applied to the `div` and not the `input`, again, I don't think there is any ill affects to having it applied to the `input`, but wanted to confirm?

layershifter:
Fixes #2274.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -9,6 +9,7 @@ import {
   customPropTypes,
   getElementType,
   getUnhandledProps,
+  htmlInputAttrs,
   makeDebugger,
   META,
   partitionHTMLInputProps,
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2294, opened 2017-11-05
test(Card): remove cruft tests

layershifter:
See #2144.


levithomason:

PR #2302, opened 2017-11-08
fix(Input): icon should be always after <input>

### LINKED ISSUE #2298: Input: animation on focus doesn't work now
✔ BUG

### Steps

```javascript
<Form.Input
  type="password"
  name="password"
  icon="lock"
  iconPosition="left"
  label="Password"
/>
```

### Expected Result

```html
<div class="field">
  <label>Password</label>
  <div class="ui left icon input">
    <input type="password" name="password" placeholder="Mot-de-passe" autocomplete="off">
    <i class="lock icon"></i>
  </div>
</div>
```

### Actual Result

```html
<div class="field">
  <label>Password</label>
  <div class="ui left icon input">
    <i class="lock icon"></i>
    <input type="password" name="password" placeholder="Mot-de-passe" autocomplete="off">
  </div>
</div>
```

I loose Semantic-UI nice animation on focus because this Semantic css:

```
.ui.icon.input > input:focus ~ i.icon {
    opacity: 1;
}
```

which doesn't work if icon is above and not below.

### Version
0.76.0

### Testcase
http://jsfiddle.net/o48pt0yf/

layershifter:
Fixes #2298.

### DIFF
--- a/src/elements/Input/Input.js
+++ b/src/elements/Input/Input.js
@@ -222,7 +222,6 @@ class Input extends Component {
     // Render Shorthand
     // ----------------------------------------
     const actionElement = Button.create(action)
-    const iconElement = Icon.create(this.computeIcon())
     const labelElement = Label.create(label, {
       defaultProps: {
         className: cx(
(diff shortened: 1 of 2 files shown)


layershifter:
@levithomason I don't merge this because I don't remember the reasons for different positions. Can you clarify?

levithomason:
The change of position here was strictly for HTML correctness with the prop value (left).  However, it appears the CSS rule was overlooked.  Your change is acceptable.

Someday, inline styles will remove all this CSS hierarchy dependency.  Long live that day!

levithomason:

PR #2306, opened 2017-11-09
feat(ElementType): new wrapper for all elements

layershifter:
Replaces #1879.

# What?

This PR offers the complex multipurpose solution for our internal problems with refs. These changes will allow to improve work with `as` prop and refs.

# Why so complex?

We cannot simply use `ref`, it' isn't a prop. It has a same behaviour as `key` - it never passed to a component. However, there were numerous issues about refs (#1740, #1602, #1849, #1732 and more) and it's time to make something in this way. `document.querySelector` is a workaround,  but it's not a solution.

Problem is a well known in React's world, there is a long-live issue facebook/react#4213, there was also the RFC facebook/react#6974 from Dan.

# Variants

| Project | Prop | Example |
| ------- | ------ | ----- |
| [formsy-react-components](https://github.com/twisty/formsy-react-components) | `componentRef` & `innerRef` | [`refs.md`](https://github.com/twisty/formsy-react-components/blob/master/docs/refs.md) |
| [office-ui-fabric-react](https://github.com/OfficeDev/office-ui-fabric-react) | `componentRef` | [office-ui-fabric-react#1356](https://github.com/OfficeDev/office-ui-fabric-react/issues/1356) |
| [react-redux](https://github.com/reactjs/react-redux)  | `connect` HOC, `withRef` | reactjs/react-redux@2d3d0be |
| [react-router](https://github.com/ReactTraining/react-router) | `innerRef` | [`Link.md`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md) |

# Solution

Several months ago :frowning_face: I've started to work on this problem. I made few iterations and checked many solutions. This PR is its result, the solution consists from two components.

### `withComputedType`

This is a HOC that allows to pass an already computed element type in `as` prop. In fact it's the most controversial part of this PR.

We need the computed type only in several components (i.e. `Button`, `Image`), it also simplifies work with default `as` values (see `Form`, `Image`). I think that we can go with it, as we plan more HOCs in future for transitions and popups.

The only valid against that I see is the increase of the render tree, **all** our components will be wrapped with it.

### `ElementType`

It has the only one task, to decide how to handle `innerRef`s. If we simply wrap it with `Ref` we will always call `findDOMNode`, even when we actually don't need it.

### P.S.

I want to ship this PR as start point and apply changes to left components iteratively. After all components will be updated, I will able to cleanup tests.


src/elements/Container/Container.js
```diff
@@ -5,18 +5,19 @@ import React from 'react'
...
   useKeyOnly,
   useTextAlignProp,
+  withComputedType,
 } from '../../lib'
 
 /**
  * A container limits content to a maximum width.
  */
-function Container(props) {
+const InnerContainer = (props) => {
```
layershifter, line 20: 1. As `Container` is wrapped with HOC, we can't user `prototype.name`. So, we can use arrow functions for components
2. `_meta` should be exported on the default export, now it's an enhanced component. However, I'm not happy about `Inner-` prefix

src/elements/Image/Image.js
```diff
@@ -197,11 +188,24 @@ Image.propTypes = {
...
+InnerImage.defaultProps = {
   as: 'img',
   ui: true,
 }
 
+const computeType = ({ children, dimmer, label, wrapped }) => {
+  if (!_.isNil(dimmer) || !_.isNil(label) || !_.isNil(wrapped) || !childrenUtils.isNil(children)) return 'div'
+}
+
+const Image = withComputedType(computeType)(InnerImage)
```
layershifter, line 177: Type computing function moved to HOC

src/elements/List/ListHeader.js
```diff
@@ -6,19 +6,19 @@ import {
...
 /**
  * A list item can contain a header.
  */
-function ListHeader(props) {
+const InnerListHeader = (props) => {
   const { children, className, content } = props
   const classes = cx('header', className)
-  const rest = getUnhandledProps(ListHeader, props)
-  const ElementType = getElementType(ListHeader, props)
+  const rest = getUnhandledProps(InnerListHeader, props, { passAs: true })
```
layershifter, line 21: Another nasty thing. `as` should be passed to `ElementType`, but not all our components are updated.

I've already add `ignoreProps` option [`babel-plugin-transform-react-handled-props` ](https://github.com/layershifter/babel-plugin-transform-react-handled-props) plugin, but we can't enable it until we will update all components

src/lib/getUnhandledProps.js
```diff
@@ -3,14 +3,18 @@
  * Useful for getting and spreading unknown props from the user.
  * @param {function} Component A function or ReactClass.
  * @param {object} props A ReactElement props object
+ * @param {Object} [options={}]
+ * @param {object|function} [options.passAs=false] Forces a pass of `as` prop to unhandled props
  * @returns {{}} A shallow copy of the prop object
  */
-const getUnhandledProps = (Component, props) => {
+const getUnhandledProps = (Component, props, options = {}) => {
```
layershifter, line 10: This function will be restored after we will update all components.

layershifter:
@levithomason I'll be glad to hear your feedback 😄

levithomason:
So sorry for the delay, I have not had enough time to fully grok this one.  Any input by @davezuko or @brianespinosa would be much help.

I don't want to block progress, but we also need to validate this approach across another min or two.

src/elements/List/ListHeader.js
brianespinosa: @layershifter do we have a list of the components that still need to get updated somewhere? I am assuming this exists in another conversation/PR somewhere.

layershifter: After this PR will be merged I will update #1321 with list for update. I don't want to do this in single PR because there will be too many changes.

Nickersoft:
Any update on the status of this?

levithomason:
I'm worried about the complexity here.  My primary concern is this obfuscates the basic definition what a component is and does so for all of our components.

Thankfully, it looks like React will now support ref forwarding which may solve this problem for us with `React.forwardRef`, see https://github.com/reactjs/rfcs/pull/30.

Per `@mxstbr`'s approval of that PR:

>This would work 100% for styled-components. `innerRef` is one of the uglier patches of our API and I'd love to get rid of it and make the wrapper components truly transparent and pass-through.

I would agree with this sentiment.  I want to try some other approaches before merging this.  I'd also be open to leaving this unsolved until React >= 16.3 ships (with `React.forwardRef`) and requiring that.

brianespinosa:
> I'd also be open to leaving this unsolved until React >= 16.3 ships (with React.forwardRef) and requiring that.

Agreed. Seems to be close (relatively speaking). If ref forwarding will solve our issue it probably makes the most sense to wait for that.

levithomason:
Hear me out before throwing stones but I'm beginning to consider extending a base component for all our HOC needs.  Why?!

I think a component library is a unique thing in terms of React components for apps.  In an application, you have many types of components with varying needs.  Whereas, in a component library, every component needs identical base functionality.

We can either bundle this into a single base component that we extend, or, go the functional approach of making several HOCs and wrap every component with every HOC to get the base feature set needed.

HOCs are hard to debug, test through, and add a lot to the render tree.  Again, in an application, it is typically easier to test dumb components and keep more complex features in a single tested HOC.  However, in a component library, it seems ludicrous to have every component be a stack of function calls and component wrappers.  Every element in a consumer's app would be a stack of function calls and its own notable render tree.  Imagine when they build more complex controls, or, imagine our own controls such as an Input that contains a Button and a Label.  Rather than rendering three React components, we'd be rendering n*h components for a single control where n is the number of components and h is the number of HOCs required for our base functionality.

A base component would make every component in the library a single flat component.  It would also very easily mean all components have all the features they need.

I know this is not the direction of the community, but, I'd love to hear thoughts and opinions on this.

levithomason:
I'll try updating a few components to use a base component and see how it goes.

layershifter:
@levithomason I'm think that we should go with `forwardRef`, I will check it and try to use with SUIR.

levithomason:
OK, I'll wait for you.  Are you thinking it will solve this problem?:

```jsx
const StatelessComponent = () => <div />

<Button as={StatelessComponent} />
```

`Warning: Stateless function components cannot be given refs.`

levithomason:
Adding this to #2747

PR #2308, opened 2017-11-09
feat(Dropdown): add support for `forceSelection` prop

### LINKED ISSUE #2299: Dropdown: add forceSelection prop
I'm not sure if this was by design, so I'm posted this question.  Sorry, if i posted this in the wrong section.

### Steps
1) https://codepen.io/anon/pen/eeBYKy?editors=1111
2) Notice by default, you see the placeholder for the dropdown
3) click on the drop down and don't select any item.  Click away from the drop down.  This will auto populated the first item in the drop down list.  OnChange Event that is triggered, it doesn't give you any context if a specific item in the list is selected vs. a user clicking anywhere on the screen

### Expected Result
1) User clicks away and does not select an item in the drop down.  Expect the placeholder to still be displayed.

### Actual Result
1) Always selects the first item in the dropdown list

### Version
0.76.0

### Testcase
https://codepen.io/anon/pen/eeBYKy?editors=1111

If this is by design, is there anyway around this?  Goal: if user accidently clicks away and not on a list item, I want nothing to be selected.  The way i determine if something is selected is onChange and examining the data.value field.  This value is always set to the first element on the list even if you don't select the item.

Thanks,
Derek

darewreck54:
Fixes #2299.

Add support for forceSeletion for dropdown.

Still need to update unit test for this.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -130,7 +130,7 @@ export default class Dropdown extends Component {
     /** A dropdown can take the full width of its parent */
     fluid: PropTypes.bool,
 
-    /** Detemine if a selection is auto selected if an item isn't directly selected */
+    /** Whether search selection will force currently selected choice when element is blurred. */
     forceSelection: PropTypes.bool,
 
     /** A dropdown menu can contain a header. */
@@ -920,7 +920,6 @@ export default class Dropdown extends Component {
     const enabledIndicies = this.getEnabledIndices(options)
 
     let newSelectedIndex
-
     const firstIndex = enabledIndicies[0]
 
     // update the selected index
@@ -946,8 +945,7 @@ export default class Dropdown extends Component {
       newSelectedIndex = _.includes(enabledIndicies, activeIndex) ? activeIndex : undefined
     }
 
-    // if no selection is made and forceSelection is true, default to the first
-    // element
+    // if no selection is made and forceSelection is true, default to the first element
     if (forceSelection && (!newSelectedIndex || newSelectedIndex < 0)) {
       newSelectedIndex = firstIndex
     }
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -361,7 +361,7 @@ export default class Dropdown extends Component {
     additionPosition: 'top',
     closeOnBlur: true,
     deburr: false,
-    forceSelection: false,
+    forceSelection: true,
     icon: 'dropdown',
     minCharacters: 1,
     noResultsMessage: 'No results found.',
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -130,6 +130,9 @@ export default class Dropdown extends Component {
     /** A dropdown can take the full width of its parent */
     fluid: PropTypes.bool,
 
+    /** Detemine if a selection is auto selected if an item isn't directly selected */
+    forceSelection: PropTypes.bool,
+
     /** A dropdown menu can contain a header. */
     header: PropTypes.node,
 
@@ -358,6 +361,7 @@ export default class Dropdown extends Component {
     additionPosition: 'top',
     closeOnBlur: true,
     deburr: false,
+    forceSelection: false,
     icon: 'dropdown',
     minCharacters: 1,
     noResultsMessage: 'No results found.',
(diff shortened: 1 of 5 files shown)


layershifter:
`forceSelection` should be `true` by default as now. I'll give more detailed review tomorrow.

src/modules/Dropdown/Dropdown.js
```diff
@@ -358,6 +361,7 @@ export default class Dropdown extends Component {
     additionPosition: 'top',
     closeOnBlur: true,
     deburr: false,
+    forceSelection: false,
```
darewreck54: oops will correct.  I had it as true initially, but will correct.

darewreck54:
I'm actually checking out the failed test and i don't think this is related to changes I made in this pull request:

✖ 1 test failed

```
FAILED TESTS:
  TransitionablePortal
    onHide
      ✖ is called with (null, data) when exiting transition finished
        Error: Uncaught AssertionError: expected spy to have been called exactly once, but it was called 0 times (node_modules/chai/lib/chai/assertion.js:141:0
```

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -474,6 +474,28 @@ describe('Dropdown', () => {
 
       instance.setSelectedIndex.should.not.have.been.calledOnce()
     })
+
+    it('will only set the selectedIndex to the first element if forceSelection is set to true (default)', () => {
```
darewreck54: I added unit test, but i'm not understanding why it's failing.

it's complaining that does not have a 'selectedIndex' state. However, I know that this isn't true. I also followed the pattern in line 1987 which has the same line of code. It's probably something i'm missing with how the test should be written.

Logic:

1. Create the dropdown component
2. initialize the initial state with setState
3. call setProps to trigger a change which will call setSelectedIndex().
4. validate the state of selectedIndex

I'm not familiar enough with the test infrastructure to debug the test. Any insights on how to configure debugging would be helpful.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -474,6 +474,28 @@ describe('Dropdown', () => {
 
       instance.setSelectedIndex.should.not.have.been.calledOnce()
     })
+
```
darewreck54: Also i notice that you make a call to common.propKeyOnlyToClassName(Dropdown, 'fluid');  Does this need to be done with forceSelection.  I'm not understanding what this class actually test.

Also any reason why the react code isn't in jsx files vs. the js files that i'm seeing

darewreck54:
associated with issue https://github.com/Semantic-Org/Semantic-UI-React/issues/2299

layershifter:
@darewreck54 I will review proposed changes tomorrow, I also will take a look on tests.

darewreck54:
@layershifter any updates?

layershifter:
@darewreck54 Sorry for delay, to much work on this week.

I'm think that `onChange` should be called, but only when `value` was changed. It looks curious to me when I have a change event when I simply close `Dropdown` and select nothing.

test/specs/modules/Dropdown/Dropdown-test.js
layershifter: > `common.propKeyOnlyToClassName(Dropdown, 'fluid')`

Yes, it tests that `fluid` CSS class is added when prop is defined.

> Also any reason why the react code isn't in jsx files vs. the js files that i'm seeing

It's our code style and we will follow it in future. And yes, I know that there is the [airbnb's rule](https://github.com/airbnb/javascript/tree/master/react#naming) about naming 😄

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -474,6 +474,24 @@ describe('Dropdown', () => {
...
+    })
+
+    it('will only not set the selectedIndex to the first element if forceSelection is set to false', () => {
+      wrapperMount(<Dropdown options={options} forceSelection={false} />)
+      wrapper.setProps({
+        options: [],
+      })
+
+      wrapper.setState({ selectedIndex: undefined })
+      wrapper.should.have.state('selectedIndex', undefined)
```
layershifter:
```jsx
wrapper.should.have.state('selectedIndex', undefined)
```

The assertion above is broken, don't have any idea why.

Assertions below work :thinking: 
```jsx
expect(wrapper.state().selectedIndex).to.equal(undefined)
wrapper.should.have.not.state('selectedIndex')
```

test/specs/modules/Dropdown/Dropdown-test.js
layershifter: @darewreck54 I've changed test quite a bit, now it passes.

> I'm not familiar enough with the test infrastructure to debug the test. Any insights on how to configure debugging would be helpful. 

All my attempts to get work debugging tools were in vain 😭  I don't have a normal solution for you there. I remove all non-related tests and debug them with `console`.

layershifter (approved):
@darewreck54 Sorry again for delay, I have many tasks with my new project at work and too less free time.

I've pushed some changes and now this PR LGTM. @levithomason we wait for your review there, please pay attention to new comments in `Dropdown.js`, they don't look clear to me

levithomason:
I believe this is solved by the existing `selectOnBlur` prop, see https://github.com/Semantic-Org/Semantic-UI-React/issues/2299#issuecomment-346953224 for more detail.  If I am mistaken, I apologize and will reopen.

ghost:
`selectOnBlur={false}` works good for me.

PR #2309, opened 2017-11-10
fix(Dropdown): display all provided values, even if they aren't in the options

jayphelps:
If you provide active values that aren't in the options you provide they're currently ignored. This PR changes that behavior so that `values` is the source of truth.

This is a "fix" to us because this is how the jQuery variant works, but this could also be seen as a feature and also potentially a breaking change if someone was relying on this. Lmk if you'd like this changed of course. 😄 

This allows the dev to not need to maintain two separate options in state; the original, and the modified one with additions. Instead they can think of options as immutable to the user. This prolly comes down to a difference in perception of what "options" is and whether a controlled component should blindly render the provided values or selectively choose.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -860,7 +860,9 @@ export default class Dropdown extends Component {
   getItemByValue = (value) => {
     const { options } = this.props
 
-    return _.find(options, { value })
+    // The added item might not be in the provided options
+    // in which case we just assume the value/text are the same
+    return _.find(options, { value }) || { text: value, value }
   }
 
   getMenuItemIndexByValue = (value, givenOptions) => {
--- a/test/specs/modules/Dropdown/Dropdown-test.js
+++ b/test/specs/modules/Dropdown/Dropdown-test.js
@@ -1361,6 +1361,22 @@ describe('Dropdown', () => {
         .find('Label')
         .should.have.prop('content', activeItem.text)
     })
+    it('displays a label for active items that are not in options', () => {
+      // select a random item, expect a label with the item's text
+      const activeItem = _.sample(options)
+      wrapperShallow(<Dropdown options={options} selection value={[activeItem.value, 'not-in-options']} multiple />)
+        .should.have.descendants('Label')
+
+      wrapper
+        .find('Label')
+        .at(0)
+        .should.have.prop('content', activeItem.text)
+
+      wrapper
+        .find('Label')
+        .at(1)
+        .should.have.prop('content', 'not-in-options')
+    })
     it('keeps the selection within the range of remaining options', () => {
       // items are removed as they are made active
       // the selection should move if the last item is made active


jayphelps:
> SUMMARY:
> ✔ 7478 tests completed
> ✖ 2 tests failed
> 
> FAILED TESTS:
> Sticky
> ✖ "after each" hook
> HeadlessChrome 0.0.0 (Linux 0.0.0)
> Error: Uncaught RangeError: Maximum call stack size exceeded (src/modules/Modal/Modal.js?e6f1:1880:0
> 
> ✖ "before each" hook
> HeadlessChrome 0.0.0 (Linux 0.0.0)
> TypeError: Cannot read property 'call' of undefined

Not a clue. "works for me" and all that 😆

layershifter:
@jayphelps Thanks for PR, however I don't think that we should merge it, because it will all more strange magic to `Dropdown`. @levithomason  what you think?

jayphelps:
Can you clarify your thinking a bit? My current view is that a controlled component choosing not to display the state it was provided seems more magical. I’m obv coming at it from a biased prospective.

jayphelps:
I’ll come up with a demo of why the existing behavior is fairly inconvenient/awkward for us, maybe it’ll help my case, or reveal an alternative we don’t see.

jayphelps:
Here's a couple examples to demonstrate the awkwardness we're experiencing. It uses redux-form to manage the state of the Dropdown.

The first one here shows the issue, that additions that aren't included in the options are ignored:

https://codesandbox.io/s/ov3jq53vxy?module=%2FExampleForm.js

When you add new additions, they are not kept because we would need to update the options to include this new addition. In our app, the options come from the server and are used in several places, so we don't want to update _that_ particular copy. Instead, we'd could maintain a separate copy that includes the additions but we also need the options to stay in sync with the originals (as they change by the server). So instead I think the solution is to only store the additions and then merge them with the options on-demand. Here's an example of that:

https://codesandbox.io/s/52woz05xlk?module=%2FExampleForm.js

This obviously works, so I'm not blocked. Just wondering if the jQuery behavior was more intuitive and helpful as a controlled component. Cheers gents!

levithomason:
Sorry for the delay getting to this!

>My current view is that a controlled component choosing not to display the state it was provided seems more magical.

Definitely agreed. However, in this case, we are doing exactly that, always displaying the state exactly as it is.

The `allowAdditions` prop adds the UI and flow for users to add options.  We do not automatically handle option that the user attempts to add (i.e updating the list of options).  We used to do this but we found most users had a unique flow required when the option is added.  They didn't want the text/value to be set the way we did, they wanted to modify the values first, or they didn't want the options to update at all because they wanted to post the item to a server instead, etc.

Instead of trying to support every use case with more configurable props, we provided the `onAddItem` prop.  This way, users can handle the additional item according to their needs.

>In our app, the options come from the server and are used in several places, so we don't want to update that particular copy.

Yep.  This is the typical optimistic UI problem.  If you're using a state management framework like Redux or MobX you can implement this experience at a shared higher level in your app.  You can optimistically add the item to the list of items in local app state, then on POST failure, your state management would rollback the addition.  Since your Dropdown would get its options from app state, it will automatically reflect the proper options.

***

Your examples are both idiomatic React and I don't see any unexpected results there.

The first example explicitly renders the Dropdown with a fixed array of options.  I would expect the Dropdown to always show those options.  Compare this to a vanilla controlled React input:
```jsx
<Dropdown options={[ { key: 1, text: 'One', value: 1 } ]} />
<input value='One' />
```
Just as the `input` will never change value, because it is rendered with a fixed value, the `Dropdown` will never change `options` because it is rendered with a fixed set of options.

I think your second example linked above is idiomatic React.  Again, compare this to a vanilla controlled React `input`.  You need to move the value to `state` and update it with an `onChange` handler.  This is symmetric with the `Dropdown` `options`.

I hope that helps!

jayphelps:
Nope it doesn’t, and I still disagree, but that’s okay. Thanks for taking the time. 👍

> Definitely agreed. However, in this case, we are doing exactly that, always displaying the state exactly as it is.

Unless we’re talking about different things this isn’t the case, see the demos above and the PR content. It only renders values which are also in options, so it doesn’t render exactly as provided.

No worries, disagreements are part of OSS. ✌️

levithomason:
Apologies if I'm not understanding.  Are you expecting the Dropdown to render something different than what is passed to its options?

I don't want to overlook a bug here.

jayphelps:
Yep, but possibly not what you’re thinking. I’m referring to the selected values. So if you add something that is not in the options it currently does not show up

jayphelps:
Yep it’s possible there’s confusion about what  referring to.

```
// will NOT render the “not-in-options” value because it is not in the options array
<Dropdown options={options} selection value={[activeItem.value, 'not-in-options']} multiple />
```

> . Are you expecting the Dropdown to render something different than what is passed to its options?

Actually the opposite. Right now it does not render what I give it. My argument was for the component to render what it is given as-is. So if it’s given a selected value that isn’t in the options it doesn’t matter it will render it as a selected value anyway.

jayphelps:
Not a big deal either way, I’ve abstracted this away with a wrapper component. 👍

levithomason:
>So if it’s given a selected value that isn’t in the options it doesn’t matter it will render it as a selected value anyway.

Ah, I understand now.  Yes, this is an interesting edge case and I think you're actually correct here.  Glad you've solved it, if it comes up again, I'd gladly update the component.

PR #2311, opened 2017-11-12
refactor(dts): add white to the list of SemanticColors in typescript dts

sammysamau:
Add white to the list of SemanticColors in the Typescript definition.


layershifter:
Thanks for the contribution, however we can't merge it.

> The color names are defined by Semantic UI css selectors, they are not actual inline style css colors.

SUI [doesn't support](https://semantic-ui.com/usage/theming.html) `white` as primary color.

sammysamau:
My mistake. I had thought 'ui inverted white' is what made an inverted button white but it turns out that's the default.

Issue #2301, opened 2017-11-08
Icon render slow depending on the word order in name property


akomm:
Looks like https://github.com/Semantic-Org/Semantic-UI-React/pull/2149 did not entirely fix https://github.com/Semantic-Org/Semantic-UI-React/issues/1184

Or is it not possible to "fix" the suggest performance via memoize entirely?

### Steps
1. Create `Icon` component
2. Provide the name property "square check" (word order is important)
3. (optional) Render it multiple times or copy & paste the same icon 10 times

### Expected Result
Rendering should be fast.

### Actual Result
Rendering is very slow. In my case 10 Icons under dev environment take about 600-800ms.
When you write "check square" it renders fast.

Did not test it with other multi-word icons.

Found the issue in suggest-usage from inside the semantic Icon component using performance profiling.

### Version
0.76.0

dvdzkwsk:
PR for this: https://github.com/Semantic-Org/Semantic-UI-React/pull/2314. I don't think memoize is enough, the lookup through an array is super expensive when there are multiple unique icons. I'm going to consider ways to optimize the word ordering, since that will always perform an expensive check, unfortunately.

akomm:
I once noticed  that the compiled CSS for icons (the one with content values) is quite large. So far it was not that critical and important, but I had the Idea to maybe create a small parser and go through FA variables.scss and generate an array of "icon name" => "content value", while "icon name" is split and ordered and joined together. Then create an Icon component, which accepts a name prop, but does split and order the words same way.  The class/style for its content could be generated inside the Icon component itself. Ordering words consistent in Icon component and the precompilation of the name-value-array should make the need of searching and guessing obsolete.

Not sure how this fits into this project, as you have the CSS and this project actually separated, since the CSS for everything beside the content-values would still have to come from CSS and version matching might be important.

brianespinosa:
@akomm At some point in the future styles may become a part of this repository, but at present, we are relying on the structure of the CSS that comes out of the build tools for SUI core.

In my current project we are actually passing the icon names as the `className` instead for performance in development. As @levithomason [mentioned on the original issue](https://github.com/Semantic-Org/Semantic-UI-React/issues/1184#issuecomment-292458271), this only effects dev environments. In production all PropType validation is not enabled.

akomm:
Thanks for the post reference. I did find it too before I'v posted this issue. This is only for feedback purpose on the PR, which missed this edge case (word order). :)

[...]

PR #2314, opened 2017-11-13
fix(customPropTypes): improve perf of `suggest`

dvdzkwsk:
~This is not ready yet because it's not solving the case presented by https://github.com/Semantic-Org/Semantic-UI-React/issues/2301, which concerns word ordering.~

Word ordering has been solved by way of key sorting.

### DIFF
--- a/src/lib/customPropTypes.js
+++ b/src/lib/customPropTypes.js
@@ -12,29 +12,6 @@ export const as = (...args) => PropTypes.oneOfType([
   PropTypes.func,
 ])(...args)
 
-/* eslint-disable max-nested-callbacks */
-const findBestSuggestions = _.memoize((propValueWords, suggestions) => _.flow(
-  _.map((suggestion) => {
-    const suggestionWords = suggestion.split(' ')
-
-    const propValueScore = _.flow(
-      _.map(x => _.map(y => leven(x, y), suggestionWords)),
-      _.map(_.min),
-      _.sum,
-    )(propValueWords)
-
-    const suggestionScore = _.flow(
-      _.map(x => _.map(y => leven(x, y), propValueWords)),
-      _.map(_.min),
-      _.sum,
-    )(suggestionWords)
-
-    return { suggestion, score: propValueScore + suggestionScore }
-  }),
-  _.sortBy(['score', 'suggestion']),
-  _.take(3),
-)(suggestions))
-/* eslint-enable max-nested-callbacks */
 
 /**
  * Similar to PropTypes.oneOf but shows closest matches.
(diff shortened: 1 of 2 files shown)


src/lib/customPropTypes.js
```diff
@@ -42,33 +42,39 @@ const findBestSuggestions = _.memoize((propValueWords, suggestions) => _.flow(
...
-      ` See \`${propName}\` prop in \`${componentName}\`.`,
-    ].join(''))
+    throw new Error('Invalid argument supplied to suggest, expected an instance of array.')
   }
-  const propValue = props[propName]
 
-  // skip if prop is undefined or is included in the suggestions
-  if (_.isNil(propValue) || propValue === false || _.includes(propValue, suggestions)) return
+  // Convert the suggestions list into
+  const suggestionsLookup = suggestions.reduce((acc, key) => {
```
dvdzkwsk: Convert O(n) lookup to O(1).

levithomason: 🙌

src/lib/customPropTypes.js
```diff
@@ -12,63 +12,80 @@ export const as = (...args) => PropTypes.oneOfType([
...
-      ` See \`${propName}\` prop in \`${componentName}\`.`,
-    ].join(''))
+    throw new Error('Invalid argument supplied to suggest, expected an instance of array.')
   }
-  const propValue = props[propName]
 
-  // skip if prop is undefined or is included in the suggestions
-  if (_.isNil(propValue) || propValue === false || _.includes(propValue, suggestions)) return
+  /* eslint-disable max-nested-callbacks */
+  const findBestSuggestions = _.memoize((str) => {
```
dvdzkwsk, line 28: Implementation is the same, just passed the base string in instead of an array so that `_.memoize` can actually do its job, since arrays are not memoized by value:

```js
let calls = 0
const fn = _.memoize((arr) => ++calls)

fn(['my', 'words'])
fn(['my', 'words'])
console.log(calls) // 2
```

Note that the previous usage of this memoized function was incorrect, since it would ignore the second argument `suggestions`, meaning the first call to it with a given propValue would have won out. This was unintentionally mitigated by the fact that `_.memoize` only tracks the first argument by default, and that argument was an array, ergo no actual memoization was occurring.

The _real_ fix would be to write our own resolver function for the memoization, but it would essentially achieve the same thing as I'm doing here by inlining the function, given the use case for this `suggest` utility.

src/lib/customPropTypes.js
```diff
@@ -12,63 +12,80 @@ export const as = (...args) => PropTypes.oneOfType([
...
+    )(suggestions)
+  })
+  /* eslint-enable max-nested-callbacks */
+
+  // Convert the suggestions list into a hash map for O(n) lookup times. Ensure
+  // the words in each key are sorted alphabetically so that we have a consistent
+  // way of looking up a value in the map, i.e. we can sort the words in the
+  // incoming propValue and look that up without having to check all permutations.
+  const suggestionsLookup = suggestions.reduce((acc, key) => {
+    acc[key.split(' ').sort().join(' ')] = true
```
dvdzkwsk, line 60: Key sorting

dvdzkwsk:
Anything else needed here?

layershifter:
@davezuko I want to hear @levithomason before merge this. Let's wait for his review.

levithomason:
I had a few conversations with @davezuko over Slack about this.  It should make a pretty big improvement.  Actual metrics would be nice, but I don't have time for it.  Will release soon.

levithomason:

Issue #2079, opened 2017-09-15
srcSet missing from Image


Maxhodges:
srcSet should probably be a property of Image
`Warning: Unknown DOM property srcset. Did you mean srcSet?`

layershifter:
All unhandled props are passed to component's container, [codepen](https://codesandbox.io/s/x32w133z4p).

Maxhodges:
why do you provide src but not srcSet? I understand it's not defined by SUI, but do you think it shoul be?

layershifter:
Because, `src` is handled and is needed by our factories. While `srcSet` is a simple prop, we don't need to expose all possible HTML props in components.

Maxhodges:
OK. thank you sir!

codekiln:
@layershifter while I agree that SemanticUI shouldn't expose *all* HTML props in components, `srcSet` seems very important. Browsers already know how to render responsive images so we can handle the difference between Retina, iPhone and normal resolution, and without `srcSet` placed directly on the `<img`, what is the proposed method? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Example_3_Using_the_srcset_attribute

Update / Workaround: I just figured out you can achieve the same thing by placing your own image inside the Image component: 
```
<Item.Image size="tiny" shape="circular">
<img 
  src={withPrefix(avatar.responsiveResolution.src)}
  srcSet={getPrefixedSrcSet(avatar.responsiveResolution.srcSet)}/>
</Item.Image>
```

That's not so bad after all.

Maxhodges:
The `picture` element is "beginning to land in browsers". Google is recommending it here. Perhaps this is something SUI should get behind. @jlukic 
https://developers.google.com/web/fundamentals/design-and-ux/responsive/images

layershifter:
I've added #2316, handling of HTML props were improved, examples were added.

PR #2316, opened 2017-11-13
feat(Image): improve handling of HTML props

layershifter:
Related to #2079.

This PR improves handling of HTML props using our `partitionHTMLInputProps` function and adds examples.

### DIFF
--- a/src/elements/Image/Image.js
+++ b/src/elements/Image/Image.js
@@ -10,6 +10,7 @@ import {
   getElementType,
   getUnhandledProps,
   META,
+  partitionHTMLInputProps,
   SUI,
   useKeyOnly,
   useKeyOrValueAndKey,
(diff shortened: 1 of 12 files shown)


src/elements/Image/Image.js
```diff
@@ -73,6 +71,7 @@ function Image(props) {
     className,
   )
   const rest = getUnhandledProps(Image, props)
+  const [imgTagProps, rootProps] = partitionHTMLInputProps(rest, { htmlProps: imageProps })
```
levithomason: This seems strange to overload an input prop parser for an image.  Should this become a generic parser?

layershifter: Agree, I will update PR and we will go with `partitionHTMLProps`.

layershifter:
`partitionHTMLInputProps` function was renamed to `partitionHTMLProps` . I will merge this as there was no another remarks.

levithomason:

PR #2320, opened 2017-11-14
feat(Embed): disable "related videos" when `brandedUI` is set to `true`

### LINKED ISSUE #2313: Embed: brandedUI setting seems backwards, and doesn't affect related videos
### Steps
Create a YouTube iframe embed element, and set `brandedUI` to `false`. Observe that the YouTube logo still displays.

The documentation for the embed modules says that the `brandedUI` setting determines "whether to show networks branded UI like title cards, or after video calls to action."

My understanding of this is that setting `brandedUI` to `false` is meant to hide the YouTube branding on the player. Looking [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Embed/Embed.js#L138) however, it seems like that's backwards, and setting `brandedUI` to `true` would remove the YouTube logo.

In addition, it would be nice if this prop disabled the "related videos" setting in the YouTube embed (`rel=0`), since it says it's also supposed to affect "after video calls to action".

### Expected Result
YouTube branding (and, ideally, related videos) are disabled when setting `brandedUI={false}`.

### Actual Result
YouTube branding is disabled when setting `brandedUI={true}`. Related videos are shown no matter what.

### Version
0.76.0

### Testcase
https://codesandbox.io/s/o10xl8qx0y

If this is in fact a bug I'd be happy to submit a PR to fix it, but I wanted to check to see if my understanding was correct first.

bennettrogers:
Fixes #2313.

### DIFF
--- a/docs/app/Examples/modules/Embed/Usage/EmbedExampleIframe.js
+++ b/docs/app/Examples/modules/Embed/Usage/EmbedExampleIframe.js
@@ -4,7 +4,6 @@ import { Embed } from 'semantic-ui-react'
 const EmbedExampleIframe = () => (
   <Embed
     autoplay={false}
-    brandedUI={false}
     color='white'
     hd={false}
     id='gJscrxxl_Bg'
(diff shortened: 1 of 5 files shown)


layershifter (approved):
@bennettrogers Thanks for PR :+1: I've added some changes, LGTM

levithomason:

PR #2329, opened 2017-11-18
fix(Portal): fix usage of eventPool

layershifter:
Fixes usage of `eventPool` parameter in `Portal`, [third argument](https://github.com/Semantic-Org/Semantic-UI-React/blob/fix/modal-eventpool/src/lib/eventStack/eventStack.js#L39) should be an `object` with options.

See #2099 more context.


levithomason:

Issue #2224, opened 2017-10-20
Dropdown: maxSelections


davioliveiira:
This element `<Dropdown />` not have support for `maxSelections`? #

layershifter:
No, we don't have support of `maxSelections`. Contributions are welcome.

patrikmolsson:
How would this work? Would it simply just not allow the user to select more items when the cap is reached?

How can we communicate this to the user? Shall we leave it to the developer how to handle when the cap is reached - maybe with a custom callback?

patrikmolsson:
@layershifter Any suggested implementation for this? I would gladly help with it.

layershifter:
@patrikmolsson you can use [these](https://semantic-ui.com/modules/dropdown.html#maximum-selections) examples as reference behaviour

patrikmolsson:
@layershifter Thanks! :grin:  I'm on it.

fazlulkarimweb:
I'm posting my workaround here for people who are trying to find a solution on google. 

At first, save the values in the state (preferably redux state) after every onChange. React state also would do fine. Then, make it disabled when a certain array length of the value is reached. 

```js
const districtData = ['Dhaka', 'Bagerhat', 'Bandarban', 
'Barguna', 'Barishal', 'Bhola']

const [districtValue, setDistrictValue] = useState();

<Dropdown
onChange={async (e, { name, value }) => {setDistrictValue(value)}}
options={districtData.map((currentValue, index) => {
    return {
    key: `${currentValue}`,
    value: `${currentValue}`,
    text: `${currentValue}`,
    disabled: districtValue.length > 2 ? true : false 
}
// Do other things here
// Max 3 values here, then options will be disabled. 
// Will be live again if any selected options are less than 3 here                

/>
```

PR #2330, opened 2017-11-19
feat(Dropdown): add support for 'maxSelections'

patrikmolsson:
Addresses https://github.com/Semantic-Org/Semantic-UI-React/issues/2224

- [x] Add tests
- [x] Add example
- [x] Add prop check (maxSelections prop should only be available with multiple and selection).
- [x] Implement subcomponent DropdownMessage, see feedback below.
- [x] Change ```maxSelections``` prop to allow strings as well, for coherence. (do not implement)
- [x] Use factory for DropdownMessage
- [x] Add specific `DropdownMessage` tests

### DIFF
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
@@ -22,8 +22,15 @@ const options = [
   { key: 'ux', text: 'User Experience', value: 'ux' },
 ]
 
-const DropdownExampleMultipleSelection = () => (
-  <Dropdown fluid multiple selection search options={options} maxSelections={3} />
+const DropdownExampleMaxSelections = () => (
+  <Dropdown
+    fluid
+    maxSelections={3}
+    multiple
+    search
+    selection
+    options={options}
+  />
 )
 
-export default DropdownExampleMultipleSelection
+export default DropdownExampleMaxSelections
--- a/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
+++ b/docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { Dropdown } from 'semantic-ui-react'
+
+const options = [
+  { key: 'angular', text: 'Angular', value: 'angular' },
+  { key: 'css', text: 'CSS', value: 'css' },
+  { key: 'design', text: 'Graphic Design', value: 'design' },
+  { key: 'ember', text: 'Ember', value: 'ember' },
+  { key: 'html', text: 'HTML', value: 'html' },
+  { key: 'ia', text: 'Information Architecture', value: 'ia' },
+  { key: 'javascript', text: 'Javascript', value: 'javascript' },
+  { key: 'mech', text: 'Mechanical Engineering', value: 'mech' },
+  { key: 'meteor', text: 'Meteor', value: 'meteor' },
+  { key: 'node', text: 'NodeJS', value: 'node' },
+  { key: 'plumbing', text: 'Plumbing', value: 'plumbing' },
+  { key: 'python', text: 'Python', value: 'python' },
+  { key: 'rails', text: 'Rails', value: 'rails' },
+  { key: 'react', text: 'React', value: 'react' },
+  { key: 'repair', text: 'Kitchen Repair', value: 'repair' },
+  { key: 'ruby', text: 'Ruby', value: 'ruby' },
+  { key: 'ui', text: 'UI Design', value: 'ui' },
+  { key: 'ux', text: 'User Experience', value: 'ux' },
+]
+
+const DropdownExampleMultipleSelection = () => (
+  <Dropdown fluid multiple selection search options={options} maxSelections={3} />
+)
+
+export default DropdownExampleMultipleSelection
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -26,6 +26,7 @@ import DropdownItem from './DropdownItem'
 import DropdownHeader from './DropdownHeader'
 import DropdownMenu from './DropdownMenu'
 import DropdownSearchInput from './DropdownSearchInput'
+import DropdownMessage from './DropdownMessage'
 
 const debug = makeDebugger('dropdown')
 
@@ -403,6 +404,7 @@ export default class Dropdown extends Component {
   static Header = DropdownHeader
   static Item = DropdownItem
   static Menu = DropdownMenu
+  static Message = DropdownMessage
   static SearchInput = DropdownSearchInput
 
   getInitialAutoControlledState() {
@@ -1244,11 +1246,11 @@ export default class Dropdown extends Component {
     const options = this.getMenuOptions()
 
     if (!_.isNil(maxSelections) && value.length >= maxSelections) {
-      return <div className='message'>Max {maxSelections} selection{maxSelections > 1 && 's'}</div>
+      return <DropdownMessage type={'maxSelections'} value={maxSelections} />
     }
 
     if (noResultsMessage !== null && search && _.isEmpty(options)) {
-      return <div className='message'>{noResultsMessage}</div>
+      return <DropdownMessage type={'noResultsMessage'} noResultsMessage={noResultsMessage} />
     }
 
     const isActive = multiple
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1253,7 +1253,6 @@ export default class Dropdown extends Component {
     const options = this.getMenuOptions()
 
     if (!_.isNil(maxSelections) && value.length >= maxSelections) {
-
       return DropdownMessage.create('maxSelectionsMessage', {
         defaultProps: {
           value: maxSelections,
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -155,6 +155,9 @@ export default class Dropdown extends Component {
     /** A dropdown can show that it is currently loading data. */
     loading: PropTypes.bool,
 
+    /** A dropdown can allow a maximum number of selections. */
+    maxSelections: PropTypes.number,
+
     /** The minimum characters for a search to begin showing results. */
     minCharacters: PropTypes.number,
 
@@ -593,7 +596,7 @@ export default class Dropdown extends Component {
 
   makeSelectedItemActive = (e) => {
     const { open } = this.state
-    const { multiple } = this.props
+    const { maxSelections, multiple } = this.props
 
     const item = this.getSelectedItem()
     const value = _.get(item, 'value')
@@ -601,6 +604,9 @@ export default class Dropdown extends Component {
     // prevent selecting null if there was no selected item value
     // prevent selecting duplicate items when the dropdown is closed
     if (_.isNil(value) || !open) return
+
+    // prevent selecting item if maxSelections is reached
+    if (!_.isNil(maxSelections) && this.state.value.length >= maxSelections) return
 
     // state value may be undefined
     const newValue = multiple ? _.union(this.state.value, [value]) : value
@@ -722,7 +728,7 @@ export default class Dropdown extends Component {
   handleItemClick = (e, item) => {
     debug('handleItemClick()', item)
 
-    const { multiple, search } = this.props
+    const { maxSelections, multiple, search } = this.props
     const { value } = item
 
     // prevent toggle() in handleClick()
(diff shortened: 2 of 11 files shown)


patrikmolsson:
I've looked into the "positive integer" prop, and no other module seems to be using it, even though they might benefit from one. Example of modules:

- [Rating module](https://react.semantic-ui.com/modules/rating): Uses strings and numbers, although only works with integers. Doesn't work with negative numbers.
- [Progress module](https://react.semantic-ui.com/modules/progress): Uses strings and numbers, works with floats as well. Doesn't work with negative numbers.

Regarding the propType, this leads to the options. Either:
1. Develop a propType that allows only positive numbers, in either number or string format (more general),
2. Develop a propType that allows only positive integer, in either number or string format (exactly what some modules use. Would be a great addition to the Rating module, since it rounds the rating down, but doesn't communicate this to either the user or developer),
3. Do nothing, keep it simple.

What are the community's thoughts on this?

-------

This also indicates that we, for coherence, should change ```maxSelections``` prop to allow strings as well. (Added to todo).

How would I implement this most easily? Convert it to a number every time I make the comparison?

layershifter:
@patrikmolsson thanks for PR 👍 

> What are the community's thoughts on this?

Let's be guided by common sense, if someone indicates a negative value, then he will be wrong 😄

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
```diff
@@ -0,0 +1,29 @@
...
+  { key: 'rails', text: 'Rails', value: 'rails' },
+  { key: 'react', text: 'React', value: 'react' },
+  { key: 'repair', text: 'Kitchen Repair', value: 'repair' },
+  { key: 'ruby', text: 'Ruby', value: 'ruby' },
+  { key: 'ui', text: 'UI Design', value: 'ui' },
+  { key: 'ux', text: 'User Experience', value: 'ux' },
+]
+
+const DropdownExampleMultipleSelection = () => (
+  <Dropdown fluid multiple selection search options={options} maxSelections={3} />
```
layershifter: Let's sort props there and expression multiline, i.e:

```jsx
<Dropdown 
  fluid 
  multiple 
  // ...
/>
```

docs/app/Examples/modules/Dropdown/Usage/DropdownExampleMaxSelections.js
```diff
@@ -0,0 +1,29 @@
...
+)
+
+export default DropdownExampleMultipleSelection
```
layershifter: And fix name 😄 Should be `DropdownExampleMaxSelections`

src/modules/Dropdown/Dropdown.js
```diff
@@ -593,6 +599,9 @@ export default class Dropdown extends Component {
     // prevent selecting duplicate items when the dropdown is closed
     if (_.isNil(value) || !open) return
 
+    // prevent selecting item if maxSelections is reached
+    if (!_.isNil(maxSelections) && this.state.value.length >= maxSelections) return
```
layershifter, line 621: If we will set default value of `maxSelections` to `0`, we will can do `maxSelections > 0` instead of `_.isNil`. BTW, SUI makes [similar thing](https://semantic-ui.com/modules/dropdown.html#multiple-select-settings)

src/modules/Dropdown/Dropdown.js
```diff
@@ -1201,10 +1214,14 @@ export default class Dropdown extends Component {
   }
 
   renderOptions = () => {
-    const { multiple, search, noResultsMessage } = this.props
+    const { maxSelections, multiple, search, noResultsMessage } = this.props
     const { selectedIndex, value } = this.state
     const options = this.getMenuOptions()
 
+    if (!_.isNil(maxSelections) && value.length >= maxSelections) {
+      return <div className='message'>Max {maxSelections} selection{maxSelections > 1 && 's'}</div>
```
layershifter: Let's create a new subcomponent `DropdownMessage`. This will allow to use factories for it, see `DropdownSearchInput` for reference.

It will accept `type` (`maxSelections` or `noResultsMessage`) and `value` (`maxSelections`). This will allow our users customize message better and localize it, too.

src/modules/Dropdown/Dropdown.js
patrikmolsson: This seems to break the propType check that I did to see if ```multiple``` and ```selection``` is also on the element. :/ Any ideas how I can still implement this propType check?

patrikmolsson: @layershifter If I remove the check if it also has ```multiple``` and ```selection``` props when using ```maxSelections```, I can get it to work. Would this be reasonable?

layershifter: Oh, I see. I had similar situation with `Embed`. Let's go with `_.isNil`.

patrikmolsson:
I've tried to implement the DropdownMessage, but I seem to be missing something. I encounter this error: 

```
[22:02:45] TypeError in plugin 'gulp-react-docgen'
Message:
    Cannot read property 'replace' of undefined
File: /home/patrik/Development/Semantic-UI-React/src/modules/Dropdown/DropdownMessage.js.
```

Any ideas?

levithomason:
Ensure your component passes conformance tests.  This is the easiest way to make sure you set it up right.  You can also copy an existing component and update the values.

My guess is you are missing the static `_meta` property, or have it improperly configured. See:

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.github/CONTRIBUTING.md#define-_meta

levithomason:
I'd recommend following the workflow for creating a new component:

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.github/CONTRIBUTING.md#workflow

It walks through each step with examples.

patrikmolsson:
@levithomason Thanks for the tips!

I managed to finally fix it, I forgot that ```PropTypes.oneOf``` takes an array, and not multiple arguments... 

Well, now it is finally up and running. As always, any feedback is appreciated!

I've also looked into the receive both string and number input for ```maxSelections``` prop, and usually we only use string and number when SUI itself can handle it. Would it be reasonable to only accept number for ```maxSelections```? If not - should I move ```maxSelections``` to state and handle changes with ```componentWillReceiveProps```?

layershifter:
@patrikmolsson In this case I think it will be better if we will go with `PropTypes.number`. `Dropdown` is enough complicated, less code is better.

patrikmolsson:
@layershifter Great! Then I think I've covered all aspects. Anything else that needs to be done before merge?

layershifter:
@patrikmolsson Great work 👍 This PR needs a review, however I didn't had enought time for it.

src/modules/Dropdown/Dropdown.js
```diff
@@ -1201,12 +1216,16 @@ export default class Dropdown extends Component {
...
     const { selectedIndex, value } = this.state
     const options = this.getMenuOptions()
 
+    if (!_.isNil(maxSelections) && value.length >= maxSelections) {
+      return <DropdownMessage type={'maxSelections'} value={maxSelections} />
+    }
+
     if (noResultsMessage !== null && search && _.isEmpty(options)) {
-      return <div className='message'>{noResultsMessage}</div>
+      return <DropdownMessage type={'noResultsMessage'} noResultsMessage={noResultsMessage} />
```
layershifter: Let's use factories there, [here is example](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.77.0/src/modules/Dropdown/DropdownSearchInput.js#L82), we will also use `type` as meanfull prop. This will allow to get following usage:

```jsx
DropdownMessage.create('noResultsMessage')
```

It will allow to use shorthanding options there. We also need to add a new test [like there](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.77.0/test/specs/collections/Menu/MenuItem-test.js#L12).

src/modules/Dropdown/Dropdown.js
```diff
@@ -1201,12 +1216,16 @@ export default class Dropdown extends Component {
   }
 
   renderOptions = () => {
-    const { multiple, search, noResultsMessage } = this.props
+    const { maxSelections, multiple, search, noResultsMessage } = this.props
     const { selectedIndex, value } = this.state
     const options = this.getMenuOptions()
 
+    if (!_.isNil(maxSelections) && value.length >= maxSelections) {
+      return <DropdownMessage type={'maxSelections'} value={maxSelections} />
```
layershifter:
```jsx
DropdownMessage.create('maxSelectionsMessage', { 
  defaultProps: { 
    value: maxSelections,
  },
})
```

The notes below.

src/modules/Dropdown/Dropdown.js
layershifter: `noResultsMessage` will become obsolete in `Dropdown`, as it will be breaking change we should show before and after usages in the PR heading.

src/modules/Dropdown/DropdownMessage.js
```diff
@@ -0,0 +1,61 @@
...
+function DropdownMessage(props) {
+  const { className, noResultsMessage, type, value } = props
+
+  const classes = cx('message', className)
+
+  let message = ''
+  if (type === 'noResultsMessage') {
+    message = noResultsMessage
+  } else {
+    message = `Max ${value} selection${value > 1 ? 's' : ''}`
```
layershifter: As I mentioned abowe, let's use predefined props there.

```jsx
DropdownMessage.defaultProps = {
  noResults: 'No results found.',
  maxSelections: 'Max {value} selection{suffix}',
}

DropdownMessage.propTypes = {
  noResults: customPropTypes.contentShorthand,
  maxSelections: customPropTypes.contentShorthand,
}
```

```jsx
import { cloneElement, isValidElement } from 'react'

const createMessage = (props, type, value) => {
  const message = props[type]

  if(isValidElement(message)) return cloneElement(message, { type, value })
  if(typeof message !== 'string') return message
  return message
    .replace('value', value)
    .replace('suffix', suffix)
}
```

Also let's add an example with localization to confirm that my suggestions are not falsy.

layershifter (changes requested):
@patrikmolsson The great work there 👍 I'm really sorry that I reached it only now.

It seems that not all my initial suggestion were correct. I've left some comments. Let's improve this PR and get it merged 👍

patrikmolsson:
@layershifter Thank you for your thorough feedback! I will look into it the coming days :smiley:

mikdatdogru:
pls merge and deploy this request... i need this feature :)

patrikmolsson:
Hey everyone,

I will try to get going with this the coming weeks. I took some time off.

@mikdatdogru In the meantime, you could try a workaround with [Custom search](https://react.semantic-ui.com/modules/dropdown#dropdown-example-custom-search-function) and render an appropriate element when you've got the number of selections that you want.

patrikmolsson:
@layershifter I've finally started working on this again.

I have two questions:

1. You mentioned breaking changes earlier when it comes to the `noResultsMessage`. Did you mean that we no longer allow strings to `noResultsMessage` prop? And only `DropdownMessage` components? Could we not allow both that component and a string? Or would that just add unnecessary complexity?
2. To get localization working properly, I would have to add the suffix as a prop as well to the DropdownMessage component. Is this what you had in mind?

I.e. My suggestion is that the final result would be:

- `Dropdown` takes prop `noResultsMessage` as either a string or a `DropdownMessage` component. 
- `Dropdown` takes prop `maxSelectionsMessage` as a `DropdownMessage` component.
- `DropdownMessage` takes props `maxSelectionsMessage`, `maxSelectionsMessageSuffix`, and `noResultsMessage`, and we render the suffix when value > 1.

patrikmolsson:
@layershifter Alright, I think I've got something that's ready for review. However, two things:

1. I would like to allow sending in a function to Dropdown for the localization. So that users can do e.g.:

```jsx
const LocalizedMessage = ({ value }) => `Max ${value} selection${value > 1 ? 's' : ''}`

<Dropdown
  [...]
  maxSelectionsMessage={LocalizedMessage}
/>
```

Do we want to allow that? If so, what are the appropriate `PropTypes`?

2. If the user does:

```jsx
<Dropdown
  [...]
  maxSelectionsMessage={<span>Hello</span>}
/>
```

It will cause this output:

```html
<div class="message">
<span type="maxSelectionsMessage" value="2">
Hello
</span>
</div>
```

Just to confirm, we don't want this flattened, right?

sagynbek:
@patrikmolsson , hey there! So what about this merge? Your requested changes did not pass Tests and we will not get this feature???

patrikmolsson:
@sagynbek Unfortunately I weren't able to implement this. And at the moment I'm afraid I do not have time... 

I might address it later this fall, but it is open for grabs if anyone wants to implement it.

patrikmolsson:
Closing this. Hopefully someone else will find time implementing it.

PR #2333, opened 2017-11-20
fix(Image): correct the type of the width prop

lafritay:
The image width prop maps directly to the img width html attribute.  It is incorrectly defined as 
SemanticWIDTHS.  It should be string | number similar to the height prop.


lafritay:
Let me know if you'd like me to file an issue for this as well

layershifter (approved):
@lafritay Thanks 👍 I will merge this, however keep in mind that #2316 will remove HTML props from `Image`

levithomason:

PR #2348, opened 2017-12-03
chore(docgen): add a friendly exception

layershifter:
This PR adds an new exception to our docgen plugin, it allows better understand problem.


levithomason:

PR #2349, opened 2017-12-03
chore(docgen|menugen): fix problems with paths on Windows

layershifter:


levithomason:

PR #2350, opened 2017-12-03
fix(Transition): fix of the choice of duration

### LINKED ISSUE #2347: Transition: Not working properly when 'show' duration is less than 'hide'
### Steps
Create a transition with a show duration smaller than hide duration
```jsx
<Transition duration={{show: 100, hide: 2000}} animation="fade">
   <h1>Nippertown</h1>
</Transition>
```
### Expected Result
The transition should take 2 seconds to hide the child.

### Actual Result
The hide transition will be interrupted at the show duration (e.g. 100 ms) without completing the intended duration (2 seconds).

### Version
0.77.0

### Testcase
This is reproducible with the example in the [documentation](https://react.semantic-ui.com/modules/transition#transition-example-duration). Configure the sliders to reflect a show value less than the show value.

layershifter:
Fixes #2347.

### DIFF
--- a/src/modules/Transition/Transition.js
+++ b/src/modules/Transition/Transition.js
@@ -160,8 +160,11 @@ export default class Transition extends Component {
 
     this.nextStatus = null
     this.setSafeState({ status, animating: true }, () => {
+      const durationType = TRANSITION_TYPE[status]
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2352, opened 2017-12-03
feat(Pagination): add new component

layershifter:
This PR adds the fully featured pagination to SUIR using existing Menu component.


layershifter:
@levithomason I'll be glad to hear your opinion there 😄

levithomason:
This is so cool and much needed!  Here are my initial thoughts when trying the component.

### showFirstAndLast

I would expect this prop to show the first and last page _number_ in the list of pages.  However, that is already shown.  This prop seems to add first and last _navigation_ buttons but those are duplicating the first and last page _number_ buttons:

image

**💡  Suggestion:**

Perhaps split this into `showFirstAndLastNav` and `showFirstAndLastPage`?

### showPreviousAndNext

Since the item prop is `prevItem` I think we should use `showPrevAndNext`.  Also, if we split the first/last nav vs page props above, then this would be `showPrevAndNextNav` for consistency.

### Item shorthand

I can't seem to turn off the first/last page number.  Passing `null`/`false` still renders them.  I think we should allow turning these off.

```jsx
<Pagination firstItem={null} lastItem={false} />
```
image

Oh, right.  You'd pass `showPreviousAndNext={false}` 😊.  Do we even need the `show*` props?  The `*Item` props could work as `<Dropdown icon={null} />` and `<Modal dimmer={false} />` which removes the shorthand items.  Thoughts?

layershifter:
@levithomason I've applied proposed changes and removed all boolean `show*` props 👍

levithomason:
Cool, great first implementation!

levithomason:

PR #2357, opened 2017-12-05
docs(README.md): add bailfacile.fr to list of websites using Semantic UI React

valentinfily:
Add www.bailfacile.fr website using Semantic UI React to list


levithomason:

Issue #406, opened 2016-08-19
RFC: Standardize width props


levithomason:
# Issue

Many components can specify a width in number of grid columns, sometimes also supporting `equal width`.  We've standardized the values we accept (number, string, 'equal').
However, the prop name used to control the width or width of child components is inconsistent.

There are three types of components to consider that implement some type of width prop.
### Group Components

The popular vote here was to use `widths`.  `Card.Group` `itemsPerRow` was used as an experiment to test it's intuitiveness over `widths`.

``` jsx
<Card.Group itemsPerRow='2' />    //=> two cards
<Statistic.Group widths='2' />    //=> two statistics
<Button.Group ___='' />          // PR in progress
<Field.Group ___='' />           // PR in progress
<Step.Group ___='' />            // Overlooked, needs PR
```
### Parent Components

`Form` supports `equal width` only.  `Grid.Column` parents use `columns` as it is most intuitive.

``` jsx
<Form widths='equal' />     //=> equal width form
<Grid columns='2' />        //=> two column grid
<Row columns='2' />         //=> two column row
```
### Individual Components

Individually sized components seem to work well using `width`.  This is clear and consistent.

``` jsx
<Column width='2' />    //=> two wide column
<Field width='2' />     //=> two wide field
```

> Heads up, close call API conflict! `<Image width='2' />` is applied to the img tag `<img width="2" />`.  `ImageGroup` however has no concept of column widths at present.
# Brainstorm & Analysis

I propose the individual component `width` prop remain as is.  This RFC considers only parent and group width props.

Here are some opinionated but flexible API guidelines:
1. **Intuitive** - Users ought to be able to immediately understand the API without explanation
2. **Consistency** - Users ought to be able to infer the API of all components after learning only a few
3. **SUI Parity** - Departures from SUI conventions should not be taken lightly
4. **Concise** - No one wants to use an obtuse API
### Plural Names

In use on Parents, not Groups. Here's what it would look like for all components:

``` jsx
<Button.Group buttons='2' />
<Card.Group cards='2' />
<Field.Group fields='2' />
<Statistic.Group statistics='2' />

<Grid columns='2' />
<Row columns='equal' />
<Form fields='equal' />

<Column width='2' />
<Field width='2' />
```
1. **Intuitive?**
   - **Groups** - _Yes_ - Though, you could argue it may imply "total number" when it is actually "number per row".
[...]

PR #2361, opened 2017-12-06
feat(Button): add `buttons` shorthand to ButtonGroup

### LINKED ISSUE #2310: RFC: Button.Group shorthand to render an array of buttons with a 'buttons' prop
Currently it doesn't look like the `Button.Group` component has a way to render an array of buttons in a similar way that a `Menu` or `List` component will allow you to set an `items` array to render child components.

I did not see anywhere upon searching issue history where it was discussed if this was a feature we should have or not. Maybe @levithomason can shed some light on this... and I am not sure if shorthand props like this are sticking around or not.

### Proposal

```
const buttonList = [
  {
    key: 'button-a',
    content: 'Button A',
    icon: 'user',
  },
  {
    key: 'button-b',
    content: 'Button B',
    icon: 'close',
  },
  {
    key: 'button-c',
    content: 'Button C',
    icon: 'trash',
  },
]

<Button.Group buttons={buttonList} />
```
This would render a `Button.Group` with three `Button` components.

layershifter:
Fixes #2310.

This PRs adds new shorthand and updates existing docs to latest style.


test/specs/elements/Button/ButtonGroup-test.js
```diff
@@ -24,8 +26,6 @@ describe('ButtonGroup', () => {
   common.propKeyOnlyToClassName(ButtonGroup, 'negative')
   common.propKeyOnlyToClassName(ButtonGroup, 'positive')
   common.propKeyOnlyToClassName(ButtonGroup, 'primary')
-  common.propKeyOnlyToClassName(ButtonGroup, 'primary')
-  common.propKeyOnlyToClassName(ButtonGroup, 'secondary')
```
layershifter, line 28: Duplicates 😺

test/specs/elements/Button/ButtonGroup-test.js
```diff
@@ -34,4 +34,15 @@ describe('ButtonGroup', () => {
 
   common.propValueOnlyToClassName(ButtonGroup, 'color', SUI.COLORS)
   common.propValueOnlyToClassName(ButtonGroup, 'size', SUI.SIZES)
+
+  describe('buttons', () => {
```
layershifter, line 38: @levithomason Should we go with `buttons` or `items` prop?

test/specs/elements/Button/ButtonGroup-test.js
levithomason: 😱

test/specs/elements/Button/ButtonGroup-test.js
levithomason: I think `buttons` for consistency.  The name should follow the sub-component or child components in this case:

```jsx
<Breadcrumb sections={...} />
<Feed events={...} />
<Modal actions={...} />
<Button.Group buttons={...} />
```

The exception is `<Card.Group items={...} itemsPerRow={...} />` which was an early exploratory API.  However, this will be replaced with standardized width props, #406.  It should be: `<Card.Group cards={...} widths={...} />`.

src/elements/Button/ButtonGroup.js
```diff
@@ -93,6 +104,9 @@ ButtonGroup.propTypes = {
   /** Groups can be less pronounced. */
   basic: PropTypes.bool,
 
+  /** Shorthand array of props for Button. */
```
levithomason: This doesn't necessarily have to be `props`, it can be any shorthand value.

levithomason:
LGTM, I've made one minor update to the doc block for `buttons`.  This is ready to merge once CI passes.

levithomason:

PR #2365, opened 2017-12-07
docs(Step): fix display of an icon in StepExampleEvenlyDivided

jeroenvervaeke:
'Truck' icon wasn't showing in the step docs due to a typo.
Somebody accidentally used the 'icon' attribute instead of the 'name' attribute.


jeroenvervaeke:
The failing test:
    onHide
      ✖ is called with (null, data) when exiting transition finished
Has nothing to do with my changes

layershifter (approved):
@jeroenvervaeke Thanks for the contribution 👍 You're right, the failed test doesn't have nothing in common with your changes.

I've retriggered build, now it's okay 😸

PR #2375, opened 2017-12-12
fix(Dropdown): UX behavior of selectedItems and scroll for filtered options.

### LINKED ISSUE #1485: Dropdown: Search + Mouse Selection

### LINKED ISSUE #2080: Dropdown: clicking outside vs. closing behavior
### ~~Issue 1~~
- **Steps**: Click on the dropdown then outside.
- **Expected Result**: No option selected. Works if we close instead of clicking outside.
- **Actual Result**: First option selected.

### Issue 2
- **Steps**: Select `Item B` then search for `Item D` (no results). Close dropdown (do not click outside).
- **Expected Result**: `Item B` displayed.
- **Actual Result**: `Item D` displayed. `Item B` is displayed back if we click outside instead.

### ~~Issue 3~~
- **Steps**: Select `Item B` then search for `Item D` (no results). Click outside, open the dropdown and click outside once more.
- **Expected Result**: `Item B` selected.
- **Actual Result**: `Item A` selected.

### Version
0.73.0

### Testcases
[https://codepen.io/anon/pen/JrdwxW](https://codepen.io/anon/pen/JrdwxW)

### LINKED ISSUE #2336: Dropdown: Invalid activeIndex after item selection in search Dropdown
**Steps to Reproduce**
1. Set the dropdown with search and enough items (e.g. 50 items) to make it scrollable.
2. Type in something in order to get one of the bottom items.
3. Use arrow-down/up to move between filtered items. 
4. Press Enter.
(
Or
3. choose an item from end of the list by mouse click
)

**Expected**
The Dropdown should select the item in the text box and selected item becomes active item.

**Result**
The Dropdown does not do this. Instead the search text stays as it was. If you clear the text from the search box, then the selected item will not be the active item. This is not the same behaviour as the original Semantic-UI with JQuery. Users are annoyed by this.

**Testcase**
https://codepen.io/kourosz/pen/LOQJBE

This is very critical for us as we are migrating from Semantic-UI/jQuery to React in a very important project.

francoiscote:
Here is an attempt to fix #1485, fix #2336 and fix #2080. Also, this would make the PR #1510 irrelevant now.

1. **fix `selectedIndex` of filtered options lists**
So, as mentioned in the other bugs, the problem comes from the fact that Dropdown is setting is`selectedIndex` using an index from a small filtered list of options when using search. I don't think it is a problem because it represents the actual state of the list at that time. Instead, I think the fix should be to set the `selectedIndex` again when the full list is shown (on open). This is what I did on line 1103

This PR also brings two small related fixes.

2. **removed `scrollSelectedItemIntoView()` call in open handler**: this does not respect the react component lifecycles and could generate bugs. Also, this is already taken care of by a check on the `state.open` in `componentDidMount()`

3. **scrollSelectedItemIntoView when searchQuery has changed**: This has been added to `componentDidUpdate` and fixes a bug where the scroll position would not get update when typing into search input. 

This is my first PR on this project and I plan to contribute in the next few weeks. Comments and change request are appreciated. Thanks! 🚀


francoiscote:
So I see that tests failed. I wanted to get your opinion first before starting to fix them.
- **Dropdown : search : filters the items based on custom search function**: I will have a look at this one.
- **Dropdown : search : sets the selected item to the first search result**: This behavior is different now. I could update the tests

levithomason:
Hi there!  Thanks for contributing and sorry for the delay.  I had a very long and crazy holiday season but I'm getting back to the grind now :)

The code and explanations look correct.  I'll try to review this one soon.

>Instead, I think the fix should be to set the selectedIndex again when the full list is shown (on open). This is what I did on line 1103

Yes, thanks!

>Dropdown : search : sets the selected item to the first search result: This behavior is different now. I could update the tests

This should be ensuring that changing the search query sets the selected index to 0.  We want to retain that behavior.

mlmassey:
Hi guys, I created PR #2523 to address the state of the selected item index when doing searches.  It does not address the scrolling items into view though.

levithomason:
I looked at the proposed fixes here and in other PRs.  I've updated this PR to fix tests.  I also changed the logic a bit as to when/why we set the selected index and scroll items into view.

I removed all the scattered calls to set selected index.  We were trying to identify all cases where we needed to set it, and call it there.  However, even a value prop change would cause a state change and therefore require setting the selected index again.

Rather than trace all these down, I've removed all the calls and just set selected index in response to state changes in value.  I've done similar with the scrolling into view.  There is one additional case for each, resetting the selected index to 0 on query change and scrolling items into view on open.  Both of those are also implemented.

This works with all current tests.  I've also ported and simplified the test from #1510 that asserts the index is properly set when dealing with clicks after a search query.

Finally, I've added a memoization optimization to `getMenuItems` to prevent unnecessary duplicate calls to the `search` function.  This was required as setting state reactively in response to other state changes causes 2 calls to that function in one test.

😅 Hope I don't have to go down that hole again in the next year!  Thanks for all the help everyone.

levithomason:
Alright, I've added a failing test for the bug in #2080.  The changes made thus far do not solve this one.

The issue is, when a search yields no results the selected index should revert to the currently active item if there is one.  Instead, it is currently disregarding the active item's index and keeping 0 as the selected index.

The test is `it('retains the selected item after searching with no results')`.  Once someone gets that to pass, we can merge this.

src/modules/Dropdown/Dropdown.js
```diff
@@ -811,7 +818,20 @@ export default class Dropdown extends Component {
...
+
+        if (lastValue === value && lastOptions === options && lastSearch === search) {
+          filteredOptions = this.getMenuOptions.lastReturn
+        } else {
+          this.getMenuOptions.lastValue = value
+          this.getMenuOptions.lastOptions = options
+          this.getMenuOptions.lastSearch = search
+
+          filteredOptions = search(filteredOptions, searchQuery)
+        }
```
mlmassey: This seems unwise to add more state management, especially outside of this.state.  I thought my PR addressed this issue quite well by ensuring this.state stayed in sync with prop changes.  Did you have a chance to review my PR?

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -1980,6 +1980,66 @@ describe('Dropdown', () => {
...
+
+      wrapper.find('DropdownItem').at(0).should.have.prop('active', false)
+      wrapper.find('DropdownItem').at(0).should.have.prop('selected', false)
+
+      // search for a non-existent item, triggering the not found message
+      const search = wrapper.find('input.search')
+      search.simulate('change', { target: { value: '__nonExistingSearchQuery__' } })
+
+      // click outside, close the dropdown
+      domEvent.click(document)
```
mlmassey, line 2185: This line isn't working really.  I noted that once you added describe.only, the call domEvent.click(document) isn't really working, and would also cause a failure in the test `closes on click outside` because it uses the same call.  If you use `document.body.click()` (like the test `blurs the Dropdown node on close by clicking outside component`) or `domEvent.click(document.body)` then things start working.  What this causes though is a call to `closeOnDocumentClick`, which causes the dropdown menu to close, but the focus is still on the input and the input search text is never cleared.   In this case, this is correct behavior, since the menu has been closed, but focus is still on the input, so the invalid search text should not be removed.  What the test should be doing is calling the `handleBlur` function which is what you're trying to test with this test case.  In summary,
1. Your checks below with the wrapper.find() calls aren't working because the dropdown still has the search text in the input and is showing the No results message.
2. Need a different way to trigger a blur event instead of document.body.click().  This applies to this test case and to the `blurs the Dropdown node on close by clicking outside component` case.  May need to include another button element to throw the focus to cause a blur to happen.  
3. Actually, I believe the dropdown is working well with your changes, its just the test case isn't correctly testing it

mlmassey (changes requested):

levithomason:
As @mlmassey has pointed out, there are some issues with the tests here.  I believe I'm on to resolving this in #1542.  There seem to be some issues with shared scope between tests.

pierrebiver:
Hey guys,
I have the same issue than  #2624 ? is there a release to fix it available soon?

Thanks :)

MarianSulgan:
Hello, any updates on this one? Can we expect fix anytime soon?

layershifter:
I merged `master` to this branch, merge conflicts are crazy 😨 

It's great that we have all issues linked to this PR, I will check what actually exists still in `master`. @francoiscote if you can help to restore your changes it will be really appreciated 🙏

layershifter:
## Issues verification

### #1485 ✅ 
Is no longer present in the latest `master` https://github.com/Semantic-Org/Semantic-UI-React/issues/1485#issuecomment-673377496. @francoiscote if you still can reproduce it, please let me know.

### #2336 ✅ 

Was also fixed, https://github.com/Semantic-Org/Semantic-UI-React/issues/2336#issuecomment-665705649

### #2080 ⚔️ 

The issue #2 which was qualified as a bug is still present, https://github.com/Semantic-Org/Semantic-UI-React/issues/2080#issuecomment-675339758

PR #2380, opened 2017-12-13
fix(Portal): onClose and onOpen don't get called on prop changes

### LINKED ISSUE #2381: Popup is incorrectly positioned if prop open is set to true on initial render
Popup onOpen and onClose handlers aren't getting called correctly on open prop, so it doesn't render properly

### Steps
Create popup component with open={true} and a trigger

### Expected Result
Expect the popup to render near the trigger

### Actual Result
Popup is rendered top right corner of the screen

### Version
0.77.1

### Testcase
https://codesandbox.io/s/py63r935qq

fracmak:
So onClose and onOpen only ever get called by the internal event handlers managing the portal. That's great unless you want to control the opening and closing of a portal using the open autocontrolled prop. While the opening and closing do work just fine, the onClose and onOpen callbacks don't get called which causes really subtle bugs like the Popup component not rendering in the right place. 

This fixes that by overriding the componentWillReceiveProps and detecting when the open prop changes and firing the appropriate callback. 

Fixes #2381

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -147,6 +147,18 @@ class Portal extends Component {
     this.renderPortal()
   }
 
+  componentWillReceiveProps(nextProps) {
+    super.componentWillReceiveProps(nextProps)
+    if (nextProps.open !== this.props.open) {
+      if (nextProps.open && nextProps.onOpen) {
+        nextProps.onOpen(null, nextProps)
+      }
+      if (!nextProps.open && nextProps.onClose) {
+        nextProps.onClose(null, nextProps)
+      }
+    }
+  }
+
   componentDidUpdate(prevProps, prevState) {
     debug('componentDidUpdate()')
     // NOTE: Ideally the portal rendering would happen in the render() function
--- a/test/specs/addons/Portal/Portal-test.js
+++ b/test/specs/addons/Portal/Portal-test.js
@@ -85,6 +85,14 @@ describe('Portal', () => {
       document.body.lastElementChild.should.equal(instance.rootNode)
       document.body.childElementCount.should.equal(1)
     })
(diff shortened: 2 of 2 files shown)


fracmak:
hmm, a better solution might be to move this code up to the popup, since that's where the bug actually is.

fracmak:
though the popup doesn't really know where the trigger is located in the dom, it only knows because of the browser event, the portal renders the trigger. I'm open to suggestions on where to go to fix this

levithomason:
Note, callbacks are only called on user action but not on programmatic action.  This is also true in vanilla React and HTML.

Setting the `value` of an HTML `<input>` via JS will not fire the `onchange` event.  Similarly changing the `value` prop of a React `<input/>` will not call the `onChange` handler.  Therefore, SUIR does not call callbacks when corresponding props change.

Example sandbox showing this behavior: https://codesandbox.io/s/ym3pnnx711

Apart from consistency with existing paradigms, there are two more reasons we don't call callbacks on prop changes:
1. If this was done it could create infinite loops since change handlers are often responsible for updating the prop that changed.  
2. If the component is controlled, then something is responsible for updating the prop and that thing, therefore, doesn't need a callback.  It can do the work of the callback at the same time it does the work of updating the prop.  There is no need to tell an owner that it just did something.

***

All said, there may be a bug here but it needs to be solved without calling callbacks on prop changes.

fracmak:
Ya, I was realizing that when I saw the other tests failed. I've been looking at the other portals to see how they handle this scenario. You can see the bug in it's most simplistic form here: https://codesandbox.io/s/py63r935qq

Not sure how to fix it though without grabbing a direct reference to the trigger to get it's location

fracmak:
closing for now while I ponder a better way to fix this

PR #2383, opened 2017-12-14
feat(Progress): Add value option, to just show the value in the bar

UnbrandedTech:
This PR adds ability to pass `value` to `Progress`:

```jsx
  <Progress progress='value' value={35} />
```

### DIFF
--- a/docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValue.js
+++ b/docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValue.js
@@ -1,0 +1,8 @@
+import React from 'react'
+import { Progress } from 'semantic-ui-react'
+
+const ProgressExampleProgressValue = () => (
+  <Progress progress='value' value={35} />
+)
+
+export default ProgressExampleProgressValue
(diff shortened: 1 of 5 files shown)


UnbrandedTech:
Hey @layershifter or @levithomason, I think this is good to go. Let me know what you think.
This is what im trying to accomplish btw. I'll deal with the css, but the value inside is what I need.
screen shot 2017-12-14 at 9 32 47 am

levithomason:
Though this is an extension of the SUI Progress settings, I think it makes sense.  If you could add some tests ~~, typings,~~ and docs to this we can merge it.

UnbrandedTech:
I'm not sure why that test it failing....

layershifter:
@levithomason I'm pushed some changes to make this working, however I think that it will be better if we will introduce `Progress.Progress` component and allow to use it as shorthand.

UnbrandedTech:
Any update on this?

levithomason:
This looks good to me.

@layershifter what do you mean about `Progress.Progress`, can you give some shorthand examples?   Maybe we do this in another PR?

levithomason:

PR #2384, opened 2017-12-14
fix(Portal): click inside detection

### LINKED ISSUE #1831: Popup: Popup in a Popup closes when you interact with the inner Popup
### Steps

- Create a `Popup` in a `Popup` both set to trigger `on` `click`
- When you click in the inner `Popup`, it closes the outer `Popup`.

### Expected Result
- Expect to interact with the inner `Popup` without closing outer `Popup`

### Actual Result
- Outer popup should remain open

### Version
0.70

### Testcase
https://codepen.io/kungfoo/pen/JJZXpQ

levithomason:
Fixes #1831 

### Problem

We've had a cryptic bug for a while where clicks inside of a popup cause it to close.  This is because clicks inside are detected by checking if the popup `.contains()` the clicked `e.target` node.

This fails if the result of the click also removes the node.  Example, a popup containing a Dropdown with removable Labels.  When the `x` is clicked on the label, the label is removed from the Dropdown.  The Popup then checks to see if the label is within the Popup node.  It has been removed, so the Popup does not `.contain()` the `e.target`.  The Popup then assumes the click originated outside of the Popup (body click) and closes.

**Test case**: https://codesandbox.io/s/53ykvz7rk4

### Solution

Instead of checking for nodes containing other nodes, we now check that the coordinates of the click event are within the bounds of the Portal.  This way, we're shielded from DOM updates within the Portal as a result of click handlers.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -5,6 +5,7 @@ import ReactDOM from 'react-dom'
 
 import {
   AutoControlledComponent as Component,
+  doesNodeContainClick,
   eventStack,
   isBrowser,
   keyboardKey,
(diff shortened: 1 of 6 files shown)


ghost:
Can you please review my PR (https://github.com/Semantic-Org/Semantic-UI-React/pull/2397) on the fix with tiny changes before submitting this?

Dealing with ClientX/ClientY is bad when another solutions is available

levithomason:
@Attrash-Islam I reviewed the other PR and it looks like there's a browser compatibility issue with that event property and method.  That would have been a nice approach!

One shortcoming here may be that this implementation doesn't consider whether or not the node is in the topmost layer.  I can imagine a rare edge case where a click on an open Modal could be considered "inside" of an underlying Portal if they were aligned.  That said, this is probably a very rare case if an issue at all.

levithomason:

PR #2389, opened 2017-12-18
fix(Popup) popup appearing in wrong position when open prop is set

### LINKED ISSUE #1065: Popup appears in wrong spot when set manually before trigger is activated.

fracmak:
Fixes #1065

When popup prop is used to open the popup, the popup isn't positioned correctly. This fixes that by saving the reference to the trigger and using that as a fallback when this.coords isn't available.


levithomason:
When I paste the example from the bug report and click the `Turn on popup` button, it still appears in the top right of the screen.

```jsx
import React from 'react'
import { Container, Button, Popup, Icon } from 'semantic-ui-react'

class PopupExample extends React.Component {
  state = {}
  render() {
    return (
      <Container>
        <Button onClick={() => this.setState({ open: !this.state.open })}>Turn on popup</Button>
        <div style={{ marginTop: '25px' }}>
          Here is the anchor
          <Popup
            open={this.state.open}
            on='click'
            onOpen={() => this.setState({ open: true })}
            onClose={() => this.setState({ open: false })}
            trigger={<Icon name='settings' />}
          >
            Here I am
          </Popup>
        </div>
      </Container>
    )
  }
}

export default PopupExample
```

I think the solution here is to use `requestAnimationFrame` to position the popup.  The trigger is always available, it is just that the initial state isn't set when the popup is opened via the `open` prop changing.

fracmak:
sorry, didn't realized there was a comment on the PR. Could you give me some more details on what you mean by using requestAnimationFrame to position the popup?

levithomason:
Sure, have a look at the Modal's use of RAF. Also the Sticky component. The idea is to animate the position so that positioning is handled each frame. I think it would solve this issue as well. The core issue seemed to be that the positioning code was not properly called on mount, however, by animating the component's position it will be called constantly.

levithomason:
Excuse brevity and typos, on mobile!

oskarhane:
+1 on this.

This looks very weird: 
oskarhane-mbpt_2018-04-03_at_13 51 00

What's stopping this from being merged?

levithomason:
@oskarhane the issues and requests are spelled out above. If you are able to fork this branch and make the necessary fixes we would certainly merge it and appreciate it.

cdaringe:
#2775 loosely lands the extra few lines requested

layershifter:
#2775 was merged, closing for housekeeping :+1:

PR #2390, opened 2017-12-19
fix(TransionablePortal): fix state handling

### LINKED ISSUE #2382: TransitionablePortal: doesn't close on closeOnDocumentClick
hi,
It works with Portal but not with Transitionable..
Note that the Portal controlled demo here https://react.semantic-ui.com/addons/portal doesn't work too.

Version 0.76.0

layershifter:
Fixes #2382.

### DIFF
--- a/src/addons/TransitionablePortal/TransitionablePortal.js
+++ b/src/addons/TransitionablePortal/TransitionablePortal.js
@@ -88,9 +88,7 @@ export default class TransitionablePortal extends Component {
   componentWillReceiveProps({ open }) {
     debug('componentWillReceiveProps()', { open })
 
-    // Heads up! We apply `open` prop only when it's defined, otherwise it will break
-    // autocontrolled Portal
-    if (!_.isNil(open)) this.setState({ portalOpen: open })
+    this.setState({ portalOpen: open })
   }
 
   // ----------------------------------------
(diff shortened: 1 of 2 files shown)


src/addons/TransitionablePortal/TransitionablePortal.js
```diff
@@ -99,13 +97,8 @@ export default class TransitionablePortal extends Component {
...
   handlePortalClose = () => {
     debug('handlePortalClose()')
-    const { open } = this.props
-    const { portalOpen } = this.state
 
-    // Heads up! We simply call `onClose` when component is controlled with `open` prop.
-    // But, when it's autocontrolled we should change the state to opposite to keep the transition
-    // queue
-    if (_.isNil(open)) this.setState({ portalOpen: !portalOpen })
+    this.setState({ portalOpen: false })
```
layershifter, line 101: I rethought the state handling there and as I see, it was overcomplicated. I've simplified it and it looks that everything is correct.

src/addons/TransitionablePortal/TransitionablePortal.js
```diff
@@ -99,13 +97,8 @@ export default class TransitionablePortal extends Component {
 
   handlePortalClose = () => {
     debug('handlePortalClose()')
-    const { open } = this.props
-    const { portalOpen } = this.state
 
-    // Heads up! We simply call `onClose` when component is controlled with `open` prop.
```
layershifter, line 105: Cruft, we don't call `onClose` and it seems that condition below doesn't make sence.

levithomason:

Issue #2370, opened 2017-12-09
Checkbox: onChange called multiple times in a row


MrSauceman:
I found that the Checkbox element calls its `onChange` event several times each time you check or uncheck the box.  It has something to do with the label that is generated alongside the checkbox because when I manually remove the label from the html, the call is only made once.

https://codesandbox.io/s/34vrxvlwqp

That link reproduces the issue if you check the console.  Notice how the number of times `onChange` is called varies for each checkbox because of its props.  For instance, adding an id prop to the Checkbox and clicking its label causes it to call its `onChange` event 3 times, while clicking the actual checkbox causes `onChange` to be called twice.  I believe this has something to do with the `<label for="id">Label</label>`.  The number of times called varies based on whether you click the actual checkbox or the label.

Seems very inconsistent and you can see that checkbox 3 in the demo cannot even set its value correctly when you click its label because `onChange` is called 3 times.

HariSoni87:
We had a same problem and we have resolved it using a wrapper component. 

const CheckBoxComponent = props => {
  const { onChange, id, ...rest } = props
  return (<Checkbox {...rest} onChange={(e, data) => onChange(e, { ...data, id })} />)
};

jkudish:
Having the same issue with 0.77.1 -- not a problem in  0.76.0

yoshatest:
I've got a similar problem on Radio (no label). Clicking <Radio toggle={true} onChange=... > fires events twice 1) checked=true, 2) checked=false, result in no toggle at all. Reverting to 0.76.0 fixes the problem.

ghost:
Added a PR, Hopefully it solves the issue :+1:

ericbock:
I'm also seeing the change handler being called twice. Thinking that it might be related to the component binding both [onChange and onClick](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Checkbox/Checkbox.js#L227) to the same handler, I tested focusing the checkbox and toggling with the spacebar. That seems to work; my change handler is only called once per toggle.

levithomason:
Fixed in #2392

PR #2392, opened 2017-12-20
fix(Checkbox): fix event handling with `id` prop

ghost:
-  onChange triggered twice when passing id attribute (Related to Checkbox3 in CodeSandbox reproduction)

Details: The Checkbox Component was rendering a label htmlFor attribute which will bind between
the label and the input tag when being clicked. The problem is that the click (change) handler
is NOT attached on the input tag but it's attached on the parent <ElementType ... /> component.

Below this ElementType we've both the input tag and the label tag (as element children), and once we click on the label
the ElementType element will get the click (change) event thanks to Event Bubbling in JavaScript,
which therefor will trigger the checkbox state change, but if htmlFor={id} is supplied then clicking on the
label tag will trigger the input tag and the ElementType, and input tag click will be bubbled up again
into ElementType which will make it trigger the onChange event twice.

Since we've the correct behavior of the label being linked to the input because of attaching the event on
the parent element then the usage of htmlFor is buggy and already implemented.

### DIFF
--- a/src/modules/Checkbox/Checkbox.d.ts
+++ b/src/modules/Checkbox/Checkbox.d.ts
@@ -25,6 +25,9 @@ export interface CheckboxProps {
   /** Removes padding for a label. Auto applied when there is no label. */
   fitted?: boolean;
 
+  /** A unique identifier. */
+  id?: number | string;
+
   /** Whether or not checkbox is indeterminate. */
   indeterminate?: boolean;
 
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -218,7 +218,6 @@ export default class Checkbox extends Component {
     const unhandled = getUnhandledProps(Checkbox, this.props)
     const ElementType = getElementType(Checkbox, this.props)
     const [htmlInputProps, rest] = partitionHTMLProps(unhandled, { htmlProps: htmlInputAttrs })
-    const id = _.get(htmlInputProps, 'id')
 
     return (
       <ElementType
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -46,6 +46,12 @@ export default class Checkbox extends Component {
     /** Removes padding for a label. Auto applied when there is no label. */
     fitted: PropTypes.bool,
 
+    /** A unique identifier. */
+    id: PropTypes.oneOfType([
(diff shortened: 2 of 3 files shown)


ghost:
Related to: https://github.com/Semantic-Org/Semantic-UI-React/issues/2370

layershifter:
This PR reverts removes changes, but it's not a solution. Please see #2274 for content of `id` prop.

I'm sure that we will not merge this PR in current condition.

ghost:
Hi @layershifter ,
Thanks for the response. 

I don't understand what the htmlFor unless to link between label clicks to inputs? The situation you have which the click in even attached to the input parent make the htmlFor useless.

Can you explain to me why htmlFor should be added as a prop in this situation? The goal of htmlFor is to do one thing and your implementation already doing it correctly

layershifter:
Linked issue has a description, we need `id` and `htmlFor` for accessibility purposes. [SO answer](https://stackoverflow.com/a/18432715/6488546) has more details about this.

In fact, we need a different event handlers for container element and input.

layershifter:
@levithomason I've pushed some changes there. Can you you take attention to them?

ghost:
@layershifter Since you're attaching a 2 event handlers, why not keep just one handler at the parent (as it was before) but to extend it using event.target to identify where the click event came from?

Also, binding the event on just the input (not the parent container) may solve the issue since the click on the label will also trigger the input thanks to "for" html attribute

layershifter:
> why not keep just one handler at the parent

There is different behaviour of `label` and `input` in cases when we have `id`, I also don't like multiple event handlers, possible @levithomason has a better idea.

levithomason:
For clarification, the input is never directly clicked by the user as it is hidden.  The box next to the label is actuall a psuedo `label::before` element.  That said, the `htmlFor` attribute is passing the click event along to the hidden input when there is an id present.  Thus, we're getting two clicks when there is an id.

The other event coming through here is due to the `onChange` prop on the root element. We actually don't need that at all.  Both clicks and spacebar will trigger the `onClick` prop.  Our click handler already calls back props for `onClick` and `onChange`, since, the only way to change a checkbox is to click it.

**EDIT** The solution previously posted was incorrect, will merge the PR as is.  Thanks!

levithomason:

NullVoxPopuli:
onClick seems to be called multiple times on the checkbox.
I'm using version 0.84.0

Issue #2386, opened 2017-12-17
Docs: Home Page template is not responsive


azias:
I am not able to say if it is "just" a problem in the example or a problem inside the implementation of Semantic UI React itself.

I thought having read that the menu of this page is not supposed to be responsive in Semantic-UI-React so I do not point this here (while it is a difference with the original one) but there is a vertical grey overlay that seems to be a bug.

### Steps
- Go to https://react.semantic-ui.com/layouts/homepage
- in chrome : F12, then Ctrl+Shift M, then choose "Nexus 5X" in the dropdown (or you can use a smartphone)

### Expected Result
- I suppose we must obtain the same result as the one at https://semantic-ui.com/examples/homepage.html

### Actual Result
-  there is a vertical grey overlay on the right that prevents to acess log in and sign  up buttons

### Version
0.77.1

layershifter:
Responsive menu was not implimented, I will make PR tomorrow.

Issue #1408, opened 2017-03-01
feat(Sidebar): Full page sidebar


UnbrandedTech:
Posting an issue for feed back.

Given the idea behind [this feature](https://github.com/Semantic-Org/Semantic-UI-React/pull/905#pullrequestreview-9709670), I can only see on small issue.

This is my use case.
```
<Sidebar.Pushable>
  <Sidebar as={Menu} visible vertical inverted >
    <Menu.Item name='home'>
      <Icon name='home' />
      Home
    </Menu.Item>
    <Menu.Item name='gamepad'>
      <Icon name='gamepad' />
      Games
    </Menu.Item>
    <Menu.Item name='camera'>
      <Icon name='camera' />
      Channels
    </Menu.Item>
  </Sidebar>
  <Sidebar.Pusher >
    //maybe another menu here for the top nav
    <Segment basic>
      {React.Children.toArray(this.props.children)}
    </Segment>
  </Sidebar.Pusher>
</Sidebar.Pushable>
```

In @levithomason's comment he said add a page prop to the `Sidebar` component. Which would be fine but what would happen if someone did this?

```
<Sidebar.Pushable>
  <Sidebar as={Menu} visible vertical inverted page>
    ...
  </Sidebar>
  <Sidebar.Pusher >
   ...
  </Sidebar.Pusher>
</Sidebar.Pushable>
```

Do we add a pushable to both the pushable div and the body? Should we throw a warning?

levithomason:
>In @levithomason's comment he said add a page prop to the Sidebar component. Which would be fine but what would happen if someone did this?
>
> [ ...shows page sidebar in pushable component ]
>
>Do we add a pushable to both the pushable div and the body? Should we throw a warning?

I think there is only so much we can do to prevent misuse.

However, upon reflection t is probably a good idea to keep the full page sidebar API usage as close to the other usage as possible as well.  Proposals welcome.

codeaid:
Any update on this? Would really like to use this library in my project but hard to get started without a decent "holy grail" full-page layout example.

rafalradomski:
@levithomason Same issue, how to force sidebar to cover the whole page?

levithomason:
Would love a PR on this, there has not been a business need to date on my end to fund the time to complete it.

Regarding above comments, you can find layout examples here:
https://react.semantic-ui.com/layouts

[...]

PR #2396, opened 2017-12-24
docs(layouts): make HomepageLayout responsive

layershifter:
Fixes: #2386

### DIFF
--- a/docs/app/Components/LayoutsLayout.js
+++ b/docs/app/Components/LayoutsLayout.js
@@ -63,9 +63,9 @@ export default class LayoutsLayout extends Component {
           <Button
             as={NavLink}
             to={`${repoURL}/blob/master/docs/app/Layouts/${filename}`}
-            color='secondary'
             icon='github'
             content='Source'
+            secondary
             target='_blank'
           />
         </div>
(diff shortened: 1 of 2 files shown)


layershifter:
@brianespinosa Your feedback is much needed there

brianespinosa:
@layershifter just getting back from the holidays. I just assigned this to myself and will take a deeper look later this afternoon. 👍

dongcai:
@layershifter & @brianespinosa I think this solution breaks the Server Side Rendering. I actually tried to do the similar thing at: https://github.com/dongcai/react-redux-semantic-ui/blob/master/src/components/Navigation/Navigation.js
And found that the code like `<responsive>{children}</responsive>` will break the SSR, which I would like `{chidren}` to be rendered by.  Any ideas?

layershifter:
@dongcai please make a minimal repro repo and open new issue.

brianespinosa:
@layershifter Sorry... my workload exploded this month. I haven't even been on github in two weeks. I will try to get some eyes on this by Monday. Apologies for the delay.

brianespinosa:
Thanks @layershifter 

Additionally... I meant to bring this up here and with @levithomason , it might be good to update the layout of the docs in general to make the side menu available to toggle on/off. Maybe just update the style of the docs in general in preparation for 1.0.

But I think this looks good for now. 👍

levithomason:
Definitely agree.  We should use the actual Sidebar component.  I think it needs #1408 before that can be done properly.

levithomason:

Issue #2388, opened 2017-12-18
Popup closes when clicking delete icon from dropdown multiple selection


fracmak:
### Steps
1: Given a popup
2: Inside the popup is a Dropdown component with the multiple property set
3: Select an item inside the dropdown
4: Click the "X" to delete it 

### Expected Result
Popup remains open

### Actual Result
Popup closes

### Version
0.77.1

### Testcase
https://codesandbox.io/s/lxp9kpxkzq

ghost:
Found the bug.. Preparing PR :smile: 

The bug is related to Portal logic, which means that the bug maybe produce in other than Select components under a specific circumstances that I'll explain in the commit message

ghost:
Submitted the PR, waiting for review

ak1394:
I've just run into the same issue. Would like to see it merged soon.

layershifter:
In fact, it's a duplicate of #1831 and will be fixed by #2384.

PR #2397, opened 2017-12-25
Fix Issue #2388

ghost:
Details: The issue was related to portals and their management of outside-clicks, where Portal
support dismissing the Portal children component once a click outside the Portal is recongnize. The
issue can be reproduce on any element (Not just the Select component) by clicking on element that will be deleted from the DOM, In this case the click on the icon-delete button (x) was doing some internal logic of the Select component and then the icon-delete will be removed from the DOM, after that the click will be bubbled to the Portal and the portal will decide if the click is in/out the Portal to handle auto-close feature.

The Portal will search for icon-delete and see if it in the this.rootNode of the Portal, and because it has already deleted the icon from the DOM tree it will not find it in the rootNode and it will decide that it's was a click outside the Portal! Also, the event.target won't keep a full reference to the top body element in case we want to search the icon-delete parents! so the way is to use event.path / event.composedPath to have the full snapshot of the bubbled event

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -11,6 +11,7 @@ import {
   makeDebugger,
   META,
 } from '../../lib'
+import isNodeInEventBubblePath from '../../lib/isNodeInEventBubblePath'
 import Ref from '../Ref'
 
 const debug = makeDebugger('portal')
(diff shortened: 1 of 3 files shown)


Attrash-Islam:
I'll check tommorrow morning why the build failed, because locally it passes

layershifter:
Does this PR propose better changes that #2384?

ghost:
@layershifter This PR is better than the one referenced absolutely. I don't know why to propose such a difficult implementation with ClientX and ClientY when you can fix it with a tiny fix like the proposed here in this PR!

Also, Compare the doesNodeContainClick.js code with my isNodeInEventBubblePath.js code! Which is easiest to maintain? and note that the issue in Portal and not in Popup and I've already tested the case and it's working as expected (Hopefully not affecting other scenarios)

ghost:
There's a problem in FF, I'm checking more things

ghost:
The cross-browser incompatibility dropped this PR in my face :disappointed: You can amend it or close it if you think about something else

Issue #2343, opened 2017-11-27
Dropdown: prop text only allows 'string'


nopol:
don't know if this is a bug; i think its a feature request

If you use a Dropdown Element with a Custom text Node
You'll receive a

        Warning: Failed prop type: Invalid prop `text` of type `object` supplied to `Dropdown`, expected `string`.

But. The render of the text element is working as intended.
It would be great to allow Node elements in text prop to render custom content for the Dropdown Element.

farebord:
**text** type should be **any**.

nopol:
currently it's string (via Docs & Code)
perhaps it should be changed to any?

layershifter:
Duplicate #1864.

PR #2398, opened 2017-12-26
The text property also takes an element.

elongl:
The text property should also take an Element as a PropType.
If you want to bold, or change color of your text, you can't do that without using an element.


ghost:
@elongl It should be PropTypes.oneOfType not PropTypes.oneOfType**s**

layershifter:
Duplicate #2343.

PR #2394, opened 2017-12-21
fix(Modal): scrolling disabled when modal unmounts

guillaumep:
Fixes #1157, fixes #2248.

Another PR was opened for this fix, #2255. This PR is an alternative approach, possibly more generic.

At Modal mount time we check if the mount node has the scrolling class, and keep that information in a class variable (scrollingWasDefinedAtMountTime). At Modal umount, we reset the scrolling class to its original value.

I will see to add tests to the PR as well, today or next week.

Issue #1443, opened 2017-03-11
Consider publishing individual components as packages


levithomason:
A tool like https://github.com/lerna/lerna could enable us to publish individual components as standalone packages.

themre:
that would be very nice. would it be possible also to split CSS into component packages?

layershifter:
CSS is splitted by SUI team, [Table](https://github.com/Semantic-Org/UI-Table) for example. At now this repo doesn't deal with CSS at all.

themre:
ok. but should we try to split CSS into separate components? 'cause now CSS bundle is more than 500 kB.

layershifter:
If you use vanilla SUI it's already done, take a look on link that I've provided:
```js
import 'semantic-ui-table/table.css';
```

If you use styled [SUI](https://github.com/Semantic-Org/Semantic-UI) you can [define](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/semantic.less) components that you want use:

```js
import './semantic/semantic.less';
// Or
import './semantic/definitions/globals/reset.less';
import './semantic/definitions/globals/site.less';
```

I don't understand your problem there. Here is a [small how-to](https://www.artembutusov.com/webpack-semantic-ui/).

themre:
ok thanks.

levithomason:
I've marked this for the v1 release.

levithomason:
We should consider shipping the commonjs, es, and umd build targets as separate packages as well.  It would make the imports much cleaner for devs.

```js
import { Button } from 'semantic-ui-react-es'
```

Versus our current practice of exposing our dist directory structure, which is leaky and brittle:

```js
import { Button } from 'semantic-ui-react/dist/es'
```

layershifter:
Sometime ago @levithomason has already proposed this. I'm think it's a good idea for V2, some utils like `eventStack` can be moved in separate packages.

PR #2407, opened 2018-01-05
feat(MountNode): add component

### LINKED ISSUE #2248: Modal: scrolling issue when using sub-modals

layershifter:
Fixes #2248.
Replaces #2394, #2255.

### Core problem

Our `Modal` component makes too many things, the main problem is the management of `classList` on `mountNode`. It made too many problem for us, I thought about it and here is solution.

### Solution

We need a declative way for `classList` management on `mountNode`, I've took a look on [react-document-title](react-document-title) and made similar component.


gulp/tasks/docs.js
```diff
@@ -29,12 +29,12 @@ task('clean:docs', (cb) => {
 // ----------------------------------------
 
 task('build:docs:docgen', () => src([
-  `${config.paths.src()}/addons/**/*.js`,
-  `${config.paths.src()}/behaviors/**/*.js`,
-  `${config.paths.src()}/elements/**/*.js`,
-  `${config.paths.src()}/collections/**/*.js`,
-  `${config.paths.src()}/modules/**/*.js`,
-  `${config.paths.src()}/views/**/*.js`,
+  `${config.paths.src()}/addons/*/*.js`,
```
layershifter, line 32: I think that we need to store component specific functionality in its directory, this will allow to manage easier in future, see #1443.

guillaumep:
@layershifter Please consider the changes made to test/specs/modules/Modal/Modal-test.js in #2394. It fixes an invisible issue in the tests that were made visible by my proposed changes.

The issue is that the modals are never unmounted, thus modals mounted in one tests interacts with modals mounted in another test. This was provoking various seemingly unrelated modal tests to fails.

guillaumep:
@layershifter Also, your approach to fix this issue looks promising!

gulp/tasks/docs.js
levithomason: Yep, looks like we already have a flat structure any how.

src/addons/MountNode/MountNode.js
```diff
@@ -0,0 +1,55 @@
...
+
+  static _meta = {
+    name: 'MountNode',
+    type: TYPES.ADDON,
+  }
+
+  shouldComponentUpdate({className: next}) {
+    const {className: current} = this.props
+
+    return next !== current
```
levithomason: I know we've started this pattern of renaming destructured keys elsewhere but I'd prefer more explicit names like `nextClassName` and `currClassName`.

src/addons/MountNode/lib/computeClasses.js
```diff
@@ -0,0 +1,11 @@
+import _ from 'lodash/fp'
+
+const computeClasses = (components) => components && _.flow(
+  _.map('props.className'),
+  _.filter(_.isString),
+  _.map(_.split(/\s+/)),
+  _.flatten,
+  _.uniq,
+)([...components])
```
levithomason: I didn't test this, but I'm pretty sure:
- Lodash will gracefully handle null/undefined args so no need to check if `components` exists first
- Lodash will not mutate the original array so no need copy if first
- There is a `_.flatMap` that will handle mapping and flattening at the same time.
- This will currently include empty string `className`s.  However, switching the filter to an `_.identity` will filter all falsy values.

This can probably be simplified to:

```js
const computeClasses = _.flow(
  _.map('props.className'),
  _.filter(_.identity),
  _.flatMap(_.split(/\s+/)),
  _.uniq,
)
```

Which works like so:
```js
computeClasses(null)      //=> []
computeClasses(undefined) //=> []

computeClasses([
  null,
  undefined,
  {},
  { props: {} },
  { props: { className: null } },
  { props: { className: undefined } },
  { props: { className: 'foo' } },
  { props: { className: 'foo bar' } },
  { props: { className: '0' } },
  { props: { className: 'false' } },
])
// => [ 'foo', 'bar', '0', 'false' ]
```

src/modules/Modal/Modal.js
```diff
@@ -227,42 +217,39 @@ class Modal extends Component {
...
-          mountNode.classList.remove('scrolling')
-        }
       }
 
-      if (Object.keys(newState).length > 0) this.setState(newState)
+      if (scrolling) classes += ' scrolling'
     }
 
+    if (this.state.mountClasses !== classes) newState.mountClasses = classes
+    console.log(classes)
```
levithomason: Whoops

src/modules/Modal/Modal.js
```diff
@@ -227,42 +217,39 @@ class Modal extends Component {
...
+    if (Object.keys(newState).length > 0) this.setState(newState)
```
levithomason: For consistency, how about this?

```js
if (!_.isEmpty(newState)) this.setState(newState)
```

src/modules/Modal/Modal.js
```diff
@@ -304,6 +294,8 @@ class Modal extends Component {
 
     return (
       <ElementType {...rest} className={classes} style={{ marginTop, ...style }} ref={this.handleRef}>
+        <MountNode className={mountClasses} node={mountNode || document.body} />
```
levithomason: See above SSR comment.

src/addons/MountNode/lib/handleChange.js
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash'
+import computeClasses from './computeClasses'
+
+let prevClasses = []
```
levithomason: This will be shared across all `MountNode` instances.  If two MountNodes are on the page, they will need their own copy of `prevClasses` to diff against, correct?

This might belong on the actual MountNode class so it can be instance specific.

levithomason:
Oh boy, just noticed this wasn't ready for review 🤦‍♂️   Well, hopefully, the comments are of some help and not too intrusive!  Sorry about that...

layershifter:
@levithomason your feedback is awesome as always, will back there soon 👍

layershifter:
@levithomason now it's ready for review 😸

src/addons/MountNode/lib/handleClassNamesChange.js
```diff
@@ -0,0 +1,18 @@
+import _ from 'lodash'
+
+import computeClassNames from './computeClassNames'
+import computeClassNamesDifference from './computeClassNamesDifference'
+
+const prevClassNames = new Map()
```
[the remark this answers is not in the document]
layershifter: Yes, it's a planned solution. We use a DOM node as the key in Map, so we will have the same classes for the same node.

Here is a simple example:
```jsx
const div = document.createElement('div')
const Tree = () => (
  <>
   <MountNode className='foo' node={div} />
   <MountNode className='bar' node={div} />
  </>
)
```

Because we have the same Map, we will have `foo bar` className on the `div` element after `Tree` will be mounted.

layershifter:
@levithomason ping

levithomason:

PR #2416, opened 2018-01-08
feat(customPropTypes): allow to pass Symbol to `as`

### LINKED ISSUE #2414: Responsive: console warning generated when rendered as={React.Fragment}
### Steps
Normally Responsive would wrap its children in extra `<div>` but with introduction of React.Fragment the following construction is now possible to avoid it:

```jsx
  <Responsive as={React.Fragment} minWidth="10">
    <Label>label1</Label>
    <Label>label2</Label>
    <Label>label3</Label>
  </Responsive>
```

### Expected Result
3 labels without aggregating `<div>` element.

### Actual Result
Visual effect is correct, but there is a warning generated in dev console that:
> Warning: Failed prop type: Invalid prop `as` supplied to `Responsive`.
>    in Responsive (created by App)
>     in App

This seems to be caused by:
https://github.com/Semantic-Org/Semantic-UI-React/blob/7ef9fdbbe46f691f52c920d032a88e5ab0038ff9/src/lib/customPropTypes.js#L10-L13
which expects `as` property to be either string or a function (while React.Fragment is a "symbol").

### Version
0.77.1

### Testcase
https://codesandbox.io/s/pjk969mym7

layershifter:
Fixes #2414.

### DIFF
--- a/src/lib/customPropTypes.js
+++ b/src/lib/customPropTypes.js
@@ -8,10 +8,10 @@ const typeOf = (...args) => Object.prototype.toString.call(...args)
  * Ensure a component can render as a give prop value.
  */
 export const as = (...args) => PropTypes.oneOfType([
-  PropTypes.string,
   PropTypes.func,
+  PropTypes.string,
+  PropTypes.symbol,
 ])(...args)
-
 
 /**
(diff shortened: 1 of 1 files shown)


levithomason:

PR #2421, opened 2018-01-10
fix(Responsive): use setState() safely

layershifter:
### Problem

`Responsive` uses RAF for event handling and it means that `setState()` will be called after component will be unmounted.

Here is repro on [codesandox](https://codesandbox.io/s/koql1pn5ro), just resize preview area multiple times and you will see:

> Warning: Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.
>
> Please check the code for the Responsive component.

### Solution

Use `safeSetState` in same manner as in the `Transition` component.

---

Based on [SO question](https://stackoverflow.com/questions/48093185/semantic-ui-react-responsive-not-working-for-table-cell).

### DIFF
--- a/src/addons/Responsive/Responsive.js
+++ b/src/addons/Responsive/Responsive.js
@@ -66,11 +66,14 @@ export default class Responsive extends Component {
   componentDidMount() {
     const { fireOnMount } = this.props
 
+    this.mounted = true
+
     eventStack.sub('resize', this.handleResize, { target: 'window' })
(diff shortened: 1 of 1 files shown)


levithomason:
We've had this in a few areas.  Seems we might need to find a permanent solution.  This works for now, thanks!

levithomason:

Issue #2418, opened 2018-01-10
Item: chance of auto-generated key collision when using shorthand props


noinkling:
### Steps

```jsx
<Item header="1984" meta="1984" />

// or even:
<Item>
  <Item.Content header="1984" meta={1984} />
</Item>
```

### Expected Result

No issues.

### Actual Result

> Warning: Encountered two children with the same key, \`1984\`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

The warning is caused by the keys being automatically set from the value, here:
https://github.com/Semantic-Org/Semantic-UI-React/blob/9ef415645fcad18b4465b869b64b4355cb2b7cc5/src/lib/factories.js#L89-L92

Note that there's no need for keys at all in this example.

One workaround that prevents the keys from being set (and therefore the error):

```jsx
<Item header={{ content: "1984" }} meta={{ content: "1984" }} />
// yuck
```

### Version
0.77.1

### Testcase

https://codesandbox.io/s/4r702vq8y4

levithomason:
If you have duplicated shorthand values, pass props objects and include unique keys.  Alternatively, you can use the full markup.

Shorthand is a convenience and it cannot cover every use case.

noinkling:
> Shorthand is a convenience and it cannot cover every use case.

I understand that, however if you're dynamically generating your props you might have no idea that there's a possibility of having duplicate keys until you happen to feed it the right data (or maybe not at all if the warning isn't shown in production, which could potentially result in insidious bugs). It's the fact that it's surprising/not easily predictable that makes it an issue.

I'm sure there's a good underlying reason for `createShorthand()` to use auto-generated keys, but...

- In this particular case keys don't serve any purpose. The elements are laid out in a static order and it's not like each prop can generate multiple elements.

  Looking at the code: https://github.com/Semantic-Org/Semantic-UI-React/blob/9ef415645fcad18b4465b869b64b4355cb2b7cc5/src/views/Item/ItemContent.js#L46-L54

  Each of those `.create()` methods can accept an `options` object as a second argument that gets passed through to `createShorthand()` - a flag could be added that disables the autogenerated key (in this particular context and others where it makes sense) like:

  ```js
  ItemHeader.create(header, { generateKey: false })  // or dontGenerateKey: true
  ```

[...]

PR #2424, opened 2018-01-11
feat(factories): add autoGenerateKey option

noinkling:
See #2418 for more details.

WIP, currently just includes the implementation and a couple of tests.


test/specs/lib/factories-test.js
```diff
@@ -395,6 +396,18 @@ describe('factories', () => {
         value: 'a string',
         mapValueToProps: () => ({ undef: undefined, nil: null, zero: 0, empty: '' }),
       })
+
```
noinkling: I wasn't completely sure if this was a good place to put the tests, but currently 'key' and 'childKey' (further up) are two separate sections that seem to deal exclusively with explicitly provided object/element props, so it didn't feel right putting it into one of those.

levithomason: Let's keep all tests for a given option under that `describe()` block.  Ideally, we only `describe(key)` once.

src/lib/factories.js
```diff
@@ -81,12 +82,13 @@ export function createShorthand(Component, mapValueToProps, val, options = {}) {
...
   if (_.isNil(props.key)) {
     const { childKey } = props
+    const { generateKey = true } = options
 
     if (!_.isNil(childKey)) {
       // apply and consume the childKey
       props.key = typeof childKey === 'function' ? childKey(props) : childKey
       delete props.childKey
-    } else if (valIsString || valIsNumber) {
+    } else if (generateKey && (valIsString || valIsNumber)) {
```
levithomason: Currently, setting `generateKey: false` can still result in creating a key if `childKey` is passed.  I think this check can be hoisted up to line 83.  If `generateKey` is false, I would expect the factory to not perform any operation related to keys. What are your thoughts?

noinkling: This is how the check was made in the previous implementation (before it was removed). I guess the question is: is there value in being able to disable an explicitly specified `childKey` in this way? I believe you can already achieve this by passing in `overrideProps: { childKey: null }`. There's also the discrepancy with `key`, which bypasses the conditional.

If the logic stays as is, the option could be renamed to `autoGenerateKey` or `allowAutoKey` (etc.) to signify the intent better. Alternatively the conditional logic could be rejigged so that passing `overrideProps: { key: null }` achieves the same thing, giving it better behaviour parity with object/element values (and removing the need for the extra option altogether), but I think it would require us to differentiate between `null` and `undefined` (potentially brittle?).

levithomason: I see your point.  The shortcoming of using `overrideProps` is that it would override the user's key also, even when they explicitly define one.  I do like `autoGenerateKey`.  This way, it can be set to false and users can still choose to pass a key or not and get intuitive results.

noinkling: > The shortcoming of using `overrideProps` is that it would override the user's key also, even when they explicitly define one.

Ah I overlooked that somehow, however there's still `defaultProps` which could work:

```js
...
} else if (props.key === undefined && (valIsString || valIsNumber)) {
...
```

```js
// At call time:
ItemHeader.create(header, { defaultProps: { key: null } })
```

The obvious downside is that the semantics of that are far less clear to the user compared to a for-purpose boolean flag (and it's a bit verbose).

noinkling:
I'm trying to modify the `implementsShortHandProp` test to accept `autoGenerateKey`, so each component that implements shorthand props can be tested, however the `.should.contain()` (or `.should.containMatchingElement()`) assertion doesn't seem to compare keys (I'm guessing because `key` isn't a normal prop). It's called here:

https://github.com/Semantic-Org/Semantic-UI-React/blob/d077a171b682b1ef2d414a0143778cd74ad6c2f0/test/specs/commonTests/implementsShorthandProp.js#L56

I'm not very familiar with enzyme/chai-enzyme, so I'm not sure of the best way to fix this. After looking at the docs the best I can come up with is something like this:

```js
const wrapper = shallow(element)
wrapper.should[assertMethod](shorthandElement)
expect(wrapper.find(ShorthandComponent).key()).to.equal(shorthandElement.key)
```

But there's an assumption here that `.find()` will return a single node, and I don't know if that's an ok assumption to make in this context? Also it's essentially doing the work of finding the same node twice. Those could both be avoided if `contain()`/ `containMatchingElement()` returned the matched node somehow, but they don't. The extra work could be avoided like so:

```js
const wrapperChild = shallow(element).find(ShorthandComponent)
expect(wrapperChild.equals(shorthandElement)).to.equal(true)
expect(wrapperChild.key()).to.equal(shorthandElement.key)
```

But that still has the same 1 node assumption, and I don't know how to make it play nice with `assertExactMatch` since `assertMethod` isn't used anymore.

Help or guidance welcome.

noinkling:
So I've implemented the flag for all the shorthand props that already have a test that uses `implementsShorthandProp`, listed below:

- **Confirm**: header, content
- **BreadcrumbDivider**: icon
- **FormField**: label
- **MenuItem**: icon
- **Message**: icon, content, header, list
- **Button**: icon, label
- **Header**: icon, image, subheader
- **Image**: label, dimmer
- **Input**: action, icon, label, input
- **Label**: icon, image, detail
- **Step**: icon
- **StepContent**: title, description
- **Checkbox**: label
- **Dropdown**: icon, header
- **DropdownHeader**: icon
- **DropdownItem**: icon, label, image, flag, description, text
- **Embed**: iframe, icon
- **Modal**: header, content
- **CardContent**: header, meta, description
- **FeedContent**: date, summary, extraImages, extraText, meta
- **FeedEvent**: icon, image
- **FeedLabel**: icon
- **FeedLike**: icon
- **FeedMeta**: like
- **FeedSummary**: date, user
- **Item**: image
- **ItemContent**: header, meta, description, extra
- **Statistic**: label, value

However there are bunch more that aren't specifically tested, at least not using `implementsShorthandProp`. I haven't touched those ones yet - if they need tests (do they? cc @levithomason) then it's probably best to do that first. Out of the components I've already modified, these are the props I've identified that fall under this category:

- **Confirm**: cancelButton, confirmButton
- **Label**: removeIcon
- **Step**: description, title
- **Dropdown**: searchInput
- **Modal**: actions, closeIcon
- **FeedLabel**: image

Edit: the others that can benefit:

- **Table**: headerRow, footerRow
- **TableCell**: icon
- **ListContent**: header, description
- **ListItem**: icon, image, header, description
- **Popup**: header, content
- **Progress**: label
- **Search**: input
- **SearchResult**: image
- **Tab**: menu
- **Card**: image
- **CommentAvatar**: src

I think that's all of them.

levithomason:
Yes, I think it is best to update all usage of factories where we are creating a single element to use `autoGenerateKey: false`.  Looking really great so far 👍

levithomason:
Fixed merge conflicts here.

noinkling:
Thanks. I'm still gradually working on this, it's only taken so long because I'm trying to unify the testing for shorthand patterns that don't fit into the simple case (e.g. nested object keys, collection shorthands). It's mostly done, just need to find the time to do the last little bit.

levithomason:
I've fixed the remaining components and merged master.  I'm OK merging this now and doubling back for the test updates.

levithomason:
Merged master and fixed conflicts again.  Ready to merge if tests pass.

levithomason:

PR #2433, opened 2018-01-14
feat(Sidebar): closing sidebar on blur

cjsheets:
This is a follow up to the abandoned [#1473](https://github.com/Semantic-Org/Semantic-UI-React/pull/1473/files). This implementation is very similar to the original PR.

The biggest remaining gap is in testing. The use of `<Ref>` breaks the standard test methods like `isConformant` as well as tests that presume to affect the root node.

I'm interested in suggestions regarding a path forward. Should standard methods be updated to support `<Ref>` or should Sidebar have its own, more specific series of tests.

### DIFF
--- a/src/modules/Sidebar/Sidebar.js
+++ b/src/modules/Sidebar/Sidebar.js
@@ -1,11 +1,14 @@
 import cx from 'classnames'
+import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React from 'react'
 
+import Ref from '../../addons/Ref'
 import {
   AutoControlledComponent as Component,
   childrenUtils,
   customPropTypes,
+  eventStack,
   getUnhandledProps,
   getElementType,
   META,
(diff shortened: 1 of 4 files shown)


src/modules/Sidebar/Sidebar.js
```diff
@@ -40,6 +43,9 @@ class Sidebar extends Component {
     /** Direction the sidebar should appear on. */
     direction: PropTypes.oneOf(['top', 'right', 'bottom', 'left']),
 
+    /** Called when the user clicks away from the sidebar. */
+    onSidebarBlur: PropTypes.func,
```
levithomason, line 47: Since this is a prop on the Sidebar component, let's leave out the word `Sidebar` in the prop name:

```jsx
<Sidebar onBlur={} />
```

src/modules/Sidebar/Sidebar.js
```diff
@@ -63,20 +69,45 @@ class Sidebar extends Component {
...
+    })
   }
 
+  handleAnimationEnd = () => {
+    if (!this.state.visible) _.invoke(this.props, 'onHidden')
+    this.setState({ animating: false })
+  }
+
+  handleDocumentClick = (e) => {
+    if (!this.state.visible || this.ref.contains(e.target)) return
```
levithomason, line 103: In the event that the result of the click removes the target node from the DOM, it will fail to be detected here.  The ref will not contain the `e.target` because the target node was removed before this is invoked.

We wrote `doesNodeContainClick` for this specific usage.  You can import it from `lib` and use it like so:

```jsx
doesNodeContainClick(this.ref, e)
```

You can search the codebase for more usage examples, or look at the tests in `doesNodeContainClick-test.js`.

levithomason:
>I'm interested in suggestions regarding a path forward. Should standard methods be updated to support <Ref> or should Sidebar have its own, more specific series of tests.

It appears the Portal also renders a `<Ref />` and our solution there was to skip `common.isConformant(Portal)` and only run `common.hasValidTypings(Portal)`.

layershifter:
Closing in favor of #2845.

PR #2444, opened 2018-01-18
feat(Popup): Provide option to handle vertical offset

bharatzvm:
## Breaking Change

Popup now supports vertical offset alongside horizontal offset, in process offset prop
had to be renamed explicitly to mention the axis.

### Before

```
  <Popup offset={1000} />
```

### After

```
  <Popup horizontalOffset={1000} />
```

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -72,8 +72,11 @@ export default class Popup extends Component {
     /** Invert the colors of the Popup. */
     inverted: PropTypes.bool,
 
-    /** Horizontal offset in pixels to be applied to the Popup. */
-    offset: PropTypes.number,
+    /** Offset in pixels to be applied to the Popup. */
+    offset: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.arrayOf(PropTypes.number),
+    ]),
 
     /** Events triggering the popup. */
     on: PropTypes.oneOfType([
@@ -135,6 +138,7 @@ export default class Popup extends Component {
   static defaultProps = {
     position: 'top left',
     on: 'hover',
+    offset: 0,
   }
 
   static _meta = {
@@ -153,7 +157,6 @@ export default class Popup extends Component {
     // Do not access window/document when server side rendering
     if (!isBrowser()) return style
 
-    const { offset } = this.props
     const { pageYOffset, pageXOffset } = window
     const { clientWidth, clientHeight } = document.documentElement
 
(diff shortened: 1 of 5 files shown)


levithomason:
Whoops, appears we also have #2450, where I've left some updated comments.  I'll let you two decide which PR to keep around.  If you should choose to update your PR, please see the comments I left there.

bharatzvm:
@levithomason I have made some updates according to the discussions in #2450 and #1146, waiting for your opinion. Sorry for delay in response, I was travelling.

src/modules/Popup/Popup.js
```diff
@@ -188,15 +192,25 @@ export default class Popup extends Component {
...
+    const { horizontalOffset, verticalOffset } = this.props
+    return {
+      ...this.applyHorizontalOffset(style, horizontalOffset),
+      ...this.applyVeriticalOffset(style, verticalOffset),
     }
+  }
+
+  applyVeriticalOffset = ({ top, bottom }, offset) => {
+    if (_.isNumber(top)) return { bottom, top: top + offset }
+    return { top, bottom: bottom + offset }
```
adam-26: @bharatzvm, the`verticalOffset` calculation should be different based on if the popup is located above or below the _trigger_.  Are you sure this logic will not work as expected when the popup is below the trigger?

src/modules/Popup/Popup.js
```diff
@@ -188,15 +192,25 @@ export default class Popup extends Component {
...
+  }
 
-    return style
+  applyHorizontalOffset = ({ right, left }, offset) => {
+    if (_.isNumber(right)) return { left, right: right + offset }
+    return { right, left: left + offset }
```
adam-26: @bharatzvm, the original code **subtracts** the `horizontalOffset` value from the left/right value, you are adding the offset. Have you made a code change somewhere else that inverts the logic?

bharatzvm: @adam-26, Consider my previous comment here too(But my bad, I didn't realise this change while creating the pr)

src/modules/Popup/Popup.js
bharatzvm: @adam-26 I followed the notion of moving the popup outwards of the screen through the centre of the screen from the explicitly positioned direction whichever it is. It was also one of the reason why I had to add the offset.

bharatzvm: @adam-26 My bad again, I was wrong thanks for spotting it.

bharatzvm: @adam-26 I think the logic works fine irrespective of the positioning because if it can't place the popup in the view port it tries all other positions until it find ones in the viewport if it cant, it falls back to the position user specified position. I will be updating the pr to fix these reviews

levithomason:
@bharatzvm PR #2450 has been updated and merged.  I will close this one.  Thank you for working on this and for collaborating with @adam-26.

PR #2450, opened 2018-01-20
breaking(popup): verticalOffset

### LINKED ISSUE #2449: Popup: add verticalOffset
Component: Popup
New prop: `verticalOffset`

Allows the vertical position of a popup to be offset, he vertical equivalent to the horizontal `offset` prop.

I'll submit a PR.

adam-26:
## Breaking Change
Popup must rename the `offset` property to `horizontalOffset`

### Before
```js
<Popup offset={10} />
```

### After
```js
<Popup horizontalOffset={10} />
```

***

Add a `verticalOffset` prop to the popup component.

Fixes #2449.

### DIFF
--- a/docs/app/Examples/modules/Popup/Variations/PopupExampleOffset.js
+++ b/docs/app/Examples/modules/Popup/Variations/PopupExampleOffset.js
@@ -6,14 +6,26 @@ const PopupExampleOffset = () => (
     <Popup
       trigger={<Icon size='large' name='heart' circular />}
       content='Way off to the left'
-      offset={50}
+      horizontalOffset={50}
       position='left center'
     />
     <Popup
       trigger={<Icon size='large' name='heart' circular />}
       content='As expected this popup is way off to the right'
-      offset={50}
+      horizontalOffset={50}
       position='right center'
+    />
+    <Popup
+      trigger={<Icon size='large' name='heart' circular />}
+      content='Way off to the top'
+      verticalOffset={50}
+      position='top center'
+    />
+    <Popup
+      trigger={<Icon size='large' name='heart' circular />}
+      content='As expected this popup is way off to the bottom'
+      verticalOffset={50}
+      position='bottom center'
     />
   </div>
 )
--- a/docs/app/Examples/modules/Popup/Variations/index.js
+++ b/docs/app/Examples/modules/Popup/Variations/index.js
@@ -36,7 +36,7 @@ const PopupVariationsExamples = () => (
     />
     <ComponentExample
       title='Offset'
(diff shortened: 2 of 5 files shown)


levithomason:
Great, thanks for this!  We had this same PR submitted some time ago and never finished.

In an effort to keep the API consistent, let's consider a prop rename/restructure here.  Currently, we have `offset` for horizontal values and `verticalOffset` for vertical values.  Preferably, these would be more consistent.

### Option 1
Rename `offset` to `horizontalOffset`.  Downside, it is a breaking change.  Upside, it is explicit, consistent, and fits the SUI tone.

```jsx
<Popup horizontalOffset={} verticalOffset={} />
```

### Option 2
Allow `offset` to accept an object.  Downside, it is a little magical and nested prop values aren't ideal.  Upside, it is concise, maps to DOM element position values, and is not a breaking change.

```jsx
<Popup offset={10} />                // horizontal
<Popup offset={{ x: 10, y: 20 }} />  // horizontal, vertical
```

***

Thoughts?

levithomason:
Heads up, this is actually a duplicate of #2444.  I'll let the PR authors decide on which PR to focus efforts.

adam-26:
@levithomason, I think option 1 fits best with SUI. Why not just add ‘horizontalOffset’ as an alias to offset? This way, it’s not a breaking change. You could add a warning about using ‘offset’ and deprecate it in a future version?

levithomason:
Option 1 it is 👍.  We don't support backward compatibility while in development (`0.x`) so no need to alias.

levithomason:
Our changelog is auto-linked to PRs.  So that users can easily identify breaking changes and upgrade accordingly, we include a very brief description of why we are breaking the API and a before/after snippet.  Example PR here #1791.

Let's update the description is this PR as well.

adam-26:
@levithomason please review changes, and that the PR description is correctly formatted for the change log.

This also resolves #2444.

levithomason:
The fourth offset example, vertical 50 from bottom center, seems to misbehave.  The Popup appears then disappears immediately afterward.  This is the snippet:

```jsx
<Popup
  trigger={<Icon size='large' name='heart' circular />}
  content='As expected this popup is way off to the bottom'
  verticalOffset={50}
  position='bottom center'
/>
```

However, when changed to -50, it behaves correctly and is positioned far below as it should be.  Seems there is a bug there.

adam-26:
@levithomason bug is now fixed and tested. Please review again.

levithomason:
Awesome, this looks great.  Thanks much!

levithomason:

Issue #2401, opened 2017-12-26
Cannot read property 'firstElementChild' of null in Portal.js line 311


eldyvoon:
✔ BUGS → This form is required:

### Steps
https://drive.google.com/file/d/1d471crHVkIgjvBSXsVMTq3B4WWau6NqR/view

### Expected Result
This is should not show an error

### Actual Result
Error on
TypeError: Cannot read property 'firstElementChild' of null
node_modules/semantic-ui-react/dist/es/addons/Portal/Portal.js:311

### Version
"semantic-ui-react": "^0.77.0"

### My code

my container

```
render() {

  const { ad, redirectTo } = this.props

  return(
    <Container>
      {redirectTo && <Redirect to={redirectTo} />}
      {ad && <AdDetail 
        data={ad}
        applyAd={()=>this.applyAd()}
        openModal={()=>this.setState({isOpen: true})}
        closeModal={()=>this.setState({isOpen: false})}
        isOpen={this.state.isOpen} />}
    </Container>
  )
}
```

AdDetail component

```
import Modal from './modal'

export const AdDetailBody = ({data}) => (
    <Card
      fluid
      className='ad'
      header={data.title}
      meta={
        <div className='meta'>
          <Label>{data.category}</Label>
          <Label>{data.language}</Label>
          <p>{data.requirements}</p>
        </div>
      }
      extra={
        <div className='extra'>
          <div className='extra'>
            <p className='left'>{`RM ${data.reward}`}</p>
            <p className='right'>{data.expiry_date && `Expiry date: ${data.expiry_date}`}</p>
          </div>
        </div>
      }
    />
)

export const AdApplyBody = ({openModal, closeModal, isOpen, applyAd}) => (
  <div>
  <center>
    <Button onClick={openModal} primary>Take this job</Button>
  </center>
  <Modal 
    isOpen={isOpen}
    closeModal={closeModal}
    onSuccess={applyAd}
    header={'Confirm'}
    content={'I have read the requirements and able to deliver the task.'}
    cancelText={'Cancel'}
    successText={'Yes'}
  />
  </div>
)

const AdDetail = ({data, openModal, closeModal, isOpen, applyAd}) => (
  <div>
      <AdDetailBody data={data} />
      <br />
      <AdApplyBody 
        openModal={openModal}
        closeModal={closeModal}
        isOpen={isOpen}
        applyAd={applyAd}
      />
  </div>
)
```

Modal.js
```
[...]

PR #2457, opened 2018-01-23
Fix(Portal) check for rootNode rendering to prevent race condition 

Brantron:
We rely heavily on Semantic-UI-React and have had our monitoring reporting the following issue
```
TypeError: Cannot read property 'firstElementChild' of null
```

We've investigated it a few times but recently stumbled upon this issue https://github.com/Semantic-Org/Semantic-UI-React/issues/2401

This commit has been able to successfully prevent this issue without introducing any new issues.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -359,6 +359,10 @@ class Portal extends Component {
       eventStack.unsub('mouseenter', this.handlePortalMouseEnter, { pool: eventPool, target: this.portalNode })
     }
 
+    // Prevent race condition bug
+    // https://github.com/Semantic-Org/Semantic-UI-React/issues/2401
+    if (!this.rootNode) return null
+
     ReactDOM.unstable_renderSubtreeIntoContainer(
       this,
       Children.only(children),
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -359,21 +359,23 @@ class Portal extends Component {
       eventStack.unsub('mouseenter', this.handlePortalMouseEnter, { pool: eventPool, target: this.portalNode })
     }
 
-    // Prevent race condition bug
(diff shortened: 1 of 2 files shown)


levithomason:
Let's add a failing test that validates the fix.  The fix location is also suspect. If `this.rootNode` were `null` or `undefined` at line 364, where the check is currently, then [line 354](https://github.com/Brantron/Semantic-UI-React/blob/5077c28ed04c8306773a48350b301b4561cb6b83/src/addons/Portal/Portal.js#L354) would throw a TypeError before ever reaching that point:
```jsx
    this.rootNode.className = className || ''
```

Brantron:
@levithomason, excellent point. I dug a little further and came across https://github.com/facebook/react/issues/9808 which highlights

> The problem is `unstable_renderSubtreeIntoContainer` is async and can be called after `componentDidUpdate`. When refs are used in `componentDidUpdate`, they can be undefined due to async nature of `unstable_renderSubtreeIntoContainer`.

Thanks for the guidance. I've refactored a bit and added a test.

Brantron:
@levithomason ping

levithomason:
Ah, nice find!  I've done some manual testing and this all looks great.  Thanks for the detective work.

aloey:
Thank you so much for resolving this issue!
Any chance you will be updating the npm package soon?

levithomason:

levithomason:
@aloey I shoot for a release every weekend.  Sometimes life prevents it.

JsGvDev:
Thanks folks to fix this error. I've spent a few hours to find out about this bug, because I only  could get this error from Portal but into 'node_modules/react/cjs/react.development.js', until I can isolate that error. Nothing into the app is broken, everything was working fine, but I got that error #2401 from console. I was using semantic-ui-react@v0.76.0, after updating semantic-ui-react@v0.78.0 that error is gone.

PR #2458, opened 2018-01-23
fix(Visibility): context changes cause memory leak and incorrect scroll events

fracmak:
There's a couple of bugs in visibility regarding custom contexts. 
* If the context changes through a prop update, we leak the original scroll and resize handlers because on unmount, we look at the current context prop and unsub that, instead of what was originally sub'd
* If we provide a null context (like if the ref={} call is delayed), we listen to document scroll events and never listen to the scroll events of the context. Along those lines, I'm assuming sending the document scroll events are incorrect and shouldn't be sent to the onUpdate handler, but I'm open to arguments otherwise

### DIFF
--- a/src/behaviors/Visibility/Visibility.js
+++ b/src/behaviors/Visibility/Visibility.js
@@ -188,11 +188,16 @@ export default class Visibility extends Component {
   // Lifecycle
   // ----------------------------------------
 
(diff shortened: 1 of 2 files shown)


levithomason:
Superb.  Thanks for the excellent PR.

PR #2480, opened 2018-02-02
docs(Grid): fix column/row typo

### LINKED ISSUE #2479: Docs: Error Description: Semantic UI React -> Grid -> Grid.Column -> only
In Semanti UI React Docs, "Grid -> Grid.Column -> only" field's description starts as "A row can ...". It should be "A column can ...".

levithomason:
Fixes #2479


levithomason:

Issue #2478, opened 2018-02-01
Accordion.Title index prop type


TomFaulkner:
https://github.com/Semantic-Org/Semantic-UI-React/blob/a740b36998401a7bdb6f86979443c022343411ee/src/modules/Accordion/AccordionTitle.js#L37

The Accordion seems to function perfectly well if the index given is a string. We use uuids on the back-end for everything, so I used a uuid as the index without thinking about it. Is there anything wrong with changing the PropType to number or string?

```
index: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number,
]),
```

levithomason:
I think changing to string or number is acceptable.  In the end, the string will be coerced to a number though, since it is actually an array index that is needed.  Strings are common the UI anyway as HTML control values are string based.

PRs welcome.

TomFaulkner:
Great. I would have submitted a PR, but didn't know if there were any other changes necessary. Glad to see @bharatzvm got to it so quickly.

PR #2483, opened 2018-02-02
fix(AccordionTitle): allow strings in index PropType

bharatzvm:
Earlier AccordionTitle's prop index used to type check only for number,
now it also supports a string.

Fixes: https://github.com/Semantic-Org/Semantic-UI-React/issues/2478
Signed-off-by: bp <bharatzvm@github.com>


levithomason:
Looks like the corresponding typings in `AccordionTitle.d.ts` were already set correctly.  We're good to merge here, thanks!

levithomason:

PR #2491, opened 2018-02-04
chore(.github): add probot configurations

levithomason:
This PR adds configurations for several bots that will help manage issues and pull requests for us.


levithomason:

PR #2494, opened 2018-02-05
fix(doesNodeContainClick): only use x/y if !e.target

### LINKED ISSUE #2493: Modal calls onClose when it opens
### Steps
Use a `Popup` with a `Button` trigger, that indirectly controls the state of a `Modal`, which is nested inside a modal wrapper. Pass a callback function to the modal wrapper component as `onClose`, which resets the state of the parent component. Click the Popup trigger button to open the `Modal`, and nothing happens. The `Modal` is calling its `onClose` event when you click the trigger.

This issue only began to occur when I upgraded to `semantic-ui-react` v. ^0.78.0.

### Expected Result
The `Modal` `onClick` event does not fire unless triggered by an `onClick` event within the context of the `Modal`, and not its parent context.

### Actual Result
The `onClose` function will be called by the underlying semantic-ui-react `Modal` when the `Popup` trigger is clicked. As such, the `Modal` never opens, because it is 'closed' as it opens.

### Version
0.78.0

### Testcase
https://codesandbox.io/s/l7p5l52n2m

levithomason:
Fixes #2493 

If the `document` contains the `e.target`, then rely on the `node.contains(e.target)` method for checking if a click was within a `node`.

Only in cases where the `e.target` is no longer in the `document` do we need to resort to measuring the click position.

### DIFF
--- a/src/lib/doesNodeContainClick.js
+++ b/src/lib/doesNodeContainClick.js
@@ -12,8 +12,19 @@ import _ from 'lodash'
 const doesNodeContainClick = (node, e) => {
   if (_.some([e, node], _.isNil)) return false
 
-  // first check if the node contains the e.target, simplest use case
-  if (node.contains(e.target)) return true
+  // if there is an e.target and it is in the document, use a simple node.contains() check
+  if (e.target) {
+    _.invoke(e.target, 'setAttribute', 'data-suir-click-target', true)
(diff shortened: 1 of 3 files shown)


levithomason:
Released in `semantic-ui-react@0.78.1`.

PR #2498, opened 2018-02-06
docs(TableExamplePagination): Update chevron names

corydeppen:


levithomason:
Note, icon names are _not_ order dependent 👍 I'll close this for now as there is no effective change.  Please let me know if I've overlooked anything and we can reopen.

corydeppen:
In the `Icon` component, [`name`](https://github.com/Semantic-Org/Semantic-UI-React/blob/e1b0e91c5d701cfad38fbfa561ff65a10b1279cf/src/elements/Icon/Icon.d.ts#L48) has the type `SemanticICONS`, which should match one of the [specific strings](https://github.com/Semantic-Org/Semantic-UI-React/blob/e1b0e91c5d701cfad38fbfa561ff65a10b1279cf/src/index.d.ts#L83). Using `<Icon name="left chevron" />` results in a TS error `Type '"left chevron"' is not assignable to type '"download" | "target" | "list" | "time" | "copy" | "move" | "tree" | "content" | "header" | "sear...'.`, however `'chevron left'` is a valid string and compiles without a problem.

levithomason:
I see.  That is a great point.  The typescript does not include order variants, however, the prop type checker we have for this will respect order variations.

PR #2499, opened 2018-02-06
breaking(Tab): change menu.aligned prop to menuPosition and infer menu.tabular from it

### LINKED ISSUE #2430: Tab component passing redundant `aligned` prop to underlying Menu > div
### Steps
Use `Tab` component with panes shortcut prop.

### Expected Result
In development mode - no errors/warnings.

### Actual Result
In development mode - react's console error: ```Warning: Unknown prop `aligned` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop```.
In production mode - all is OK.

### Version
0.77.0 and above

### Testcase
To spot the problem one needs to open react-dev-tools and navigate to nodes `Tab` > `div` > `Menu` > `div` > `aligned` prop passed.
https://codesandbox.io/s/vy7kq4npp7 using 0.77.2 - there is a redundant prop passed to `div`
https://codesandbox.io/s/4w9z82qj8x using 0.76.0 - no redundant props passed to `div`

Also, here is the screenshot from official website, which is using 0.77.2:
tab-bug-suir

#### Investigation summary
The `aligned` prop was introduced in #2265 
It seems like `Tab` component passes this prop (as if it was unhandled) down to `Menu` component. `Menu` has no such prop also, so again it's being passed to underlying `div` as unhandled.

amankkg:
# BREAKING CHANGE

The `menu.aligned` prop has been replaced with `menuPosition`.  This keeps the `menu` prop strictly equal to the `<Menu />`'s props.

## Before
```jsx
const menu = {  fluid: true, vertical: true, aligned: 'right' }

<Tab panes={[...]} menu={menu} />
```

## After
```jsx
const menu = {  fluid: true,  vertical: true }

<Tab panes={[...]} menu={menu} menuPosition="right" />
```

***

The current implementation of `Tab`'s `menu`.`aligned` prop triggers React's design-time warning on redundant `aligned` prop being passed down to `div` of underlying `Menu`.
In this change new dedicated `menuPosition` prop is introduced and old `menu`.`aligned` prop is removed.

Remaining behavior is not changed:

- supported values is `PropTypes.oneOf(['left', 'right'])`
- non-required and default value is `'left'`

Resolves #2430


amankkg:
Any feedback on prop name? `menuPosition` is OK too, IMO, sounds similar to `Input`'s `label` and `labelPosition` props.

levithomason:
Ah, yes, `menuPosition` is more fitting.

amankkg:
Done!

levithomason:
### Vertical Example

This example shows the tabs on the wrong side of the pane.

http://localhost:8080/modules/tab#tab-example-vertical-true

image

It seems we also have conflicting configuration now between the props `menu.tabular` and `menuPosition`.  The `menu.tabular` prop should only take a boolean now that we have `menuPosition`.  The direction of a `tabular` menu should be derived from the `menuPosition`.

levithomason:
I see, we'll still need `menuPosition` for the cases where you need position a non-tabular menu, such as a `secondary` menu.

I think it would be ideal to infer the `tabular` value from the `menuPosition`, if provided.  So, that `menuPosition='right'` would automatically use a `tabular='right'` menu.  Otherwise, the user has to configure both to "right" to make it appear correctly.

amankkg:
@levithomason 
> This example shows the tabs on the wrong side of the pane.

But why? If there is no `menuPosition` prop provided - it defaults to `'left'`. And `tabular="right"` is only applied on `Menu` itself.

> It seems we also have conflicting configuration now between the props menu.tabular and menuPosition. The menu.tabular prop should only take a boolean now that we have menuPosition. The direction of a tabular menu should be derived from the menuPosition.

> I think it would be ideal to infer the `tabular` value from the `menuPosition`, if provided. So, that `menuPosition='right'` would automatically use a `tabular='right'` menu. Otherwise, the user has to configure both to "right" to make it appear correctly.

OK. We ~can~ should infer menu's `tabular` value from the `menuPosition` prop.
But, as user, I'd expect that `Tab`'s `menu` API to align `Menu`'s API. And `Menu`'s API on `tabular` prop is expecting one of `false | true | 'right'`. So `menu.tabular` shouldn't accept boolean only.
And, if we're going to infer menu's `tabular` value, then `Menu`'s `tabular` should accept `'left'` also, this way we can specify `Tab`'s `menu.tabular` as `'left'` and `menuPosition` as `'right'`.

src/modules/Tab/Tab.js
```diff
@@ -110,9 +110,13 @@ class Tab extends Component {
   }
 
   renderMenu() {
-    const { menu, panes } = this.props
+    const { menu, panes, menuPosition } = this.props
     const { activeIndex } = this.state
 
+    if (menu.tabular === true) {
+      menu.tabular = menuPosition === 'left' || 'right'
+    }
```
amankkg: `menu.tabular = menuPosition === 'left' || 'right'`
since there is no support for value `'left'` in `menu.tabular`, we cannot set `menu.tabular` as `'left'` AND `menuPosition` as `'right'`

amankkg:
There is the logic of inferring of `menu.tabular` at 4d6a9c4edde29d7b4c833fc0ccbfbf9996002e6d (comments are visible [here](https://github.com/Semantic-Org/Semantic-UI-React/pull/2499/commits/4d6a9c4edde29d7b4c833fc0ccbfbf9996002e6d)).
If it's OK we need to update `Menu` propTypes and handle its `tabular`'s value `'left'` to be similar as `true`.
And fix tests, docs, etc.

src/modules/Tab/Tab.js
amankkg: this way we can specify `menu.tabular` as `true` if we just want to infer it from `menuPosition`
and as `left` or `right` if we want to specify ourselves

src/modules/Tab/Tab.js
```diff
@@ -114,7 +114,7 @@ class Tab extends Component {
     const { activeIndex } = this.state
 
     if (menu.tabular === true) {
-      menu.tabular = menuPosition === 'left' || 'right'
+      menu.tabular = menuPosition
     }
```
amankkg: `menu.tabular = menuPosition`
this way we can specify `menu.tabular` as `true` if we just want to infer it from `menuPosition`
and as `left` or `right` if we want to specify ourselves

amankkg:
Now `menu.tabular` supports `'left'` and it derives final value only if set to `true`. This way we can specify `menu.tabular: 'left'` and `menuPosition: 'right'` simultaneously (which will became `tabular: true`, not 'right').

amankkg:
I'd like to see `Menu`'s `tabular` accepting `'left'` also, even if its behavior remains similar to `false`. If so, `Tab`'s `menu` prop would match `Menu`'s props.

amankkg:
@levithomason I'll update the rest of the code if it's OK now.

levithomason:
>I'd like to see Menu's tabular accepting 'left' also, even if its behavior remains similar to false. If so, Tab's menu prop would match Menu's props.

While I agree, let's save this for another PR. There are many cases of props taking boolean values along with a "right" or "bottom" but not accepting a "left" or "top". 

The only other comment I have here is that I think we need to simplify the prop computation. The prop description has too many considerations given the scope of the feature (positioning the menu). Ideally, the user wouldn't have to think or make decisions beyond a left/right decision for the position. Anything more than this I feel reduces the utility of the component.

I apologize I as I don't have time to offer suggestions at the moment. Feel free to offer your thoughts on this.

levithomason:
@amankkg How's your availability looking for wrapping this one up?

amankkg:
Oh, sorry for delay. I'll continue this asap 😎

amankkg:
@levithomason 

> While I agree, let's save this for another PR.

Yep, sure.

amankkg:
@levithomason there is an ugly prop computation because of `Menu`'s `aligned` prop API is one of...
> There are many cases of props taking boolean values along with a "right" or "bottom" but not accepting a "left" or "top"


Not sure if this can simplified without decreasing `Menu`'s customizeability.
Otherwise, there is no way to specify `menuPosition` to be right and `menu.tabular` to be left at the same time, as example.

Any comments and proposals are welcome, especially on props description in documentation.

upd: example case

amankkg:
Updated test cases and added new example for `menu.tabular` prop in docs.

levithomason:
This looks great!  I will do some cleanup and merge this today.

levithomason:
Added/updated docs, updated logic a bit and tests.  Now `menuPosition` always wins, but, `menu.tabular` can also be used to set the position, just as with a `<Menu />`.

amankkg:
@levithomason what is the status of this PR? What can I do to help?

levithomason:
Sorry for the delay.  Merging master and running tests and we'll ship this!

layershifter:
Removed cruft example, tests passed :+1:

levithomason:

PR #2501, opened 2018-02-06
Accept HTMLFormElement form event

### LINKED ISSUE #2500: Form onSubmit should accept HTMLFormElement event
It's not a bug, but rather a request to change the [TS definition](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/collections/Form/Form.d.ts#L38) for the `event` parameter for `onSubmit` from `React.FormEvent<HTMLElement>` to `React.FormEvent<HTMLFormElement>` in the `Form` component. This seems more appropriate and will allow it to handle events from other libraries like Formik without a compilation error. I don't mind submitting the PR to make this quick change if there is agreement.

corydeppen:
Fixes #2500

### DIFF
--- a/src/collections/Form/Form.d.ts
+++ b/src/collections/Form/Form.d.ts
@@ -35,7 +35,7 @@ export interface FormProps {
   loading?: boolean;
 
   /** The HTML form submit handler. */
-  onSubmit?: (event: React.FormEvent<HTMLElement>, data: FormProps) => void;
+  onSubmit?: (event: React.FormEvent<HTMLFormElement>, data: FormProps) => void;
 
   /** A comment can contain a form to reply to a comment. This may have arbitrary content. */
   reply?: boolean;


levithomason:
Released in `semantic-ui-react@0.78.2`.

levithomason:
Released in `semantic-ui-react@0.78.2`.

/cc @corydeppen

polc:
Hello, should not the event type be : `React.SyntheticEvent<HTMLFormElement>` ? compile with `react-final-form`. Here is the function type in popular libraries : 
- https://github.com/final-form/react-final-form/pull/52/files#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R37
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redux-form/v6/lib/reduxForm.d.ts#L488

Issue #2502, opened 2018-02-07
Progress: if progress is set to `value` and both value and total are set, display progress bar at the percentage of value/total


rachelslurs:
**Steps to Reproduce**
1. Render a `<Progress/>` component with these props
```
<Progress progress='value' value={35} total={50} />
```

**Expected**
The Progress bar should show it's 70% complete (percent value 35 divided by total 50). I want it to behave the same way `progress='ratio'` does, but without the `/{total}`

**Result**
The Progress bar sets it to 35% complete

**Testcase**
https://codepen.io/rachelslurs/pen/ZrpMYJ

levithomason:
Fixed in https://github.com/Semantic-Org/Semantic-UI-React/pull/2503

PR #2503, opened 2018-02-07
fix(Progress): use correct width for value/total

rachelslurs:
This is a fix for #2502

### DIFF
--- a/docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValuePercentageOfTotal.js
+++ b/docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValuePercentageOfTotal.js
@@ -1,0 +1,35 @@
+import React from 'react'
+import { Divider, Icon, Item, Progress } from 'semantic-ui-react'
+
+const ProgressExampleProgressValuePercentageOfTotal = () => (
+  <Item.Group divided relaxed>
+    <Item>
+      <Item.Image size='small' src='/assets/images/avatar/large/rachel.png' />
+      <Item.Content>
+        <Item.Header>Rachel</Item.Header>
+        <Item.Meta><Icon circular name='book' /> last read an article on January 26th</Item.Meta>
+        <Item.Meta><Icon circular name='newspaper' /> read 20 articles in January</Item.Meta>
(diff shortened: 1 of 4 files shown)


docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValuePercentageOfTotal.js
```diff
@@ -0,0 +1,35 @@
...
+    <Item>
+      <Item.Image size='small' src='/assets/images/avatar/large/zoe.jpg' />
+      <Item.Content>
+        <Item.Header>Zoe</Item.Header>
+        <Item.Meta><Icon circular name='book' /> last read an article on January 10th</Item.Meta>
+        <Item.Meta><Icon circular name='newspaper' /> read 7 articles in January</Item.Meta>
+        <Divider hidden />
+        <Item.Description>
+          <Progress progress='value' value={5} total={7} content='Fashion' color='purple' />
+          <Progress progress='value' value={2} total={7} content='Cats' color='yellow' />
```
levithomason: Let's simplify this example down to the minimum amount code required to show the feature.  It makes the docs a little quicker to scan and the doc code editor easier to play with.

levithomason:
Just one comment there. Looks great otherwise!

docs/app/Examples/modules/Progress/Content/ProgressExampleProgressValuePercentageOfTotal.js
rachelslurs: Sure, good idea. I wasn't sure if an example use case might be useful. Would there be somewhere else more appropriate to add the example I did?

I'll update this with a simpler example tomorrow.

levithomason: Right now we don't have a formal place for more cohesive examples.  It may be something we add in the future.

rachelslurs:
Updated with a simpler example.

rachelslurs:
@levithomason ready to review 👍

levithomason:

Issue #2277, opened 2017-11-01
Dropdown: Cannot read property 'disabled' of undefined


stelioschar:
This is reproducible in [dropdown multiple search selection](https://react.semantic-ui.com/modules/dropdown#dropdown-example-multiple-search-selection) 

### Steps
1. Go to [dropdown multiple search selection](https://react.semantic-ui.com/modules/dropdown#dropdown-example-multiple-search-selection) 
2. Type "afghanistan"
3. Press "Enter" - "Afghanistan" has been selected 
4. Press "Up"(or "Down") arrow then an error `Cannot read property 'disabled' of undefined` appears in console.

This first came up to my local project and I reproduced in demo. This happens everytime you type to step 2 the first item of the list. E.g if you have the below list, you need to type "Angular"
- Angular
- Ember
- React

This is also reproducible with or without the `flag` option.

### Expected Result
After step 4 I expected to navigate up & down and be able to select another option.

### Actual Result
JS error and cannot proceed with the selection of extra options.

### Version
0.75.1

lsouzaoliveira:
This bug already solved ? I'm facing this too, i would like to help to solve this, how can i help?

sudheerDev:
@stelioschar I cannot replicate this on latest release can u cross check this once?

stelioschar:
@sudheerDev yes. You can go here https://react.semantic-ui.com/modules/dropdown#dropdown-example-multiple-search-selection 
The key is that you need to type as many keys as only one option is filtered.
1. Type "greec"
2. Select "Greece" then once "Greece" is selected
3. Press Up/Down key

sudheerDev:
@stelioschar Thanks i will give this a shot

gabimoncha:
is it closed or still open?

afalahi:
I'm having this issue with the following code
``` javascript
import React from 'react';
import { Form, Item } from 'semantic-ui-react';


const DropDown = ({options, selectedColor, onSelectedColorChange}) =>{
  const renderedOptions = options.map( (option) =>{
    if (option.value === selectedColor.value) {
      return null;
    }
    return (
      <Item 
        key={option.value}
        onClick={() => onSelectedColorChange(option)}
      >
        {option.label}
      </Item>
    )
  });

  return (
    <Form>
      <Form.Dropdown 
        label='Select a Color'
        selection
        options={renderedOptions}
        className='active visible' text={selectedColor.label}
      />
    </Form>
  )
}

[...]

PR #2505, opened 2018-02-07
fix(Dropdown): Arrow navigation after unique search result select

sudheerDev:
[WIP]  `enabledIndicies` can be an empty array setting `enabledIndicies[0]` as selectedIndex state causes this issue.

Fixes: #2277

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -917,6 +917,8 @@ export default class Dropdown extends Component {
     this.trySetState({ value })
   }
 
+  checkForEnabledIndices = enabledIndicies => enabledIndicies[0] || 0
+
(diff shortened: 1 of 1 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -950,7 +952,7 @@ export default class Dropdown extends Component {
     }
 
     if (!newSelectedIndex || newSelectedIndex < 0) {
-      newSelectedIndex = enabledIndicies[0]
```
sudheerDev: `newSelectedIndex` is undefined in the case above causing this to set with enabledIndicies[0].

sudheerDev:
Closing in favour of this https://github.com/Semantic-Org/Semantic-UI-React/pull/2523

PR #2507, opened 2018-02-08
fix(Input): fix icon order to ensure rounded border is kept

### LINKED ISSUE #2506: Input: adding icon to Input with action clears border radius on the action button
**Steps to Reproduce**
use this page to reproduce: https://react.semantic-ui.com/elements/input#input-example-action

without icon:

image

with icon:
notice that button now doesn't have the border radius
image

Intregrisist:
Fixes #2506 

screen shot 2018-02-07 at 5 50 38 pm


ilyastam:
ship it!

brianespinosa:
Thanks @Intregrisist! 

Can you also add a test to this PR to make sure this order of elements is maintained?

I will take a look at this today and look through all of the other `Input` element examples and make sure this change does not break anything else.

src/elements/Input/Input.js
```diff
@@ -237,8 +237,8 @@ class Input extends Component {
         {actionPosition === 'left' && actionElement}
         {labelPosition !== 'right' && labelElement}
         {createHTMLInput(input || type, { defaultProps: htmlInputProps })}
-        {actionPosition !== 'left' && actionElement}
         {Icon.create(this.computeIcon())}
+        {actionPosition !== 'left' && actionElement}
```
levithomason: I would prefer a test that the render order is as it should be.  However, a comment will do in this case.  We use this convention when encountering gotchas like this:

```
// Heads up!
// A short comment about something very important.
```

Let's notify our future selves that SUI CSS requires the icon to be between these two elements.

levithomason:
Whoops, sorry @brianespinosa!  Didn't see your comment before I posted 😊 Obviously, I agree with you 👍

Intregrisist:
Sorry, wasn't available to make the changes until now. @levithomason & @brianespinosa please see a28a0aa.

levithomason:
Looking much better.  There is technically an infinite number of places the icon should not be.  Can we update the tests to assert where it should be instead of where it should not be?  This will make for stronger tests.

levithomason:
@Intregrisist ^

levithomason:
I've updated tests and merged master.  This can be merged on pass.

levithomason:

PR #2510, opened 2018-02-08
docs(Pagination): add missing prop in PaginationExampleOptions

elliottspirit:
Add the missing 'siblingRange' props of Pagination


layershifter:
Thanks :+1:

levithomason:

Issue #2114, opened 2017-09-24
Accordion: component should support transitions


ryardley:
### Steps
Use the Accordion component here from the docs: https://react.semantic-ui.com/modules/accordion#accordion-example-standard

```jsx
import React from 'react'
import { Accordion, Icon } from 'semantic-ui-react'

const AccordionExampleStandard = () => (
  <Accordion>
    <Accordion.Title>
      <Icon name='dropdown' />
      What is a dog?
    </Accordion.Title>
    <Accordion.Content>
      <p>
        A dog is a type of domesticated animal. Known for its loyalty and faithfulness,
        {' '}it can be found as a welcome guest in many households across the world.
      </p>
    </Accordion.Content>
    <Accordion.Title>
      <Icon name='dropdown' />
      What kinds of dogs are there?
    </Accordion.Title>
    <Accordion.Content>
      <p>
        There are many breeds of dogs. Each breed varies in size and temperament.
        {' '}Owners often select a breed of dog that they find to be compatible
        with their own lifestyle and desires from a companion.
      </p>
    </Accordion.Content>
    <Accordion.Title>
      <Icon name='dropdown' />
      How do you acquire a dog?
    </Accordion.Title>
    <Accordion.Content>
      <p>
        Three common ways for a prospective owner to acquire a dog is from pet shops,
        {' '}private owners, or shelters.
      </p>
      <p> A pet shop may be the most convenient way to buy a dog.
        {' '}Buying a dog from a private owner allows you to assess the pedigree and
        {' '}upbringing of your dog before choosing to take it home. Lastly, finding your dog
        {' '}from a shelter, helps give a good home to a dog who may not find one so readily.
      </p>
    </Accordion.Content>
  </Accordion>
)

export default AccordionExampleStandard
```

### Expected Result

The user should see an Accordion with content expanding in and out such as the one here: https://semantic-ui.com/modules/accordion.html 

### Actual Result

Accordion content does not animate and instead is immediately visible or hidden. 

### Version
0.73.1

levithomason:
Please note, no Transitions are included in this library, yet.

Transitions are a work in progress.  We've recently merged `<Transition />` and now `<Accordion />` refactoring that should make this possible.

I'll leave this issue open for tracking its implementation here.

Maxhodges:
tracked here now
https://github.com/Semantic-Org/Semantic-UI-React/issues/2114

balramsinghindia:
[...]

PR #2517, opened 2018-02-11
[WIP] Animated Accordion API

moaxaca:
Hey guys, 

I've made this in response to issue: https://github.com/Semantic-Org/Semantic-UI-React/issues/2114

Disclaimer: This isn't completed I want to talk to the maintainers on API design. I haven't included test coverage but I will when the API design is solidified.  

I am opening up this PR to promote a conversation on how to expose decorated animated interfaces. I was looking for examples of modules being decorated by Transitions and couldn't find a convention. 

I've included with this PR two ways of accomplishing the behavior one from a new top layer decorator of the Accordion -> AccordionAnimated. Secondly, I've also included a more intrusive example where I define the animation props against the AccordionContent. I don't expose the full transition API in the current implementation and would like to discuss a strategy on doing that gracefully. 

Looking forward to some feedback when you guys get the chance!

Also kudos on the codebase this being my first exposure it was extremely easy to bootstrap and contribute.

### DIFF
--- a/docs/app/Examples/addons/Pagination/Usage/PaginationExampleOptions.js
+++ b/docs/app/Examples/addons/Pagination/Usage/PaginationExampleOptions.js
@@ -37,6 +37,7 @@ export default class PaginationExampleCustomization extends Component {
             boundaryRange={boundaryRange}
             onPageChange={this.handlePaginationChange}
             size='mini'
+            siblingRange={siblingRange}
             totalPages={totalPages}
             // Heads up! All items are powered by shorthands, if you want to hide one of them, just pass `null` as value
             ellipsisItem={showEllipsis ? undefined : null}
--- a/docs/app/Examples/modules/Accordion/Advanced/AccordionExampleAnimated.js
+++ b/docs/app/Examples/modules/Accordion/Advanced/AccordionExampleAnimated.js
@@ -1,0 +1,51 @@
+import React from 'react'
+import { Accordion } from 'semantic-ui-react'
+
+const ShortContent = (
+  <div>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut varius,
+    nulla quis ornare suscipit, lacus diam lacinia tellus, quis pharetra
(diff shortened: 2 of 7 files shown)


brianespinosa:
Thanks for putting this together @moaxaca. I think we will want some input from @layershifter on this since I believe he had some vision on how we could reduce boilerplate when he created the `Transition` component. Ideally we could have one decorator component that could be reused on other components that is not specific to `Accordion`. So the `AccordionAnimated` component would be something more generic. I'll let him chime in here in case I am recalling his comments incorrectly from when he was taking his second stab at adding a `Transition` component.

moaxaca:
@brianespinosa 
That sounds good, I am looking forward to hearing from him. One that I discovered while implementing this the lack of CSS3 transitions against auto-calculated values. Here is a document describing the problem and some workarounds. I should note this only applies to the container of the accordion, not the content that's being animated. 

https://css-tricks.com/using-css-transitions-auto-dimensions/

I will explore the current implementation of Transition and see if there's a graceful solution.

GCrispino:
Good to know that there’s someone working on this. I’m new to the Semantic UI codebase, but I’m interested in contributing to this enhancement, if needed

moaxaca:
@GCrispino Thanks, were currently waiting for the maintainers to engage.

GCrispino:
@moaxaca Ok cool!

tayupov:
What's the status on this?

moaxaca:
@tayupov  need author feedback. @brianespinosa could we get the label "needs author feedback"

Whipgit:
How can we get a follow-up to this?

layershifter:
The right way was described for the implementation [there](https://github.com/Semantic-Org/Semantic-UI-React/pull/2001#issuecomment-326613796).

moaxaca:
@layershifter I will review and take a look at it this weekend.

hugomn:
@moaxaca Why was this closed? Can I help somehow?

PR #2523, opened 2018-02-13
Fixed Dropdown setSelectedIndex and getMenuOptions

mlmassey:
The Dropdown is not handling the following properly:
1) If the parent is managing the prop searchQuery, the componentWillReceiveProp updates the state value using the AutoControlledComponent.  The value is not actually set in this.state though, so the later call to setSelectedIndex will produce incorrect results.  This is because that call uses the state.searchQuery value to index the list.
2) When clearing the state.searchQuery, the selected index value is not updated appropriately.  When using state.searchQuery, the menu options will be different than the full menu options list, hence the selected index is a different value.  To fix this, simply call setSelectedIndex after clearing searchQuery in the clearSearchQuery function.  This address #2375.

The issue being seen was the following:
When performing a search, select an item from the filtered list of items per the search term.  If you then drop the menu list again, it should show the selected item highlighted, but it actually is showing a different selected item.  This is because the selected item index value is being computed based on the filtered list, but once the filtered list is reset (searchQuery is removed), the index never gets updated based on the currently selected value.

This PR passes both yarn lint and yarn test.  Let me know if you need anything else

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -726,8 +726,8 @@ export default class Dropdown extends Component {
     this.setValue(newValue)
     this.setSelectedIndex(value)
 
-    const optionSize = _.size(this.getMenuOptions())
-    if (!multiple || isAdditionItem || optionSize === 1) this.clearSearchQuery()
+    const optionSize = _.size(this.getMenuOptions(value))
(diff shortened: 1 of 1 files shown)


levithomason:
We've received a few duplicate PRs for this bug.  I am working to get #2375 merged as it also addresses a few related issues I've been wanting to fix.  Thank you for the investigation and contribution.

PR #2530, opened 2018-02-16
Fixed the redundant aligned prop passing to div issue

Abdul-Moiz-Ansari:
# Fix to Redundant 'aligned' props passing to div issue - #2430 .

Issue was introduced in #2430  that in development while using Tab component, React console gives this  `Warning: Unknown prop aligned on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop.`

This PR fixes this issue, with a separate 'tab' prop having a 'menuAligned' property, same as the solution stated in #2430 . Since tab prop is not passing in the Menu component, it is not being rendered as unused props. 

So in order to make the menu right aligned, change will be as below:

## Before

`const panes = [ ...{} ]
const menu = {
  fluid: true,
  vertical: true,
  secondary: true,
  aligned: 'right'
}

<Tab menu={menu} panes={panes} />`

## After
`const panes = [ ...{} ]
const menu = {
  fluid: true,
  vertical: true,
  secondary: true
}
const tab = {
  menuAligned : 'right'
}

<Tab menu={menu} panes={panes} tab={tab} />`

tab props will not be passed in Menu component, so will not be rendered in any case.

### DIFF
--- a/src/modules/Tab/Tab.js
+++ b/src/modules/Tab/Tab.js
@@ -73,7 +73,8 @@ class Tab extends Component {
 
   static defaultProps = {
     grid: { paneWidth: 12, tabWidth: 4 },
-    menu: { attached: true, tabular: true, aligned: 'left' },
+    menu: { attached: true, tabular: true },
+    tab : {menuAligned : 'left'},
     renderActiveOnly: true,
   }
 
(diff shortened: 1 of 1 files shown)


levithomason:
Thanks for the contribution!  Unfortunately, this is a duplicate of https://github.com/Semantic-Org/Semantic-UI-React/pull/2499.

I will leave both open to see which one we can get merged.  However, you may want to take a look there to get up to speed on the latest convo.

levithomason:
Closing in favor of #2499.

PR #2531, opened 2018-02-16
fix(Responsive): use root element client width

autarc:
There is currently an issue with the [Responsive](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/addons/Responsive/Responsive.js) component regarding the expected rendering behavior. While the element should render if it's visible the `fitsMinWidth` and `fitsMaxWidth` can return miss matches depending on the state of mobile devices. Zooming on webkit based browser at iOS (mobile Safari, Google Chrome) triggers the `resize` event at which `window.innerWidth` returns an incorrect value.

By using the measurements of the root element of the document ([document.documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement)) rather than the window viewport the detection is more reliable. Instead of capturing the modified width during gestures like pinch to zoom the actual one of the document will be used.


levithomason:
I wish we had mobile tests :)  Thanks for the investigation and contribution.

src/addons/Responsive/Responsive.js
```diff
@@ -60,7 +60,7 @@ export default class Responsive extends Component {
   constructor(...args) {
     super(...args)
 
-    this.state = { width: isBrowser() ? window.innerWidth : 0 }
+    this.state = { width: isBrowser() ? document.documentElement.clientWidth : 0 }
```
levithomason: This is probably worth adding an inline comment with the GitHub issue URL and a one-line description as to why this is being used.

autarc: Okay added a comment and reference to the PR. Will be squashed if it's fine :)

levithomason:

Issue #2476, opened 2018-02-01
Dropdown: option to disable "infinity" list


Unsfer:
Can you add some flag option to disable jumping to first/last item when selecting items by up and down arrow keys and reached the end of list.
It's a defauld behavior of an <select> html element and I want to reproduce it.

levithomason:
I'm typically opposed to adding more branch logic to the Dropdown as it is quite massive and complicated as is.  However, I would accept a PR for this as I prefer we mimic HTML controls when possible.

How about `wrapSelection`?

Unsfer:
Sure. Something like this?
`
<Dropdown 
        selection 
        options={friendOptions}
        wrapSelection
/>
`
And will it concern only Semantic-UI-React or pure js version too?

levithomason:
Yep, that seems great to me.  It will affect only the React version.

GCrispino:
Hello guys. I went ahead and tried to implement this. It was basically just change a couple of lines inside Dropdown component's moveSelectionBy() method, and it works. However, 4 Dropdown component's tests failed. If I understood that correctly, they all failed because the tests relied on the behavior of "infinite" selection, as @Unsfer named it. 

The failed tests were "× moves up on arrow up when open", "× scrolls the selected item into view", "× updates value on up arrow", "× keeps the selection within the range of remaining options". 

What should be done? Should the tests be updated to not rely on that behaviour? I'm not too experienced with unit testing in Javascript, but I think I can take care of this, with your help, if needed

EDIT: I just realized again that the feature has to be activated by a flag, which means that the tests would probably not fail, then. But I think that new tests should be added, right?

PR #2533, opened 2018-02-17
feat(Dropdown): 'wrapSelecion' prop: option to disable "infinity" list

GCrispino:
As discussed on issue [#2476](https://github.com/Semantic-Org/Semantic-UI-React/issues/2476), I have implemented the commented functionality, with the use of the 'wrapSelection' prop on the Dropdown component. I've also added simple tests and an example to the documentation


GCrispino:
Forgot to do the linting thing, now it's all good and passing

src/modules/Dropdown/Dropdown.d.ts
```diff
@@ -254,6 +254,12 @@ export interface DropdownProps {
 
   /** A dropdown can open upward. */
   upward?: boolean;
+
+  /**
+   * A dropdown won't go to the last element when ArrowUp is pressed on the first,
+   *  nor go to the first when ArrowDown is pressed on the last(aka infinite selection)
```
levithomason: nit: extra space before `nor` and missing space after `last(`

src/modules/Dropdown/Dropdown.js
```diff
@@ -352,6 +352,11 @@ export default class Dropdown extends Component {
 
     /** A dropdown can open upward. */
     upward: PropTypes.bool,
+
+    /** A dropdown won't go to the last element when ArrowUp is pressed on the first,
+     *  nor go to the first when ArrowDown is pressed on the last(aka infinite selection)
```
levithomason: See above comment on whitespace

src/modules/Dropdown/Dropdown.js
```diff
@@ -994,10 +999,18 @@ export default class Dropdown extends Component {
...
+    const { wrapSelection } = this.props
     // next is after last, wrap to beginning
     // next is before first, wrap to end
     let nextIndex = startIndex + offset
-    if (nextIndex > lastIndex) nextIndex = 0
+
+    /*
+    * if 'wrapSelection' is set and selection
+    * is after last or before first, it just does not change
+    */
```
levithomason: `//` for inline comments, please.  This should also fit on one line.  The linter allows 120 columns I believe.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -672,6 +672,58 @@ describe('Dropdown', () => {
         .at(options.length - 1)
         .should.have.prop('selected', true)
     })
+    it('does not move up on arrow up when first item is selected when open when \'wrapSelection\' is on', () => {
```
levithomason: nit: Double quotes when string contains single quotes (prevents escaping)

Side note, we really need to move to `prettier` :)

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -672,6 +672,58 @@ describe('Dropdown', () => {
...
+      wrapper
+        .find('DropdownItem')
+        .first()
+        .should.have.prop('selected', true)
+      wrapper
+        .find('DropdownItem')
+        .at(options.length - 1)
+        .should.have.prop('selected', false)
+    })
+    it('does not move down on arrow down when last item is selected when open when \'wrapSelection\' is on', () => {
```
levithomason: We try to organize our tests by prop name.  Let's add a new `describe('wrapSelection')` block.  It helps find tests, ensure coverage, and prevents duplication in test `it()` descriptions.

test/specs/modules/Dropdown/Dropdown-test.js
```diff
@@ -672,6 +672,58 @@ describe('Dropdown', () => {
...
+      // selection should not move to first item
+      // should keep on last instead
+      wrapper
+        .find('DropdownItem')
+        .first()
+        .should.have.prop('selected', false)
+      wrapper
+        .find('DropdownItem')
+        .at(options.length - 1)
+        .should.have.prop('selected', true)
```
levithomason: These are superb tests, thanks much!

docs/app/Examples/modules/Dropdown/Usage/index.js
```diff
@@ -119,6 +119,11 @@ const DropdownUsageExamples = () => (
       description='A dropdown can open its menu upward.'
       examplePath='modules/Dropdown/Usage/DropdownExampleUpwardSelection'
     />
+    <ComponentExample
+      title='Wrap Selection'
+      description='A dropdown can wrap its selection to disable "infinite" selection'
```
levithomason: Should this say `enable` rather than `disable`?  Also, what do you think about adding the word "scrolling" into here?  Something "enable infinite scrolling the selection".  Or, "A dropdown selection can wrap to the start when it reaches the end and vice versa"?  Could just be me but I wasn't quite sure what this was described on the first read.

Thoughts?

levithomason:
I see, I think we need to invert the treatment of the value and add a default value of true here.

```jsx
<Dropdown
  wrapSelection         // the selection will wrap
  wrapSelection={false} // the selection will not wrap
/>
```

All else is good by me.

GCrispino:
OK, but to make sure, what do you mean by ‘wrap’? Does it mean that the selection is able to go from the last option to the first and vice-versa, or that it’s not able?

levithomason:
Wrap to me means something is going from the end to the start.  In this case, wrap means the selection goes from the end to the beginning.  I compare this to "soft wraps" in a text editor where the text wraps from the end of the line to the start of a new line.  Also, in CSS where white-space wraps by default but can be set to `nowrap` where the text does not wrap but overflows.

I think this maps here where `wrapSelection` (i.e. `wrapSelection={true}` is the equivalent of "enabled wrapping behavior".

GCrispino:
Ooh I see. Now I understand why I confused you, I implemented things the opposite way. In my current implementation it does not wrap when wrapSelection is true. So I’ll just invert things to have a better meaning, if that’s ok for you. Also, should I default wrapSelection to true?

levithomason:
>So I’ll just invert things to have a better meaning, if that’s ok for you. Also, should I default wrapSelection to true?

Yes and yes  😄

GCrispino:
Cool! Will do that soon, together with the other changes. Thanks and sorry for the confusion, I’m kind of new to contributing :D

Em 18 de fev de 2018, à(s) 16:56, Levi Thomason <notifications@github.com<mailto:notifications@github.com>> escreveu:


So I’ll just invert things to have a better meaning, if that’s ok for you. Also, should I default wrapSelection to true?

Yes and yes 😄

levithomason:
Sounds great.  No worries, there are no truly right answers here.  I'm super grateful for the PR and the rapid replies.

GCrispino:
Awesome!

Em 18 de fev de 2018, à(s) 16:59, Levi Thomason <notifications@github.com<mailto:notifications@github.com>> escreveu:


Sounds great. No worries, there are no truly right answers here. I'm super grateful for the PR and the rapid replies.

docs/app/Examples/modules/Dropdown/Usage/index.js
```diff
@@ -120,9 +120,9 @@ const DropdownUsageExamples = () => (
       examplePath='modules/Dropdown/Usage/DropdownExampleUpwardSelection'
     />
     <ComponentExample
-      title='Wrap Selection'
-      description='A dropdown can wrap its selection to disable "infinite" selection'
-      examplePath='modules/Dropdown/Usage/DropdownExampleWrapSelection'
+      title='Unwrap Selection'
```
levithomason: Let's keep this named after the prop, "Wrap Selection".

GCrispino: Ok cool. What about the description and the component name, are they fine?

levithomason:
```jsx
<ComponentExample
  description={[
    'This is one line of description that is quite long, however, there is',
    ' even more to go.  Luckily, we can join an array of strings!',
  ].join('')}
/>
```

Admittedly, not the greatest solution, but, this is how we've been handling it by convention.

GCrispino: Oh, I see, no problem ;P

Will update that later today

levithomason:

Unsfer:
I got "too much recursion" error when the first/last item is disabled and I move in that direction

levithomason:
OK, please file a new issue and we'll get it resolved.

PR #2534, opened 2018-02-18
docs(TableExamplePagination): order icon names per docs

searene:
According to https://react.semantic-ui.com/elements/icon, `chevron left` and `chevron right` should be used.


levithomason:
Icon names are not order dependent.  However, the typescript is not as forgiving as the prop type checker and the CSS.

Since we've had this PR twice in the last week or two, I'll go ahead and merge this one 👍

levithomason:

PR #2537, opened 2018-02-19
fix(Pagination): allow to override props in PaginationItem

aivins:
This PR permits explicit passing in of the disabled prop to a PaginationItem.


src/addons/Pagination/PaginationItem.js
```diff
@@ -66,7 +66,7 @@ class PaginationItem extends Component {
 
   render() {
     const { active, ariaLabel, type, ...rest } = this.props
-    const disabled = type === 'ellipsisItem'
+    const disabled = rest.disabled || type === 'ellipsisItem'
```
levithomason: This sort of makes me wonder why `rest` props aren't spread _after_ all other props below, see lines 72-79.  Usually, the user's props should always "win".

@layershifter is there a reason for this or what is it an oversight?  Also, shouldn't the props passed to the `MenuItem` factory be passed as `{ defaultProps: { ... } }` so that they don't override the users `rest` props?

layershifter: Yep, it's my miss. I will check this PR on this weekend

src/addons/Pagination/PaginationItem.js
```diff
@@ -64,19 +67,26 @@ class PaginationItem extends Component {
...
+      defaultProps: {
+        active,
+        disabled,
+        'aria-current': active,
+        'aria-label': ariaLabel,
+        onClick: this.handleClick,
+        onKeyDown: this.handleKeyDown,
+        tabIndex: disabled ? -1 : 0,
+      },
+      overrideProps: this.handleOverrides,
```
layershifter, line 89: @levithomason I've pushed some changes and added some new tests, ready for your review :+1:

layershifter (approved):
@levithomason I've pushed

src/addons/Pagination/PaginationItem.js
levithomason: You're awesome ❤️

levithomason:
Thanks!

levithomason:

PR #2538, opened 2018-02-19
test(Portal): fix tests with timeouts

levithomason:
`TransitionablePortal` tests are intermittently failing due to unpredictable animation durations and set timeout times.

This PR adds a new util `assertWithTimeout` which repeatedly makes the assertion until it passes or until 2s have passed.  In the event 2s is reached without the assertion passing, the error will be thrown.  This allows us to execute tests as fast as possible but also account for unknown execution times.


layershifter:
Merged as it was marked as `ready for merge`, @levithomason :+1:

levithomason:

Issue #2559, opened 2018-02-22
Example #3 is not running


progrookie:
Hello everyone,

I using the webpack-Example "3".
When I use the command "npm start" then the following error occured: 
_"Error: babel-plugin-direct-import: contains unknown keys { modules }"_

I have the start.js or other script not touched!

Can everyone help me?

many thanks in advance

Sven

KajSzy:
Just edit .babelrc file:

`plugins : [
 "direct-import",
    [
      "module:semantic-ui-react",
      {
        "name": "semantic-ui-react",
        "indexFile": "semantic-ui-react/dist/es/index.js"
      }
    ]
]`

But sadly still got some problems:

`Module build failed: Error: Must export a default export when using ES6 modules`

progrookie:
Thank you for your quick answer!

Exists an executable (small) example with
- React
- Semantic UI React
- WebPack (or Rollup ;))

Thanks a lot!

layershifter:
Guys, sorry something went wrong there :crying_cat_face:

I've shipped some updates, now boilerplate should be fully working. Please check again :+1:

KajSzy:
There is still missing "react-hot-loader" as devDep

layershifter:
@KajSzy thanks, check #2567.

PR #2567, opened 2018-02-25
fix(examples): add `react-hot-loader` to deps in webpack3 example

layershifter:
Rel #2559.


levithomason:

PR #2568, opened 2018-02-25
fix(Accordion): update signature of `onTitleClick` in typings

### LINKED ISSUE #2557: Accordion: onTitleClick typings are incorrect
### Steps

Use Accordion's onTitleClick property according to the typings, which look like this

```typescript
onTitleClick?: (event: React.MouseEvent<HTMLDivElement>, index: number | number[]) => void;
```

### Expected Result

get a `number` or `number[]` as the second argument

### Actual Result

got an `object`

### Version
0.78.2

### Testcase
https://codesandbox.io/s/vjzyjjqr13

layershifter:
Fixes #2557.


levithomason:

PR #2569, opened 2018-02-25
feat(Popup): add keepInView prop (#2472)

### LINKED ISSUE #2472: Popup: add prop to optionally find an "in viewport" position
problem:
when I set the popup to a certain position,
https://react.semantic-ui.com/modules/popup
such as: 
`bottom left`
but the window is too small to allow bottom left direction,
the popup goes back to automatic position.

This should not happen! you should give an option to force a position direction,
the position is set with the intent to be enforced, not a suggestion,
sometimes the content inside the popup is too big, and the only way to ensure that the full content is  seen is to direct it to the bottom, because the web browser will add scrollers anyway.
please consider either enforcing the position, or adding a parameter to make it automatically.

### Steps

1. create popover on any page with a large amount of content, such as a form, or an image or just 20 lines of text for example.
2. set position `bottom left`
3. before you open the popup, resize the browser window to the button that opens the popup.
4. click the buttton that opens the popup.

### Expected Result
the popup opens as expected, towards bottom left, and the web browser adds scrollers anyway... to accomodate this effect.
### Actual Result
the popup opens to any direction which has space... usually top left, this is not supposed to happend,
it makes the setting "bottom left" not very useful.
### Version
0.71.5

please provide any hints, i can do a pull request if neccresary.

thank you!

paulcmason:
Fixes #2472

### DIFF
--- a/test/specs/modules/Popup/Popup-test.js
+++ b/test/specs/modules/Popup/Popup-test.js
@@ -192,6 +192,54 @@ describe('Popup', () => {
     })
   })
 
+  describe('keepInViewPort', () => {
+    it('will not alter the position and render outside the viewport if set to false', () => {
+      wrapperMount(
+        <Popup
+          content='_'
+          position='top center'
+          trigger={<button>foo</button>}
+          on='click'
+          keepInViewPort={false}
+        />,
+      )
+      wrapper.find('button').simulate('click')
+
+      const rect = document.querySelector('.popup.ui').getBoundingClientRect()
+      const { top } = rect
+
+      const selectedPosition = wrapper.state('position')
+
+      expect(selectedPosition).to.equal('top center')
+      expect(top).to.be.below(0)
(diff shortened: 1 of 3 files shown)


test/specs/modules/Popup/Popup-test.js
```diff
@@ -192,6 +192,29 @@ describe('Popup', () => {
     })
   })
 
+  describe('keepInViewPort', () => {
+    it('will not alter the position and render outside the viewport if set to false', () => {
```
levithomason, line 196: Perfect!  We just need to add tests asserting that:
- `it('is enabled by default')`
- `it('alters the position when true')`

levithomason:
Looks great, just need a couple more tests here.

paulcmason:
Hi @levithomason - thanks for taking a look - extra couple of tests have been added now.

PR #2571, opened 2018-02-25
Visibility: an update after component's unmount throws exception

### LINKED ISSUE #2570: Visibility: Cannot read property 'getBoundingClientRect' of null

### Steps
I could not track it down.
It is available on your site:
https://react.semantic-ui.com/layouts/homepage

### Expected Result
Responsiveness
### Actual Result
Crashing between 768px and 992px width and less than 320 px.
### Version

```js
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at Visibility.computeCalculations (Visibility.js:238)
    at Visibility._this.update (Visibility.js:72)
computeCalculations @ Visibility.js:238
Visibility._this.update @ Visibility.js:72
requestAnimationFrame (async)
Visibility._this.handleUpdate @ Visibility.js:67
(anonymous) @ EventTarget.js:31
arrayEach @ _arrayEach.js:15
forEach @ forEach.js:38
(anonymous) @ EventTarget.js:30
(anonymous) @ _createBaseFor.js:17
baseForOwn @ _baseForOwn.js:13
(anonymous) @ _createBaseEach.js:17
forEach @ forEach.js:38
(anonymous) @ EventTarget.js:24
```

VikVelev:
The responsiveness on your doc page was broken and wasn't considering devices other than mobile and desktop.

Fixes #2570

### DIFF
--- a/docs/app/Layouts/HomepageLayout.js
+++ b/docs/app/Layouts/HomepageLayout.js
@@ -69,7 +69,7 @@ class DesktopContainer extends Component {
     const { fixed } = this.state
 
     return (
-      <Responsive {...Responsive.onlyComputer}>
+      <Responsive {...Responsive.onlyComputer} minWidth={768}>
         <Visibility once={false} onBottomPassed={this.showFixedMenu} onBottomPassedReverse={this.hideFixedMenu}>
           <Segment inverted textAlign='center' style={{ minHeight: 700, padding: '1em 0em' }} vertical>
             <Menu
@@ -114,7 +114,7 @@ class MobileContainer extends Component {
     const { sidebarOpened } = this.state
 
     return (
-      <Responsive {...Responsive.onlyMobile}>
+      <Responsive {...Responsive.onlyMobile} minWidth={0}>
         <Sidebar.Pushable>
           <Sidebar as={Menu} animation='uncover' inverted vertical visible={sidebarOpened}>
             <Menu.Item as='a' active>Home</Menu.Item>


layershifter:
To team: don't merge this, the problem is in suddenly unmounted Visibility component and should be fixed there.

docs/app/Layouts/HomepageLayout.js
```diff
@@ -69,7 +69,7 @@ class DesktopContainer extends Component {
     const { fixed } = this.state
 
     return (
-      <Responsive {...Responsive.onlyComputer}>
+      <Responsive {...Responsive.onlyComputer} minWidth={768}>
```
levithomason, line 72: I'm not sure on this.  The props being spread in `Responsive.onlyComputer` include a minWidth of 992.  Why is is that we need to override this to 768?

https://github.com/Semantic-Org/Semantic-UI-React/blob/8ccee0f75fb2a9802bbaf33394182dc97e8b5d2e/src/addons/Responsive/Responsive.js#L56

If 768 is indeed correct, then we should not be spreading the other value and then overridding it.

docs/app/Layouts/HomepageLayout.js
```diff
@@ -114,7 +114,7 @@ class MobileContainer extends Component {
     const { sidebarOpened } = this.state
 
     return (
-      <Responsive {...Responsive.onlyMobile}>
+      <Responsive {...Responsive.onlyMobile} minWidth={0}>
```
levithomason, line 117: This one also is specifically spreading a different value for minWidth.  Why is it that we need to override this to 0?

https://github.com/Semantic-Org/Semantic-UI-React/blob/8ccee0f75fb2a9802bbaf33394182dc97e8b5d2e/src/addons/Responsive/Responsive.js#L54

As above, if minWidth 0 is correct then we should not be spreading `onlyMobile` here, or, we need to update the values of `onlyMobile`.

levithomason (commented):
@layershifter Can you take a look here?

layershifter:
See #2635 for details.

PR #2576, opened 2018-02-26
fix(Transition): add slide left

### LINKED ISSUE #2573: React Semantic : why slide left transition is not available in <Transition> ? Any alternative to use custom animation in this?

levithomason:
Fixes #2573


levithomason:

PR #2582, opened 2018-02-26
Added 'left' option to ribbon on Label.js

leandrocruz:


brianespinosa:
Thanks for taking the time to open a PR, @leandrocruz. If you [take a look at the markup examples for the ribbon type Label in SUI](https://semantic-ui.com/elements/label.html#ribbon), you will notice that there is no class name of `left` for the left aligned ribbon label. The ribbon style is implicitly aligned left. Adding this as a possible value for the prop does not do anything as far as the styles and classnames are concerned.

leandrocruz:
That is correct. It works. I was led to believe that I should have to choose between left and right because the documentation does not state the default value is left. Thanks for your help.

brianespinosa:
@leandrocruz I see your point. It might not be clear to someone not as familiar with the markup/classes generated that the boolean version is the default left, and you can optionally make it go right.

image

I think in the example area it would be beneficial to add a comment that the boolean/default position of a ribbon is left. Feel free to open up a PR for that!

levithomason:
I've considered allowing balanced options like this in all prop types and typings just for developer sanity and intuition.

Unfortunately, it would make a bit of a tangled mess out of our prop handlers for generating class names. We're tied to the CSS pretty heavily (repeat, unfortunately).

PR #2594, opened 2018-03-03
chore(package): switch to `babel-preset-env`

thetric:
Use `babel-preset-env` instead of `babel-preset-es2015` as this is the recommended by the Babel dev team, see https://babeljs.io/env/.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "babel-plugin-transform-react-handled-props": "^0.2.5",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.8",
     "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-1": "^6.24.1",
     "babel-register": "^6.24.1",
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    "./build/preset-es2015",
+    "./build/preset-env",
     "react",
     "stage-1"
   ],
--- a/build/preset-env.js
+++ b/build/preset-env.js
@@ -11,7 +11,7 @@ const plugins = env === 'build' || env === 'build-es' ? [
 
 module.exports = {
   presets: [
-    ['es2015', options],
+    ['env', options],
   ],
   plugins,
 }
--- a/build/preset-env.js
+++ b/build/preset-env.js
@@ -1,5 +1,5 @@
 const env = process.env.NODE_ENV
-const options = env === 'build-es' ? { modules: false } : {}
+
 const plugins = env === 'build' || env === 'build-es' ? [
   ['filter-imports', {
     imports: {
@@ -8,10 +8,19 @@ const plugins = env === 'build' || env === 'build-es' ? [
     },
   }],
 ] : []
+const browsers = [
+  'last 8 versions',
+  'safari > 8',
+  'firefox > 23',
+  'chrome > 24',
+  'opera > 15',
+  'not ie < 11',
+  'not ie_mob <= 11',
+]
 
 module.exports = {
-  presets: [
-    ['env', options],
-  ],
+  presets: [['env', {
+    targets: { browsers },
+  }]],
   plugins,
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -686,7 +686,7 @@ babel-plugin-transform-async-generator-functions@^6.24.1:
     babel-plugin-syntax-async-generators "^6.5.0"
     babel-runtime "^6.22.0"
 
-babel-plugin-transform-async-to-generator@^6.24.1:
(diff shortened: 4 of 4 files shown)


layershifter (changes requested):
We should define browser requirements for `preset-env`:

```
"targets": {
  "browsers": ["last 2 versions", "ie >= 11"]
}
```

thetric:
Suggestion for the browserlist: `last 2 versions, not ie < 11, not ie_mob <= 11`

fracmak:
Should throw in Firefox ESR

thetric:
Ok, so `last 2 versions, not ie < 11, not ie_mob <= 11, firefox esr, not op_mob < 37` should cover the most commonly used modern browsers. I additionally excluded older Opera Mobile versions (~2012).
Here a link to the [targeted browsers](http://browserl.ist/?q=last+2+versions%2C+not+ie+%3C+11%2C+not+ie_mob+%3C%3D+11%2C+firefox+esr%2C+not+op_mob+%3C+37)

package.json
```diff
@@ -46,6 +46,18 @@
...
+      "last 2 versions",
+       "not ie < 11",
+       "not ie_mob <= 11",
+       "firefox esr",
+       "not op_mob < 37"
+    ],
+    "test": [
+      "last 2 chrome versions"
+    ]
+  },
```
layershifter: Let's keep the Babel's config in single place and move this to `preset-env.js`.

thetric:
@layershifter I've amended the last commit to move the env config to `build/preset-env.js`

layershifter (approved):
I've update the browsers list.

### Why?

1. We're making the component library there that should work in the most part of modern browsers, [updated config](http://browserl.ist/?q=chrome+%3E+24%2C+safari+%3E+8%2C+opera+%3E+15%2C+firefox+%3E+23%2C+not+ie+%3C+11%2C+not+ie_mob+%3C%3D+11%2C+last+8+versions) gives 97.48% coverage.
2. I choosed these versions because they [natively support](https://caniuse.com/#search=requestAni) `requestAnimationFrame()` that we use widely.

@thetric thanks again for your work there
@levithomason your opinion?

thetric:
IMO it doesn't make sense for Babel to target browsers that aren't supported by the semantic ui css, see https://github.com/Semantic-Org/Semantic-UI#browser-support

thetric:
BTW, I cleaned up the commit history of this Pr

levithomason:
>BTW, I cleaned up the commit history of this Pr

Heads up, we squash all PRs to one commit automatically 👍

thetric:
What about my comment about the configured browser target of `babel-preset-env`? With the current config we would target _more_ and _older_ browsers than Semantic UI (chrome 25-61, firefox 24-51, android < 4.4, ...). Why shouldn't we align with them?

levithomason:
I know of large enterprise consumers that rely on IE 11.  The standard here is coverage and consumers opposed to alignment with SUI.

If anything, I would be prone to increase support.  However, we'll wait and see what our users say.

thetric:
Did you see the effectively supported browsers in [browsersl.ist](http://browserl.ist/?q=chrome+%3E+24%2C+safari+%3E+8%2C+opera+%3E+15%2C+firefox+%3E+23%2C+not+ie+%3C+11%2C+not+ie_mob+%3C%3D+11%2C+last+8+versions)? The current babel config transpiles even for **IE 5.5** and other old browsers with non-existing global usage (Opera mobile 10-12, Android browser 2.3 - 4, Opera 16 - 48)

Oops, just noticed I accidentally dropped `modules: false`.

levithomason:
Got it.  Here's the [list of browsers](http://browserl.ist/?q=last+8+versions%2C+safari+%3E+8%2C+firefox+%3E+23%2C+chrome+%3E+24%2C+opera+%3E+15%2C+not+ie+%3C+11%2C+not+ie_mob+%3C%3D+11).  We're still leaving out 3.25% of browsers which seems a little large to me.

brianespinosa:
> Got it. Here's the list of browsers. We're still leaving out 3.25% of browsers which seems a little large to me.

@thetric based on the above list of browsers that @levithomason shared, that will not work for the enterprise project I currently maintain. I am sure there will be others with similar issues. We have to support (unfortunately) back to IE10.

levithomason:

levithomason:
@brianespinosa if we screwed up support for you there, please open a PR restoring it.  We don't want to lose users, especially you!

thetric:
I compared the (unminified) build with the current browserslist and another build with my restricted one for more recent browsers and couldn't spot a difference - even when _explicitly including_ IE10. @brianespinosa So the current build should work fine even under IE10 :+1:

brianespinosa:
We have not updated to the latest version yet, partially because there's going to be a lot of work to update some custom styles for modals that we use that are going to significantly break from the upstream changes. So this is not breaking anything for us yet. I won't have any eyes on how this will function in IE10 for at least a few weeks.

@thetric if you've already tested this for IE10 then 👍

thetric:
Just compared the generated sources, not actually tested in a real browser

PR #2596, opened 2018-03-04
chore(package): dep updates

thetric:
Brings some dependency updates.

**DO NOT MERGE!** The updated eslint code style isn't fully applied, so this is WIP.

PR #2597, opened 2018-03-04
chore(package): Prettier integration

thetric:
As suggested by @layershifter in https://github.com/Semantic-Org/Semantic-UI-React/pull/2596#pullrequestreview-100998473 I have added `prettier` as a code formatter.

The [prettier docs](https://prettier.io/docs/en/options.html#print-width) recommends _against_  a max line length > 80, but this code base uses a print width of 120.

> **For readability we recommend against using more than 80 characters:**
>
>In code styleguides, maximum line length rules are often set to 100 or 120. However, when humans write code, they don't strive to reach the maximum number of columns on every line. Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.
>
>Prettier, on the other hand, strives to fit the most code into every line. With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,16 @@
+.github/
+
+# editor configs
+.idea/
+.vscode/
+
+
+coverage/
+
+dist/
+dll/
+docs/app/docgenInfo.json
+docs/app/menuInfo.jso
+docs/build/
+
+package.json
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "lint-staged": "^7.0.2",
     "mocha": "^3.5.0",
     "node-sass": "^4.5.3",
-    "prettier": "1.11.1",
+    "prettier": "1.12.0",
     "puppeteer": "^0.10.1",
     "raw-loader": "^0.5.1",
     "react": "^16.0.0",
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "leven": "^2.1.0",
     "mocha": "^3.5.0",
     "node-sass": "^4.5.3",
+    "prettier": "^1.11.1",
     "puppeteer": "^0.10.1",
     "raw-loader": "^0.5.1",
     "react": "^16.0.0",
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "leven": "^2.1.0",
     "mocha": "^3.5.0",
     "node-sass": "^4.5.3",
-    "prettier": "^1.11.1",
+    "prettier": "1.11.1",
     "puppeteer": "^0.10.1",
     "raw-loader": "^0.5.1",
     "react": "^16.0.0",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
     "deploy:docs": "gh-pages -d docs/build -m 'deploy docs [ci skip]'",
     "lint": "cross-env NODE_ENV=production eslint .",
     "lint:fix": "yarn lint --fix",
+    "prettier": "prettier --list-different \"**/*.{js,jsx,ts,tsx}\"",
+    "prettier:fix": "prettier --write \"**/*.{js,jsx,ts,tsx}\"",
+    "precommit": "lint-staged",
     "prerelease": "yarn lint && yarn tsd:lint && yarn tsd:test && yarn test && cross-env NODE_ENV=production yarn build",
     "postrelease": "yarn deploy:docs",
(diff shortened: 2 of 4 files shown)


layershifter:
Let's go with single quotes on JSX, it's a part of our codestyle and it will allow to reduce changes.

thetric:
Prettier doesn't support [single quotes in JSX](https://prettier.io/docs/en/options.html#quotes):

> Quotes in JSX will always be double and ignore this setting.

.eslintrc
```diff
@@ -1,27 +1,20 @@
 {
   "parser": "babel-eslint",
-  "extends": "airbnb",
+  "extends": ["airbnb", "prettier"],
```
layershifter: I think that we should stop use airbnb's preset

.eslintrc
```diff
@@ -34,6 +27,7 @@
     "react/jsx-filename-extension": [2, { "extensions": [".js"] }],
     "react/no-unused-prop-types": 0,
     "react/sort-comp": 0,
-    "react/require-default-props": 0
+    "react/require-default-props": 0,
+    "prettier/prettier": "error"
```
layershifter: Let's use numbers for rules config

.eslintrc
thetric: What are your reasons that you want to remove the `airbnb` preset?
I think that this is out of scope of the PR

levithomason:
Let's keep this PR just adding the config.  We can test it locally.  This will help in reviewing.  We can merge the reformat later.

Also, other PRs can merge master, get the config update, and format their branches.

.eslintrc
levithomason: Prettier only covers stylistic issues, we'll still need eslint and a good preset for non-style issues such as unused vars.

levithomason:
>Let's go with single quotes on JSX, it's a part of our codestyle and it will allow to reduce changes.

Yea, that sucks but it is what it is.  I pushed the point and got approval to submit a PR for the `singleQuote` option to respect JSX, however, jest was failing with cryptic errors and I gave up.  In the end, this is point of `prettier`.  Style issues just aren't worth the time.

thetric:
> Let's keep this PR just adding the config. We can test it locally. This will help in reviewing. We can merge the reformat later.

Updated the PR to only add prettier and two config files

.prettierignore
```diff
@@ -0,0 +1,16 @@
+.github/
```
levithomason, line 1: _Deleted my comment, not applicable._

package.json
```diff
@@ -28,6 +28,9 @@
     "deploy:docs": "gh-pages -d docs/build -m 'deploy docs [ci skip]'",
     "lint": "cross-env NODE_ENV=production eslint .",
     "lint:fix": "yarn lint --fix",
+    "prettier": "prettier --list-different \"**/*.{js,jsx,ts,tsx}\"",
+    "prettier:fix": "prettier --write \"**/*.{js,jsx,ts,tsx}\"",
```
thetric: Did you test it under Windows? In another project I had trouble with glob patterns as they tried to match files from `C:\` and not from the current directory. The fix was to prefix the glob pattern with `./`.

thetric, line 32: Did you test it under Windows? In another project I had trouble with glob patterns as they tried to match files from `C:\` and not from the current directory. The fix was to prefix the glob pattern with `./`.

levithomason: I did not.  Do you have a windows box for testing?  Perhaps @layershifter?

Note that I've escaped double quotes here, which means the globs will not be expanded by your shell (mac vs win) but rather by the node module that prettier uses under the hood, node-glob.  That module _should_ be platform agnostic, but I'm not 100% sure this is all correct.

Let me know how a Win test goes:

- edit a JS file, leaving a format mistake
- commit it (don't push)
- expect only that JS file to go through eslint and prettier
- repeat with a `ts` file, expecting tslint and prettier

thetric: Works on a Windows machine, too.

My trouble with the glob pattern under windows seems to be related to the quoting of the glob pattern. Using double quotes seems to fix it

levithomason:
Are we good to merge here @layershifter and @thetric?

thetric:
We could [upgrade prettier](https://prettier.io/blog/2018/04/11/1.12.0.html) before merging

thetric:
I've nothing to add. What about you, @layershifter?

layershifter:
@thetric Marked as ready to merge, great work :+1:

levithomason:

PR #2607, opened 2018-03-05
breaking(Pagination): rename `ariaLabel` prop `aria-label`

### LINKED ISSUE #2601: Pagination: invalid ARIA attribute `ariaLabel`
### Steps
Create a small pagination component.
When the component mounts, a warning is displayed.

### Expected Result
Not show the warning.

### Actual Result
Shows the warning.

### Version
* semantic-ui-react: 0.78.3
* react: 16.2.0
* react-dom: 16.2.0

### Testcase
https://codesandbox.io/s/9l90v9xnwp

AndreiEnache:
# BREAKING CHANGE

This PR renames `ariaLabel` prop that used in `Pagination` component to the native `aria-label` prop.

### Before

```jsx
<Pagination ariaLabel='A label' />
<Pagination prevItem={{ ariaLabel: 'A label' }} />
<Pagination.Item ariaLabel='A label' />
```

### After

```jsx
<Pagination aria-label='A label' />
<Pagination prevItem={{ 'aria-label': 'A label' }} />
<Pagination.Item aria-label='A label' />
```

----

Fixes #2601.

It also requires update to latest version of [`babel-plugin-transform-react-handled-props`](https://github.com/layershifter/babel-plugin-transform-react-handled-props) because previos version have problems with parse of string properties.

### DIFF
--- a/docs/app/Layouts/HomepageLayout.js
+++ b/docs/app/Layouts/HomepageLayout.js
@@ -107,6 +107,12 @@ DesktopContainer.propTypes = {
 class MobileContainer extends Component {
   state = {}
 
+  handlePusherClick = () => {
+    const { sidebarOpened } = this.state
+
+    if (sidebarOpened) this.setState({ sidebarOpened: false })
+  }
+
   handleToggle = () => this.setState({ sidebarOpened: !this.state.sidebarOpened })
 
   render() {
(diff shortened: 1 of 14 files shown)


layershifter:
Let's modify `Pagination` component, it should use `aria-label` instead of `ariaLabel`.

AndreiEnache:
I might be wrong, but the problem is not with the `Pagination` component.
The pagination component uses aria-label instead of ariaLabel here:
*Taken from `Pagination.js` line 138*
```
<Menu {...rest} aria-label={ariaLabel} pagination role='navigation'>
```
While going further down the road, I saw that PaginationItems are created. Looking into `PaginationItem.js`, I saw that it actually creates a modified `MenuItem`
*Taken from `PaginationItem.js` line 79*
```
MenuItem.create(this.props, {
      defaultProps: {
        active,
        disabled,
        'aria-current': active,
        'aria-label': ariaLabel,
```
So the PaginationItem creates a `MenuItem `and passes props to it.
In `MenuItem`, I saw that the unhandled props are taken `const rest = getUnhandledProps(MenuItem, this.props)` and spread as props to the element 
*Taken from `MenuItem.js` line 134/141*
```
<ElementType {...rest} className={classes} onClick={this.handleClick}>
```
After modifying the above line to handle ariaLabel as a special prop and manually inserting it into the element as aria-label [(as per react docs)](https://reactjs.org/docs/dom-elements.html), the github pages build stopped giving the warning.

Again, I might be wrong, it's my first pull request, but I did try to understand what happens.

layershifter:
@AndreiEnache in fact, problem is with-in `Pagination` component. It uses `ariaLabel`  prop, but it should use the native `aria-label`, as we downpass all unhandled props `ariaLabel` will be passed to `Menu.Item`.

AndreiEnache:
@layershifter You're right!
Looking through your commit now I understand what the actual issue was.
I identified the fact that unhandled props are being spread into the component but I thought it could be handled into the component in which they are passed rather than in the pagination; your approach is indeed much cleaner.

levithomason:

PR #2615, opened 2018-03-07
fix(Tab): Added a prop menuAligned to Tab and removed an invalid prop

iannbing:
As proposed in #2430, a new prop menuAligned with default value 'left' is added to Tab, and the invalid prop aligned passed to the Menu inside of Tab is removed.


iannbing:
Apparently there's been some discussion and contribution happening.

Issue #2614, opened 2018-03-07
Responsive: incompatible change using document.documentElement.clientWidth


andrewferk:
### Steps

1. Ensure scrollbars are enabled for your OS or browser
2. Render `<Responsive maxWidth={767}>Only mobile</Responsive>`
3. Set the window width to 768px

### Expected Result

The "Only mobile" text **does not** render/display.

### Actual Result

The "Only mobile" text **does** render/display.

### Version

0.78.3

### My analysis for the issue

This is due to #2531, which updated `Responsive` to use `document.documentElement.clientWidth`, instead of `window.innerWidth`. They are [very technically different measurements](https://stackoverflow.com/a/7205786/651983), and the change has introduced the unexpected result I showed above. This is because `clientWidth` does not include the width of the scrollbar in the measurement, so it appears the width is `768px - 17px`. Also, I have many tests that rely on stubbing `window.innerWidth`, and 0.78.3 has broken them all.

### My proposal to fix the issue and more

1. Revert #2531 and use `window.innerWidth` as the default way of measuring width for the `Responsive` component. Semantic UI's responsive breakpoints should take into account the scrollbar, which `document.documentElement.clientWidth` does not. Without reverting back, the use of `<Responsive {...Responsive.mobileOnly} />`, etc. will not work as expected.
2. Add a new property to `Responsive` that allows developers to use a custom measurement for the width: `<Responsive getWidth={() => document.documentElement.clientWidth} />`. This would allow @Autarc measure what he wants without introducing the backwards incompatible change.
3. Fix the false-positive tests in `Responsive-test.js`. I have noticed some of the `Responsive` tests aren't testing the right thing, and they are passing when they should not. I'll fix this and explain what was wrong.
4. Only set state when width actually changes. Right now, every `window.resize` event causes `Responsive` to set the width, causing unneeded renders. This is mostly notable on mobile devices which hide and show the URL bar at the top on vertical scrolling. Scrolling on mobile devices that cause the URL bar to show/hide, triggers a `window.resize` event and `Responsive` renders, when not needed.

layershifter:
Looks awesome, feel free to open PR.

levithomason:
Fixed in #2621

PR #2621, opened 2018-03-08
feat(Responsive): add getWidth prop, fix perf

andrewferk:
Implements #2614 

1. Revert #2531, as it breaks `Responsive`'s integration with Semantic UI's breakpoints. This ensures `Responsive` uses `window.innerWidth` as the default calculation for screen width. `document.documentElement.clientWidth` does not use the scrollbar in its calculation.

2. Fix false-positive tests in `Responsive-test.js`. [`chai-enzyme`](https://github.com/producthunt/chai-enzyme) provides assertion helpers [`blank()`](https://github.com/producthunt/chai-enzyme#blank) and [`present()`](https://github.com/producthunt/chai-enzyme#present), which were being used as if they were opposites. They are not opposites. For a wrapper to be `empty()`, it must also be `present()`. Tests have been updated to use `not.empty()` and `empty()`, which are opposites. Also, the resizing test was using enzyme's `shallow()`, but `mount()` was needed to support state change and updating render.

3. Add optional `getWidth` property to `Responsive`. The property is a function that returns the width of the screen. This property defaults to a function that returns `window.innerWidth` for browsers and `0` for non-browsers. @Autarc or anyone else wanting to use `document.documentElement.clientWidth` as discussed in #2531 can do so using `<Responsive getWidth={() => document.documentElement.clientWidth} />`.

4. Prevent `Responsive` from updating render when `window.resize` is triggered, but width does not change. `Responsive` currently updates on mobile devices when scrolling and the url bar displays and hides. This happens because the height of the window is changing, triggering `window.resize`. This update allows `Responsive` to ignore these events. `onUpdate` will still be invoked.

### DIFF
--- a/src/addons/Responsive/Responsive.js
+++ b/src/addons/Responsive/Responsive.js
@@ -60,9 +60,7 @@ export default class Responsive extends Component {
   constructor(...args) {
     super(...args)
 
-    // Measure the root element dimension to handle gesture transitions on iOS safely
-    // https://github.com/Semantic-Org/Semantic-UI-React/pull/2531
-    this.state = { width: isBrowser() ? document.documentElement.clientWidth : 0 }
+    this.state = { width: isBrowser() ? window.innerWidth : 0 }
   }
 
   componentDidMount() {
@@ -114,7 +112,7 @@ export default class Responsive extends Component {
 
   handleUpdate = (e) => {
     this.ticking = false
-    const width = document.documentElement.clientWidth
+    const width = window.innerWidth
 
     this.setSafeState({ width })
     _.invoke(this.props, 'onUpdate', e, { ...this.props, width })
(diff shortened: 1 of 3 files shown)


andrewferk:
Maybe the way I did the default `getWidth` doesn't look the best in the docs.

image

I'd be more than welcome to remove it from the `defaultProps` and implement it another way.

layershifter (approved):
I've added some styling fixed. LGTM

@andrewferk thanks for the awesome work there :+1:

src/addons/Responsive/Responsive.js
```diff
@@ -46,6 +53,10 @@ export default class Responsive extends Component {
     onUpdate: PropTypes.func,
   }
 
+  static defaultProps = {
+    getWidth: () => (isBrowser() ? window.innerWidth : 0),
```
levithomason, line 57: I would think if there's no browser at all, `null` might be more appropriate.  Thoughts?

levithomason:

PR #2623, opened 2018-03-08
fix(Modal): fix positioning when "as" prop is a component

trevorharwell:
I found that when I use `<Modal as={SomeComponent} />` in React 16, the modal positioning broke.  This is an easy and straight forward way to remedy the problem.  Let me know if there is an alternative approach that I should be taking.

Re Testing: I didn't see any tests that covered this particular function so I didn't add any.  If that is desired I can try to figure something out.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -21,6 +21,7 @@ import ModalHeader from './ModalHeader'
 import ModalContent from './ModalContent'
 import ModalActions from './ModalActions'
 import ModalDescription from './ModalDescription'
+import Ref from '../../addons/Ref'
 
 const debug = makeDebugger('modal')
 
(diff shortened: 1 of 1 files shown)


trevorharwell:
Now that I have had some time to think about this, I believe this could be easily solved by wrapping the content element in a `Ref` component and using that to set the ref.  Would that be something you guys would approve?

trevorharwell:
@layershifter Thanks for getting back to me.  I updated the PR per your request.  Let me know if there is anything else I should do.

FYI we use SUIR for a large scale web application at my company and we love it! Thanks for such a great library!

levithomason (approved):

levithomason:
@layershifter this looks OK to me for a short-term fix, merge at will.  Perhaps we should add these TODOs to #2306 so we remember to clean them up.

PR #2632, opened 2018-03-11
fix(Pagination): call `onPageChange` only when change is occurred

### LINKED ISSUE #2627: fix(Pagination): disable active pagination.item
**Steps to Reproduce**
1. Do something
2. Do something else.

**Expected**
Pagination should disable the current active pagination.item so that additional clicks won't trigger additional method calls for the same content. 

**Result**
A check can be made in the handler but it would be good to have a visual representation, to denote there's no use in clicking on a page you are already on.

**Testcase**
undefined

layershifter:
Fixes #2627.

### DIFF
--- a/src/addons/Pagination/Pagination.js
+++ b/src/addons/Pagination/Pagination.js
@@ -112,9 +112,14 @@ export default class Pagination extends Component {
 
   static Item = PaginationItem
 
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2633, opened 2018-03-11
fix(Responsive): fix types in typings

### LINKED ISSUE #2608: Responsive: wrong type of `width` in ResponsiveOnUpdateData
In Responsive.d.ts, the type of the width property in the ResponsiveOnUpdateData interface is "string" but this should be "number".

layershifter:
Fixes #2608.

### DIFF
--- a/src/addons/Responsive/Responsive.d.ts
+++ b/src/addons/Responsive/Responsive.d.ts
@@ -28,12 +28,12 @@ export interface ResponsiveProps {
 }
 
 export interface ResponsiveOnUpdateData extends ResponsiveProps {
-  width: string;
+  width: number;
 }
 
 export interface ResponsiveWidthShorthand {
-  minWidth: number;
-  maxWidth?: number;
+  minWidth?: number | string;
+  maxWidth?: number | string;
 }
 
 declare class Responsive extends React.Component<ResponsiveProps, {}> {


src/addons/Responsive/Responsive.d.ts
```diff
@@ -28,12 +28,12 @@ export interface ResponsiveProps {
...
 
 export interface ResponsiveOnUpdateData extends ResponsiveProps {
-  width: string;
+  width: number;
 }
 
 export interface ResponsiveWidthShorthand {
-  minWidth: number;
-  maxWidth?: number;
+  minWidth?: number | string;
```
layershifter, line 35: `minWidth` is actually optional

src/addons/Responsive/Responsive.d.ts
```diff
@@ -28,12 +28,12 @@ export interface ResponsiveProps {
...
+  maxWidth?: number | string;
```
layershifter, line 36: `minWidth` and `maxWidth` can be strings, see props definitions

src/addons/Responsive/Responsive.d.ts
MJLHThomassen-Eurocom: This is a potentially breaking change (everyone who expects minWidth to be there and has coded against it will now potentially run into linter errors stating that minWidth is optional and should change their code to check for this).

levithomason:

PR #2647, opened 2018-03-15
fix(Input): add missing accept prop

fastman:
Fixes `accept` not being passed down to input elements by adding it to `htmlInputAttrs`.


layershifter:
Awesome, thanks :+1:

levithomason:

PR #2650, opened 2018-03-16
docs(Search): debounce search change handler

ryanflowers:
Added a debounce around the search change event as to not execute the callback on every key press. Dropped the timeout to be just under the debounce. This includes standard and category search examples.


levithomason:

PR #2654, opened 2018-03-17
fix(Search): strict validation to avoid errors when not using title

eightnoteight:
Hi,
at here
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Search/Search.js#L375

we are using setValue(result.title), if the title is not a mandatory field then value will be set to undefined, and for my use case i'm getting error here, for checking length attribute of undefined.

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Search/Search.js#L496

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -71,7 +71,7 @@ export default class Search extends Component {
      */
     results: PropTypes.oneOfType([
       PropTypes.arrayOf(PropTypes.shape(SearchResult.propTypes)),
-      PropTypes.object,
+      PropTypes.shape(SearchCategory.propTypes),
     ]),
 
     /** Whether the search should automatically select the first result after searching. */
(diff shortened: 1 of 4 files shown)


levithomason:
I'm afraid I don't quite understand the problem that is being solved.  Could you add a failing test or fork https://codesandbox.io/s/2l3n74j9y and demonstrate the issue?

eightnoteight:
https://codesandbox.io/s/y29ol2l22z

video of issue being reproduced:
https://youtu.be/0_vlkmuIxIM

enter any letter and something0 will be search result, now if you select the first result i.e "something0"
internally the value is set to `result.title` which is `undefined` as I haven't defined in my result object, the next time when i clicked on the input it tries to rerender at which time we are trying to check `currentValue.length` this raises an error.

the issue as a whole occurred as we are expecting title to be mandatory field of result object, as you can render a result without mentioning title. 

having a stricter validation to make the title as mandatory field will avoid such issues in future.

levithomason:
I see, thanks!

PR #2663, opened 2018-03-19
chore(package): restore babel option `modules: false` for ESM builds

### LINKED ISSUE #2661: build: restore `modules: false` for the esm build
#2594 accidentally removed the `modules: false` option from the `babel-preset-env`. Restore the option to preserve the old behavior for esm builds

thetric:
Fixes #2661.


layershifter:
Thanks :+1:

levithomason:

PR #2678, opened 2018-03-24
fix(Dropdown): SearchCleared not called when multiple

jdetle:
Hi!
I was running into an issue where a setState was getting called onEnter key presses, this gets rid of the error and all the tests still pass. Can anyone with a better understanding of the codebase help me?

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -617,17 +617,18 @@ export default class Dropdown extends Component {
 
   selectItemOnEnter = (e) => {
     debug('selectItemOnEnter()', keyboardKey.getName(e))
-    const { search } = this.props
+    const { search, multiple } = this.props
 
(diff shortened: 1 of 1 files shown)


levithomason:
I have a pretty good understanding of the codebase.  However, I'd need a completed issue template to better understand the issue.  Want to post one here?

jdetle:
Okay, @levithomason. I've spent a fair bit of time trying to minimally reproduce this but I'm having some problems and need to get to working on other things.

The problem I'm running into isn't acutally related to the `multiple` prop so this is a bad fix. The problem is that I'm using this dropdown as a connected component with ReduxForm. When a value is selected, my redux state is changed and a rerender is triggered. This causes the dropdown list to unmount, and clearSearchQuery performs a setState on an unmounted component. 

I don't know what the most succinct solution is, I dont think there is a way to get around this error using the lifecycle methods of Dropdown. But I'm closing this and happy to move this over to an issue later when I have more time to produce a succinct reproducible example.

PR #2682, opened 2018-03-26
docs(readme): add lawreview to built with list

barexam:
# Added "- https://www.lawreview.com" to line 82 of https://github.com/Semantic-Org/Semantic-UI-React/README.md; it's frontend is being built using Semantic UI React.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ These great products are built on Semantic UI React. Add yours [here][22].
 - https://www.hurriyetoto.com
 - https://173app.com
 - https://disten.se
+- https://www.lawreview.com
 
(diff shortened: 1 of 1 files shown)


levithomason:
Thanks for using Semantic-UI-React, @jdhakert!  When I navigate to https://www.lawreview.com, I get the following error page:

image

Let's get it resolved before merging.

thetric:
The linked page still doesn't work, I got an Error 525: SSL Handshake failed

levithomason:
Same here.  Will close for now.  Please open a new PR if/when the site is up.  Thanks for using SUIR!

Issue #2550, opened 2018-02-20
Update SUIR to be compatible with SUI 2.3


brianespinosa:
Creating this as a container for referencing other issues that are likely to pop up with the release of SUI 2.3 which happened today. Feel free to edit this and add related issues below for tracking. 

Notes about what is new in this release:
https://semantic-ui.com/introduction/new.html

Full CHANGELOG:
https://github.com/Semantic-Org/Semantic-UI/blob/master/RELEASE-NOTES.md

Known issues so far. Please edit this list to add or link to issues:
- [x] Fix modal placement (#2549) 
- [x] Update `Icon` component to use new FontAwesome 5.0 icons.
- [x] Add support for new Transition names `zoom` and `glow`
- [x]  Modal support to disable vertical centering ([missing from the above linked notes. See examples.](https://semantic-ui.com/modules/modal.html#/examples))
- [ ] Update Search logic for `fullTextSearch: 'exact'` vs `fuzzy`

brianespinosa:
Updated the list with a few more and also added a link to the full CHANGELOG. There may be one or two I missed. And with how we render popups via a Portal, I am not sure we have to worry about the update allowing you to trigger a popup from other blocks of markup.

levithomason:
I think we're ready to go here.  Changing labels.

davidhenley:
Any update on this? Modal is not usable.

brianespinosa:
@davidhenley any issues tagged with `help wanted` are open for anyone to submit PRs to fix. You are welcome to do so. Any info related to this and all issues will be shown above in the issue history. Based on reading this history, it looks like there are no open PRs yet.

Errolyn:
Any chance your Modal documentation can be updated to reflect the modal placement issue?

amoerie:
In any case, it seems wise to add a warning on the docs homepage that Semantic UI React is not officially compatible with Semantic UI 2.3 yet, and that using the 2.3 css in combination with the latest react version is at your own risk until this issue is resolved.

brianespinosa:
Valid point @Riddlerat, @amoerie. Since this issue was not picked up quickly for work, I have submitted a quick PR to update the README.md in #2644 linking here to alert people until we get a patch on this.

Fingers crossed, I am hoping to at minimum get the modal working with 2.3 over the weekend.

layershifter:
[...]

PR #2690, opened 2018-03-28
breaking(Modal): Fix vertical position with SUI 2.3

brianespinosa:
This is part of the work for #2550. There are a few things here to look at.

1. Modal now has a `centered` prop that can be set to false which positions the modal to the top of the viewport instead of the center. The default centered modal follows the same logic as SUI.
2. Modal now renders in a flexbox wrapper the same way SUI does with modals, and no longer applies negative top margin divided by height. The logic for this calculation was left commented out since we may have to support legacy browsers... inline comment about this in the code with link for reference.
3. Portal had to get updated to also pass style prop to the portal that we create. This means we have to manually convert a style object to a CSS string since we are creating the portal with vanilla JS. Instead of taking the time to try to handle all cases here manually, I added a minimal package that does only this.

There may be more to do but I at least wanted to get this up so we could get some eyes on it. I think we may be able to ship out an update that moves us for supporting 2.3.x soon after this one.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -2,6 +2,7 @@ import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React, { Children, cloneElement } from 'react'
 import ReactDOM from 'react-dom'
+import toInlineStyle from '@f/to-inline-style'
 
 import {
   AutoControlledComponent as Component,
@@ -122,6 +123,9 @@ class Portal extends Component {
 
     /** Controls whether the portal should be prepended to the mountNode instead of appended. */
     prepend: PropTypes.bool,
+
+    /** Any inline styles to the Portal container. */
+    style: PropTypes.object,
 
     /** Element to be rendered in-place where the portal is defined. */
     trigger: PropTypes.node,
@@ -344,7 +348,7 @@ class Portal extends Component {
     if (!this.state.open) return
     debug('renderPortal()')
 
-    const { children, className, eventPool } = this.props
+    const { children, className, eventPool, style } = this.props
 
     this.mountPortal()
 
@@ -352,6 +356,7 @@ class Portal extends Component {
     if (!isBrowser()) return null
 
     this.rootNode.className = className || ''
+    this.rootNode.style = style ? toInlineStyle(style) : ''
 
     // when re-rendering, first remove listeners before re-adding them to the new node
     if (this.portalNode) {
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -2,7 +2,6 @@ import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React, { Children, cloneElement } from 'react'
 import ReactDOM from 'react-dom'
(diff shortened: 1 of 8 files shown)


brianespinosa:
Ran out of time tonight to update the tests. Will try to get some eyes on them early tomorrow... but I likely will not have time to get back in on them until the evening.

src/addons/Portal/Portal.js
```diff
@@ -344,14 +348,15 @@ class Portal extends Component {
...
-    const { children, className, eventPool } = this.props
+    const { children, className, eventPool, style } = this.props
 
     this.mountPortal()
 
     // Server side rendering
     if (!isBrowser()) return null
 
     this.rootNode.className = className || ''
+    this.rootNode.style = style ? toInlineStyle(style) : ''
```
layershifter: Nice catch!

src/modules/Modal/Modal.js
```diff
@@ -364,6 +372,7 @@ class Modal extends Component {
         onMount={this.handlePortalMount}
         onOpen={this.handleOpen}
         onUnmount={this.handlePortalUnmount}
+        style={{ display: 'flex !important' }}
```
layershifter: I see a problem there, React doesn't support `important` keyword for inline styles, see facebook/react#1881. Take a look on `Dimmer`'s PR #2656 for a workaround.

layershifter (changes requested):

src/modules/Modal/Modal.js
brianespinosa: Ah. I guess it is working in this case because this is being applied to a rootNode of a portal. Or possibly because the `toInlineStyle` package is handling it. I'll take a look today.

brianespinosa: @layershifter Updated. Technically it was working because we were passing those attributes as a string. This will keep the semantics the same way as we are handling this use case in `Dimmer`.

src/addons/Portal/Portal.js
brianespinosa: We also no longer need this external library based on the refactor requested below.

src/addons/Portal/Portal.js
```diff
@@ -344,14 +347,15 @@ class Portal extends Component {
...
-    const { children, className, eventPool } = this.props
+    const { children, className, eventPool, style } = this.props
 
     this.mountPortal()
 
     // Server side rendering
     if (!isBrowser()) return null
 
     this.rootNode.className = className || ''
+    this.rootNode.style = style || ''
```
levithomason, line 358: I think this is susceptible to users passing style objects to the `<Portal style={...} />`.  Let me know if I'm misreading here.

brianespinosa: @levithomason correct. It is. Is there a reason you see that we should not allow that? If we do not want to allow that, is there another way you would prefer we do this?

levithomason: Hm, I think I was tripping out that day... Apologies, continue!

levithomason:
@layershifter @brianespinosa this looks good to me.  Please merge at will if you are good with it 👍

Issue #2684, opened 2018-03-27
fix(Container): on the website error: Container is not defined


Abusoss:
**Steps to Reproduce**
1. Go to https://react.semantic-ui.com/elements/container
2. Edit any exemples with the edit code widget(just a "Space" somewhere is enough)

**Expected**
Changes in the view

**Result**
error: 
"Container is not defined "

At the bottom of any example

**Testcase**
Not needed the probleme is on the the website.
Ps: no adblockers

levithomason:
It seems the recent update to the regex for the code editor has broken things.  This comment at the top of the file is confusing the parser:

```jsx
/* eslint-disable max-len */
```

levithomason:
See the change [here](https://github.com/Semantic-Org/Semantic-UI-React/pull/2481/files).  We need to update this to properly grab the body of the example.

levithomason:
Fixed in #2706

levithomason:
Released in `semantic-ui-react@0.80.0`.

PR #2706, opened 2018-04-06
fix(docs): Container is not defined (#2684)

rtmruczek:
Looks like there was an issue with the regex that pulls the imports out of the component source. The previous regex assumed the imports would always be the first line, but in some cases we have some eslint disables as the first line. I modified the regex to look for previous lines, if there are any.

The editors will still break if someone is a jerk and types an import statement in the middle of the editor, though.


levithomason:
Thanks, this seems better than what we currently have.

levithomason:

PR #2707, opened 2018-04-06
chore(package): use keyboard-key package

levithomason:
We'll be extracting more of our `/lib` to separate packages for use in other frameworks.  This PR abstracts the `keyboard-key` util.


brianespinosa:
@levithomason Is the plan to make this into a monorepo that exports read-only versions of these lib packages so they can be imported elsewhere?

levithomason:
Yes, that is the plan for the components.  I hadn't considered it for the lib, but that also may make sense for some things.

Did you have a proposal on this by chance?

brianespinosa:
> Did you have a proposal on this by chance?

No proposal on this yet. I am still exploring options on doing this in my company's current projects. Depending on how this gets set up there's potentially a lot of infrastructure to set up... but once it is done you get the benefits of both the many/singe repo approach. I don't have any experiences to share at this time but once I've done some trial runs and a POC in my own project I can open up a RFC for this on our project and we can have a discussion about our needs and what further info I might have from my experiences.

layershifter:
@levithomason I've finished this PR, take a look :cat:

levithomason:
Excellent, we can merge this now 💪 

Moving forward, I think we might want to consider making our lib a mono repo with yarn workspaces rather than having literal separate repos.  Thoughts?

levithomason:

PR #2724, opened 2018-04-17
docs(List): fix duplicate Types sections

levithomason:
The List's `Variations` section is mislabeled and is duplicating `Types`.


levithomason:

PR #2739, opened 2018-04-24
docs(Table): rename TableExampleStriped example

### LINKED ISSUE #2737: Clicking on examples in docs moves and shows different documentation component
Clicking on this particular documentation component moves you to another component - which has conflicting html id (also for `Striped`). That's because files have same name but reside in different folders. We should either rename those files like `TableExampleStriped` -> `TableTypesExampleStriped` & `TableVariationsExampleStriped` with corresponding fixed in names, or we could add folder names in anchorName generator [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js#L63)

image

layershifter:
Fixes #2737.


levithomason:

Issue #2600, opened 2018-03-05
Dropdown Search onOpen fires 2 times when clicking on Icon


DennisKo:
### Steps
Create a Dropdown with search component. Add onOpen handler. Click on the arrow on the right to open it. 

### Expected Result
It fires once.
### Actual Result
It fires twice.
### Version
0.78.2

### Testcase
https://codesandbox.io/s/6nx916460r
click on right arrow to open - look in the console

willb335:
This seems to be related to the search prop.  I'll look a bit deeper into it.  It seems to me that it may have something to do with this bit of code in componentDidUpdate
```js
  const openable = !search || (search && minCharacters === 1)

        debug('mouse is not down, opening')
        if (openOnFocus && openable) this.open()
```

willb335:
Working on this at #2744

DennisKo:
fixed in https://github.com/Semantic-Org/Semantic-UI-React/pull/2744

PR #2744, opened 2018-04-26
fix(Dropdown): stop onOpen from firing twice on icon click

willb335:
Connects #2600

onOpen is firing twice when Dropdown icon is clicked with a ```search``` prop

The change is at line 426.  The other changes are from Prettier

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -43,10 +43,7 @@ export default class Dropdown extends Component {
     as: customPropTypes.as,
 
     /** Label prefixed to an option added by a user. */
-    additionLabel: PropTypes.oneOfType([
-      PropTypes.element,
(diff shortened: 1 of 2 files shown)


levithomason:
Great fix, thanks!

Let's also get a test that asserts the callback is `.calledOnce()`.  You can find tests in `Dropdown-test.js`.  We `describe()` tests by prop, so look for `describe('onOpen', ...)` (around line 1280).  

There are many examples of using `.calledOnce` in the file you can refer to.  We only need one test saying that `it('is called once when the icon is clicked', ...)`.  It would also be superb to leave a comment link to the GitHub issue there.

willb335:
The new test in Dropdown-test.js starts at line 1249.  The new test fails as expected without the changes in the first commit.

levithomason:

PR #2748, opened 2018-04-28
fix(Checkbox): match DOM checkbox behavior

### LINKED ISSUE #2587: Checkbox: doesn't toggle its state, when using only the keyboard
### Steps
1) Create a toggle function with a setState that uses a function as updater.
2) Bind the function to a Checkbox onChange
3) Use tab and space bar to select and check the checkbox

### Expected Result
Checkbox should call toggle and toggle the check state.
### Actual Result
Toggle function is called twice, so the check state is not being toggled.
### Version
0.78.3

### Testcase
https://codesandbox.io/s/23wm962zkn

Thank you!

levithomason:
Fixes #2587 

A controlled Checkbox did not fire onClick / onChange when pressing spacebar after focus. This was strictly when a functional state updater was used.

### Changes

I've commented on relevant changes to help separate the prettier updates from the actual code changes.

1. Change Checkbox handlers so that mouse down, mouse up, click, and change events are identical to the DOM checkbox
1. Add example comparing a vanilla DOM checkbox to our Checkbox with an event log
1. Rearrange "Remote" checkbox example from "States" to "Usage" and rename "Controlled"

### TODO

Help much appreciated!  Make PRs against `fix/checkbox-spacebar` instead of master.

- [x] Add/fix tests _(Note, the new example works but the tests don't seem to think so.)_


src/modules/Checkbox/Checkbox.js
```diff
@@ -242,17 +226,17 @@ export default class Checkbox extends Component {
...
+        onMouseUp={this.handleMouseUp}
       >
         <input
           {...htmlInputProps}
           checked={checked}
           className='hidden'
           id={id}
           name={name}
-          onClick={this.handleInputClick}
+          onChange={this.handleInputClick}
```
levithomason: Relevant change, remove duplicate click firing.

src/modules/Checkbox/Checkbox.js
```diff
@@ -242,17 +226,17 @@ export default class Checkbox extends Component {
       <ElementType
         {...rest}
         className={classes}
-        onClick={this.handleContainerClick}
         onChange={this.handleContainerClick}
         onMouseDown={this.handleMouseDown}
+        onMouseUp={this.handleMouseUp}
```
levithomason, line 249: Relevant change

src/modules/Checkbox/Checkbox.js
```diff
@@ -194,6 +181,15 @@ export default class Checkbox extends Component {
...
   }
 
+  handleMouseUp = (e) => {
+    debug('handleMouseUp()')
+    const { checked, indeterminate } = this.state
+
+    _.invoke(this.props, 'onMouseUp', e, { ...this.props, checked: !!checked, indeterminate: !!indeterminate })
+
+    this.handleClick(e)
+  }
```
levithomason, line 196: Relevant change, click is fired on mouse up.

src/modules/Checkbox/Checkbox.d.ts
```diff
@@ -61,6 +61,14 @@ export interface CheckboxProps {
...
   onMouseDown?: (event: React.MouseEvent<HTMLInputElement>, data: CheckboxProps) => void;
 
+  /**
+   * Called when the user releases the mouse.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and current checked/indeterminate state.
+   */
+  onMouseUp?: (event: React.MouseEvent<HTMLInputElement>, data: CheckboxProps) => void;
+
```
levithomason, line 73: Relevant change

layershifter:
I've checked this branch, `Checkbox` doesn't work at all, at example's log I have:
```
15:25:52: MouseUp       { checked: false } // click again
15:25:52: MouseDown     { checked: false }
15:25:51: MouseUp       { checked: false } // first click
15:25:51: MouseDown     { checked: false }
```

layershifter:
@levithomason I've fixed tests and replaced some of them with new, also updated example.

There is a single failing conformance test. I think that we can disable it all, what you think?

levithomason:
I've fixed the last broken test.  The Checkbox now has full parity with the HTML checkbox.  See the new example for confirmation: http://localhost:3000/modules/checkbox/#usage-dom-comparison

The onChange, onClick, onMouseDown, and onMouseUp events are fired in the same order and with the same arguments as a vanilla HTML checkbox.

PR #2750, opened 2018-04-30
chore(package): use Babel 7

### LINKED ISSUE #2603: build: upgrade to Babel 7
Upgrade to babel 7, see https://github.com/Semantic-Org/Semantic-UI-React/pull/2596#pullrequestreview-100998473

layershifter:
Fixes #2603.

This PR:
- switch package to use Babel 7 
- use release version of Gulp 4


browser/index.html
```diff
@@ -1,45 +0,0 @@
...
-      </Container>
-    )
-
-    const mountNode = document.createElement('div')
-    document.body.appendChild(mountNode)
-
-    ReactDOM.render(<App />, mountNode)
-  </script>
-</body>
-</html>
```
layershifter, line 45: Cruft file

package.json
```diff
@@ -113,10 +113,9 @@
     "express": "^4.15.4",
     "faker": "^4.1.0",
     "gh-pages": "^1.0.0",
-    "gulp": "gulpjs/gulp#4.0",
-    "gulp-html-replace": "^1.6.2",
```
layershifter, line 118: Cruft plugin

browser/index.html
levithomason: This file allowed manual testing to ensure the UMD build worked.  It was originally created to verify that the codepen would work as codepen used the UMD bundle.  Since we now use codesandbox and it doesn't need a UMD, it is probably safe to remove.

In fact, I wonder if we anyone actually uses the UMD bundle at all.  It would simplify and speed things up if we removed it.

docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -15,7 +15,7 @@ import ComponentControls from '../ComponentControls'
 import ComponentExampleTitle from './ComponentExampleTitle'
 
 const babelConfig = {
-  presets: ['es2015', 'react', 'stage-1'],
+  presets: ['es2015', 'react', ['stage-1', { decoratorsLegacy: true }]],
```
levithomason: Can we use `babel-preset-env` in the browser too?

docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -442,18 +437,10 @@ class ComponentExample extends Component {
...
           </Grid.Row>
 
-          <Grid.Row columns={1}>
-            {children && (
-              <Grid.Column style={childrenStyle}>
-                {children}
-              </Grid.Column>
-            )}
-          </Grid.Row>
+          <Grid.Row columns={1}>{children && <Grid.Column style={childrenStyle}>{children}</Grid.Column>}</Grid.Row>
```
levithomason: Perhaps we should shorten our line length.  This is quite hard to read.

```jsx
<Grid.Row columns={1}>{children && <Grid.Column style={childrenStyle}>{children}</Grid.Column>}</Grid.Row>
```

This line is 118 columns, maybe go with 100 columns?

```jsx
<Grid.Row columns={1}>
  {children && <Grid.Column style={childrenStyle}>{children}</Grid.Column>}
</Grid.Row>
```

levithomason (approved):
See suggestions, good to merge by me!

browser/index.html
layershifter: I'm think that we will drop UMD build in near future :+1:

levithomason:

PR #2752, opened 2018-04-30
feat(Button): add role prop

### LINKED ISSUE #2751: Button role="button" is not configurable
When i try to use a button as an anchor link using the "as" property `<Button as="a" role="link" ...>` the role of the generated anchor tag is still button.

in Button.js we see below code , may be role can also be accepted as a prop
` ref={this.handleRef}
        role='button'
        tabIndex={tabIndex}`

levithomason:
Fixes #2751

We did not allow setting the `role` prop.  Now we do.

### DIFF
--- a/src/elements/Button/Button.d.ts
+++ b/src/elements/Button/Button.d.ts
@@ -1,12 +1,6 @@
 import * as React from 'react';
 
-import {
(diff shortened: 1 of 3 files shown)


levithomason:

Issue #2747, opened 2018-04-28
Update to React 16


levithomason:
React 16 offers some updates that we should implement, however, we'll need to drop support for React 15 when doing so.  Let's batch these changes into an upcoming major release:

- [x] #2746 Portal to use `ReactDOM.createPortal` (https://reactjs.org/docs/portals.html)
- [ ] Refactor and simplify Accordion to return an array of panels (https://reactjs.org/docs/react-component.html#render)
- [ ] Refactor and simplify Breadcrumb to return an array of sections/dividers (https://reactjs.org/docs/react-component.html#render)
- [ ] #2306 Fix ref issues using `forwardRef` (https://reactjs.org/docs/forwarding-refs.html)
- [x] #2842 Use `PureComponent` for `Icon` and `Flag`

@Semantic-Org/react-maintainers feel free to add to this list at will...

layershifter:
Added `Breadcrumb` to list, there is quite [unfriendly](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.79.1/src/collections/Breadcrumb/Breadcrumb.js#L43) code.

packetstracer:
Maybe you should add a Fix for unsafe lifecycle methods for asynchronous rendering. 

In the project I'm working currently, since we activated the new React's Strict Mode, we are getting a lot of console warnings about using unsafe lifecycle methods y some of the components. 

For example:

`componentWillMount: Please update the following components to use componentDidMount instead: Dropdown`

`componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Dropdown, Menu, Tab`

`componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: AccordionAccordion`

`componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Checkbox`

layershifter:
@packetstracer see #2732, feel free to contribute 👍

PR #2755, opened 2018-05-01
breaking(Portal): use createPortal() API

### LINKED ISSUE #2746: Update Portal to use ReactDOM.createPortal
### Steps
Mount a react component that has a modal form using enzyme and try searching for children nodes.

### Expected Result
A call to `wrapper.find(<content>)` should return content with specified selector.

### Actual Result
It's not possible to get modal content using enzyme wrapper.

### Version
0.79.1

layershifter:
# BREAKING

### Why?

Our `Portal` component relies on `unstable_renderSubtreeIntoContainer`, this API is deprecated and will be removed in future releases of React. Also, this causes problems with tests.

### 1. Requirement of React 16

React 16 was released in September 2017, the update process is simple and I don't know any supported application that still uses React 15. It's time to say goodbye to React 15.

Please keep in mind, that in near future we will possibly require React 16.3.

### 2. Removal of `className` prop

As we use the new API (_createPortal()_) we don't need wrapper nodes as was before, so we don't have a target where we can apply `className`.

#### Before

```jsx
<Portal className='foo'>
  <button>A button</button>
</Portal>
```

#### After

```jsx
<Portal>
  <div className='foo'>
    <button>A button</button>
  </div>
</Portal>
```

### 3. Removal of `prepend` prop

Now, we don't control DOM manipulations, so this prop is useless now.

### 4. Removal of `closeOnRootNodeClick`

Now, we don't have `rootNode` at all, `createPortal` is responsible for DOM manipulations.

---

Fixes #2746.


src/addons/Portal/PortalInner.js
```diff
@@ -0,0 +1,60 @@
...
+  componentWillUnmount() {
+    _.invoke(this.props, 'onUnmount', null, { ...this.props, node: this.ref })
+  }
+
+  handleRef = c => (this.ref = c)
+
+  render() {
+    const { children, mountNode = isBrowser() ? document.body : null } = this.props
+
+    return createPortal(<Ref innerRef={this.handleRef}>{children}</Ref>, mountNode)
```
layershifter, line 56: We need a valid `ref` to get working `closeOnDocumentClick` as now we don't have `rootNode` and `portalNode` in fact

levithomason:
What do you think about making a `react-16-3` branch for this?  I know only 16.0 is required, but we can also bundle the ref changes into that branch.  This way, we can gather all of our changes that require React >=16.3 and make one breaking release once it is all ready.

Linking #2747.

layershifter:
> I know only 16.0 is required, but we can also bundle the ref changes into that branch.

In this case I think it will be better if we ship this PR now. `forwardRef` is more complicated thing in fact.

levithomason:
OK, let's merge this into a `react-16` branch.  I don't want to drop React 15 users on the next release.

layershifter:
@levithomason I've changed base to `react-16` branch and merged.

PR #2760, opened 2018-05-03
fix(docs): Fixing doc navigation link issues

msrikanth508:
Fixed https://github.com/Semantic-Org/Semantic-UI-React/issues/2737 issue. Please review and let me know your comments.


docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -441,7 +441,7 @@ class ComponentExample extends Component {
     }
 
     return (
-      <Visibility once={false} onTopPassed={this.handlePass} onTopPassedReverse={this.handlePass}>
```
levithomason: The right menu no longer updates as the user scrolls.  These handlers enabled that functionality.

docs/app/Components/ComponentDoc/ComponentSidebar/ComponentSidebarSection.js
```diff
@@ -52,7 +62,7 @@ class ComponentSidebarSection extends Component {
         <Accordion.Content as={Menu.Menu} active={active}>
           {_.map(examples, ({ title, path }) => (
             <ComponentSidebarItem
-              active={activePath === path}
+              active={activePath === _.kebabCase(path.split('/').join(' '))}
```
levithomason: The utility function would be used here as well.

docs/app/Components/ComponentDoc/ComponentSidebar/ComponentSidebarSection.js
```diff
@@ -35,7 +39,13 @@ class ComponentSidebarSection extends Component {
 
   handleItemClick = (e, itemProps) => _.invoke(this.props, 'onItemClick', e, itemProps)
 
-  handleTitleClick = () => this.setState(prevState => ({ isActiveByUser: !prevState.isActiveByUser }))
+  handleTitleClick = () =>
+    this.setState(prevState => ({ isActiveByUser: !prevState.isActiveByUser }))
+
+  isActiveAccordion = (props = this.props) =>
+    (props.examples || []).findIndex(
+      item => _.kebabCase(item.path.split('/').join(' ')) === props.activePath,
```
levithomason: The utility function would be used here.

docs/app/Components/ComponentDoc/ComponentDoc.js
```diff
@@ -68,14 +76,28 @@ class ComponentDoc extends Component {
 
   handleSidebarItemClick = (e, { path }) => {
     const { history } = this.props
-    const aPath = _.kebabCase(_.last(path.split('/')))
+    const aPath = _.kebabCase(path.split('/').join(' '))
```
levithomason: Let's pull this into a utility function in `docs/apps/utils`.  Something like `createComponentHash()`.  It would be ideal to be able to pass an example filename or file path to a single function that knows how to create the hash.  This way it can be used everywhere we are constructing the hash.

Also, we need to add redirects from the old hashes to the new hashes.  A utility function could be used for transforming the old hashes to the new hashes as well.

docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -71,7 +71,7 @@ class ComponentExample extends Component {
     const { examplePath, location } = this.props
     const sourceCode = this.getOriginalSourceCode()
 
-    this.anchorName = _.kebabCase(_.last(examplePath.split('/')))
+    this.anchorName = _.kebabCase(examplePath.split('/').join(' '))
```
levithomason: This would be the utility function as well.

docs/app/Components/ComponentDoc/ComponentDoc.js
```diff
@@ -47,8 +47,16 @@ class ComponentDoc extends Component {
...
+  constructor(props) {
+    super(props)
+    let activePath
+    if (location.hash) {
+      activePath = _.last((location.hash || '').split('#'))
+    }
+    this.state = {
+      activePath,
+    }
+  }
```
levithomason: Prefer property initializer for state unless `props` are needed to calculate state:

```jsx
state = {
  activePath: _.last((location.hash || '').split('#'))
}
```

levithomason (changes requested):

levithomason:
We have some unnecessary duplication in the hash.  Let's minimize these in this PR.  We can remove the pathname and `${component}-example`:

**Current**

```
/collections/table# | ${pathname}       | ${section} | ${exampleName}
                    |                   |            |          
/collections/table# | collections-table | variations | table-example-collapsing
```

**Proposed**

```
/collections/table# | ${section} | ${exampleName without "component-example"}
                    |            |          
/collections/table# | variations | collapsing
```

**Comparison**
```
/collections/table#collections-table-variations-table-example-collapsing
/collections/table#variations-collapsing
```

docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
msrikanth508: Oh! I didn't check that. Thank you for pointing it out. I'll look into it. 
And, all your comments make sense to me, I'll fix them.

msrikanth508:
Regarding unnecessary duplication in the hash, do you want me to rename examplePath property in ```ComponentExample``` examples folder or through util function? 

As we are proposing minimized hash string, we should make sure that there won't be any collisions. And, I'm wondering about redirection utility, I think there is no way to map from old hash to new hash as there is no single source of truth. For ex, old hash is ```/collections/table#table-example-collapsing```
new hash would be ```/collections/table#collections-table-variations-table-example-collapsing -> collections/table#variations-collapsing```, here we won't know that old hash belongs to ```variations``` category.

levithomason:
>Regarding unnecessary duplication in the hash, do you want me to rename examplePath property in ComponentExample examples folder or through util function?

I don't quite understand this one.  The `examplePath` is set to match the directory structure and should remain as-is.  The folder structure, `examplePath` values, and `docgenInfo.json` will all remain the same.  We'll only use the util function when creating hashes.  I might need more info here.

>As we are proposing minimized hash string, we should make sure that there won't be any collisions. 

We only have duplications when there is an example in more than one section for the same prop.  By adding the section (type, state, variation, etc) to the hash we are avoiding any duplicates.  Any duplicate examples for the same prop within the same category are true duplicate examples and should be merged.

>I think there is no way to map from old hash to new hash as there is no single source of truth

`DocsLayout` might be a good place to handle this.  Have a look at `resetPage`.  We would convert old hashes to new ones here.

>For ex, old hash is /collections/table#table-example-collapsing new hash would be /collections/table#collections-table-variations-table-example-collapsing -> collections/table#variations-collapsing, here we won't know that old hash belongs to variations category.

The old hashes are just the example's filename.  We can `_.startCase(hash).replace(/ /g, '')` to get the filename again.  Then, we can look it up in the `exampleContext` to see what section it is from.  This gives us all the necessary parts to make the new hash.  If we cannot lookup the hash in the `exampleContext` then we know it is not an old hash name and we do nothing.

We will not be able to resolve old and conflicting hashes (e.g. table collapsing) to know which section is correct (e.g. variations or types) but we can resolve all others correctly.  Even in the case of conflicting example names, we will still resolve it to the new location, we just will have the wrong section sometimes.

Regarding the new hash.  We should be able to create `collections/table#variations-collapsing` on the first pass without creating `/collections/table#collections-table-variations-table-example-collapsing` as a prerequisite.

msrikanth508:
I don't quite understand this one. The examplePath is set to match the directory structure and should remain as-is. The folder structure, examplePath values, and docgenInfo.json will all remain the same. We'll only use the util function when creating hashes. I might need more info here.
> I got it, will create util func.

Regarding the new hash. We should be able to create collections/table#variations-collapsing on the first pass without creating /collections/table#collections-table-variations-table-example-collapsing as a prerequisite.
> I meant the same, will replace current logic to generate proposed hash. I mentioned  ```/collections/table#collections-table-variations-table-example-collapsing -> collections/table#variations-collapsing``` for giving more context :)

levithomason:
Got it, looking forward to it.

msrikanth508:
@levithomason I've fixed code review comments and also created necessary utility functions to

1. Shorten hash name
2. Create new hash from old
3. Check whether give hash is old or new

msrikanth508:
@levithomason  could you please review this PR?

levithomason:
Apologies for the delay. I've merged master and I am reviewing this now.

levithomason (approved):

levithomason:

PR #2657, opened 2018-03-18
BREAKING(package): update SUI to 2.3

layershifter:
# BREAKING

### This PR requieres usage of Semantic UI CSS >= 2.3.x

----

Fixes #2550. This PR should:

- [x] update `Dimmer` to be compatible with SUI 2.3 (#2656)
- [x] update `Modal` to be compatible with SUI 2.3 (#2690)
- [x] update `Icon` to use FA5 (#2665)
- [x] add support for new `Transition` names: `zoom` and `glow` (#2658)

PR #2766, opened 2018-05-06
feat(icon): Adding missing cog icon alias, fix SUI.js

srbala:
Cog icon (alias of single gear) is missing in current list


srbala:
Just realized that, it may be need SUI 2.3.x

levithomason:
>Just realized that, it may be need SUI 2.3.x

Yep.  This is done in our `next` branch.  Closing as a dupe of #2657.

PR #2768, opened 2018-05-07
feat(Provider): stub v2 Provider component

levithomason:
We will iterate on this as we merge the first few v2 components.  Merging.


hubertguillemain:
Not sure this is the right place for this comment ....
Can you please clarify wether V2 will be backwards compatible with projects build with V1?
Especially, what about projects that managed theming via external standalone custom SUI theme files ?

PR #2769, opened 2018-05-08
Chore: Gut v1 components

dvdzkwsk:
Massive PR, but it's fairly restrictive in its scope. I'm sure at least a thing here or there is broken, but the `HelloWorld` component example works and the tests pass.


docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -206,7 +206,7 @@ class ComponentExample extends Component {
     const FAKER = require('faker')
     const LODASH = require('lodash')
     const REACT = require('react')
-    const SEMANTIC_UI_REACT = require('semantic-ui-react')
+    const STARDUST = require('stardust')
```
dvdzkwsk, line 209: Changing the name now will make it way more convenient to work with the docs and old library in parallel.

.circleci/config.yml
```diff
@@ -49,12 +49,6 @@ jobs:
             - ~/.cache/yarn
             - .yarn-cache
             - node_modules
-      - run:
-          name: Lint TypeScript
-          command: yarn tsd:lint
-      - run:
-          name: Test TypeScript
-          command: yarn tsd:test
```
levithomason: Why remove these?

PR #2775, opened 2018-05-08
fix(Popup) popup appearing in wrong position when open prop is set 

cdaringe:
# problem

- please see #2389

# solution

- add in RAF behavior per @levithomason's recommendation.
  - i _didn't_ actually properly animate it in, but this still has the added benefit of keeping the `<Popup />` correctly attached to the trigger if the trigger moves

## demo

here i use `<Popup />` for form validation.

with fix, the `<Popup />` immediately renders when the field is dirty and empty:

good mov

without fix.  the `<Popup />` is out of the viewport.  I have to fidget with the `trigger={<Input />}` to get the Popup to eventually appear where it has been commanded to go:

bad mov


src/modules/Popup/Popup.js
```diff
@@ -341,6 +344,21 @@ export default class Popup extends Component {
...
+      this.triggerRef = triggerRef
+    }
+  }
+
+  setPosition = () => {
+    if (this.triggerRef) {
+      this.setPopupStyle(this.props.position)
+    }
+
+    this.animationRequestId = requestAnimationFrame(this.setPosition)
```
fracmak, line 360: haven't you introduced an infinite loop now? When does this ever get short circuited?

cdaringe: good point.  fixed in 7c2104f

cdaringe:
hi all, kind and cordial bump!

levithomason:
Such a gentleman 😉 reviewing!

levithomason:
Just merged master, we can ship this once tests pass again.

cdaringe:
codecov seems to be all hung up!

levithomason:

cdaringe:
nice!  ps, if you havent seen already, the latest version of semantic-release auto posts back into PRs what versions PRs get released in. so great!

aminland:
With this commit, scrolling the window causes very strange behavior.

semantic-ui-popup-bug

src/modules/Popup/Popup.js
miroslavstastny: Unfortunately the infinite loop is still there - all the time the popup is displayed `setPosition()` is called in a loop, it then calls `setPopupStyle` which re-sets component's state which causes re-render - takes ~20% CPU when the popup is displayed.

prawdziwy-sok:
Hi, this issue is resolved in v0.80.2, but in the newest (v0.82.1 at the moment) it is still present. Did the fix get removed in the meantime somewhere? Could it be put back?

PR #2790, opened 2018-05-11
feat(factories): support functions

levithomason:
As part of the v2 manifesto, we are allowing users to construct the render tree shorthand props themselves.  More details to come.  This is now possible:

image

```jsx
  <Button
    content='Nice!'
    icon={(Icon, props) => <Icon name='thumbs up' {...props} />}
  />
```

We simply give the user the component, props, and children of what ever the shorthand prop was going to render.  This allows the user to control the render tree, previously not possible with shorthand props.


levithomason:

Issue #2318, opened 2017-11-14
Visibility: no callback when filtering causes the component to be shown


lafritay:
### Steps

This is similar to [Issue 1911](https://github.com/Semantic-Org/Semantic-UI-React/issues/1911).  Here is the scenario:

1. You display a grid of images that all have the title.  The goal is to only download the image when the image is on screen.
2. Each image has a title and you have a filter box for filtering on that title.
3. If you scroll up and down onScreen fires reliably.
4. If you use the filter to expose images that started off page though, there is no callback to know that those images are now visible.

### Expected Result

When an element appears on screen because of an event like a filter, there is a callback that will be fired to indicate that it is on screen.

### Actual Result

There is no callback that is fired.

### Version
0.75

### Testcase

For this repro:
1. The GitHub button is now below the litany of messages.  It's initial button image is 'users'.  When it is scrolled onto the screen, it's image becomes the github image.
2. If you click the "Show GitHub Button" button, it'll hide the messages, causing the button to appear on screen.

[Repro](https://codepen.io/anon/pen/pdwYJO?editors=0010)

### Thoughts
It seems the issue here is that the Visibility element is only reacting to the resize and scroll events.  The only thing I've been able to come up with is to allow the Visibility component to be passed an additional event that it will listen for and then have systems like filter boxes fire that event when it changes.  I'm certainly open to other suggestions though.

layershifter:
I don't sure that it's actually bug, but I fully agree that your issue it valid.

I think that better solution to get something like observer for window size. As workaround, you can dispatch `resize` event on `window`.

levithomason:
Observing the window size won't work in all cases either.  It is possible that an element comes into view due to a reflow and the window size does not change.  Imagine a scroll container where items are removed from the top and the bottom items become visible.  Also, an absolute position, fixed position, or transformed element coming into view will not cause a window resize.

The core issue here is that the element would need to have some kind of loop for checking if it is in the viewport.  If I were implementing this, I'd consider using `requestAnimationFrame`.  This way, it always aware of when it enters and leaves the viewport regardless of why or how it did so.

### Proposal

Looking at the Visibility component, we already use `requestAnimationFrame` to throttle `resize` and `scroll` events.  We could add a prop for choosing whether to `update()` on scroll/resize events or with an animation frame.

[...]

PR #2791, opened 2018-05-13
feat(Visibility): add `updateOn` prop

layershifter:
Refs #2318.


test/specs/behaviors/Visibility/Visibility-test.js
```diff
@@ -537,4 +537,36 @@ describe('Visibility', () => {
       })
     })
   })
+
+  describe('updateOn', () => {
```
levithomason, line 541: We should add a test here that `it()` defaults to `events`.  Also, it would be great to have an event based assertion to ensure that events are still working.

levithomason:
❤️

test/specs/behaviors/Visibility/Visibility-test.js
layershifter: I added a new test to test the default value. All other tests work with events, so I don't think that we need to duplicate any of them.

PR #2792, opened 2018-05-13
fix(ComponentExample): fix href in copy link button

### LINKED ISSUE #2770: Docs - Copy link buttons don't work
## Bug Report
Docs - Copy link buttons don't work
### Steps
When I click on a link button the tooltip says 'Copied link!'

### Expected Result
Link address to be in clipboard

### Actual Result
Clipboard remains the same

layershifter:
Fixes #2770.

### DIFF
--- a/docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
+++ b/docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
@@ -109,9 +109,8 @@ class ComponentExample extends Component {
   }
 
   handleDirectLinkClick = () => {
-    const { location } = this.props
     this.setHashAndScroll()
-    copyToClipboard(location.href)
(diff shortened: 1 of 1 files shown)


docs/app/Components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -109,9 +109,8 @@ class ComponentExample extends Component {
   }
 
   handleDirectLinkClick = () => {
-    const { location } = this.props
     this.setHashAndScroll()
-    copyToClipboard(location.href)
+    copyToClipboard(window.location.href)
```
layershifter, line 113: We made push to history in `setHashAndScroll`, so we can simply grab `href` from `window.location`

levithomason:

PR #2794, opened 2018-05-13
chore(package): update enzyme, puppeteer, sinon

layershifter:
This PR:
- updates `enzyme`
- updates `puppeteer` to `1.3`, the latest version has problems with Enzyme's event mocking 
- updates `sinon` and `sinon-chai`


layershifter:
Needs investigation why tests are failed.

levithomason:
In v2 we are attempting to move to Jest.  It would be ideal if we could bring parity between the two.  Not only would it help us prove out v2 patterns with the v1 community and set of components, we can also share findings and updates.

@davezuko chime in?  Also, how is progress on the Jest front?

We should also be considering multiple browser testing.  Something we don't do now but absolutely should in the future, definitely in v2.

dvdzkwsk:
No progress just yet, been busy with other mandatory tasks. We can talk more about it tomorrow, and I'll update the post accordingly.

levithomason:

levithomason:
The stardust fork is no running on Jest.  There are no component tests yet, just the `/lib` tests.  Once we get component tests going, we can consider updating.

PR #2796, opened 2018-05-14
fix(ComponentExample): clipboard and active state

### LINKED ISSUE #2770: Docs - Copy link buttons don't work

levithomason:
Fixes #2770 

This PR is largely copied from v2:

- fix copy direct link clipboard
- fix example active state when clicking code/html/direct and using direct links
- add better hover state style and improve listener performance


levithomason:

PR #2800, opened 2018-05-15
fix(checkbox): pass disabled prop to input

ofir-shapira-como:
There is an issue with the input element that is created by `<Checkbox>`.

When the `<Checkbox>` is disabled, the [input ](url)is not disabled and it allows to focus on the element.
See an example in: https://codesandbox.io/s/3x9rn7lo5

The proposed solution is to add _disabled_ attribute to the `input `in case the user asks the `<Checkbox>` to be _disabled_.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -47,10 +47,7 @@ export default class Checkbox extends Component {
     fitted: PropTypes.bool,
 
     /** A unique identifier. */
-    id: PropTypes.oneOfType([
-      PropTypes.number,
-      PropTypes.string,
-    ]),
+    id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
 
     /** Whether or not checkbox is indeterminate. */
     indeterminate: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/modules/Checkbox/Checkbox.js
```diff
@@ -250,6 +237,7 @@ export default class Checkbox extends Component {
           {...htmlInputProps}
           checked={checked}
           className='hidden'
+          disabled={disabled}
```
ofir-shapira-como, line 240: This is the actual change, passing disabled attribute to the input if set as a prop to Checkbox

levithomason:
Thanks much!  Let's add a test for this.  The `<Input />` has an identical test for asserting disabled is properly passed.  You can see it here:

https://github.com/Semantic-Org/Semantic-UI-React/blob/447f96b6709e5a4a7163abbe2df0db33bd5c7dd1/test/specs/elements/Input/Input-test.js#L225

test/specs/modules/Checkbox/Checkbox-test.js
```diff
@@ -154,17 +142,24 @@ describe('Checkbox', () => {
...
+
     it('cannot be unchecked', () => {
       wrapperShallow(<Checkbox defaultChecked disabled />)
 
       wrapper.simulate('click')
-      wrapper
+      wrapper.find('input').should.be.checked()
+    })
+
+    it('is applied to the underlying html input element', () => {
```
ofir-shapira-como, line 155: These are the tests, better to review in Split mode as GitHub did a diff mess because of the Lint

ofir-shapira-como:
@levithomason - done, added two tests, one check for when disabled property is truthy, and the other when it's false.

levithomason:
Thank you so much!

levithomason:

Issue #889, opened 2016-11-17
RFC: Advanced Dropdown menu shorthand


jeffcarbs:
http://react.semantic-ui.com/modules/dropdown#search-in-menu

I'd like to use this variation of the dropdown in my app but we currently don't support it. I'm happy to take a stab at implementing it, but don't really have a good idea how it should work.

The issue is that the markup can become pretty complex, with variations on headers, dividers, the search box, options, etc. See this example for another variation: http://react.semantic-ui.com/modules/dropdown#input

It would be nice to give ultimate control to the user, but I'm struggling to see how that's possible since we need to know:

1. the options so that we can filter them. Currently we don't support options + children.
2. where the search box is so that we can
 - detect change
 - allow the `searchQuery` to be set as an auto-controlled prop (this isn't an autocontrolled prop now but I kinda think it should be. If it's too hard to do with search-in-menu we can forego it though.)

One kinda crazy idea is to allow the user to specify a "schema". Something like:
```js
const nestedMenu: [
  { component: Dropdown.Header, content: 'Search Issues' },
  { component: Dropdown.Input, value: this.props.searchQuery },
  { component: Dropdown.Header, icon: 'tags', content: 'Filter by tag' },
  { component: Dropdown.Divider },
  { component: Dropdown.Menu }
]

<Dropdown options={options} nestedMenu={nestedMenu} />
```

Then we can parse out which is the search input, and which is the nested menu where we should render the options. We can validate the schema such that there's exactly one menu and at most one input (theoretically you can make a nested menu that's not searchable).

@levithomason @layershifter - would appreciate your thoughts!

jeffcarbs:
@levithomason - any thoughts here? I think we're again faced with a similar issue as https://github.com/Semantic-Org/Semantic-UI-React/pull/893#issuecomment-262409730, where we want to give the user complete control of rendering while being able to listen for events emitted by those user-rendered components.

@davezuko - Since Levi mentioned you on that related comment, would love your thoughts as well.

levithomason:
Sorry, haven't had the brain cycles this one needs.  Giving it a little thought my intuition is that we provide shorthand for the most probable use case, for now.  Something that adds the search and a callback when it is changed?

```jsx
<Dropdown options={options} onSearchInMenuChange={} />
```

We could render a minimal search like so:

![image](https://cloud.githubusercontent.com/assets/5067638/20698734/eec386e4-b5b7-11e6-8441-19ea691adfec.png)

[...]

PR #2807, opened 2018-05-18
feat(Dropdown): add managed state support for children

sinangun:
Managed state support added for Dropdown's Dropdown.Item children.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -52,7 +52,10 @@ export default class Dropdown extends Component {
      * Allow user additions to the list of options (boolean).
      * Requires the use of `selection`, `options` and `search`.
      */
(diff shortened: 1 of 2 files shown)


levithomason:
Thank you for the PR.  We want to avoid inspecting children, however, as it is brittle and leads to unexpected behavior.  Example, a user can wrap an `DropdownItem` in an HOC for navigation purposes and we would no longer be able to handle that case even though it is valid usage.

Our preferred solution is to extend the capabilities of the shorthand that the dropdown handles to enable more use cases, see #1757.  Also #889.

Our solution to allowing the user to control children entirely _and_ have access to the state of the dropdown is to use a render function pattern.  We've extended the factories to allow functions, #2790.  This means you can now have access to the computed state of the Dropdown menu while explicitly defining the menu as JSX.

More examples and docs to come.

PR #2813, opened 2018-05-19
fix (Datetime): Fix bugs in Datetime component

geoff-wasilwa:
Fixes:
- [x] Navigation on calendar pop-up
- [x] Truncation of days in a month (e.g. September 2018 with Sunday as first day of the week) 
- [x] Display hours in calendar pop-up
- [x] Display minutes in calendar pop-up
- [x] Update calendar when user clicks on a month or year in header

### DIFF
--- a/src/addons/Datetime/Datetime.js
+++ b/src/addons/Datetime/Datetime.js
@@ -249,6 +249,7 @@ export default class Datetime extends Component {
   }
 
   handleChangeMode = (e, { mode }) => {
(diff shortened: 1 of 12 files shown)


levithomason:
Great, thank you!

levithomason:
I think a good next step would be to remove all the non-working bits.  We can ship a minimal version and iterate on it.

Example, remove the Range example as we don't have a range control in here, yet.

PR #2815, opened 2018-05-20
change from "right chevron" to "chevron right" for typescript

searene:
To conform with typescript definition files


levithomason:
Thanks much!

levithomason:

PR #2817, opened 2018-05-21
feat(Dropdown): open to side with most space left #720

mihai-dinculescu:
#720

I have added a new prop, "keepInViewPort", so dropdown is consistent to Popup. However, this one is `false` by default, so the current behavior doesn't change.

Please let me know your thoughts.


src/modules/Dropdown/Dropdown.js
```diff
@@ -1056,6 +1092,29 @@ export default class Dropdown extends Component {
...
+    const spaceAtTheBottom =
+      document.documentElement.clientHeight - dropdownRect.y - dropdownRect.height - itemsHeight
+    const spaceAtTheTop = dropdownRect.y - itemsHeight
+
+    if (spaceAtTheBottom < 0 && spaceAtTheTop > 0) {
+      this.setState({ shouldOpenUpward: true })
+    } else {
+      this.setState({ shouldOpenUpward: false })
+    }
+  }
```
levithomason: An if/else which results in true/false statements can be simplified to using the condition as the value:

```jsx
this.setState({ 
  shouldOpenUpward: spaceAtTheBottom < 0 && spaceAtTheTop > 0,
})
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -1056,6 +1092,29 @@ export default class Dropdown extends Component {
...
+
+    const { keepInViewPort } = this.props
+    if (!keepInViewPort) return
+
+    const menu = this.ref.querySelector('.menu.visible')
+
+    if (!menu) return
+
+    const dropdownRect = this.ref.getBoundingClientRect()
+    const itemsHeight = menu.clientHeight
```
levithomason: nit: Let's keep the naming consistent and use `menuHeight`.

src/modules/Dropdown/Dropdown.js
levithomason: In the case that both top and bottom are blocked, perhaps we should open to the side with the most available space.  I think this math is wrong, but something like:

```jsx
this.setState({ shouldOpenUpward: spaceAtTheBottom > spaceAtTheTop })
```

src/modules/Dropdown/Dropdown.js
```diff
@@ -1273,7 +1338,7 @@ export default class Dropdown extends Component {
       useKeyOnly(selection, 'selection'),
       useKeyOnly(simple, 'simple'),
       useKeyOnly(scrolling, 'scrolling'),
-      useKeyOnly(upward, 'upward'),
+      useKeyOnly(upward || shouldOpenUpward, 'upward'),
```
levithomason: When we have props that need to be controlled, we move to them to being `autoControlledProps`.  See Dropdown.js around line 346.

- add `upward` to `static autoControlledProps = []`
- add `defaultUpward` as a `propType` (since it is now stateful, we accept a default value for it)
- move all usages of `upward` from `this.props.upward` to `this.state.upward`
- to change the value of `upward` use `this.trySetState({ upward: <new value> })`
- remove `shouldOpenUpward` as it is no longer needed.

Auto controlled props automatically defer to the user's prop value, if any.  Also, `this.trySetState()` will not update the state value if the user is using that prop.

levithomason:
Looking great!  I've added some feedback for you here.  Thanks much for tackling this one.

samboylett:
I've been looking at this too. I think if it doesn't change direction, it should only render once, so the search function will only be called once. Otherwise it will have to render twice (there's not really any way around this) so the search function will be called twice too.

mihai-dinculescu:
Thank you for  your input, @samboylett.

I've optimized the code so it updates the state only when there is a relevant change.

levithomason:
This looks great, thanks!  The last thing we need here is a test.  The test would resize the browser window and assert the dropdown opened upward.  We can merge after this update.

levithomason:
@mihai-dinculescu You've contributed here before and you've shown care and initiative for ensuring SUIR becomes better. We appreciate this and want to enable you to do more. I've added you as a collaborator. You're now able to manage issues and PRs as well as create branches here in the main repo. You no longer need to work from a fork.  Merging to master is the only reserved permission.

We are very select about who we add and when, so know that we trust your abilities and your judgment. Please feel free to weigh in on any issues and PRs that you have an opinion about. You have a share of ownership and say here now.

Once you accept the invite (see your email), you might want to show the Semantic Org badge publicly on your profile. To do that, navigate to the "people" page and change the "private" dropdown to "public" for your user in the table. Here's a [direct link](https://github.com/orgs/Semantic-Org/people?utf8=%E2%9C%93&query=mihai-dinculescu) to your user in the table.

/cc
@Semantic-Org/react-collaborators
@Semantic-Org/react-maintainers

Please welcome @mihai-dinculescu to our ranks 😄 🎉

mihai-dinculescu:
Thank you @levithomason. I'll do my best to be helpful.

levithomason:

PR #2819, opened 2018-05-21
docs(Knobs): add knobs and improve docs ux

levithomason:
This PR:

Adds "Knobs" capability to component examples.  Allows you to create controls for users to play with examples.  Knob code does not clutter the example code but is passed in to the example component as `props.knobs`.

Adds acceleration to doc link scrolling.  This reduces the jarring effect when switching examples.

"Deactivates" an example when closing the code editor.  This removes it from the hash and removes the active style.  Currently, examples look like they "get stuck" in active mode when closing or switching between examples.

Removes the GitHub bug link from each example code editor. This was a fun idea but only ever produced accidental issues on our repo.

Simplifies some doc site styles.


brianespinosa:
I love this.

PR #2820, opened 2018-05-21
feat(docs): improve ux

levithomason:
Back-ported from v2.  This PR makes improvements to anchor handling, scrolling, styles, and the experience around editing/closing live edit examples.

See the v2 counterpart #2819


levithomason:

PR #2823, opened 2018-05-23
docs(Form): fix group variations typo

elliottcarlson:
The message info references `widths='even'` which is an invalid widths type - the correct value should be `equals`.


levithomason:
Thank you!  Fixed typo `equals => equal`, good to merge once tests pass.

levithomason:

PR #2837, opened 2018-05-25
refactor(eventStack): make eventStack immutable

layershifter:
# Why?

Sometime ago I tried to attack #1473, I've fully reworked the `Sidebar` component and docs. But then I found a strange problem when tested the component: event handlers simply called before they should be called. I've found the problem, it was in the mutable array of `handlers` in the `EventTarget`.

This PR performs the deep refactoring and adds more test coverage for internals of the `eventStack`.

The easiest way to show problem: just clone `feat/sidebar-closable-property ` branch and try to run the `Sidebar` examples without & with these changes.


src/lib/eventStack/EventSet.js
```diff
@@ -0,0 +1,65 @@
...
+   */
+  dispatchEvent(event, dispatchAll) {
+    if (dispatchAll) {
+      this.handlers.forEach((handler) => {
+        handler(event)
+      })
+      return
+    }
+
+    const recentHandler = [...this.handlers].pop()
```
levithomason, line 40: **EDIT** - Oh, this is a Set, not an array... you can ignore my original
***
Rather than cloning the entire array to get the last handler, you can directly access the last item:

```jsx
const recentHandler = this.handlers.slice(-1)[0]

// or

const recentHandler = this.handlers[this.handlers.length - 1]
```

Note, the spread approach, the slice approach, and the length approach all avoid mutating the `this.handlers` array, however, the `recentHandler` is not a unique instance.  It is the same instance as was originally provided to the array of handlers.  I'm not sure if this matters to you or not.

src/lib/eventStack/EventStack.js
```diff
@@ -0,0 +1,70 @@
+import isBrowser from '../isBrowser'
+import EventTarget from './EventTarget'
+import normalizeHandlers from './normalizeHandlers'
+import normalizeTarget from './normalizeTarget'
+
+export default class EventStack {
```
levithomason, line 6: It would be helpful to have a description above the EventSet, EventStack, and EventPool.  It is not immediately clear how the 3 operate together.

levithomason (changes requested):
See comments.  This looks good, just a request for a little more documentation on the classes.

src/lib/eventStack/EventStack.js
layershifter: @levithomason I've added a README file with description

levithomason (approved):
I've updated the EventStack README a bit.  If it looks good, this can be merged.  Thanks for the excellent write up!

levithomason:
Merging for progress.  Any doc tweaks can happen in another PR 💪

levithomason:

Issue #2836, opened 2018-05-25
export element props without [key: string]: any;


evollu:
### Problem description
We use typescript and HOC in our project. There are many cases where we wrap a semantic element and need to remove a property. 
For example, 
- we have a HOC which wraps Dropdown component and passes `option` though redux. Then we need to remove `option` from props
- we introduced a default property but the only clean way to define the typescript is using HOC to change a required prop to be optional

In both cases, we need to utilize `Omit` ([reference](https://github.com/piotrwitek/utility-types)) operation to remove a key from Semantic element's prop type definition. However each type definition from Semantic contains `[key: string]: any;` which makes `Omit` not working

### Proposed solution
export a type definiton without `[key: string]: any;` for HOC to use.

### MVP
Take button as example
Current code:
```typescript
export interface ButtonProps {
  [key: string]: any;

  /** An element type to render as (string or function). */
  as?: any;

  /** A button can show it is currently the active user selection. */
  active?: boolean;
}
```

Proposal
```typescript
export interface ButtonPropsStrict {
  /** An element type to render as (string or function). */
  as?: any;

  /** A button can show it is currently the active user selection. */
  active?: boolean;
}

export interface ButtonProps = ButtonPropsStrict & { [key: string]: any; }
```

levithomason:
This seems like a great proposal to me.  This still meets the goals of our [earlier discussion](https://github.com/Semantic-Org/Semantic-UI-React/issues/1072) while also giving type safety when needed.

PRs welcome!

evollu:
@levithomason I've created a PR for Dropdown. If you a fine with the change, I will apply the change to more components.

PR #2839, opened 2018-05-25
Strict type dropdown

evollu:
add StrictDropdownProps based on discussion on #2836
somehow precommit step is reformatting the whole file

more PRs coming


levithomason:
Sorry for the huge delay here!  I think this change looks good.  If we start this work, we'll need to complete it for all components.  That is a large effort.  Is this something you'd like to start up?

evollu:
the PR is too old, I open a new one with all props updated @levithomason

PR #2844, opened 2018-05-26
feat(forwardRef): add forwardRefFactory

layershifter:
Replaces #2306.

---

React 16.3 gives the wonderful `forwardRef` API, this PR implements the `forwardRefFactory` that allows to use it.

Some ideas are taken from [klimashkin/react-forwardref-utils](https://github.com/klimashkin/react-forwardref-utils).

### DIFF
--- a/src/lib/componentUtils.js
+++ b/src/lib/componentUtils.js
@@ -1,25 +1,4 @@
-/**
- * Asserts that the passed Component is a React class component.
- *
- * @param {Function|Component} Component A Component to check
- * @return {Object}
- */
-export const isClassComponent = Component =>
-  typeof Component === 'function' && Component.prototype && !!Component.prototype.isReactComponent
+export const isForwardRef = Component =>
+  (Component && Component.$$typeof && typeof Component.render === 'function') || false
 
-/**
- * Asserts that the passed Component is a React function component.
- *
- * @param {Function|Component} Component A Component to check
- * @return {Object}
- */
-export const isStatelessComponent = Component =>
-  typeof Component === 'function' && Component.prototype && !Component.prototype.render
-
-/**
- * Asserts that the passed Component supports the refs API.
- *
- * @param {Function|Component} Component A Component to check
- * @return {Object}
- */
-export const supportsRef = Component => typeof Component === 'string' || isClassComponent(Component)
+export const isBasic = Component => typeof Component === 'string'
--- a/src/lib/componentUtils.js
+++ b/src/lib/componentUtils.js
@@ -1,0 +1,11 @@
+/* eslint import/prefer-default-export: 0 */
+/* This file will contain more functions in the future */
+
+/**
+ * Asserts that the passed Component is a React class component.
+ *
+ * @param {Function|Component} Component A Component to check
+ * @return {Object}
+ */
+export const isClassComponent = Component =>
+  typeof Component === 'function' && Component.prototype && !!Component.prototype.isReactComponent
--- a/src/lib/componentUtils.js
+++ b/src/lib/componentUtils.js
@@ -1,6 +1,3 @@
-/* eslint import/prefer-default-export: 0 */
-/* This file will contain more functions in the future */
-
 /**
  * Asserts that the passed Component is a React class component.
  *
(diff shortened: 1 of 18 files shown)


src/lib/forwardRefFactory.js
```diff
@@ -0,0 +1,47 @@
...
+  // eslint-disable-next-line react/prop-types
+  if (_.isUndefined(props.as) || _.isString(props.as) || isClassComponent(props.as)) {
+    return <Component {...{ [forwardRefSymbol]: ref, ...props }} />
+  }
+
+  return (
+    <Ref innerRef={ref}>
+      <Component {...props} />
+    </Ref>
+  )
```
layershifter: This factory allows to use stateless components too :smiley_cat:

src/lib/getUnhandledProps.js
```diff
@@ -11,6 +13,10 @@ const getUnhandledProps = (Component, props) => {
 
   return Object.keys(props).reduce((acc, prop) => {
     if (prop === 'childKey') return acc
+    if (prop === forwardRefSymbol) {
+      acc.ref = props[forwardRefSymbol]
+      return acc
+    }
```
layershifter, line 19: This hack allows to don't map `ref` in the every component

layershifter:
@levithomason this is an initial pass, no tests, just PoC. Will be interesting to hear your thoughts.

mihai-dinculescu:
Could add `displayName`, it's quite useful in Dev Tools.

src/lib/forwardRefFactory.js
```diff
@@ -0,0 +1,47 @@
...
+ * like so: { $$type: Symbol(), render: function }. Assigns (hoists) static methods of passed component to result
+ * forward component using 'hoist-non-react-statics' module.
+ *
+ * @param {Function|Component} Component A Component to wrap with forwardRef()
+ * @return {Object}
+ */
+export const forwardRefFactory = (Component) => {
+  const forwarder = forwardRef(forwardFunctionFactory(Component))
+
+  hoistStatics(forwarder, Component, { $$typeof: true, render: true })
```
levithomason: Why do we hoist `{ $$typeof: true, render: true }` here?  Also, the values are conflicting with the docblock on this function.

src/lib/componentUtils.js
```diff
@@ -0,0 +1,11 @@
...
+/* This file will contain more functions in the future */
+
+/**
+ * Asserts that the passed Component is a React class component.
+ *
+ * @param {Function|Component} Component A Component to check
+ * @return {Object}
+ */
+export const isClassComponent = Component =>
+  typeof Component === 'function' && Component.prototype && !!Component.prototype.isReactComponent
```
levithomason: Perhaps we also add:

```jsx
export const isStatelessComponent = Component =>
  typeof Component === 'function' && Component.prototype && !Component.prototype.render

export const supportsRef = Component => 
  typeof Component === 'string' || isClassComponent(Component)
```

It may help clarify the factory function, see below.

src/lib/forwardRefFactory.js
```diff
@@ -0,0 +1,47 @@
...
+ * Creates a function that will choose how to pass a ref.
+ *
+ * @param {Function|Component} Component A Component to wrap
+ * @return {Function}
+ */
+export const forwardFunctionFactory = Component => (props, ref) => {
+  // eslint-disable-next-line react/prop-types
+  if (_.isUndefined(props.as) || _.isString(props.as) || isClassComponent(props.as)) {
+    return <Component {...{ [forwardRefSymbol]: ref, ...props }} />
+  }
```
levithomason: What do you think of explicitly handling the ref for supported use cases and returning the Component in all other cases?:

```jsx
import { isStatelessComponent, supportsRef } from './componentUtils.js'

export const forwardFunctionFactory = Component => (props, ref) => {
  /* eslint-disable react/prop-types */

  if (isStatelessComponent(props.as)) {
    return (
      <Ref innerRef={ref}>
        <Component {...props} />
      </Ref>
    )
  }

  if (supportsRef(props.as)) {
    return <Component {...{ [forwardRefSymbol]: ref, ...props }} />
  }

  return Component
}
```

levithomason:
This implementation is far better, thanks!

Also, this pattern should definitely be implemented in the v2 fork.  I'm about to post a set of specification proposals for the `lib` directory.  This directory is becoming a core lib called "Stardust".  Our library and others will be built on the standards there.  Dealing with refs is one of those standards.

layershifter:
We decided to stop our work there, https://github.com/stardust-ui/react/pull/587#issuecomment-445905932. 

There is a RFC https://github.com/reactjs/rfcs/pull/97 that possibly will simplify all.

Issue #2387, opened 2017-12-18
Message type warning not working


Dscheksn:
**Steps to Reproduce**
1. Import Message from "semantic-ui-react"
2. Use Message with type warning

**Expected**
The Message should be orange

**Result**
The Message does not appear

**Testcase**
Tested with 'positive'. Result: Appears as expected
Tested with 'negative'. Result: Appears as expected
Tested with 'warning'. Result: Doesn't appear at all!

layershifter:
Please fill issue fully and make a repro on codesandbox.

PR #2845, opened 2018-05-26
feat(Sidebar): add lifecycle handlers

layershifter:
Requires #2837.
Replaces #2433, #1473.

-----

This PR:
- adds `onHide`, `onHidden`, `onShow` and `onVisible` callbacks to allow the deep control of `visible` prop
- reworks `Sidebar` docs, examples are much better and clear now :+1: 
- improves code coverage

----

Especially thanks to @cjsheets and @mchirkin for their work in previous PRs!


levithomason:
#2387 merged, unblocked this PR.

levithomason:
Heads up, I've reworked the directories in `/docs`.  This PR needs updated.  There is a detailed upgrade guide in the PR, #2863.

layershifter:
@levithomason marked as ready for review 👍

levithomason:
In http://localhost:8080/modules/sidebar#types-sidebar, clicking outside does not close the Sidebar.  It seems `onHide` is called twice, resulting in a double toggle.

If I create a `handleHide` method that sets visible to false only and assign that to `onHide`, then it works.  However, we should ensure the `onHide` is not called twice.

layershifter:
Will check, sounds stange 🤔

layershifter:
I made some updates, now it works as expected.

levithomason:

PR #2850, opened 2018-05-28
chore(package): remove node-sass

layershifter:
There is a strange historical dep, it was actually never used. Let's drop it :+1:


levithomason:

PR #2854, opened 2018-05-30
fix(Sticky): prevent npe error after unmount of component

fracmak:
This fixes a simple problem where the event listeners call handleUpdate() which calls requestAnimationFrame to call the update function in the next tick, but if the component unmounts between the handleUpdate call and the requestAnimationFrame firing, we get an NPE error because the handleTriggerRef function has been fired with null as the ref ( reference: https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element ). 

This prevents that from happening by saving the requestId from requestAnimationFrame and clearing it on componentWillUnmount

### DIFF
--- a/src/modules/Sticky/Sticky.js
+++ b/src/modules/Sticky/Sticky.js
@@ -130,7 +130,7 @@ export default class Sticky extends Component {
 
     if (active) {
       this.removeListeners()
-      cancelAnimationFrame(this.requestId)
+      cancelAnimationFrame(this.frameId)
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/modules/Sticky/Sticky.js
```diff
@@ -217,7 +220,7 @@ export default class Sticky extends Component {
   didReachContextBottom = () => {
     const { offset } = this.props
 
-    return (this.stickyRect.height + offset) >= this.contextRect.bottom
```
fracmak, line 220: I believe these lines changed because of prettier/eslint --fix

layershifter (approved):
Made some style changes, LGTM 👍 @fracmak thanks for PR!

levithomason:

PR #2859, opened 2018-06-01
test(Search): fix tests on second run

mihai-dinculescu:
This fixes two similar tests that are failing in watch mode, on the second run.


levithomason:

PR #2862, opened 2018-06-01
fix(getComponentInfo): subcomponents from js files

levithomason:
Currently, we consider any file in a component directory to be a subcomponent.  However, there was a stray markdown file in `/src/modules/Dropdown/TODO.md` that resulted in a "subcomponent" called "TODO" on the Dropdown.

This PR strictly limits subcomponent identification to JS files in component directories.


levithomason:

PR #2866, opened 2018-06-03
fix(ComponentExample): remove filter effect

### LINKED ISSUE #2865: Weird looking docs
### Problem description
What is this strange new "feature" in the docs - that I have to -hover- over a section to see normally the look of the component? Everything else is opaque-ed

### Proposed solution
Could you please revert to the old scheme? When everything was visible instantly - so I could scan for the needed component layout w\o having to hover over each and every variant availble?

levithomason:
Fixes #2865 

The grayscale filter effect was intended to bring focus to an example when editing code.  However, it is also made active when clicking an example link in the right menu.  This is confusing and distracting, ironic as it is.


levithomason:

PR #2882, opened 2018-06-05
fix(Modal): remove dimmer={false}

### LINKED ISSUE #2873: Modal:  `dimmer={false}` stops the modal from opening
### Steps
A Modal with `dimmer={false}`.

### Expected Result
The modal opens.

### Actual Result
The modal doesn't open. This can be observed in the docs.

### Version
0.81.1

### Testcase
https://react.semantic-ui.com/modules/modal#variations-dimmer
https://react.semantic-ui.com/modules/modal#types-multiple

mihai-dinculescu:
Fixes #2873.

In fact, this change should be done in `v0.81.0` when we performed update to SUI 2.3 (#2657). Without a `dimmer`, it is no longer able to figure out the position of the modal.

> A Special Message about Flex Modals
> 
> There will be an update shortly to resolve issues related to flex modals when using multiple modals and detachable: false, in order to not hold up this release, we've decided to move forward without a fix.
> 
> A general solution will most likely require branching code for IE11 which will disable flex (as IE11 doesnt correctly implement the latest spec for absolute positioned flex containers).

### Before

You able to hide a dimmer with: `<Modal dimmer={false} />`.

### After

This behaviour is deprecated and removed, you need to apply custom styling for dimmers if you actually need this feature.

-----

#### 1. Multiple example
Remove the now controversial `dimmer={false}`. It adds nothing to the example anyway.

#### 2. Close Config example
Make it so that the "Yes" and "No" buttons inside the Modal can close it. The "No Close on Dimmer Click" example is currently unclosable and requires browser refresh.

#### 3. Dimmer examples

Removed useless example with `dimmer={false}`.

#### 4. Remove useless tests and update propTypes for the `dimmer` prop

### DIFF
--- a/docs/src/examples/modules/Modal/Types/ModalExampleMultiple.js
+++ b/docs/src/examples/modules/Modal/Types/ModalExampleMultiple.js
@@ -12,7 +12,6 @@ class NestedModal extends Component {
 
     return (
       <Modal
-        dimmer={false}
         open={open}
         onOpen={this.open}
         onClose={this.close}
--- a/docs/src/examples/modules/Modal/Variations/ModalExampleCloseConfig.js
+++ b/docs/src/examples/modules/Modal/Variations/ModalExampleCloseConfig.js
@@ -22,15 +22,22 @@ class ModalExampleCloseConfig extends Component {
           open={open}
           closeOnEscape={closeOnEscape}
           closeOnRootNodeClick={closeOnRootNodeClick}
-          onClose={this.close}
         >
           <Modal.Header>Delete Your Account</Modal.Header>
           <Modal.Content>
             <p>Are you sure you want to delete your account</p>
           </Modal.Content>
           <Modal.Actions>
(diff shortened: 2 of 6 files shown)


layershifter:
@mihai-dinculescu thanks for PR 👍 I've pushed some updates and update the PR's description

mihai-dinculescu (commented):
Whops. It was already done. Sorry about that.

levithomason:
I believe this is actually a bug fix to 0.80.1, correct?  It should have been shipped as a breaking change in the 0.81.0 release, but now that we've missed that I am inclined to say we're "fixing" the 0.81 release.

Is this sane to others?

layershifter:
In fact, `dimmer={false}` already is not working, so we can ship this as a fix. I updated title :+1:

layershifter:
Performed a review again, merged. @mihai-dinculescu thanks :+1:

levithomason:

PR #2883, opened 2018-06-06
chore(getComponentInfo): handle HOC default exports

levithomason:
This PR extends our usage of react-docgen to handle exported HOCs.


levithomason:

PR #2884, opened 2018-06-06
fix(Docs): Add babel-polyfill so that the Docs work in IE11

mihai-dinculescu:


fracmak:
Can you throw in 

useBuiltIns: 'entry'

Into the .babelrc file for the babel-preset-env plugin?

package.json
```diff
@@ -90,6 +90,7 @@
     "babel-plugin-lodash": "^3.2.11",
     "babel-plugin-transform-react-handled-props": "^1.0.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.13",
+    "babel-polyfill": "^6.26.0",
```
layershifter: We use Babel 7

mihai-dinculescu (commented):
@fracmak, like this?

fracmak:
yup

build/preset-env.js
```diff
@@ -26,6 +26,7 @@ const browsers = [
 
 const envOptions = {
   targets: { browsers },
+  useBuiltIns: 'entry',
```
layershifter: This change will affect also `dist` build that will be published to NPM. We also using [transform runtime](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.babelrc#L16).

I am not sure that proposed changes are safe.

mihai-dinculescu: I'm not confident that this setting helps the docs. Albeit crude, my tests showed no difference in network traffic. I'll take it out for now.

levithomason:
This is an easy merge, however, what feature are we using that is not IE 11 compliant?  This is a target browser.

mihai-dinculescu:
`Promise`

levithomason:

PR #2885, opened 2018-06-06
fix(Dropdown): automatic upward logic in IE11

mihai-dinculescu:
Replacing `DOMRect.y` with `DOMRect.top` because the latter seems to work in all browsers, including IE11.


levithomason:

PR #2890, opened 2018-06-07
feat(Sticky): use SUI CSS classes

pferreir:
The `<Sticky />` component currently uses inline styles, which makes it difficult to take advantage of CSS to do some tricks like [this one](https://semantic-ui.com/modules/sticky.html#sticking-to-own-context).
This PR adds the standard Semantic UI classes meant for Sticky elements. Since it seems impossible to have the same element structure as with vanilla SUI, I'm adding `stuck-container`, `fixed-container` and `bound-container` classes that would allow for styling of the actual top-level container in its sticky state and "sub-states" (bound and fixed).


layershifter:
Marked as ready for review, I want to spend some time there in next few days.

src/modules/Sticky/Sticky.js
```diff
@@ -299,13 +311,36 @@ export default class Sticky extends Component {
...
+
     return (
-      <ElementType {...rest} className={className}>
+      <ElementType {...rest} className={containerClasses}>
         <div ref={this.handleTriggerRef} />
-        <div ref={this.handleStickyRef} style={this.computeStyle()}>
+        <div
+          ref={this.handleStickyRef}
+          className={cx(...elementClasses)}
+          style={this.computeStyle()}
```
layershifter: Please order props in alphabetical order

src/modules/Sticky/Sticky.js
```diff
@@ -299,13 +311,36 @@ export default class Sticky extends Component {
...
 
+    const containerClasses = cx(
+      className,
+      'ui',
+      sticky && 'stuck-container',
+      sticky && bound && 'bound-container',
+      sticky && !bound && 'fixed-container',
+    )
+
+    const elementClasses = ['ui', 'sticky']
```
layershifter: Let's use the `classnames` package there:
```jsx
const elementClasses = (
  'ui',
  bound ? 'bound bottom' : 'fixed',
  // etc
  'sticky',
)
```

src/modules/Sticky/Sticky.js
```diff
@@ -299,13 +311,36 @@ export default class Sticky extends Component {
 
   render() {
     const { children, className } = this.props
+    const { sticky, bound, bottom } = this.state
```
layershifter: Please order destructed props in alphabetical order

src/modules/Sticky/Sticky.js
```diff
@@ -299,13 +311,36 @@ export default class Sticky extends Component {
...
     const rest = getUnhandledProps(Sticky, this.props)
     const ElementType = getElementType(Sticky, this.props)
 
+    const containerClasses = cx(
+      className,
```
layershifter: The `className` prop is always last in `cx()` in our codebase

src/modules/Sticky/Sticky.js
```diff
@@ -255,33 +266,34 @@ export default class Sticky extends Component {
...
+    this.bound(false)
     this.pushing(false)
   }
 
   stickToScreenBottom = (e) => {
     const { bottomOffset: bottom } = this.props
 
     this.stick(e)
+    this.bound(false)
     this.setState({ bottom, top: null })
```
layershifter: I'm not sure that we really need the `bound()` method, this case shows it because we have duplicate call of `setState()` there

src/modules/Sticky/Sticky.js
```diff
@@ -199,13 +200,17 @@ export default class Sticky extends Component {
...
-    const { bottom, sticky, top } = this.state
-
+    const { bottom, bound, sticky, top } = this.state
     if (!sticky) return {}
+    if (bound) {
+      return {
+        bottom: 0,
+        width: this.triggerRect.width,
+      }
+    }
```
layershifter: This condition is strange, I think that we should just have `bottom: null` from the component's state there.

layershifter (changes requested):
Thanks for PR :+1: I've left some comments, but it's not all

layershifter (approved):
Added some styling changes, @pferreir thank you for PR 👍

levithomason:

pferreir:
Cheers! 👍

PR #2896, opened 2018-06-09
[WIP] docs(generator): use NextJS to generate prerendered pages

### LINKED ISSUE #2826: Prerender docs on the server side
## Feature Request

Improve the SEO for Google etc. by adding a sitemap and a robots.txt

### Problem description

When we search e.g. for "semantic ui react label", no result for the react label will appear:

image

### Proposed solution

- add a sitemap.xml
- add a robots.txt which refer to the sitemap

### MVP

Best way would be to not hard code the sitemap. So maybe we could generate the links?

I'm also not sure, if this repo is the correct place for such an issue. I haven't found the repository for the semantic ui react documentation page yet.

## Edit:

 I tried to find the URLs in Google by searching for `site:https://react.semantic-ui.com/elements/label`.
=> Google has indexed the result, so maybe just the SEO is very bad on these pages? idk

layershifter:
# WIP

Fixes #2826.


layershifter:
@levithomason can you take a look and give some feedback?

This is an early POC, you will need to run:
```bash
yarn next:build
yarn next:serve
```

I don't integrated it deeply to our build process, yet. There are only two working pages:
- theming: http://localhost:3000/theming
- usage: http://localhost:3000/usage

The generated pages will be also in the `out` directory, so you will can review them.

----

If we will use NextJs we should make changes in the directory structure:
```
docs/src/views => pages
docs/src/assets => static
```

layershifter:
I tried `react-static` and seems that it feets better, working on new PR.

Issue #1668, opened 2017-05-11
 in input field: iconPosition doesn't accept the value'Right'.


riyagarg:
Warning: Failed prop type: Invalid prop `iconPosition` of value `right` supplied to `Input`, expected one of ["left"].

layershifter:
It doesn't accept `right` because it's right by default:

```jsx
<Input /> // equals to <Input iconPosition='right' /> 
```

So, it's doesn't make sence. Also, [SUI's css](https://github.com/Semantic-Org/UI-Input/blob/2.2.10/input.css#L280) doesn't have class for `right`.

riyagarg:
but how to deal with the iconPosition in RTL?
As if we are giving the iconPosition = 'left' in the input then its moving to right in RTL. 
 if we are using the default position for icons in the rightside then its position is not changing in RTL,  ideally it should move to left.
but if I am providing the iconPosition ='right', it works fine in the RTL but I am getting the above logged warning.

layershifter:
I don't understand your problem. If you need icon on right, don't set `iconPosition` prop.

However, possible there is another behaviour in RTL SUI.

### LTR

```jsx
<Input icon='users' />
```

![_854](https://cloud.githubusercontent.com/assets/14183168/25953437/0a874ad4-366c-11e7-8ed9-2358934740e2.png)

```jsx
<Input icon='users' iconPosition='left' />
```

![_855](https://cloud.githubusercontent.com/assets/14183168/25953430/0313cb42-366c-11e7-9ac1-385b991ccc97.png)

### RTL

```jsx
<Input icon='users' />
```

![_858](https://cloud.githubusercontent.com/assets/14183168/25953498/3d99df2c-366c-11e7-86a1-d5b180dccf4d.png)

```jsx
<Input icon='users' iconPosition='left' />
```

![_856](https://cloud.githubusercontent.com/assets/14183168/25953490/384e48d2-366c-11e7-90df-f42ffb80d859.png)

riyagarg:
I want the same icon which has a default position (right) in LTR to shift to Left in RTL. which is not happening untill I provide the iconPosition = 'right' in LTR.

levithomason:
I think there might be a valid issue here.  Could you create a codepen that demonstrates the problem?  Fork this to get started: codepen.io/levithomason/pen/ZpBaJX

mcfoton:
[...]

PR #2912, opened 2018-06-13
Fix(Input): include 'right' on action and icon position

kevinpiske:
The 'right' position is currently working, but a warning message said it was invalid.


layershifter:
@kevinpiske Thanks for PR, however we can't merge it 😢 

Some time ago, we had similar issue #1668. SUI CSS doesn't has definitions for a `right` value, it's default.

PR #2928, opened 2018-06-20
feat(Input): expose select() method from ref

kierenj:
(Just as "focus" is currently)


levithomason:
Great, let's also add a test for this.  You can reference the focus test as an example:

https://github.com/Semantic-Org/Semantic-UI-React/blob/43793ae59df9504131d1c9af4762a75321aa3ccc/test/specs/elements/Input/Input-test.js#L145

levithomason:
Merged master and added a test.  This is ready for merge once CI passes.

levithomason:

kierenj:
@levithomason thanks for seeing this one through properly - I had intended to get to the test but other things got in the way. Much appreciated :+1:

PR #2561, opened 2018-02-22
fix(examples): fix in webpack3 examples

layershifter:
Fixes #2559.

PR #2929, opened 2018-06-20
fix(examples): webpack 3 App direct import description

bmbarker90:
It appears that in #2561 the decision was made to switch to using babel-plugin-lodash, but the description of how direct imports are handled was not updated


levithomason:

PR #2934, opened 2018-06-21
feat(Popup): add context prop

kohlikohl:
The property supports setting an existing element as context which will be used to anchor the
Popup to.

e.g
```jsx
<div class="myElement">My Element</div>

// .jsx file
<Popup context={document.getQuerySelector('.myElement')} />
```


src/modules/Popup/Popup.js
```diff
@@ -350,8 +353,8 @@ export default class Popup extends Component {
 
   handleTriggerRef = (triggerRef) => {
     debug('triggerMounted()')
-    if (triggerRef) {
-      this.triggerRef = triggerRef
+    if (this.props.context || triggerRef) {
+      this.triggerRef = this.props.context || triggerRef
```
levithomason: Let's use a default arg here:

```jsx
handleTriggerRef = (triggerRef = this.;props.context) => {
  ...
}
```

test/specs/modules/Popup/Popup-test.js
```diff
@@ -324,6 +324,14 @@ describe('Popup', () => {
     })
   })
 
+  describe('context', () => {
+    it('should set context as triggerRef if passed as prop', () => {
+      const context = {}
+      const popup = wrapperMount(<Popup context={context} />)
+      expect(popup.instance().triggerRef).to.equal(context)
+    })
+  })
```
levithomason: Our tests will be more robust if they test the public behavior.  Since triggerRef is an implementation detail, we want to avoid testing it as it won't validate that the feature works as expected by the user.  The public expectation is that the popup is positioned on the context node.

Suggest doing something like:

1. Create an element and positioning it fixed in the browser somewhere
1. Pass that element to context and open the Popup
1. Capture Popup position information and cleanup the DOM
1. Assert that the Popup was in the correct position

You can reference a positioning test here:

https://github.com/Semantic-Org/Semantic-UI-React/blob/0d00166bd55e9cdab5208678b340179ab4c548c7/test/specs/modules/Popup/Popup-test.js#L142

kohlikohl:
@levithomason Thanks 🙌

levithomason:

PR #2925, opened 2018-06-19
docs(build): use react-static for prerender

layershifter:
Fixes #2826.

PR #2935, opened 2018-06-22
docs(assets): move images to /public

layershifter:
@levithomason I need this for #2925, also as I see images in `stardust` were also moved.


levithomason:

PR #2938, opened 2018-06-22
[WIP] Doc site as a theme designer and sharing app

levithomason:
This PR is exploring the ability for the doc site to become a theme designer.  It would also enable users to save, share, and spin off of other user's themes.

See the RFC #2939


levithomason:
This is now blocked.  The issue is that in browser LESS import statements do not resolve variables the same as when running in node.

In order to support this, we'll need to either:

1. move to CSS in JS and compile the JS on changes (preferred)
2. precompiled a single doc site friendly LESS file that includes all components and variables in order

I vote number 1.  It is future proof and lets us do lots of cool things with persisted themes.  It is also way faster to compiled than LESS.

Number 2 is a smaller short term step for us, but it doesn't lead anywhere great.  I do not think we would be able to provide a very good experience to users in the editor nor in how themes are created and edited.

PR #2940, opened 2018-06-23
chore(package): use shallowequal instead of fbjs and use alias in docs

layershifter:
Picked from #2925.

---

I also used `shallowequal` instead of `fbjs`, reasons are same as there facebook/react#13069.


levithomason:

PR #2941, opened 2018-06-23
chore(package): add `sideEffects: false` for Webpack 4

### LINKED ISSUE #2589: RFC: add `sideEffects: false` to package.json
Webpack 4 now supports a `sideEffects: false` value in package.json which indicates to the tree shaking algorithms that there are no side effects from loading modules. This is fairly important to semantic-ui-react because defaultProps are essentially side effects as far as babel is concerned. BUT, if that's the only side effects semantic-ui-react uses, since they're only relevant to the component itself, they should be safe to shake out. Not a great explanation, but the TL:DR; is that we should add sideEffects: false to the package. Agreed? I can make a PR if that's desired.

https://github.com/webpack/webpack/tree/master/examples/side-effects

layershifter:
Fixes #2589.

### DIFF
--- a/docs/src/views/Usage.js
+++ b/docs/src/views/Usage.js
@@ -151,16 +151,16 @@ const Usage = () => (
         recipes with some of them. You can use them as start point for your projects.
       </p>
 
-      <Header as='h3'>Webpack 3</Header>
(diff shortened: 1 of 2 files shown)


levithomason:

PR #2942, opened 2018-06-23
chore(typings): update TypeScript and use tslint-config-airbnb

layershifter:
This PR:
- updates `typescript` and `tslint` to latest
- uses `tslint-config-airbnb` and uses same config as `stardust-ui`


levithomason:

PR #2947, opened 2018-06-25
fix(Icon): fix handling of aria-label

### LINKED ISSUE #2946: Icon: aria-label is not applied
## Bug Report

#2666 

> Prevent setting aria-hidden attribute when aria-label property is passed. For all other cases aria-hidden="true" is set.

Although this pull request did fix the issue raised, it also consumed the `aria-label` so `getUnhandledProps` will no longer return the `aria-label` and therefore won't be spread in to the component.

https://github.com/Semantic-Org/Semantic-UI-React/blob/caee404b194ef3dbde00282958b9528d736d4667/src/elements/Icon/Icon.js#L83-L92

https://github.com/Semantic-Org/Semantic-UI-React/blob/caee404b194ef3dbde00282958b9528d736d4667/src/elements/Icon/Icon.js#L129-L133

### Steps
```js
<Icon aria-label="Help" name="help" />
```

### Expected Result
```html
<i class="..." aria-label="help" />
```

### Actual Result
```html
<i class="..." />
```

### Version
0.81.1

### Testcase
https://codesandbox.io/s/9l2roxvl0p

layershifter:
Fixes #2946.

Also improves handling of `aria-hidden`.

### DIFF
--- a/src/elements/Icon/Icon.d.ts
+++ b/src/elements/Icon/Icon.d.ts
@@ -55,6 +55,9 @@ export interface IconProps {
 
   /** Icon can have an aria label. */
   'aria-label'?: string
+
+  /** Icon can have an aria hidden. */
+  'aria-hidden'?: string
 }
 
 declare class Icon extends React.Component<IconProps, {}> {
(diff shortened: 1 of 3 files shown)


levithomason:

PR #2948, opened 2018-06-25
docs(Search): update layout for search example to accommodate code

crashuniverse:
Fake data used in a pre tag in search example makes content overflow
& leaves a place for improvement. The example at SUI doesn't face
such problems because the state is not displayed in examples.

I made the following layout decisions to box in the content well and not break existing conventions:
1. Use a `Segment` to contain `pre` like [button#usage-attached-events](https://react.semantic-ui.com/elements/button#usage-attached-events)
2. Have a column with combinations of 6 & 10, and 8 & 8 based on how wide the content of search results need to be displayed.


levithomason:

PR #2954, opened 2018-06-26
UX and styling improvements to Datetime

sarendipitee:
First PR here. Attempted to implement input typing but thought I'd at least try
some more basic things first.  I took some time and looked at several other
Datepickers out there including my own ractive-datepicker, Antd, MUI, 
react-datepicker, react-date, react-times, etc. 

Most of the changes here are cosmetic to improve the UX
of what is already there. There are two *minor* inline styles that I deliberated
over but couldn't avoid. Without them it looks very lobsided due to Popover's
default padding and the Divider's margin.

There is currently an issue where clicking on hours 5pm - 11pm don't show 
the minutes mode after. Any idea about that? I was having trouble finding
the flow after the _.invokeARgs.

Let me know your thoughts.

### DIFF
--- a/src/addons/Datetime/Datetime.js
+++ b/src/addons/Datetime/Datetime.js
@@ -353,13 +353,17 @@ export default class Datetime extends Component {
           />
         )}
       >
-        <DatetimeMenu
-          mode={mode}
-          onDateChangeMode={this.handleChangeMode}
-          onNextPage={this.handleNextPage}
-          onPreviousPage={this.handlePreviousPage}
-          value={value}
-        />
+        {mode != 'hour' && mode != 'minute' &&
+            <DatetimeMenu
+              compact
+              size='small'
+              mode={mode}
+              onDateChangeMode={this.handleChangeMode}
+              onNextPage={this.handleNextPage}
+              onPreviousPage={this.handlePreviousPage}
+              value={value}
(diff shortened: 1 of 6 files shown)


sarendipitee:
I also added localization support using `navigator.language` and the internationalization api (https://caniuse.com/#feat=internationalization good support).

Set your browser to a different language and the months/days of weeks should be changed. Doesn't change any other strings like `'Hour'`, `'Minutes'` that's probably the domain of another localization method.

screen shot 2018-06-26 at 4 46 14 pm

alanmynah:
@JonDum, could this also be place to add some support for edge users? Currently there is no way for them to clear the date field, unless it's a manually added button. 

https://stackoverflow.com/questions/32629330/clear-value-of-input-type-date-in-microsoft-edge

might solve quite a few headaches, as it's a longstanding issue with Edge

sarendipitee:
Ideally I'd like both a little `x` button inside the input and to be able to manually clear it with Backspace or Delete. The X shouldn't be hard to add though à la antd style.

I'd really like to get feedback from @levithomason before continuing on this though

layershifter:
https://github.com/Semantic-Org/Semantic-UI-React/pull/2821#issuecomment-500816353

PR #2957, opened 2018-06-27
feat(Flag): export names in typings

### LINKED ISSUE #2956: Flag: types of property "name" are incompatible
I'm using TypeScript, and I'm getting this error when trying to enter a name into the "name" property

`<Flag name={ip.countryCode} />`

In the definition file, the name property is a huge list of allowed strings:
`name: 'ad' | 'andorra' | 'ae' | 'united arab emirates' | 'uae' .....`

The proper way to do this is to export that huge list, such as "FlagNameValues"

`name: FlagNameValues'

So I can mark the field as countryCode: string | FlagNameValues

Then we would not get this error when trying to use a dynamic value (which would be 99.99999% of use cases)

Will this be fixed?

layershifter:
Fixes #2956.


levithomason:

PR #2958, opened 2018-06-27
fix(Portal): remove non-working `style` prop

### LINKED ISSUE #2886: Portal: recent change breaks in IE11
## Bug Report

A change introduced in #2657 to the portal component causes an error in IE11 when trying to open anything using a portal.

The change: 
https://github.com/Semantic-Org/Semantic-UI-React/blob/500df9ae073dc99b3425fda89fe6b3a3d8f56760/src/addons/Portal/Portal.js#L355

### Steps
Use anything with a portal in IE11 and try to open it.

### Expected Result
Should open the portal.

### Actual Result
Throws the following error:
```
SCRIPT5045: Assignment to read-only properties is not allowed in strict mode
```
Which points to the changed line.

### Version
`semantic-ui-react@0.81.1`

mgandley:
Instead of setting the style property of a dom node (which is a read-only attribute), we set the cssText property instead.

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style for additional documentation.

Even so, most browsers allow setting style directly.  However, doing this in IE causes a crash.

This fixes #2886.

### DIFF
--- a/src/addons/Portal/Portal.d.ts
+++ b/src/addons/Portal/Portal.d.ts
@@ -101,9 +101,6 @@ export interface PortalProps {
   /** Controls whether the portal should be prepended to the mountNode instead of appended. */
   prepend?: boolean
 
-  /** Any inline styles to the Portal container. */
-  style?: object
-
   /** Element to be rendered in-place where the portal is defined. */
   trigger?: React.ReactNode
 }
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -352,7 +352,7 @@ class Portal extends Component {
     if (!isBrowser()) return null
 
     this.rootNode.className = className || ''
-    this.rootNode.style = style || ''
+    this.rootNode.cssText = style || ''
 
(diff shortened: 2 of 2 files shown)


layershifter (changes requested):
I am not sure, that this fixes something. Try to pass to `style` prop to the Portal.

layershifter:
I also don't understand why actually need to assign something to the `style` there, because we don't use this in Portal.

I think that we can safely drop this line and `style` prop on Portal.

mgandley:
Yeah - I made the faulty assumption that the style prop would be a string, but checking the docs it's an object.  Could probably swap to set style via setAttribute which automatically converts the input value to a string.

That said, if not used, makes more sense to just purge the line of code.

layershifter:
Let's try to remove `style` prop fully as non-working.

layershifter:
You need also remove it in `Portal.d.ts`.

layershifter (approved):

mgandley:
You beat me by seconds.  🥇

levithomason:

PR #2959, opened 2018-06-27
 fix(Dropdown): fix double onClose() after blur

### LINKED ISSUE #2953: Dropdown: triggers double onClose event when clicking the icon
## Bug Report
A dropdown ([like this one](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/src/examples/modules/Dropdown/Usage/DropdownExampleRemote.js)) with search, multiple, and selection will trigger the onClose event twice if i click the arrow icon to close, but only once if i click away from the dropdown.

### Steps
open dropdown, select an item, or 2 , or not.
close the dropdown by clicking on the arrow
notice console 

### Expected Result
only one call to onClose

### Actual Result
Double call to onClose 💣

### Version
0.81.2

### Testcase
[Forked codesandbox:](https://codesandbox.io/s/wqqyrv5xjl)
https://codesandbox.io/s/wqqyrv5xjl

layershifter:
Fixes #2953.


src/modules/Dropdown/Dropdown.js
```diff
@@ -1110,31 +1111,35 @@ export default class Dropdown extends Component {
...
   close = (e) => {
-    debug('close()')
-
-    const { onClose } = this.props
-    if (onClose) onClose(e, this.props)
+    const { open } = this.state
+    debug('close()', { open })
 
-    this.trySetState({ open: false })
+    if (open) {
```
layershifter, line 1130: 🌿

levithomason:

Issue #2960, opened 2018-06-27
Search: resultRender function typings mismatch


jymboche:
Shouldnt the resultRender function return a single ReactElement object? Creating a component as per the documentation results in typescript errors:
```
Types of property 'resultRenderer' are incompatible.
Type '(props: SearchResultProps) => Element' is not assignable to type '(props: SearchResultProps) => ReactElement<any>[]'.
Type 'Element' is not assignable to type 'ReactElement<any>[]'.
Property '[Symbol.unscopables]' is missing in type 'Element'. 
```

**Official Documentation**
```
import PropTypes from 'prop-types'
import React from 'react'
import { Label } from 'semantic-ui-react'

import SearchExampleStandard from './SearchExampleStandard'

const resultRenderer = ({ title }) => <Label content={title} />

resultRenderer.propTypes = {
  title: PropTypes.string,
  description: PropTypes.string,
}

const SearchExampleStandardCustom = () => <SearchExampleStandard resultRenderer={resultRenderer} />

export default SearchExampleStandardCustom
```

Typings in Search.d.ts show this however:
https://github.com/Semantic-Org/Semantic-UI-React/blob/e2e9c1e59831a8361a40d75c7d7f27ad19dd1337/src/modules/Search/Search.d.ts#L68-L75

I'm happy to create a PR if this is indeed the case. Thanks!

levithomason:
Fixed in #2962

PR #2962, opened 2018-06-28
fix(Search): fix definition of resultRenderer prop in typings

jymboche:
I believe the resultRenderer expects a single react element not an array of them. See related issue:
#2960


levithomason:

PR #2971, opened 2018-07-02
chore(package): switch to Jest

layershifter:
# WIP


layershifter:
@levithomason it's interesting, but Jest is much slower than our existing solution. I've pulled `stardust` and seen that there are same problems:
```
 PASS  test/specs/lib/eventStack/EventTarget-test.ts (8.016s)
 PASS  test/specs/lib/htmlInputPropsUtils-test.ts
 PASS  test/specs/lib/doesNodeContainClick-test.ts (11.209s)
 PASS  test/specs/lib/eventStack/eventStack-test.ts (11.271s)
 PASS  test/specs/lib/factories-test.tsx (11.314s)
```

I think that it's quite slow. Did I miss something?

levithomason:
Strange, I have no idea.  @davezuko?

dvdzkwsk:
Maybe a Windows issue? The test suite runs in 8s for me on a cold start, and 4.75 on subsequent runs.

layershifter:
### SUIR

I will try to update tests for `views` fully and then will followup again.

Our existing suite is running 02:26: https://circleci.com/gh/Semantic-Org/Semantic-UI-React/5649
Current suite (`/lib` only) runs 00:54: https://circleci.com/gh/Semantic-Org/Semantic-UI-React/5644

Current suite (`/lib` and `/views`) locally:

```
Test Suites: 5 failed, 60 passed, 65 total
Tests:       13 failed, 1789 passed, 1802 total
Snapshots:   0 total
Time:        92.138s (01:32)
```

There are many failed tests, but it looks very poorly.

layershifter:
### Stardust

```
 PASS  test/specs/lib/doesNodeContainClick-test.ts (9.266s)
 PASS  test/specs/lib/factories-test.tsx (9.416s)
```

```
Test Suites: 24 passed, 24 total
Tests:       6 skipped, 351 passed, 357 total
Snapshots:   0 total
Time:        32.258s
```

> Ubuntu 16.04
> i7 2637m

Looks bad to me, too.

layershifter:
I'm worried that after switching to Jest, our tests will be performed substantially longer.

----

If I omit Babel and write code/tests in raw JS, tests are running blazing fast. Looks like Babel has a cold start on each test :thinking:

.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           command: yarn lint
       - run:
           name: Test JavaScript
-          command: yarn test
+          command: yarn test --runInBand
```
thymikee: This is the cause of slow tests, as they are run serially. Try to leverage number of cores you can use on your CI, try different ones, e.g. `yarn test -w=4` for 4 parallel workers. Some CIs are better and some worse in using multiple cores, that's why you need to play with it a bit

layershifter: https://circleci.com/gh/Semantic-Org/Semantic-UI-React/5711

Removal of this option causes build failure. Also, stardust-ui/react#53

thymikee: That's why I was talking about using certain number of workers, not all available (depending on CI, it can tell Jest it has even 128 cpus). You should be good to run jest without flags locally, as you probably have up to 8 cores and enough memory.

PR #2972, opened 2018-07-03
fix(Dropdown): allow `defaultValue` to be a boolean

leonardosul:
A previous fix allowed using booleans as options for dropdowns. This adds the ability to also use a boolean as the defaultValue for a dropdown.

Booleans currently work as default value for dropdowns, but this change removes annoying console errors about propType errors.


levithomason:

PR #2982, opened 2018-07-05
fix(Input): add lang prop for input fields

### LINKED ISSUE #2979: Input: lang prop
## Feature Request
As of now, `<Input lang="fr"/>` will add `lang="fr"` on the wrapping div element instead of the input element itself. It would be great if we could set the lang prop to input elements for localization purposes.

### Proposed solution
I can implement the additional property and submit a PR if someone from the team can point me in the right direction. Will it work if I just add the `lang` prop in this file: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Input/Input.d.ts ?

foysalit:
Lang attribute on input elements help the browser provide better localization support. Right now, adding lang prop to input adds that as an attribute to the wrapping div which doesn't serve the localization purpose. 

Fixes #2979.

### DIFF
--- a/src/lib/htmlPropsUtils.js
+++ b/src/lib/htmlPropsUtils.js
@@ -8,7 +8,6 @@ export const htmlInputAttrs = [
 
   // LIMITED HTML PROPS
   'accept',
-  'lang',
   'autoCapitalize',
   'autoComplete',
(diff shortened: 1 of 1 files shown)


levithomason:

PR #2985, opened 2018-07-06
docs(views): wrap each view with divs

layershifter:
This PR is a part of preparement for #2925 and will allow to clarify changes there.


levithomason:

PR #2986, opened 2018-07-06
docs(plugins): add a plugin for collect sources of examples

layershifter:
Picked from #2925.


levithomason:

PR #2990, opened 2018-07-09
chore(Sidebar): use EventStack component

layershifter:
We have a cool API called EventStack. It's time to make it more awesome 🐱  This PR offers a new component exposes the EventStack API as public and provides a declarative way to manage it.

This will allow to reduce complexility with subscribtions to events and will allow our users to work with our API deeply.


src/modules/Sidebar/Sidebar.js
```diff
@@ -191,6 +162,7 @@ class Sidebar extends Component {
       <Ref innerRef={this.handleRef}>
         <ElementType {...rest} className={classes}>
           {childrenUtils.isNil(children) ? content : children}
+          {visible && <EventStack name={'click'} on={this.handleDocumentClick} />}
```
layershifter: 30 lines vs 1 line, exciting result ✌️

levithomason: Very cool, I'll give this a proper review before merging.  It looks like a fantastic idea :D

levithomason:
I assume the future of this component will change now that we're publishing EventStack from elsewhere (eventually Stardust)?

layershifter:
Yep, I will move this component to a package.

levithomason:

PR #2993, opened 2018-07-10
fix(Icon): add 'file text outline' type

paulkoegel-wessels:
Icon "file text outline" exists in the code but not in the `SemanticICONS` type.

This PR adds the missing string literal to the `SemanticICONS` type as well as `SUI.js`.

Left (highlighted):
`<Icon name="file text outline" />`

Right:
`<Icon name="file text" />`

sh-602-prod-fr2 2018-07-10 12-38-17

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1892,6 +1892,7 @@ export type SemanticICONS =
   | 'feed'
   | 'female homosexual'
   | 'file text'
+  | 'file text outline'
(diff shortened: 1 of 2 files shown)


layershifter (changes requested):
Please add it into `SUI.js`, too.

layershifter (approved):

levithomason:

PR #3008, opened 2018-07-14
docs(Button): update ButtonExampleAnimated to match SemanticICONS type

tuol:
Icon name was not confirm SemanticICONS type


levithomason:

PR #3015, opened 2018-07-17
fix(Portal): use portalNode to check focus

### LINKED ISSUE #3005: Popup: closed on trigger blur even when portal content is focused
### Steps
* create popup with input as trigger (with on="focus") and second input as popup content
* click on input to show popup
* click on input in popup

### Expected Result
* second input focused, popup is still open

### Actual Result
* popup closed

### Version
* works correctly on 0.81.1, 0.81.3
* does not work correctly on 0.82.0

### Testcase
https://codesandbox.io/s/9yv949y13o

layershifter:
Fixes #3005.


levithomason:

PR #3016, opened 2018-07-17
fix(EventStack): use `capture` option in addEventListener()

### LINKED ISSUE #3006: Popup: when multiple popups present, click on the other trigger doesn't close original popup
## Bug Report
This seems to be the same issue that was reported #1251 But it got back with new versions.

### Steps
Create several popups with a click trigger.
Click first trigger --> first popup opens, then click second trigger --> the second opens, but the first stays open.

### Expected Result
The first popup should be closed when the second trigger clicked.

### Actual Result
The first popup stays open.

### Version
0.82.0
### Testcase
https://codesandbox.io/s/3rx96jzjr6

layershifter:
Fixes #3006.

### DIFF
--- a/src/lib/eventStack/EventTarget.js
+++ b/src/lib/eventStack/EventTarget.js
@@ -83,7 +83,6 @@ export default class EventTarget {
     this.handlers.set(eventType, handler)
     this.target.addEventListener(eventType, handler, {
       capture: true,
-      passive: true,
     })
   }
 
--- a/src/lib/eventStack/EventTarget.js
+++ b/src/lib/eventStack/EventTarget.js
@@ -22,10 +22,10 @@ export default class EventTarget {
   addHandlers(poolName, eventType, eventHandlers) {
     this.removeTargetHandler(eventType)
 
-    if (!this.pools.has(poolName)) {
-      this.pools.set(poolName, EventPool.createByType(poolName, eventType, eventHandlers))
-    } else {
+    if (this.pools.has(poolName)) {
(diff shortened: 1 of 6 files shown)


src/lib/eventStack/EventTarget.js
```diff
@@ -22,10 +22,10 @@ export default class EventTarget {
   addHandlers(poolName, eventType, eventHandlers) {
     this.removeTargetHandler(eventType)
 
-    if (!this.pools.has(poolName)) {
-      this.pools.set(poolName, EventPool.createByType(poolName, eventType, eventHandlers))
-    } else {
+    if (this.pools.has(poolName)) {
       this.pools.set(poolName, this.pools.get(poolName).addHandlers(eventType, eventHandlers))
+    } else {
+      this.pools.set(poolName, EventPool.createByType(poolName, eventType, eventHandlers))
```
layershifter, line 28: Revert condition to omit `!`.

src/lib/eventStack/EventTarget.js
```diff
@@ -81,7 +81,10 @@ export default class EventTarget {
     const handler = this.createEmitter(eventType, this.pools)
 
     this.handlers.set(eventType, handler)
-    this.target.addEventListener(eventType, handler)
+    this.target.addEventListener(eventType, handler, {
+      capture: true,
+      passive: true,
```
layershifter: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

We need `capture` to fix the issue.

layershifter:
This fixes issue, but creates another issues. I need a better idea there.

PR #3020, opened 2018-07-18
fix(Popup): fix issue with open prop

### LINKED ISSUE #2991: Popup: not working with open prop
### Steps
Create a Popup with a trigger and then set it to have open=true.
### Expected Result
I expect that always open popups will display where they are meant to.
### Actual Result
The always open popup shows at the top-right corner of the page (top=0, right=0) rather than at the button.
### Version
0.82.0
### Testcase
https://codesandbox.io/s/oj2o7mmqky

msrikanth508:
**Issue:** After rendering `trigger` prop there is no way to calculate popup position styles, hence it is rendering somewhere in the DOM
**Fix:**  Calling `setPopupStyle` func in trigger handler to calculate popup positions based on `trigger` prop element

Fixes #2991.


src/modules/Popup/Popup.js
```diff
@@ -357,6 +357,11 @@ export default class Popup extends Component {
   handleTriggerRef = (triggerRef) => {
     debug(`handleTriggerRef(${triggerRef})`)
     this.triggerRef = triggerRef
+    // if it is auto popup
+    // then set popup coordinates when triggerref is available
+    if (_.has(this.props, 'open')) {
+      this.setPopupStyle()
+    }
```
levithomason, line 378: Should this not only be set if `open` is truthy?  Also, let's link the issue here in the code:

```jsx
handleTriggerRef = (triggerRef) => {
  const { open } = this.props
  // ...

  // https://github.com/Semantic-Org/Semantic-UI-React/issues/2991
  if (open) this.setPopupStyle()
}
```

test/specs/modules/Popup/Popup-test.js
```diff
@@ -147,19 +147,28 @@ describe('Popup', () => {
...
+          expect(element.style.position).to.equal('absolute')
+          expect(top).to.not.equal('')
+          expect(left).to.not.equal('')
+          expect(bottom).to.not.equal('')
+          expect(right).to.not.equal('')
+
+          // element.style.should.have.property('top', '')
+          // element.style.should.have.property('left', '')
+          // element.style.should.have.property('bottom', '')
+          // element.style.should.have.property('right', '')
```
levithomason: Let's cleanup cruft.

test/specs/modules/Popup/Popup-test.js
```diff
@@ -147,19 +147,28 @@ describe('Popup', () => {
...
-        element.style.should.have.property('bottom', '')
-        element.style.should.have.property('right', '')
+        setTimeout(() => {
+          const element = document.querySelector('.popup.ui')
+          const { top, left, bottom, right } = element.style
+          expect(element.style.position).to.equal('absolute')
+          expect(top).to.not.equal('')
+          expect(left).to.not.equal('')
+          expect(bottom).to.not.equal('')
+          expect(right).to.not.equal('')
```
levithomason, line 196: Would be ideal if these made more specific assertions about where it should be positioned opposed to where it should not be positioned.

levithomason (changes requested):
This PR is made from `msrikanth508:master` so I cannot push fixes.  I will branch from it, make the udpates I've noted and probably make a release soon.

levithomason:
It seems there are some deeper issues here with InnerPortal and ReactDOM.createPortal.  See #3029.

levithomason:
I should also note, the tests here are passing but the feature is not working.

src/modules/Popup/Popup.js
msrikanth508: Yes, and we should also add `open` in propTypes list.

msrikanth508: I'm not sure I understand inherited props. In `popup.d.ts` file we are already inheriting from PortalProps.

test/specs/modules/Popup/Popup-test.js
msrikanth508: Added new assertion `expect(element.parentNode).to.equal(document.body)`

msrikanth508:
@levithomason I've fixed code review comments. Here is the working example
#### 1. Default open prop
```javascript
<Popup trigger={<Button >Opened</Button>} open >test</Popup>
```
#### 2. Change status dynamically
```javascript
class Example extends React.Component {
  state ={
    open: false,
  }
  componentDidMount() {
    setTimeout(() => {
      this.setState({
        open: true,
      })
    }, 5000)
  }
  render() {
    return (
      <Container id='introduction-page' style={{ paddingTop: 100 }}>
        <Popup trigger={<Button >Opened</Button>} open={this.state.open} >test</Popup>
      </Container>
    )
  }
}
```
I'm facing one issue though, I'm trying to create an example in docs with `open` prop but it's not working. Need your help here.

levithomason:
>I'm facing one issue though, I'm trying to create an example in docs with open prop but it's not working. Need your help here.

This is the same issue I was having.  I'll have to come back to this one and do some digging.

msrikanth508:
@levithomason Did you get a chance to look into this?

levithomason:
I did not, apologies for the delays.  It is on my higher priority list, along with the Dropdown issues we've had recently.  Working on these a bit today as well.

fc:
For those who need this functionality, this is a regression and if you rollback to this version it'll work: `semantic-ui-react@0.81.1`

levithomason:
This needs tested in a browser to ensure it is working as expected.  Last I pulled this fork and updated one of the examples to have `<Popup open />`, the Popup was still incorrectly positioned in the upper right of the body.

Are there any others who are getting the correct results from this PR?

zlove:
For others affected by this issue, I was able to work around it by setting a flag in `componentDidMount` indicating the `context` component mounted, and then render the controlled `Popup`. You're mileage may vary.

```javascript
class SomeComponent extends React.Component {
    state = {
        element: null,
        mounted: false,
    };

    componentDidMount() {
        this.setState({ mounted: true });
    }

    setRef = element => this.setState({ element });

    render() {
        // ... some other code
       return (
            <React.Fragment>
                {popupContent && mounted && (
                    <Popup
                        context={element}
                        content={popupContent}
                        position="bottom left"
                        hoverable
                        open
                        wide
                    />
                )}
                <div ref={this.setRef}>
                    I am the context component that the Popup is attached to.
                </div>
            </React.Fragment>
        );
}
```

layershifter:
I've reached some progress in https://github.com/Semantic-Org/Semantic-UI-React/pull/3029
We should check why tests are broken, manual test was passed.

msrikanth508:
@levithomason  I've created new branch from my commits.
https://github.com/msrikanth508/Semantic-UI-React/tree/fix/pop-up

layershifter:
Fixed in #3029.

PR #3029, opened 2018-07-22
fix(Popup): default open position

### LINKED ISSUE #2991: Popup: not working with open prop

levithomason:
Fixes #2991.

---

This PR was branched from #3020 to perform some cleanup and updates since the original PR was from a fork's master and couldn't be updated.

I've also added a "Default Open" example to the docs for this use case.

It seems there is a bug with the use of `ReactDOM.createPortal` in `InnerPortal`.  When rendering a Portal with `open` set to true on first render, an infinite loop is entered upon rendering the `InnerPortal`.  If `createPortal` is bypassed, there is no infinite loop.  I do not yet know why this is :/


layershifter:
@levithomason Seems that it's really an bug in ReactDOM, it tried to debug this. The issue comes from the `renderToStaticMarkup`, similar issue: reactjs/react-modal#593.

The simpliest and stupid solution, disable html preview for specific examples. However, I'm not sure, that it's the best solution.

levithomason:

Issue #2919, opened 2018-06-18
[DOC] The example for using shorthand / renderBodyRow for the Table Component is somehow hidden


dvriv:
### Problem description
The only example using `renderBodyRow` (and generally using shorthand) I could find is this:
https://react.semantic-ui.com/collections/table#states-warning-shorthand

I guess it is considered a "shorthand", but is still really hard to find it. I was trying to figure out how to use `renderBodyRow` and I read that I needed to use the shorthand, so I search for _shorthand_ and found the example by pure luck. This example made it clear how I had to use it.

### Proposed solution
Create another section about shorthand / getting data from a object and put it in there

layershifter:
Our examples should be refactored like [`Button`s](https://react.semantic-ui.com/elements/button) where each example has examples with children and shorthand APIs.

However, it's the titanic work 💣 Help is highly appricated.

hissalht:
Working on it here #3056

hissalht:
I started working on it here #3056. Help and guidance would be very appreciated :D

husseyexplores:
@layershifter I suppose this still needs work. I'd love to help improve the docs with shorthand examples like [Buttons](https://react.semantic-ui.com/elements/button). 
I just have one question, how would I go about this? Should I keep opening pull requests as I add examples, or just create one giant pull request for each section? I'm new to open source contributions, that's why I'm asking what's the norm. Thanks!

PR #3056, opened 2018-07-31
[DOC] Add Table examples using the shorthand API

hissalht:
Trying to improve on #2919.
WIP

Also, I'm pretty new to this contribution thingy. Please let me know if I do anything wrong.

### DIFF
--- a/docs/src/examples/collections/Table/States/TableExampleActiveShorthand.js
+++ b/docs/src/examples/collections/Table/States/TableExampleActiveShorthand.js
@@ -2,16 +2,16 @@ import React from 'react'
 import { Table } from 'semantic-ui-react'
 
 const tableData = [
-  { name: 'Jamie', status: 'Approved', notes: 'Requires call' },
-  { name: 'John', status: 'Selected', notes: undefined },
-  { name: 'Jamie', status: 'Approved', notes: 'Requires call' },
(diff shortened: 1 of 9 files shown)


hissalht:
So, I'm trying to make a shorthand version of the [structured table example](http://react.semantic-ui.com/collections/table/#types-structured). However, I'm facing two issues here :

- From what I saw in the source, it is not possible to render a multiple-row header using the `headerRow` shorthand prop.

- Rendering rows containing cells with `rowSpan > 1` is not as easy as I thought... Still, I came up with the following :
https://github.com/hissalht/Semantic-UI-React/blob/53c37612ba593c68b7ec83bc3c8c921daee2b699/docs/src/examples/collections/Table/Types/TableExampleStructuredShorthand.js#L15-L55
However, in order for this to work, we need a third parameter `data` which is the array of data used to generate rows. To make it work, we have to change the `_.map()` call in `Table.js` to use the third parameter like so : https://github.com/hissalht/Semantic-UI-React/commit/53c37612ba593c68b7ec83bc3c8c921daee2b699#diff-cc5fbf946099977c9d1f74f87091c8cb

So, should I just ignore this particular example ? Open an issue about that "no multiple-row header" shorthand thing  ? I'm very new to github contributing so I dont really know what is the procedure here  ¯\_(ツ)_/¯

PR #3069, opened 2018-08-06
docs(perf): refactor ComponentControls

layershifter:
### Before

image

### After

image

---

This PR grabs controls from Stardust. I asked my collegues about this, before this change they don't know about editor 🤔 

This also allows us to omit `Popup`s creation for each example, it costs over 50-100 ms in development mode.


docs/src/components/ComponentDoc/ComponentControls/ComponentControlsEditCode.js
```diff
@@ -2,20 +2,16 @@ import PropTypes from 'prop-types'
...
+  </Menu.Item>
 )
 
 ComponentControlsEditCode.propTypes = {
   active: PropTypes.bool,
   onClick: PropTypes.func,
 }
 
-export default updateForKeys(['active'])(ComponentControlsEditCode)
+export default ComponentControlsEditCode
```
layershifter, line 17: Now, there is no sence in this HOCs.

levithomason:
>This PR grabs controls from Stardust. I asked my collegues about this, before this change they don't know about editor 🤔

This is exactly what prompted the refactor here!  I was quite surprised and bummed out that people didn't know about the editor :(

levithomason:
Should we also reverse the order of the links as in Stardust?  This makes the first link "Try it".

levithomason (approved):
See question, make a decision, and merge 👍

levithomason:
On second thought, we can always reorder them after this is merged 😄

Issue #3071, opened 2018-08-07
Message component type definition is inconsistent and missing 'medium'. Propose to import and use common SemanticSIZES type


ltfschoen:
### Steps
A clear and concise description of steps to reproduce the problem.

I'm implementing a React component as follows:
```
import { BareProps, UnlockI18n } from './types';

import React from 'react';

import Message from 'semantic-ui-react/dist/commonjs/collections/Message/index';
import { SemanticSIZES } from 'semantic-ui-react/dist/commonjs/index';

type Props = BareProps & {
  error?: UnlockI18n | null,
  size?: SemanticSIZES | 'medium' // hack until PR accepted by semantic-ui-react
};

export default class Notification extends React.PureComponent<Props> {
  render () {
    const { error, size = 'tiny' } = this.props;

    if (!error) {
      return null;
    }

    const renderMessageBody = (error: UnlockI18n) => {
      return (
        <div>
          {/* <Message.Header>{error.key}</Message.Header> */}
          <p>{error.key}: <b>{error.value}</b></p>
        </div>
      );
    };

    return (
      <div className='ui--Notifications'>
        { error.key === 'error' ? <Message size={size} error>{ renderMessageBody(error) }</Message> : null }
        { error.key === 'warning' ? <Message size={size} negative>{ renderMessageBody(error) }</Message> : null }
      </div>
    );
  }
}
```

I'm having to manually add the available types and omit `'medium'` from the type definition of `size?: 'mini' | 'tiny' | 'small' | 'large' | 'big' | 'huge' | 'massive'`, when I'd rather it just be `size?: SemanticSIZES`, and have `| 'medium'` included.

The reason why I have to do this is because I'm importing from 'semantic-ui-react/dist/commonjs/index', which includes: `export type SemanticSIZES = 'mini' | 'tiny' | 'small' | 'medium' | 'large' | 'big' | 'huge' | 'massive';`, and then I'm passing that `size` type into the component `<Message size={size} ...`, whose type definition is in 'semantic-ui-react/dist/commonjs/collections/Message/Message.d.ts' as follows:
```
...
import {
  SemanticCOLORS,
  SemanticShorthandCollection,
  SemanticShorthandContent,
  SemanticShorthandItem
} from '../..';
...
export interface MessageProps {
...
  /** A message can have different sizes. */
  size?: 'mini' | 'tiny' | 'small' | 'large' | 'big' | 'huge' | 'massive';
```

[...]

PR #3072, opened 2018-08-07
fix(Message): Fixes Issue #3071 by reusing common type `SemanticSIZES` to resolve missing `'medium'`

ltfschoen:


layershifter:
Please see, https://github.com/Semantic-Org/Semantic-UI-React/issues/3071#issuecomment-411137810

PR #3075, opened 2018-08-07
 docs(build): fix reloading of examples sources

layershifter:
Title is obvious.

### DIFF
--- a/docs/src/staticUtils/getExampleSources.js
+++ b/docs/src/staticUtils/getExampleSources.js
@@ -1,3 +1,8 @@
-const getExampleSources = () => require('../exampleSources')
+import fs from 'fs'
+import path from 'path'
+
+const filename = path.resolve(__dirname, '..', 'exampleSources.json')
+
+const getExampleSources = () => JSON.parse(fs.readFileSync(filename))
 
 export default getExampleSources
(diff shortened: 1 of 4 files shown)


docs/src/staticUtils/getExampleSources.js
```diff
@@ -1,3 +1,8 @@
-const getExampleSources = () => require('../exampleSources')
+import fs from 'fs'
+import path from 'path'
+
+const filename = path.resolve(__dirname, '..', 'exampleSources.json')
+
+const getExampleSources = () => JSON.parse(fs.readFileSync(filename))
```
layershifter, line 6: `require()` will cache an object, so it will not changed on after rebuild.

gulp/tasks/docs.js
```diff
@@ -15,7 +15,8 @@ const { paths } = config
 const g = loadPlugins()
 const { log } = g.util
 
-const handleWatchChange = e => log(`File ${e.path} was ${e.type}, running tasks...`)
+const handleWatchChange = filename =>
+  log(`File ${path.basename(filename)} was changed, running tasks...`)
```
layershifter, line 19: `gulp`'s API was changed, now message works ^_^

Issue #3081, opened 2018-08-13
docs: display error of React Component Item


starsyork:
## Bug Report
For the React [Item](https://react.semantic-ui.com/views/item/#variations-link) component examples -- Metadata, Link, Extra Content, Divided, the example preview cannot show up properly.

layershifter:
Thanks, the valid report 👍

PR #3082, opened 2018-08-13
docs(Item): cleanup examples

starsyork:
This PR is to correct the document preview issue of React Item Component #3081 

The incorrect name of 
`<ImageComponent src='/images/wireframe/short-paragraph.png' />` 
Should be 
`<Image src='/images/wireframe/short-paragraph.png' />`


levithomason:
Thank you!

PR #3083, opened 2018-08-13
fix(Message): export size prop in typings

layershifter:
Related #3071.


levithomason:
Released in `semantic-ui-react@0.82.3`.

PR #3085, opened 2018-08-13
chore(build): fixes for UMD build

### LINKED ISSUE #3079: Publish a cdn that supports stack snippets
Hi, sometimes when i want to help others with semantic-ui-react on stackoverflow i want to write a snippet using the built in stack snippets. 
This requires a umd / cdn version. but none of the cdn's out there are supported by stack snippets editor because of cors or other security issues.

It would be great if we had an official cdn for semantic-ui-react.

layershifter:
Fixes #3079.


.babel-preset.js
```diff
@@ -31,7 +32,8 @@ const plugins = [
   [
     'transform-react-remove-prop-types',
     {
-      mode: 'wrap',
+      mode: isUMDBuild ? 'remove' : 'wrap',
+      removeImport: isUMDBuild,
```
layershifter, line 36: Remove `propTypes` fully from UMD build

package.json
```diff
@@ -15,8 +15,8 @@
   "scripts": {
     "build": "cross-env NODE_ENV=production gulp build",
     "build:changelog": "github_changelog_generator --no-issues --no-unreleased --release-branch master --since-tag $(git describe --abbrev=0 --tags $(git rev-parse HEAD~300))",
-    "build:docs": "gulp --series dll build:docs",
-    "build:dist": "gulp --series dll build:dist",
```
layershifter, line 19: We don't need dll in `dist` and `docs`

webpack.config.babel.js
```diff
@@ -88,13 +55,6 @@ if (!__TEST__) {
   )
 }
 
-if (__DEV__) {
-  webpackConfig.plugins.push(
-    new webpack.HotModuleReplacementPlugin(),
-    new webpack.NoEmitOnErrorsPlugin(),
-  )
-}
-
```
layershifter, line 97: We are using `react-static` now, so we don't use `__DEV__` variable.

PR #3104, opened 2018-08-23
fix(Dropdown): fix deepEqual bug

### LINKED ISSUE #3000: Dropdown: componentWillReceiveProps performs deep equal on React component
### Steps
Create a Dropdown with content, for example like in https://react.semantic-ui.com/modules/dropdown#usage-item-content. The content must contain a React component, for example:
```JS
content: <Header icon='desktop' content='Desktop' subheader='The largest size' />,
```

### Expected Result
App should run normally, display the header, update the dropdown as needed, etc.

### Actual Result
When `Dropdown`'s `componentWillReceiveProps` runs, the app freezes for a minute or more. The issue is in line https://github.com/Semantic-Org/Semantic-UI-React/blob/972b999b7e9f85fb6722a1749c4c121c03e5613d/src/modules/Dropdown/Dropdown.js#L426. The deep equals recurses into the `Header` object, which has cyclical references (for example `nextProps.options[0].content._owner === nextProps.options[0].content._owner.return.child`) as well as recursive references through the entire app treee (like `_owner._debugOwner._debugOwner._debugOwner...` all the way to the root) and thus the app hangs. 

The code in question came in via https://github.com/Semantic-Org/Semantic-UI-React/pull/2252, where performance risks of a deep equal were already mentioned.

### Version
0.82.0

### Testcase
The bug won't reproduce in a small sample app, because in that case lodash.isEqual can traverse the entire app quickly. You'll need a fairly large app to get a freeze of several seconds and it gets worse in debug mode, because then React components have the _debugOwner property set, which can be followed all the way to the root component.

pedromtorres:
Fixes #3000.


mvanlonden:
This fixes the same issue I'm running into

levithomason (approved):

levithomason:
I linked the issue to the code change.  Tests just need to pass.  Thanks for the PR!

mvanlonden:
@levithomason when do you think this will be merged? The codecov check has taken over 48hrs, is that typical? 

Thanks for maintaining this project!

joetuson:
Would love to see this merged. Seems like codecov is stalled?

Thanks for all the work you do!

PR #3107, opened 2018-08-26
chore(typings): add strict props

### LINKED ISSUE #2836: export element props without [key: string]: any;

evollu:
Fixes #2836.

---

update of Strict type dropdown #2839 , covers all props
based on discussion on #2836 
@levithomason


evollu:
any updates?

layershifter:
@evollu thank you for cool your there 👍  I've added some changes and going to merge this 🚀

levithomason:

PR #3109, opened 2018-08-27
fix(Popup): missing CSS style update

### LINKED ISSUE #3050: Popup no longer updates when horizontalOffset changes
## Bug Report

I just updated from 0.80.2 to 0.82.1.

Now when you change `horizontalOffset` in a Popup component, it will no longer re-calculate the style of the portal, it will only calculate once on mount of the portal.

I believe `setPopupStyle` needs to be called again when it detects a change in the `horizontalOffset` property.

### Steps

Have a component that updates the `horizontalOffset` while the popup is open, eg. when the mouse moves. (see testcase)

### Expected Result
When the mouse moves, the popup moves with the mouse.

### Actual Result
The popup only gets the position when you start hovering, and will not update.

### Version
0.82.1

### Testcase

Broken on `0.82.1`: https://codesandbox.io/s/23j0n0mkrp
Working on `0.81.1`: https://codesandbox.io/s/p7rj6z961x

pferreir:
This PR should fix #3050. I suspect this was introduced by 06ab95f39bf3f60b9b9461a42d58da60a98dd730, with the removal of `requestAnimationFrame(...)`.

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -152,6 +152,17 @@ export default class Popup extends Component {
     this.mounted = true
   }
 
+  componentDidUpdate(prevProps) {
+    // if horizontal/vertical offsets change, re-calculate the CSS style
+    const { horizontalOffset, verticalOffset } = this.props
+    if (
+      horizontalOffset !== prevProps.horizontalOffset ||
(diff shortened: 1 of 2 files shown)


levithomason:
Superb!  Thanks for the excellent PR.

levithomason:

Issue #3062, opened 2018-08-02
Dropdown: doesn't close when selecting another dropdown


evollu:
### Steps
A clear and concise description of steps to reproduce the problem.

1. go to link https://react.semantic-ui.com/modules/dropdown/#usage-close-on-blur
2. select 1st drop down, select 2nd dropdown, select 1st dropdown again

### Expected Result
The result that you expected.
only one dropdown should open at a time

### Actual Result
The actual result that happened 💣
both dropdowns are open
<img width="513" alt="screen shot 2018-08-02 at 5 19 59 pm" src="https://user-images.githubusercontent.com/9213224/43611795-51dea72a-9678-11e8-8fb6-c29686e3e997.png">

### Version
0.82.1

### Testcase
[Fork, update, and replace this pen to show the bug]:
https://react.semantic-ui.com/modules/dropdown/#usage-close-on-blur

annjawn:
I can confirm that this is really an issue. I am encountering the same after using 3 dropdowns in a Header bar. It's really inconsistent. Sometime the menu closes when another one opens sometimes it doesn't.

![menu](https://user-images.githubusercontent.com/26977294/43665699-ec8e057e-9725-11e8-93ea-d5df178caf17.gif)

layershifter:
@annjawn please make sure that you're using the latest version.

layershifter:
The described issue comes from the `closeOnBlur` prop, I need to perform an additional checkes, it can be expected behaviour.

annjawn:
I can confirm that this issue not only happens when using Subcomponent API `Dropdown.Menu` but also using shorthand with `options`. I just updated the module about an hour ago hoping for a fix but the issue persists. As a workaround I had to manage individual states using `open`, `onOpen`, and `onBlur` for each `<Dropdown>` component.

levithomason:
This could be related to #3006.  Multiple popups have the same issue.  Perhaps this is an event handling problem in EventStack.

crowdozer:
If it helps at all, the process is actually _very_ consistent, exact steps:
1. Open a dropdown
2. Open a new dropdown
3. Open any dropdown but the one from step 2 
4. The dropdown from step 2 is no longer being managed, repeat from step 2 with a new dropdown to open as many as you want 
5. Click the dropdowns that are unmanaged to manually close them, they work as intended until you do this to them again
[...]

PR #3124, opened 2018-09-04
fix(EventStack): fix compatibility with IE11

### LINKED ISSUE #3043: Uncaught exception on IE 11 (Edge Compatibility)
### Steps
Clone [this repo](https://github.com/rjb8682/test-react) (my own sample project):
* Note: I'm using yarn, but of course this should work with npm with minimal changes

**Run these steps**
1. Run `yarn install`
2. Run `yarn start`
3. Go to `localhost:8080`
4. Click button on webpage
5. Type in input box 

### Expected Result
No errors in console

### Actual Result
The following errors show in the console every time an event fires (mouse click, or keyboard press)

> SCRIPT5022: Exception thrown and not caught
> File: bundle.js, Line: 40029, Column: 3

When clicking the file that is shown in the error, it links to the _nonIterableSpread function from the `@babel-runtime`

### Version
Semantic UI React version: `0.82.1`
Internet Explorer update version: `11.0.70`

layershifter:
Fixes #3043.

---

# What was done?

### Move to a separate repo

EventStack was moved to separate package and was published as [`@semantic-ui-react/event-stack`](https://github.com/layershifter/event-stack).

_Motivation:_ this package requires a specific test config for browsers. However, I'm not happy with this decision, I prefer monorepos :smiley_cat: Possible, [Stardust](https://github.com/stardust-ui) will adopt it.

### Browser compability

EventStack is now fully compatible with IE11. To avoid regresions I added browser tests:
- Chrome & Firefox: https://circleci.com/gh/layershifter/event-stack
- IE11: https://ci.appveyor.com/project/layershifter/event-stack

### Performance :rocket: 

I and my сolleagues carefully profiled this part of the lib, performance was drasticaly improved. The most slow was [`EventSet`](https://github.com/layershifter/event-stack/blob/master/src/lib/EventSet.ts):
- [`constructor()`](https://jsperf.com/suir-eventset-constructor) - almost 10x
- [`addHandlers()`](https://jsperf.com/suir-eventset-addhandlers) - almost 10x
- `dispatchEvent()` - [10x](https://jsperf.com/suir-eventset-dispatchsingle) for `default` pool, [1.5x](https://jsperf.com/suir-eventset-dispatchmultiple2) for named pool
- [`removeHandlers()`](https://jsperf.com/suir-eventset-removehandlers) - almost 5x

Feel free to edit benchmarks and propose more performant solution.

-----

There I also couple serious bugs with `EventStack` event handling, I will concentrate on them after merge. I'm also planning to move #2990 to this package.


levithomason:
>Possible, Stardust will adopt it.

Yes please!

levithomason:
We have a fork of this in Stardust as well.  Yes, I would love to have a single utility here.  It is also duplicated in Fabric (Microsoft).  One of the goals of Stardust is to normalize these kinds of utilities between frameworks.  Let's merge this for now, then we can port it to Stardust and publish it as a monorepo.

There might still be a bug here, not sure.  Here's what I think I've found.  I noticed the bug in #3062 goes away if we fix the `sub` method.  Currently, the entire pool is removed then re-added on sub.  However, the pool shares handlers with other components under the same pool name.  When a second Dropdown opens it calls `sub` for close on document click, which removes the entire pool (including the first Dropdown's close on document click), then adds only the second Dropdown's close on document click.

When I use `document.addEventListener` instead, all works as expected.  When I remove the line in EventStack that removes the entire pool on sub, all works fine as well.  So, I think we need to **skip** sub if that handler already exists and **not remove/add the entire pool**.

levithomason:

PR #3125, opened 2018-09-04
feat(List): allow overriding role prop

johannao76:
Update List and ListItem components to move ...rest to the end of the props list during render. This allows the user to override props like the aria role (which I need in my case).


levithomason:

PR #3131, opened 2018-09-05
chore(package): update deps

layershifter:
This PR updates projects dependencies and adds some changes to tests.

### `autoGenerateKey`

Is now `true` by default in shorthand tests because it's `true` in factories, this allows make our tests more strict.

### `AccordionPanel`

The update of `enzyme` allows us enable some tests for `AccordionPanel` :+1:


levithomason:

Issue #3138, opened 2018-09-08
Invalid prop `as` supplied to `MenuItem` when using a custom component.


tomitrescak:
## Bug Report

Hi, previously I could specify the "as" property to MenuItem as a custom component. This is a key functionality if I want to use custom Link components, such as those from React router or Reach Router.

### Version
0.82.3

### Testcase

Here you go (check out the console):

https://codesandbox.io/s/q71qy7kw6q

layershifter:
Fixed in #3139.

PR #3139, opened 2018-09-08
fix(customPropTypes): allow to pass an object to "as"

tomitrescak:
See #3138 for the problem described. 
Solves problem when custom links are used


layershifter:
A valid change because [`forwardRef` API](https://reactjs.org/docs/forwarding-refs.html) returns an object.

layershifter:
@tomitrescak thanks 👍

levithomason:

PR #3143, opened 2018-09-10
[RFC] feat(Accordion): add transitions

### LINKED ISSUE #2114: Accordion: component should support transitions

layershifter:
Fixes #2114.

# RFC

This PR is an early POC of animations support.

works

### `withTransition` HOC

This HOC should be implement more early. Can add transition functionality to any component.

### Height, height, HEIGHT!

SUI heavily utilizes jQuery's [`slideDown()`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/accordion.js#L208) and [`slideUp()`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/accordion.js#L269). For obvious reasons we can't use them 😼  As temporary solution I've added [`react-animate-height`](https://github.com/Stanko/react-animate-height), however it doesn't have even tests (https://github.com/Stanko/react-animate-height/issues/54).

Need to think more about it, @levithomason do you have any ideas?

# TODO

- [x] add propType for `transition`
- [ ] finilize `withTransition()` hoc
  - [x] add typings
  - [ ] add tests


layershifter:
There are issues:
- our tests doesn't like HOCs :cat: 
- `react-animate-height` has own markup and it creates two nested `div`s

@levithomason can you please take a look and give your opinion? Should I continue to work on this PR?

levithomason:
I think I would prefer a more scalable and consistent solution to our transitions.  Adding this dependency would mean that most transitions would be CSS powered from SUI while some would have a different API and depend on another source.

Maybe we can have a video call sometime soon-ish to discuss SUI / Stardust and future plans around motion and animation.  That is a big topic for us across the board.

levithomason:
Wow, these look really promising:

https://popmotion.io/pure/
https://popmotion.io/pose/

layershifter:
### Status

Needs to be re implemented with React Pose.

arozd:
Looking forward to get it fixed ;)

micahbule:
Any particular update on this one in context of #3190?

PR #3149, opened 2018-09-11
Make searchable dropdown accessible

arthurdenner:
**Current behaviour:**

A `Dropdown`/`Select` component with the `search` prop inside a Form is not accessible since the `id` prop is passed to the top level component instead of the `input` element.

**New behaviour:**

The `id` prop is passed to the `input` element.

It was added an example of an accessible searchable Dropdown.

If you want tests for this or have any changes to suggest, please let me know.

### DIFF
--- a/docs/src/examples/collections/Form/Shorthand/FormExampleFieldControlId.js
+++ b/docs/src/examples/collections/Form/Shorthand/FormExampleFieldControlId.js
@@ -1,5 +1,10 @@
 import React from 'react'
-import { Form, Input, TextArea, Button } from 'semantic-ui-react'
+import { Form, Input, TextArea, Button, Select } from 'semantic-ui-react'
+
+const genderOptions = [
+  { key: 'm', text: 'Male', value: 'male' },
+  { key: 'f', text: 'Female', value: 'female' },
+]
 
 const FormExampleFieldControlId = () => (
   <Form>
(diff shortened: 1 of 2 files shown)


layershifter:
@arthurdenner thanks for PR but I don't want to merge it because this behaviour can be reached with existing functionality.

The `searchInput` prop already does it:

```jsx
<Dropdown search searchInput={{ id: 'foo', type: 'number' }} options={options} />
```

And example is also there: http://react.semantic-ui.com/modules/dropdown/#usage-search-input

layershifter:
Possible, we should update `FormExampleFieldControlId.js` to show this functionality, this will be the best option.

arthurdenner:
Hm, sorry, I didn't notice this. However, inside a form, a searchable dropdown still not being accessible because the `id` value is passed to the top level `div` also, so clicking in the `label` doesn't work.

See this: https://codesandbox.io/s/x2lz8qwlrw

Maybe only ignoring the `id` from `getUnhandledProps` can do the trick?

layershifter:
It's a little unclear, but works:
https://codesandbox.io/s/48r8928v1x

----

```jsx
<Form.Field
  control={Select}
  options={genderOptions}
  label={{ children: "Gender", htmlFor: 'form-select-control-gender' }}
  placeholder="Gender"
  search
  searchInput={{ id: "form-select-control-gender" }}
/>
```

arthurdenner:
Yep, that's very unclear because is totally different from the `Form.Field` with `Input` as the `control` prop, for example. :confused: 

Thanks for the help!

arthurdenner:
Can I open a PR adding this as an example to the `FormExampleFieldControlId` though?

layershifter:
@arthurdenner yes, please! :+1:

Issue #3087, opened 2018-08-15
Accordion should support renderActiveOnly prop


loopmode:
### Problem description
Accordion renders all panels, regardless of whether they are expanded or collapsed (`active`).
This prevents usage of content components that do heavy-lifting or network requests in their `componentDidMount` - all of them are rendered right away, even if none of them is visible to the user.

### Proposed solution
There should be a `renderActiveOnly` prop, very much like [in the Tab component](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Tab/Tab.js#L68).
If its value is true, Accordion should not render content of inactive items.

Noteworthy: Having the default value be `false` would provide better legacy compatibility to the current behaviour, and allow release via minor version bump or patch, but it would create inconsistency as the default value in Tab is `true`.

brianespinosa:
@loopmode I like this. Definitely a good thing for us to add. I don't think there are technical limitations to this. I am marking this as an enhancement. PRs are welcome if anyone wants to submit anything.

loopmode:
Any opinions on to whether it should be default true?
I suggest to make it default false for now.

brianespinosa:
@loopmode I think for consistency we should do the same as the current use that you referenced where default is actually true. @levithomason @layershifter thoughts? Theoretically that might be a breaking change for someone which we should document as such, but I think we should keep the implementations consistent throughout the library.

levithomason:
>Theoretically that might be a breaking change for someone which we should document as such, but I think we should keep the implementations consistent throughout the library.

Agreed on both semver and default value  Please make the PR `BREAKING(Accordion): add renderActiveOnly prop` and we'll ship it correctly.

loopmode:
I did give it a try, but to be honest, I had a bumpy start and struggled with some aspects of the workflow and coding principles/paradigms of the project. However, this issue is not a good place to clarify my questions.
I'd appreciate if somebody experienced would find some time to guide me a little. I am user "loopmode" in https://gitter.im/Semantic-Org/Semantic-UI-React.

Meanwhile, if somebody feels like solving this, please go ahead. 
And otherwise, I'll give it another try next weekend.

anandrikka:
@loopmode  @SimonKinds is the issue still open?  I'm happy to help this will be my first step for open source contribution

skindstrom:
[...]

PR #3154, opened 2018-09-13
accordion initial PR

fabriceTOUSSAINT:
Opening PR for https://github.com/Semantic-Org/Semantic-UI-React/issues/3087

Updating Accordion logic to adhere to the same logic as Tabs with using `renderActiveOnly` prop for only rendering active AccordionPanels instead of adding a class to hide.


layershifter:
Closing if favor of #3190, which has much more progress.

PR #3157, opened 2018-09-14
docs(Form): add an example of accessible Form.Select

arthurdenner:
As discussed in #3149, this PR adds an example of how to accomplish an accessible select inside a form.


levithomason:

PR #3167, opened 2018-09-19
docs(Grid): fix typo in GridExampleStretched example description

sgarrity:
The description of the `stretched` property of a grid row mentioned 'width', but should have mentioned 'height'.


levithomason:

Issue #3175, opened 2018-09-27
Dropdown: invalid PropType noResultMessage


lazazivkovic:
### Steps
When using the `Dropdown` component with `selection` & `search` enabled, I want to be able to pass a React component as a `noResultMessage` to allow for links. This works as expected (it displays fine), but results in proptype warning in the console.

**NOTE:** - Same as #1664, only for other component (`Dropdown` instead of `Search`).

### Expected Result
No warning in console

### Actual Result
Failed prop type: Invalid prop `noResultsMessage` of type `object` supplied to `Dropdown`, expected `string`.

### Version
0.77.1

### Testcase
```
<Dropdown
    search
    selection
    options={[
        { value: 1, text: '1' },
        { value: 2, text: '2' },
        { value: 3, text: '3' },
    ]}
    noResultsMessage={(
        <div>
            <div>No results found</div>
            <Link onClick={addManually}>Add manually</Link>
        </div>
    )}
/>
```

yuritoledo:
@lazazivkovic why you dont use this `props`?

![image](https://user-images.githubusercontent.com/18474538/46445785-e8af9e00-c74e-11e8-98e8-1ae9dde72f9c.png)

[mode details here, you can use custom components too](http://react.semantic-ui.com/modules/dropdown/#usage-addition-label-component)

lazazivkovic:
Hi @yuritoledo 

Thanks for the suggestion, but that unfortunately does not solve my problem. I'm not trying to add an option to the existing ones, rather, I would like to add a link which would lead me to another form, where that data could be entered manually. So the _noResultsMessage_ is the actual prop I need, only it doesn't support React components as value at the moment. Actually, it works, but triggers a warning in the console. Thanks for the suggestion anyway. 😃

lazazivkovic:
Hi @layershifter, could you please take a look at #3177? Thanks 😃

levithomason:
Fixed in #3177

PR #3177, opened 2018-09-28
feat(Dropdown): noResultsMessage prop can be a node

lazazivkovic:
Relates to #3175


levithomason:
Awesome, thanks for the well-documented issue and PR!

Issue #3164, opened 2018-09-18
Missing component in example site


ghost:
### Steps
visit https://react.semantic-ui.com/views/advertisement/

### Expected Result
The Advertisement component

### Actual Result
Getting a "404, How about some good old Atari?". But the latest clone works on my local installation, so I wish I could help fix it, but I don't know how you are deploying.

### Version
i.d.k

### Testcase
[Fork, update, and replace this pen to show the bug]:
N/A

layershifter:
You need to disable Adblock.

ghost:
That was it, thanks.

layershifter:
@garajo I made #3179, it will update the example to omit problems with Adblock :+1:

PR #3179, opened 2018-09-29
 docs(Advertisement): update a base Advertisement example

layershifter:
### Why?

AdBlock blocks loading of `Advertisement` examples and docs will throw 404, see #3164. I've created a fully working example on Codesandbox (https://codesandbox.io/s/6yo47owp9n).


levithomason:

PR #3180, opened 2018-09-29
docs(Dropdown): update suggested markup for nested dropdowns

eragon512:
Update for the suggested markup in the Dropdown Examples for Nested Dropdowns

screen shot 2018-09-29 at 11 50 25 pm
screen shot 2018-09-29 at 11 50 30 pm

### DIFF
--- a/docs/src/examples/modules/Dropdown/Types/DropdownExamplePointing.js
+++ b/docs/src/examples/modules/Dropdown/Types/DropdownExamplePointing.js
@@ -7,22 +7,20 @@ const DropdownExamplePointing = () => (
     <Dropdown text='Shopping' pointing className='link item'>
       <Dropdown.Menu>
         <Dropdown.Header>Categories</Dropdown.Header>
-        <Dropdown.Item>
-          <Dropdown text='Clothing'>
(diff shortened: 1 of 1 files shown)


levithomason:
Thank you for the investigation and work.  The markup for the original code is actually correct.  The difference between a `Dropdown.Item` and a `<Dropdown item />` is that the first expresses the `item` "part" of the Dropdown while the second is a type of Dropdown component.

See the original markup here:
https://semantic-ui.com/modules/dropdown#pointing

Compare:
1. `class="item"` _(`<Dropdown.Item />`)_
1. `class="ui item dropdown"` _(`<Dropdown item />`)_

The issue with the example overall is that the React port doesn't support using Children to define the Dropdown items.  This is one of the main issues preventing our 1.0 release.  You can see more in the [1.0 Milestone](https://github.com/Semantic-Org/Semantic-UI-React/milestone/1).

You can contribute CSS fixes to https://github.com/Semantic-Org/Semantic-UI.  Once merged, we consume the changes here in the React library as well.

I will close this for now.  Let me know if I've misread anything here and we can reopen this PR.  Thanks again for your time and work!

PR #3183, opened 2018-09-30
fix(customPropTypes|typings): allow to pass a function to itemShorthand, update typings

### LINKED ISSUE #3181: Pagination: Functions work for pageItem, etc. props, but warning is raised for invalid prop
### Steps
```jsx
return <Pagination pageItem={someFunction} />
```

### Expected Result
Functionality-wise, I get exactly what I expected: the result of `someFunction` is displayed for page items; I do not expect any warnings.

### Actual Result
Despite working as expected, I also get a warning for an invalid prop, because the type for `pageItem` etc. do not include `function`.

### Version
0.82.3

### Testcase
https://codesandbox.io/s/0po3po6530

layershifter:
# POSSIBLE BREAKING CHANGE

`src/index.d.ts` was renamed to `src/generic.d.ts` I hope that noone used it from `dist/commonjs/index.d.ts` because all types are reexported from the root. However I think that this thing should be noticed.

## Fixes and improvements

1. In #2790 we introduced a small update to shorthands, however `customPropTypes.itemShorthand` wasn't updated. This PR does this. Fixes #3181.

2. Typings were improved, you will get working autocomplete there :+1: It also satisfies `noImplicitAny` :v: 

```tsx
  <Button
    content="Foo"
    icon={(Component, props) => (
      <div className="bar">
        <Component name={props.name} />
      </div>
    )}
  />
```

peek 2018-09-30 13-26

3. Some typings tests were added.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -369,7 +369,7 @@ export {
 } from './dist/commonjs/elements/Step/StepTitle'
 
 // Generics
-export * from './dist/commonjs'
+export * from './src/generic'
 
 // Modules
 export {
--- a/src/addons/Confirm/Confirm.d.ts
+++ b/src/addons/Confirm/Confirm.d.ts
@@ -1,6 +1,6 @@
 import * as React from 'react'
 
-import { SemanticShorthandItem } from '../..'
+import { SemanticShorthandItem } from '../../generic'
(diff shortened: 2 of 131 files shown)


levithomason:

PR #3184, opened 2018-09-30
fix(Grid): improve propTypes check for width props

### LINKED ISSUE #2869: Grid: Column "tablet" value applied to other sizes
## Bug Report

This may be a bug for SUI-LESS.

### Steps
```jsx
<Grid>
  <Grid.Column
    tablet={4}
    width={2}
  />
  <Grid.Column
    tablet={12}
    width={14}
  />
</Grid>
```

### Expected Result

I'm not sure if this is a bug or an unsupported mixing of properties (which is not documented), but it seems like a bug. Based on the documentation, I think bug:

>**`tablet`**
>A column can specify a width for a tablet device.

>**`width `**
>Represents width of column.

That says to me that the value from `tablet` applies only to tablet (it's basically an exception); the value for `width` applies to all sizes for which no exception has been specified.

### Actual Result
The values of `tablet` and `width` conflict: sometimes `tablet` wins and sometimes `width` wins. This is due to both having the same specificity (with no media query) and the last-wins rule.

### Version
`SUIR` `0.80.0`
`SUIL` `2.3.1`

### Testcase
https://codesandbox.io/s/nn1n03o32p

Notice that the `tablet` value is used for the first column and the `width` value is used for the second column (which causes the total to be greater than 16 aka 100%).

layershifter:
Fixes #2869.

### DIFF
--- a/docs/src/examples/collections/Grid/ResponsiveVariations/index.js
+++ b/docs/src/examples/collections/Grid/ResponsiveVariations/index.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { Icon, Message } from 'semantic-ui-react'
 
 import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
 import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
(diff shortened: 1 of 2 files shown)


levithomason:

PR #3185, opened 2018-09-30
docs(props): improve parsing of customPropTypes.every()

layershifter:
This PR improves parsing of `customPropTypes.every()` :v: 

### Before

image

image

### After

image
image


levithomason:

PR #3187, opened 2018-10-01
fix(Input): pass `title` down to an input element

### LINKED ISSUE #3186: Input: field should take title as props
## Bug Report
I'm not sure why the rest of props are sent to Form.Input is sent to the surrounding div, and not the input itself. I mean, the input should be what we're really dealing with.

### Steps
A clear and concise description of steps to reproduce the problem.
Here's an example:
```jsx
 <Form.Input
   type="text"
   maxLength={2}
   title="Month field must be a 2 digit number"
   pattern="[0-9]{2}"
   placeholder="Month..."
   name="month"
/>
```

### Expected Result
It should display the error message(i.e. title) if the user's input don't match the pattern

### Actual Result
The message didn't get displayed

### Version
0.80.0

### Testcase
https://codesandbox.io/s/q4zpo7l234

layershifter:
Fixes #3186.

### DIFF
--- a/src/lib/htmlPropsUtils.js
+++ b/src/lib/htmlPropsUtils.js
@@ -29,6 +29,7 @@ export const htmlInputAttrs = [
   'readOnly',
   'required',
   'step',
+  'title',
   'type',
   'value',
 ]


levithomason:

PR #3189, opened 2018-10-02
BREAKING(eventStack): fix issues with event subscriptions

### LINKED ISSUE #3006: Popup: when multiple popups present, click on the other trigger doesn't close original popup

layershifter:
Fixes: #3006.
Details: https://github.com/layershifter/event-stack/pull/4

# BREAKING CHANGES

This PR fixes existing issues with multiple Modals, Popups and Dropdown's. However, it changes the behaviour of the Sidebar component that was introduced in [v0.81.2](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/CHANGELOG.md#v0812-2018-06-26). This update is strongly recommend if you're using `0.81.0` or higher.

You also can control the behaviour of Sidebar with the `target` prop, #3191.

### Manual checks

- [x] Modal: Multiple, https://react.semantic-ui.com/modules/modal/#types-multiple
- [x] Dropdown: Close on blur, https://react.semantic-ui.com/modules/dropdown/#usage-close-on-blur
- [x] Popup: Clickable sequence, https://codesandbox.io/s/3rx96jzjr6

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleRenderSource.js
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleRenderSource.js
@@ -28,6 +28,7 @@ const externals = {
   faker,
   lodash: require('lodash'),
   react: React,
+  'prop-types': PropTypes,
   'semantic-ui-react': SUIR,
 }
 const commons = {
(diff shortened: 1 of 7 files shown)


levithomason:
>However, it changes the behaviour of the Sidebar component that was introduced in v0.81.2.

Can we get a before / after update in the description?  Thanks!

levithomason:

PR #3190, opened 2018-10-02
BREAKING(Accordion): add renderActiveOnly prop (#3087)

### LINKED ISSUE #3087: Accordion should support renderActiveOnly prop

skindstrom:
This PR fixes #3087. Namely it allows an Accordion to not render content that is not active, and sets not rendering inactive content to be the default behavior.

The solution differs from `Tab` due to the surrounding API. Namely that `Accordion` supports the `children` prop, requiring `renderActiveOnly` to possibly be passed all throughout `Accordion` -> `AccordionPane` -> `AccordionContent` when the user utilizes the `panes` prop. It's possible that the `Accordion` API should be changed to match the `Tab` API, where the `panes` prop accept a render function. However I felt as if this was out of scope for this issue/PR and should be discussed elsewhere.

Note that I had to set `active` to be a required prop for the `AccordionContent`. I could not figure out how to get it to pass the tests otherwise. Let me know if this is a rash decision.

Furthermore, I chose to render an empty `<div class="content" />` instead of nothing at all in `AccordionContent`. Rendering _something_ seems to be a requirement to be conformant, but let me know if you would've implemented it in another way.

### DIFF
--- a/src/modules/Accordion/AccordionAccordion.d.ts
+++ b/src/modules/Accordion/AccordionAccordion.d.ts
@@ -37,6 +37,9 @@ export interface StrictAccordionAccordionProps {
 
   /** Shorthand array of props for Accordion. */
   panels?: SemanticShorthandCollection<AccordionPanelProps>
+
+  /** An Accordion will only render content from active panels */
+  renderActiveOnly?: boolean
 }
 
 declare const AccordionAccordion: React.ComponentClass<AccordionAccordionProps>
--- a/src/modules/Accordion/AccordionAccordion.js
+++ b/src/modules/Accordion/AccordionAccordion.js
@@ -60,10 +60,14 @@ export default class AccordionAccordion extends Component {
         }),
       ),
     ]),
+
+    /** An Accordion will only render content from active panels */
+    renderActiveOnly: PropTypes.bool,
   }
 
   static defaultProps = {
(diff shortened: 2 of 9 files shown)


layershifter:
@SimonKinds awesome work 👍 I will take a timeout for few days, need to think about API there.

layershifter:
The delay is caused by #3143, I'm currently researching how we can integrate `react-pose` and combine transitions with this PR 🤔

skindstrom:
I appreciate the transparency! I'm sorry to say that transitions are way above my head, so I can't aid you with that one.

PR #3191, opened 2018-10-03
feat(Sidebar): add `target` prop

layershifter:
### Why?

The `target` prop allows to pass element to which will be attached event listener. This allows to customize closable area, defaults to `document`.


levithomason:

Issue #3192, opened 2018-10-03
Accordion: activeIndex must be array if exclusive = false


skindstrom:
### Steps
Give an `Accordion` the prop `activeIndex` (or `defaultActiveIndex`) as a number instead of an array when `exclusive` is false, e.g. 
```jsx
<Accordion defaultActiveIndex={0} exclusive={false} panes={...} />
```

### Expected Result
The given index to be active.

### Actual Result
1. No index is active, and therefore no `AccordionContent` is shown.
2. If a title is clicked, the app crashes (in Firefox Nightly)

### Version
0.82.5

### Testcase
https://codesandbox.io/s/jpj97jrr6w

### Additional information
Apparently this has been previously closed in #2440 as not an issue, as one is expected to give an array if `exclusive` is false. However there is no mention of this in the documentation.

I would argue that either the API should be updated (my preference, to keep a more consistent and easy-to-use API), or the docs should be updated. I'll submit a PR shortly with an API fix, and then you can decide.

PR #3193, opened 2018-10-03
fix(Accordion): always allow numeric activeIndex (#3192)

skindstrom:
Cleans up the API a bit for `Accordion`, always allowing a numeric `activeIndex` prop instead of requiring an array when `exclusive` is true.

More information in #3192

### DIFF
--- a/src/modules/Accordion/AccordionAccordion.js
+++ b/src/modules/Accordion/AccordionAccordion.js
@@ -73,13 +73,27 @@ export default class AccordionAccordion extends Component {
   }
 
   computeNewIndex = (index) => {
-    const { exclusive } = this.props
-    const { activeIndex } = this.state
+    if (this.props.exclusive) {
+      return this.computeNewIndexIfExclusive(index)
+    }
 
-    if (exclusive) return index === activeIndex ? -1 : index
+    return this.computeNewIndexIfNotExclusive(index)
+  }
 
-    // check to see if index is in array, and remove it, if not then add it
-    return _.includes(activeIndex, index) ? _.without(activeIndex, index) : [...activeIndex, index]
+  computeNewIndexIfExclusive = (index) => {
+    if (this.state.activeIndex === index) {
+      return -1
+    }
+
+    return index
+  }
+
+  computeNewIndexIfNotExclusive = (index) => {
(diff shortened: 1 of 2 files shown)


layershifter:
@SimonKinds thank you, contributions are always welcome :+1: 

However, my opinion that we don't need to these changes.

* we have other places were is similar API, i.e. Dropdown
* this change will make inconsistent `onChange` callback, you pass a number as value, but you will receive an array. It's will be more confusing than current API :cat: 

Possibly, the better solution will be to add example of controlled Accordion with `exclusive=false` and some info messages to clarify the current API.

Please let me know what you think.

/cc @levithomason @brianespinosa

skindstrom:
I hadn't considered the impact on the `onChange` callback, nice catch. With that in mind, I do agree that maybe it's better to keep the current behavior, but with a clear error message as well as well as an example of the behavior. 

I gave the repository a quick look regarding how error messages are presented, and I feel as if having something similar to whatever is used in `Dropdown` is a good way of doing it for the `Accordion` as well.
```js
    /* eslint-disable no-console */
      // in development, validate value type matches dropdown type
      const isNextValueArray = Array.isArray(nextProps.value)
      const hasValue = _.has(nextProps, 'value')

      if (hasValue && nextProps.multiple && !isNextValueArray) {
        console.error(
          'Dropdown `value` must be an array when `multiple` is set.' +
            ` Received type: \`${Object.prototype.toString.call(nextProps.value)}\`.`,
        )
      } else if (hasValue && !nextProps.multiple && isNextValueArray) {
        console.error(
          'Dropdown `value` must not be an array when `multiple` is not set.' +
            ' Either set `multiple={true}` or use a string or number value.',
        )
      }
    }
```

Feel free to close this PR if you see fit, and I can send another one with the updated documentation etc. if you believe that's the better way forward.

layershifter:
> I gave the repository a quick look regarding how error messages are presented, and I feel as if having something similar to whatever is used in Dropdown is a good way of doing it for the Accordion as well.

Nice catch :+1: This will be very clear.

> and I can send another one with the updated documentation etc.

Yes, sir, please :cat: I will close this PR and will wait for the new :+1:

PR #3199, opened 2018-10-07
fix(Accordion): log error if incorrect activeIndex type

### LINKED ISSUE #3192: Accordion: activeIndex must be array if exclusive = false

skindstrom:
As discussed in PR #3193, this fix ensures that an Accordion will log an error if the incorrect type is given for `activeIndex`.

The expected types are:
- `number` if `exclusive`
- `array` if not `exclusive`

It will not log in production.

At first I considered doing the logging in the same lifecycle method that is used in `Dropdown`, however it uses the now deprecated `componentWillReceiveProps`. That's why I chose to log it in `componentDidUpdate`

In #3193 we also discussed further documentation, but I noticed that there is already an example of how to use an Accordion when `exclusive` is false.

I chose to not fix the crash that might occur when clicking a panel if the incorrect type for `activeIndex` is given, as the warning will be printed beforehand. If you feel this should be fixed, let me know.

You may go ahead and close #3192 , as this logging fixes the issue I had with it.


skindstrom:
No worries, thanks for being a wonderful maintainer!

PR #3204, opened 2018-10-09
fix(typings): fix definition for SemanticShorthandItemFunc

layershifter:
As Levi mentioned in https://github.com/stardust-ui/react/pull/328#discussion_r223225907, typings should be correct. This stuff wasn't released yet, so there are no breaking changes.


levithomason:

PR #3209, opened 2018-10-11
docs(LabelExamplePointingColored): Remove basic property from labels

ahacop:
Prior to this commit, label text was illegible as both the text and label background color were red.

Before:
screen shot 2018-10-11 at 15 17 29
After:
screen shot 2018-10-11 at 15 17 55


layershifter:
https://semantic-ui.com/elements/label.html#pointing

https://github.com/Semantic-Org/Semantic-UI/blob/master/RELEASE-NOTES.md
Semantic-Org/Semantic-UI#6440

This was fixed in SUI 2.4 and already fixed in `master` branch. The next release of SUIR will also add new features from SUI 2.4, however feel free to use the latest version of SUI, it's fully compatible.

PR #3213, opened 2018-10-12
docs(Search): fix resolver imports

layershifter:
Adds missing imports for Seach examples.


levithomason:

PR #3215, opened 2018-10-12
docs(Ads): fix multiple ads

layershifter:
## Issue source

There was a strange tweet about issue with ads, https://twitter.com/ARedHerring_/status/1041044450288590848

I haven't ever seen this, looks really ugly :-1: I tried to find the issue and it seems that it comes from the Carbon's code, it adds multiple adds sometimes.

image

### DIFF
--- a/docs/public/style.css
+++ b/docs/public/style.css
@@ -80,7 +80,7 @@ code:not(.hljs)::after {
 }
 
 #carbonads {
-  display: block;
+  line-height: 1.4;
 }
 
 #carbonads span {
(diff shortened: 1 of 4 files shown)


layershifter:
@levithomason can you please take a look? Possible there is a solution without dirty tricks?

levithomason:
Thanks, this was also reported to me via email.  The refresh method removes the ad then creates a new one.  If there is no ad in the div, it will insert one.  What is happening is that `componentWillUpdate` is called in rapid succession calling refresh multiple times rapidly.  This removes and replaces the ad on the first call, but before it is loaded the second refresh is also called which fetches a second ad.

I will take a look at the changes here and confirm the fix or update it.

levithomason:
- Added check to ensure we don't duplicate fetch ads.
- Made it smaller and better styled, was taking up a lot of real estate.
- Improved sidebar ux and layout to help with real estate.

levithomason:

PR #3219, opened 2018-10-15
chore(componentInfo): omit context

layershifter:
This PR:
- removes `componentInfoContext` that uses `require.context()`, it will be replaced with generated index
- picks ideas with POSIX paths and `gulp-remember` from Stardust


levithomason:
Heads up, we're planning to share the doc experience between Semantic UI, Stardust UI, Fabric, and some internal doc systems at various companies.

As we iterate on these kinds of features, let's keep in mind both TypeScript and JavaScript support as well as generalizing solutions to support plug n' play scenarios.

layershifter:
I will left this for Stardust. The plugin will play nicely with TS, too.

gulp/plugins/gulp-component-info.js
```diff
@@ -0,0 +1,36 @@
...
+        cb(null, file)
+        return
+      }
+
+      if (file.isStream()) {
+        cb(new gutil.PluginError(pluginName, 'Streaming is not supported'))
+        return
+      }
+
+      files[file.basename] = JSON.parse(file.contents)
```
levithomason, line 24: I think we should generate separate files for each component.  This will help us with code splitting.  It also helps for navigating around the generated code.

See how Stardust creates JSON files for various concerns, per component, in the `docs/src` dir after running `yarn build:docs`.

gulp/plugins/gulp-example-source.js
```diff
@@ -4,51 +4,38 @@ import path from 'path'
...
+        cb(new gutil.PluginError(pluginName, 'Streaming is not supported'))
+        return
+      }
+
+      const sourceName = _.split(file.path, path.sep)
         .slice(-4)
         .join('/')
         .slice(0, -3)
 
       exampleSources[sourceName] = file.contents.toString()
```
levithomason, line 29: Also here, separate files would be great.

levithomason:
>I will left this for Stardust. The plugin will play nicely with TS, too.

Perhaps we should add a `stardust` label to identify which issues / PRs we're handling there?

gulp/plugins/gulp-component-info.js
layershifter: A little confusing, but this plugin just generates `index.js` in `docs/src/componentInfo` which will really help us with code splitting :) It replaces `require.context()` with valid imports.

```js
import Confirm from "./Confirm.info.json";
import MountNode from "./MountNode.info.json";
import Pagination from "./Pagination.info.json";
import PaginationItem from "./PaginationItem.info.json";

// etc...
```

gulp/plugins/gulp-example-source.js
layershifter: Yep, for next iteration of this plugin it will be great.

PR #3221, opened 2018-10-15
feat(Tab): Add support for Tab.Pane as child to Tab

fcwheat:
Add support for passing Tab.Pane as a child node to Tab. Prop `menuItem` on Tab.Pane is used to describe clickable header that switches tabs.

Simple example that I want to enable:
https://codesandbox.io/embed/op4wk2nyy

Additionally I added a small fix for using the `render` prop when specifying `renderActiveOnly=false`, a problem which is demonstrated here:
https://codesandbox.io/s/9o9ynlvvyp

I feel like having both `pane` and `render` as shorthand options seems a bit redundant, but that can be a separate discussion. For now, I wanted to update the Tab component to make the usability similar to `Menu` and `Menu.Item` when `Menu.Item` is perfectly fine as a child component.

### DIFF
--- a/src/modules/Tab/Tab.js
+++ b/src/modules/Tab/Tab.js
@@ -4,6 +4,7 @@ import React from 'react'
 
 import {
   AutoControlledComponent as Component,
+  childrenUtils,
   customPropTypes,
   getElementType,
   getUnhandledProps,
@@ -23,11 +24,17 @@ class Tab extends Component {
     /** An element type to render as (string or function). */
     as: customPropTypes.as,
 
+    /** Index of the currently active tab. */
+    activeIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+
+    /** Primary content. */
+    children: PropTypes.node,
+
     /** The initial activeIndex. */
     defaultActiveIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
 
-    /** Index of the currently active tab. */
-    activeIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    /** Shorthand props for the Grid. */
+    grid: PropTypes.object,
 
     /**
      * Shorthand props for the Menu.
@@ -37,9 +44,6 @@ class Tab extends Component {
 
     /** Align vertical menu */
     menuPosition: PropTypes.oneOf(['left', 'right']),
-
-    /** Shorthand props for the Grid. */
-    grid: PropTypes.object,
 
     /**
      * Called on tab change.
@@ -51,7 +55,7 @@ class Tab extends Component {
     onTabChange: PropTypes.func,
 
     /**
-     * Array of objects describing each Menu.Item and Tab.Pane:
+     * Shorthand array of objects describing each Menu.Item and Tab.Pane:
      * { menuItem: 'Home', render: () => <Tab.Pane /> }
      * or
      * { menuItem: 'Home', pane: 'Welcome' }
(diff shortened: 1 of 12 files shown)


src/modules/Tab/Tab.js
```diff
@@ -88,31 +92,52 @@ class Tab extends Component {
...
+
     const { activeIndex } = this.state
 
     if (menu.tabular === true && menuPosition === 'right') {
       menu.tabular = 'right'
     }
 
+    const items = childrenUtils.isNil(children)
+      ? _.map(panes, 'menuItem')
+      : _.map(children, child => child.props.menuItem)
```
layershifter: Sorry, but this change cannot be merged because it violates our API.

The parent component should never know anything about the child component. This is one of reasons for the shorthand API existance.

layershifter (changes requested):
I will additionally review this PR. This PR can't be merged as is

However, we need to do some breaking changes there for #2168 and it can be a great opportunity to do them.

src/modules/Tab/Tab.js
fcwheat: I understand. Can I take another crack at this where the `menuItem` isn't pulled from the child? Something like this perhaps:
```
<Tab menuItems={['Tab 1', 'Tab 2']}>
    <Tab.Pane>Tab 1 Content</Tab.Pane>
    <Tab.Pane>Tab 2 Content</Tab.Pane>
</Tab>
```
Where `menuItems` is an array of strings, nodes, or menu shorthands.

layershifter: I'm not sure that you're actually need this.  These can be done with `Menu` component.

```jsx
    <Menu attached tabular>
      <Menu.Item active>Tab 1</Menu.Item>
      <Menu.Item>Tab 2</Menu.Item>
      <Menu.Item>Tab 3</Menu.Item>
    </Menu>

    <Tab.Pane active>Tab 1 Content</Tab.Pane>
    <Tab.Pane active={false}>Tab 2 Content</Tab.Pane>
```

https://codesandbox.io/s/66z9j2l3k

fcwheat: I see what you are saying, but your example would require the user to implement the state that manages the active panes and such. My goal was to not have the user have to worry about that and still have `Tab.Pane` be a valid child of `Tab`, similar to how other components work.

fcwheat:
I updated my PR to remove `menuItem` from `Tab.Pane` and instead have a `menuItems` prop on `Tab` instead. @layershifter if you feel like this isn't how you want this component to work then that's okay, I can close this PR. I was just trying to add functionality that other developers may benefit from.

src/modules/Tab/Tab.js
```diff
@@ -88,31 +102,50 @@ class Tab extends Component {
...
-    if (renderActiveOnly) return _.invoke(_.get(panes, `[${activeIndex}]`), 'render', this.props)
-    return _.map(panes, ({ pane }, index) =>
-      TabPane.create(pane, {
+    const isMissingChildren = childrenUtils.isNil(children)
+
+    const items = isMissingChildren ? panes : children
+
+    if (renderActiveOnly) {
+      if (!isMissingChildren) {
+        return _.get(items, `[${activeIndex}]`)
```
levithomason, line 114: Even though we don't support reading data from children, I thought I would note that React `children` is not always an array.  You would want to use the `React.Children` API for handling children in a more consistent way.

levithomason:
The reason we do not support iterating children in most cases is because it is an anit-pattern to "read" data from children in React.  Consider a few problem cases as well:

### Wrapping Components
```jsx
<Tab>
  <Tab.Pane />    // works

  <Transition>    // oops....
    <Tab.Pane />
  </Transition>
</Tab>
```

Even though the `Transition` component does not render anything to the DOM, it still breaks the Tab because the Tab doesn't know how to dive through children to find the right component.  Attempting to do so is also problematic.  Same can be said for other wrapping components, like `Ref`.

### HOCs

This is made more clear by the case of HOCs:

```jsx
const StyledPane = withStyle(Tab.Pane)

<Tab>
  <Tab.Pane />    // works
  <StyledPane />  // oops...
</Tab>
```

### State

The difficulty with the children API is that we cannot apply state to the children or read data back from the children safely.  The reason we use shorthand is so that we can keep a data representation of the children which we can safely iterate and read from.  At the time we render the data to child elements, we can also apply any additional props or state to the children.  We cannot do this safely if we are parsing JSX child elements (see above).

Why can the Menu get away with this?  If you have a look at the Menu and MenuItem code, or the Menu doc examples, you'll notice we don't handle state when you use children.  This is the trade off for the user.  If you use shorthand, we can handle state for you.  If you use children, you need to also handle events and state for those children.

Since the Tab is strictly a state manager for a Menu and Segment, it cannot support children.  Otherwise, as @layershifter pointed out, you would just use the Menu and Segment directly and manage the state yourself.

***

Hopefully this explanation was helpful!  I see you've submitted a number of useful PRs so I thought it would be very worth while to give some details behind our thinking and patterns.

Speaking of, would you like to be a member here?  We're always looking to add people who are talented and are making considerable investments into SUIR.  I think you fit this category well 😄

layershifter:
@levithomason thanks for the detailed response 👍 Closing for housekeeping.

PR #3223, opened 2018-10-16
docs(ComponentDoc): Fix clicking sidebar item doesn't make it active bug

fcwheat:
Fix a bug within the documentation site where when a sidebar item is selected that is below the current selection or there is no current selection, the item above the selected will be active not the selected item.

Steps to reproduce issue:

- Go here: http://react.semantic-ui.com/elements/button/
- Expand the `Types` section on the sidebar on the right
- Click `Expand` right below `Button`

Expected Behavior: `Expand` will be emphasized as active within the sidebar
Current Behavior: `Button`, above `Expand` is emphasized as active

This stems from the `onTopPassed` and `onTopPassedReversed` behavior on the `Visibility` component that wraps the example. These features work fine when the user is doing the scrolling, but not as well when the scrolling is being done programmatically, ie. `scrollToAnchor()`.

My solution is simply to set a flag when the sidebar item is selected and `scrollToAnchor()` is called such that the `onPassed` events don't fire for a short time while the animation is occurring.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentDoc.js
+++ b/docs/src/components/ComponentDoc/ComponentDoc.js
@@ -33,7 +33,9 @@ class ComponentDoc extends Component {
     sidebarSections: docTypes.sidebarSections.isRequired,
   }
 
-  state = {}
+  state = {
+    autoScrollingToAnchor: false,
+  }
 
   componentWillMount() {
     const { exampleKeys, history } = this.props
@@ -58,7 +60,9 @@ class ComponentDoc extends Component {
   }
 
   handleExamplePassed = (e, { examplePath }) => {
-    this.setState({ activePath: examplePathToHash(examplePath) })
+    if (!this.state.autoScrollingToAnchor) {
+      this.setState({ activePath: examplePathToHash(examplePath) })
+    }
   }
 
   handleExamplesRef = examplesRef => this.setState({ examplesRef })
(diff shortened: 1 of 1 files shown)


levithomason:
First, thanks for diving in on this!  It has been bothering me for a while now 😅 

Let's avoid use of timeouts as they introduce hidden implicit dependencies between scrollToAnchor.js and ComponentDoc.js.  If one of those were to change, scroll would be broken again.  Let's find a fix that doesn't introduce coupling.

I have a feeling that this could be due to an off-by-one issue.  The scroll to anchor scrolls precisely to the element, but doesn't actually "pass" it.  I'm betting the scrollToAnchor logic for stopping scroll on arrival at the element and the Visibility logic for calling `onTopPassed` are 1px off from each other.

levithomason:
@fcwheat That sounds very reasonable to me on the surface.  Do you want to give a shot and see if it works out?

You'd add the propType and typings, add a doc site example for each prop, and write a test for each prop.  You can copy the patterns that exist there for the other props as well.

It is worth noting that I'm not sure if this will fix the issue or not.  You could try the theory by playing the calculations in `scrollToAnchor` to make it scroll a few pixels passed the anchor and see if that resolves the problem.  If so, then we know it is an off by one error we need to address here.

In either case, the props you mention sound pretty useful.

PR #3227, opened 2018-10-18
docs(Layouts): fix examples with incorrect props

AlanFoster:
Hey, I noticed that some of the docs were using invalid React props. I've got ahead and fixed the issues I could spot within `docs/layouts`, and if I notice any other examples being wrong in the future I'll update them too :+1:


layershifter:
@AlanFoster thank you, great job :+1:

levithomason:

PR #3230, opened 2018-10-20
fix(Input): fix typings for onChange

### LINKED ISSUE #3228: How to add associated label to semantic ui dropdown
I am using Semantic-UI with plan javascript while testing site auditing i am getting error input doesn't have associated label. On checking i found this is injected dynamic so how can i add associated label to dropdown menu.

### LINKED ISSUE #3229: Input: has incorrect typings for `onChange`
Hi, 

0.82.5 comes with a major breaking change that makes all of our typescript codebases error out on event handlers. Previously you had event handler on form elements correctly typed, now there appears only a SyntheticEvent.

For example, this:

```ts
export interface InputProps { 
...
onChange?: (event: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => void
}
```

changed to this:

```ts
export interface InputProps { 
...
onChange?: (event: React.SyntheticEvent<HTMLInputElement>, data: InputOnChangeData) => void
}
```

Is this a bug or a feature? Shall we start changing our code bases? I hope we can bring the old functionality back as getting type safe handlers just become very difficult.

layershifter:
Fixes #3229.

### DIFF
--- a/src/elements/Input/Input.d.ts
+++ b/src/elements/Input/Input.d.ts
@@ -59,10 +59,10 @@ export interface StrictInputProps {
   /**
    * Called on change.
    *
-   * @param {SyntheticEvent} event - React's original SyntheticEvent.
-   * @param {object} data - All props and proposed value.
+   * @param {ChangeEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and a proposed value.
    */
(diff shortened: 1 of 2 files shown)


levithomason:

PR #3235, opened 2018-10-22
fix(Layouts): fix typo in StickyLayout

xiaohanyu:
Make it compatible with https://semantic-ui.com/examples/sticky.html#


xiaohanyu:
image

https://semantic-ui.com/examples/sticky.html#

vs 

image

http://react.semantic-ui.com/layouts/sticky

PR #3237, opened 2018-10-23
docs(Modal|Confirm): sort `size` prop in sizes order

fcwheat:
Simple fix to correct the size property enumeration in the documentation for Modal and Confirm components. Previously the ordering was not uniform (smallest -> largest or largest -> smallest).


layershifter:
> More broadly I am curious why the Modal component doesn't implement the onApprove and onDeny behavior that the Modal element has in the core library

Because there are no real ability to implement this feature for children API without over complications (i.e. Context API) while Shorthand API allows to define callbacks with `actions`.

fcwheat:
Gotcha, thanks for the feedback.

layershifter:
@fcwheat thank you 👍

PR #3243, opened 2018-10-27
TextArea must be wrapped with From

Puspendert:
It is nowhere written in the docs that a `TextArea` must be wrapped with `Form`. Though it is required.
https://react.semantic-ui.com/addons/text-area/


layershifter:
Oops, it seems that your branch is outdated. Can you please clean up it?

layershifter:
I did changes in docs in #3269 👍

Issue #2732, opened 2018-04-22
Prepare for upcoming react deprecations introduced in 16.3


justinobney:
Suggestion: begin migration to [new react lifecycles](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) while retaining backwards compatibility via `react-lifecycles-compat`

Example: https://www.youtube.com/watch?v=G9S1IghlkCI

### Steps

* Use new lifecycle methods introduced in react 16.3 in a component that renders `react-semantic-ui` components.

### Expected Result

* React does not log errors/warning about upcoming deprecations to the console.

### Actual Result

* React IS logging errors/warning about upcoming deprecations to the console.

### Version
0.79.1

### Testcase
[Fork, update, and replace this pen to show the bug]:
https://codesandbox.io/s/2l3n74j9y

levithomason:
@justinobney this is great, could you start a PR for this?

joshuaobrien:
Is this up for grabs? I'm happy to do it.

justinobney:
@joshuaobrien that would be great. I doubt I could get to it for a few weeks.

willb335:
@joshuaobrien if you need any help let me know.

joshuaobrien:
I'm on it. Will get started on the weekend :)

msrikanth508:
I would like to start on this, Couple of life cycle methods are marked as UNSAFE in 16.3 and two new methods getDerivedStateFromProps, getSnapshotBeforeUpdate got introduced. @levithomason what exactly do we need to do?

joshuaobrien:
Hi @msrikanth508, I am working on this at the moment. What we need to do is refactor the components that use the deprecated methods so that they no longer use them. Functionality that was in these deprecated lifecycle methods needs to be moved into other lifecycle methods. 

My experiences so far is that these changes are not always clean cut. Also, changes to tests are likely required to accommodate the new methods.

gavinmeier25:
Is this still needed I would be happy to hop on it this upcoming weekend and take a gander?

layershifter:
Yep, feel free to pickup 👍

tri-hoang:
I would love to work on this if it's still needed?

joshuaobrien:
It's still needed. I haven't got around to it. There is a PR here: https://github.com/Semantic-Org/Semantic-UI-React/pull/2757

kaiz123:
[...]

PR #3245, opened 2018-10-28
chore(AutoControlledComponent): Remove unsafe lifecycles

kohakukun:
Remove unsafe lifecycles from `AutoControlledComponent` and affected
components:
- `componentWillMount`
- `componentWillRecieveProps`
- `componentWillUpdate`

Relates to #2732

### DIFF
--- a/src/lib/AutoControlledComponent.js
+++ b/src/lib/AutoControlledComponent.js
@@ -86,11 +86,15 @@ export default class AutoControlledComponent extends Component {
         const defaultProp = getDefaultPropName(prop)
         // regular prop
         if (!_.has(propTypes, defaultProp)) {
-          console.error(`${name} is missing "${defaultProp}" propTypes validation for auto controlled prop "${prop}".`)
+          console.error(
+            `${name} is missing "${defaultProp}" propTypes validation for auto controlled prop "${prop}".`,
+          )
         }
         // its default prop
         if (!_.has(propTypes, prop)) {
-          console.error(`${name} is missing propTypes validation for auto controlled prop "${prop}".`)
+          console.error(
+            `${name} is missing propTypes validation for auto controlled prop "${prop}".`,
+          )
         }
       })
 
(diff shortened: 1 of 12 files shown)


kohakukun:
Many changes where introduced by the formater in defined in the hooks. Sorry for the extra work.

This PR cannot be breaked further as inherited components are affected

src/lib/AutoControlledComponent.js
```diff
@@ -185,11 +197,13 @@ export default class AutoControlledComponent extends Component {
...
-          'State will not be set.',
-          'Only props in static autoControlledProps will be set on state.',
-        ].join(' '))
+        console.error(
+          [
+            `${name} called trySetState() with controlled props: "${illegalKeys}".`,
+            'State will not be set.',
+            'Only props in static autoControlledProps will be set on state.',
+          ].join(' '),
+        )
```
layershifter, line 206: We cannot simply rename `componentWillReceiveProps()` to `componentDidUpdate()` there because we will have an additional `render()` after `setState()`, just compare

```
componentWillReceiveProps() 
render() 
// vs
render() 
componentDidUpdate 
render() 
componentDidUpdate 
```

Example: https://codesandbox.io/s/4x35y87rl4

---

The right way is to use there [`getDerivedStateFromProps()`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops). This will be more complicated refactor :(

layershifter (changes requested):
I've left a comment about the changes, we need to make more work there.

Thank you for the start 👍 

----

However, we can safely make changes in `MountNode` for example: https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.83.0/src/addons/MountNode/MountNode.js#L38

kohakukun:
Sure, I'll try to work on define the same logic with `getDerivedStateFromProps`.

> However, we can safely make changes in MountNode for example: https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.83.0/src/addons/MountNode/MountNode.js#L38
What do you mean by that?

layershifter:
> What do you mean by that?

We can update `MountNode` component and replace there `componentWillMount` with `componentDidMount`.

layershifter:
Hey @kohakukun 

I recently merged #3303, it has a very cool simplification of `componentWillReceiveProps()`. We can migrate to `getDerivedStateFromProps()` without any problems. Are you still interested in this PR?

kohakukun:
Sorry for the slow reply

PR #3257, opened 2018-11-05
fix(Portal): hover behavior fixed in Portal's event handlers

### LINKED ISSUE #3239: Popup: hover behaviour unreliable
### Steps
In the Semantic UI & Semantic UI React examples, go to Popup and try the Flowing Example
- You should find that on SUI the Popup on hover closes only when your mouse leaves the bounds of the popup content, but on SUIR the popup content closes unpredictably, even when your mouse is within the popup content bounds.
- I am finding cases in my website where buttons and other interactable components are not able to be interacted with reliably when wrapped in a Popup.

SUI: https://semantic-ui.com/modules/popup.html
SUIR: https://react.semantic-ui.com/modules/popup/

### Expected Result
The Popup hover behaviour on SUIR should match SUI.

### Actual Result
The SUIR Popup on hover mode unreliably closes, even when the mouse is within the bounds of the popup content.

### Version
0.83.0

Fabianopb:
Originally `Portal` closes on `mouseleave`. However, `mouseleave` is fired when "the pointer has exited the element and all of its descendants" (ref: https://developer.mozilla.org/en-US/docs/Web/Events/mouseleave).

For that reason the `Popup` was closing when the pointer exited the inner elements, but not necessarily the `Portal` itself (which is used to render the `Popup`).

Gif of the fixed behavior:
popup-onmouseleave-fixed

The solution was to cancel the `handlePortalMouseLeave` event in case the event's target doesn't match the portal node. A test was also added. It seemed a clean enough solution for this bug, let me know what you think!

Cheers!
(fixes #3239)

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -173,6 +173,9 @@ class Portal extends Component {
 
     if (!closeOnPortalMouseLeave) return
 
+    // Do not close the portal when 'mouseleave' is triggered by children
+    if (e.target !== this.portalNode) return
+
     debug('handlePortalMouseLeave()')
     this.mouseLeaveTimer = this.closeWithTimeout(e, mouseLeaveDelay)
   }
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -14,7 +14,6 @@ import PortalInner from './PortalInner'
 
 const debug = makeDebugger('portal')
 
-
 /**
  * A component that allows you to render children outside their parent.
  * @see Modal
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -14,6 +14,7 @@ import PortalInner from './PortalInner'
 
 const debug = makeDebugger('portal')
 
+
 /**
  * A component that allows you to render children outside their parent.
  * @see Modal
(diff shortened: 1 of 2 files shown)


layershifter:
@Fabianopb thank you for the investigation, fix looks correctly to me. I will test it on weekend.

levithomason:
Fantastic, thanks!  I've updated the branch to the latest and approved it for merge.  We can merge once tests pass.

layershifter:
@levithomason Codecov will never pass if you made "Update branch" on Github:

image
https://circleci.com/gh/Semantic-Org/Semantic-UI-React/6167

wangjitai:
Hi there, Is there any idea to resolve please?

should I update the Semantic ui react version? Thanks!

Fabianopb:
@wangjitai the team has been [cutting releases very often](https://github.com/Semantic-Org/Semantic-UI-React/releases) and last one was 19 days ago, I'd say it shouldn't take too long for the next one.

avinashjoshi:
Exactly what i was looking for - Hopefully the new version releases soon!

flq:
Any idea when this will go to npm?

ekeren:
Hey, Any idea when this will be released as part as 0.84?

Issue #3255, opened 2018-11-02
Error during Jest unit test run after upgrading to react 16.6.0: Invariant Violation: Unable to find node on an unmounted component


laplacesdemon:
### Steps
I have upgraded to react 16.6.0 from 16.5.0. My unit tests were working fine on `16.5.0` but after the upgrade I received the following error caused by the component called `Ref` in `Popup` (see the entire stack trace below).

The code under test looks like this:
```
const reason = '...';
const outcomeView = (<Label size={size} >{text}</Label>);
return <Popup trigger={outcomeView} content={reason} />;
```

And the test code is like this:
```
const component = renderer.create(<OperationDetail {...data} />);
    const tree = component.toJSON();
    expect(tree).toMatchSnapshot();
```

I just included the relevant parts.

### Expected Result
Tests run as normal.

### Actual Result
Below error happened during the tset run.

### Version
0.83.0

### Testcase
[Fork, update, and replace this pen to show the bug]:
https://codesandbox.io/s/2l3n74j9y

```
 FAIL  src/components/shared/buildSummaries/InstallContentTask/OperationDetail.spec.jsx
  ● <OperationDetail /> › should display data of failure

    Invariant Violation: Unable to find node on an unmounted component.

      at invariant (../node_modules/react-dom/cjs/react-dom.development.js:55:15)
      at findCurrentFiberUsingSlowPath (../node_modules/react-dom/cjs/react-dom.development.js:4161:30)
      at findCurrentHostFiber (../node_modules/react-dom/cjs/react-dom.development.js:4269:23)
      at findHostInstanceWithWarning (../node_modules/react-dom/cjs/react-dom.development.js:18296:21)
      at findDOMNode (../node_modules/react-dom/cjs/react-dom.development.js:18774:14)
      at Ref.componentDidMount (../node_modules/semantic-ui-react/dist/commonjs/addons/Ref/Ref.js:49:56)
      at commitLifeCycles (../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:6955:22)
      at commitAllLifeCycles (../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:8164:7)
      at HTMLUnknownElement.callCallback (../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2264:14)
      at invokeEventListeners (../node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:219:27)
[...]

PR #3258, opened 2018-11-05
fix(Ref) Add a workaround for findDomNode to work with React 16.6.0

laplacesdemon:
This is a workaround for the following issue: https://github.com/Semantic-Org/Semantic-UI-React/issues/3255

React 16.6.0 throws an error if `findDOMNode` is called on an unmounted component which is the case in `enzyme`'s `shallow` component testing. 

Please consider this change as a temporary workaround until there's a better way to handle `findDOMNode`. React states that it's an antipattern to use it in `componentDidMount`.

### DIFF
--- a/src/addons/Ref/Ref.js
+++ b/src/addons/Ref/Ref.js
@@ -24,8 +24,11 @@ export default class Ref extends Component {
 
     // Heads up! Don't move this condition, it's a short circuit that avoids run of `findDOMNode`
     // if `innerRef` isn't passed
-    // eslint-disable-next-line react/no-find-dom-node
-    if (innerRef) innerRef(findDOMNode(this))
+    try {
+      // eslint-disable-next-line react/no-find-dom-node
+      if (innerRef) innerRef(findDOMNode(this))
+      // eslint-disable-next-line no-empty
+    } catch (_) {}
   }
 
   render() {


src/addons/Ref/Ref.js
```diff
@@ -24,8 +24,11 @@ export default class Ref extends Component {
 
     // Heads up! Don't move this condition, it's a short circuit that avoids run of `findDOMNode`
     // if `innerRef` isn't passed
-    // eslint-disable-next-line react/no-find-dom-node
-    if (innerRef) innerRef(findDOMNode(this))
+    try {
+      // eslint-disable-next-line react/no-find-dom-node
+      if (innerRef) innerRef(findDOMNode(this))
+      // eslint-disable-next-line no-empty
+    } catch (_) {}
```
laplacesdemon, line 31: Silencing the error is horrible and I'm really not happy with this line. There are 2 options: either report it with `console.error` which will pollute the unit test runs, or only do a try catch in development environment (i.e. check for `process.env.NODE_ENV === 'development`). What do you think?

layershifter:
I almost sure that there is a different issue, need to check it. If you will check our [`package.json`](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/package.json) you will see that we're using React 16.6:
```json
"react": "^16.6.0",
"react-dom": "^16.6.0",
```

And our test suite passed for it.

laplacesdemon:
@layershifter you won't hit this issue if you don't do snapshot testing via `react-test-renderer`. If your unit tests don't have that, then it's OK to use React 16.6.

src/addons/Ref/Ref.js
levithomason: @layershifter Given your updates in Stardust, what's your thought here?

layershifter: I will backport changes from https://github.com/stardust-ui/react/pull/491, this will resolve this issue fully 👍

laplacesdemon: Thanks for addressing this issue.

PR #3260, opened 2018-11-06
fix(Comment): spread image props on CommentAvatar

### LINKED ISSUE #3250: Comment: CommentAvatar has no alt field
### Problem description

CommentAvatar has not way to setup any `alt` field, much important for any accessibility concerns

### Proposed solution

Provide a way to have an `alt`. Cheers!

layershifter:
Fixes #3250.

---

### Before
```html
<div alt='Example' class='avatar'>
  <img src='matt.jpg' />
</div>
```

### After

```html
<div class='avatar'>
  <img alt='Example' src='matt.jpg' />
</div>
```

### DIFF
--- a/src/elements/Image/Image.js
+++ b/src/elements/Image/Image.js
@@ -9,6 +9,7 @@ import {
   customPropTypes,
   getElementType,
   getUnhandledProps,
+  htmlImageProps,
   partitionHTMLProps,
   SUI,
   useKeyOnly,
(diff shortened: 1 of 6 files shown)


src/elements/Image/Image.js
```diff
@@ -20,8 +21,6 @@ import Dimmer from '../../modules/Dimmer'
 import Label from '../Label/Label'
 import ImageGroup from './ImageGroup'
 
-const imageProps = ['alt', 'height', 'src', 'srcSet', 'width']
```
layershifter, line 23: Moved to `lib` because is reusable.

PR #3274, opened 2018-11-12
feat(Responsive): re render only on visibility or props change

danielr18:
Instead of computing `isVisible()` on render, it's better to have its value in the state, as it avoids:

- `shouldComponentUpdate` logic
- re rendering on every `width` change, when its value is only used to check if it's visible

By having `isVisible` in state, the component only renders when you update `isVisible` and when props change. I also made sure that `isVisible` is recomputed if the props it depends on change.


layershifter:
@danielr18 I made some changes, can you please check?

danielr18:
Changes look good

PR #3286, opened 2018-11-19
feat(Ref): add support of `createRef` API

### LINKED ISSUE #3253: Ref: support `createRef()` API
This is my implementation, but it gives error. `innerRef is not a function. innerRef(findDOMNode(this))`

     function() {
        let el = useRef(null);
        return (
        <Ref innerRef={el}>
            <Table />
        </Ref>
        );
    }

layershifter:
Fixes #3253.

### `handleRef()`

This PR backports `handleRef()` utility from Stardust UI: https://github.com/stardust-ui/react/pull/459 That allows correctly work with `createRef` API that was introduced in React 16.3.

#### Before

image

#### After 

image

### DIFF
--- a/docs/src/examples/addons/Ref/Types/RefExampleRef.js
+++ b/docs/src/examples/addons/Ref/Types/RefExampleRef.js
@@ -37,13 +37,15 @@ export default class RefExampleRef extends Component {
           {isMounted && (
             <Table>
               <Table.Header>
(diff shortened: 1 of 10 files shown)


vanya2h:
Seems like this patch doesn't appear in npm's semantic-ui-react@0.83.0 distribution.

layershifter:
Yes, this stuff was not released, yet 😭 I hope that @levithomason will cut a release on this week.

PR #3294, opened 2018-11-21
Stronger typings for Dropdown and Select components

### LINKED ISSUE #3001: [Feature] Generic dropdown component
I'm toying around with making Dropdown generic. It will allow us to remove a lot of boiler plating from our app. Is this the right place to share this to spark a discussion? Obviously there's some performance and feature issues to work - but this proves that it _can_ work.

```ts
import * as React from "react";
import { Dropdown as SemanticDropdown, SemanticShorthandContent } from "semantic-ui-react";

interface IProps<T> {
    value: T | null | undefined;
    options: T[] | null | undefined;
    mapValue: (v: T) => boolean | number | string;
    mapText: (v: T) => SemanticShorthandContent;
    onChange: (v: T | undefined) => void;
}

export class Dropdown<T> extends React.Component<IProps<T>, any> {
    public render(): JSX.Element {
        const props = this.props;
        const options = props.options && props.options.map((o) => ({
            value: props.mapValue(o),
            text: props.mapText(o),
        }));
        return (
            <SemanticDropdown
                options={options || []}
                value={props.value && props.mapValue(props.value) || undefined}
                onChange={(e, d) => props.onChange(props.options && props.options.find((o) => props.mapValue(o) === d.value) || undefined)} />
        );
    }
}
```

Usage is incredibly easy - especially since typescript 2.9:

```ts
interface Props {
    animals: Animal[];
}
interface State {
    selectedAnimal: Animal;
}
interface Animal {
    id: number;
    name: string;
    color: string;
}
<Dropdown<Animal>
    value={this.state.selectedAnimal}
    options={this.props.animals}
    mapValue={(animal) => animal.id}
    mapText={(animal) => (<span style={{ color: animal.color }}>{animal.name}</span>)}
    onChange={(animal) => { this.setState({ selectedAnimal: animal }); }} />
```

Fabianopb:
As requested and then suggested by @levithomason in #3001, it would be nice to have improvements to typings. I've chosen not to bring [JSX generics](http://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#generic-type-arguments-in-jsx-elements) at first. We can discuss if that's necessary as part of this story.

So, there are two main improvements to `Dropdown` (and consequently `Select`) typings, mainly related to callbacks, which is where we actually access the data.

First is in `Dropdown.Item`. The callback's data argument is a generic type that can receive one argument with the shape of your data:
dropdownitem
Edit: the prop `legs={4}` appears as `number`, but that's inferred and doesn't come from the generic type passed to the props.

Second is to `Dropdown` itself (also extended to `FormDropdown`, `Select` and `FormSelect`). Callback's data is also a generic type that can receive one argument that will be passed all the way down to `options`:
shorthanddropdown

There's a gotcha with the tests, though, that's why three are failing. `hasValidTypings` checks only for `interfaces`, but the mapped type in `DropdownItemProps` is in fact a `type` (apparently it can't be an `interface` otherwise it can't extend `StrictDropdownItemProps`).

But I decided to open the PR anyway to first get feedback before deciding what to do about the use of `type` and the tests.

Cheers!
(fixes #3001)

### DIFF
--- a/src/addons/Select/Select.d.ts
+++ b/src/addons/Select/Select.d.ts
@@ -6,13 +6,13 @@ import DropdownHeader from '../../modules/Dropdown/DropdownHeader'
 import DropdownItem, { DropdownItemProps } from '../../modules/Dropdown/DropdownItem'
 import DropdownMenu from '../../modules/Dropdown/DropdownMenu'
 
-export interface SelectProps extends StrictSelectProps {
+export interface SelectProps<T = any> extends StrictSelectProps<T> {
   [key: string]: any
 }
 
-export interface StrictSelectProps extends StrictDropdownProps {
+export interface StrictSelectProps<T> extends StrictDropdownProps<T> {
   /** Array of Dropdown.Item props e.g. `{ text: '', value: '' }` */
-  options: DropdownItemProps[]
+  options: DropdownItemProps<T>[]
 }
 
 interface SelectComponent extends React.StatelessComponent<SelectProps> {
--- a/src/collections/Form/FormDropdown.d.ts
+++ b/src/collections/Form/FormDropdown.d.ts
@@ -3,11 +3,11 @@ import * as React from 'react'
 import { StrictDropdownProps } from '../../modules/Dropdown'
 import { StrictFormFieldProps } from './FormField'
 
-export interface FormDropdownProps extends StrictFormDropdownProps {
+export interface FormDropdownProps<T = any> extends StrictFormDropdownProps<T> {
   [key: string]: any
 }
 
-export interface StrictFormDropdownProps extends StrictFormFieldProps, StrictDropdownProps {
+export interface StrictFormDropdownProps<T> extends StrictFormFieldProps, StrictDropdownProps<T> {
   /** An element type to render as (string or function). */
   as?: any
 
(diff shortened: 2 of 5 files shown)


Fabianopb:
@levithomason and perhaps also @layershifter would be nice to hear some feedback whenever you have some time to put your brains in that. A couple of questions buzzing around this topic:

For using JSX Generics we would need class declaration to allow generic props, and at the moment we're exporting an instantiated type. The problem is that the class can't have subcomponents (or at least I couldn't find a way to solve that), so we wouldn't be able to have the top component called a `Dropdown` and at the same time allow it to have subcomponents (`Item` being a subcomponent for example). So I left that out because it could disrupt the components' semantics. What are your thoughts on that? Or maybe you have some ideas how to tackle that?

So, does this approach I've proposed here, allowing a generic type within the component's props look reasonable?

If yes I probably need to work in some kind of `type` validation (not only `interface`) in `hasValidTypings` to get valid and correct tests.

levithomason:
Yes, these changes look good to me on first pass.  @layershifter What are your thoughts on how these changes might intersect the work we're doing in the Stardust Specifications for generating types?

Fabianopb:
So it seems that we can have two approaches:
* either we pause this story, correct the shorthand API, and then come back to fix the types (wouldn't that break the current type implementation too?)
* or we accept this current solution for the types and then make a requirement in the story that will fix the shorthand API to also make sure the types will be corrected along.

layershifter:
> either we pause this story, correct the shorthand API, and then come back to fix the types (wouldn't that break the current type implementation too?)

We definitely want to go with this option and then fix types because I understand the usecase, however I am not ready to merge these changes right now. @Fabianopb thank you for your effort on this.

PR #3301, opened 2018-11-25
fix(Popup): remove logic with setSafeState()

layershifter:
This PR removes useless logic with `setSafeState()`, we just need to clear properly intervales and frames.

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -148,10 +148,6 @@ export default class Popup extends Component {
 
   state = {}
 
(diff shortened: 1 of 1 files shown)


src/modules/Popup/Popup.js
```diff
@@ -164,7 +160,7 @@ export default class Popup extends Component {
   }
 
   componentWillUnmount() {
-    this.mounted = false
+    clearTimeout(this.timeoutId)
```
levithomason, line 163: 👍

PR #3308, opened 2018-11-26
fix(Rating): conditionally set tabIndex when element is disabled

### LINKED ISSUE #3252: Rating: RatingIcon has tabindex=0 when parent Rating component is disabled
### Steps
`<Rating rating={1} maxRating={5} disabled size="large" />`

### Expected Result
Since the parent `<Rating>` component is disabled, it doesn't make sense to be able to use the keyboard to focus the individual rating icons.

### Actual Result
The individual rating icons have tabindex=0 so they can be reached via the keyboard.

### Version
0.83.0

### Testcase
https://codesandbox.io/s/7kww573lvx

Fabianopb:
As raised in #3252, RatingIcons can be individually focused when navigating with Tab, which doesn't make sense as users cannot select disabled elements anyway.

However, thinking a bit further, still makes sense to be able to focus the whole element so it can still be accessible (i.e.: users with vision impairment can still be aware of its existence).

So the solution was to set `RatingIcon` `tabIndex = -1` ([reference](https://www.w3.org/TR/html5/editing.html#the-tabindex-attribute)) and at the same time keeping `tabIndex = 0` in the `Rating` component, when the whole element is in `disabled` state.

The demo:
rating-element

Also added test cases.

PS.: I just noticed that some lines in the test file where unintentionally reformatted, I guess we have new lint/prettier rules?

(closes #3252)

### DIFF
--- a/src/modules/Rating/Rating.js
+++ b/src/modules/Rating/Rating.js
@@ -126,6 +126,7 @@ export default class Rating extends Component {
       >
         {_.times(maxRating, i => (
           <RatingIcon
+            tabIndex={disabled ? -1 : 0}
             active={rating >= i + 1}
             aria-checked={rating === i + 1}
             aria-posinset={i + 1}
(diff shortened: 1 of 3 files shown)


levithomason:
Fantastic, thanks for the work, write up, and tests ❤️

PR #3316, opened 2018-12-01
docs(Layouts): update FixedMenuLayout

xiaohanyu:
Keep maximum compatibility with https://semantic-ui.com/examples/fixed.html


xiaohanyu:
Code from https://semantic-ui.com/examples/fixed.html

image

image

VS 

image

PR #3317, opened 2018-12-03
fix(Dropdown): clear searchQuery on item selection by mouse click

### LINKED ISSUE #3306: Dropdown: multiple does not clear on mouse click
### Steps
Use a Multiple-search dorpdown, such as the one in this example
https://react.semantic-ui.com/modules/dropdown/#types-multiple-search-selection
Type "A". A few values will be offered
With your **mouse**, select any of the values
### Expected Result
The "a" searchQuery is cleared

### Actual Result
The "a" searchQuery remains

### Version
0.83.0

### Note
If you use keyboard only to select a value from the list, then the proper behavior occurs. The issue is only when using the mouse

kohakukun:
- Remove check to have consistent behavior with `selectItemOnEnter`

Fixes #3306.

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -1,6 +1,6 @@
 import cx from 'classnames'
 import keyboardKey from 'keyboard-key'
-import _ from 'lodash'
+import _ from 'lodash' 
 import PropTypes from 'prop-types'
 import React, { Children, cloneElement } from 'react'
 import shallowEqual from 'shallowequal'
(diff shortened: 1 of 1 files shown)


src/modules/Dropdown/Dropdown.js
```diff
@@ -729,8 +729,7 @@ export default class Dropdown extends Component {
     this.setValue(newValue)
     this.setSelectedIndex(value)
 
-    const optionSize = _.size(this.getMenuOptions())
-    if (!multiple || isAdditionItem || optionSize === 1) this.clearSearchQuery()
+    this.clearSearchQuery()
```
layershifter, line 732: This change is absolutely valid, we did this in #2162 and even had an comment about this https://github.com/Semantic-Org/Semantic-UI-React/pull/2162#issuecomment-355130565.

See our decision https://github.com/Semantic-Org/Semantic-UI-React/pull/2162#issuecomment-334096165. I checked examples manually, LGTM.

layershifter:
@kohakukun thanks for the investigation 👍

PR #3324, opened 2018-12-05
chore(build): drop DLL stage

layershifter:
This PR removes `dll` stage of build because it doesn't provide any performance benefits and massively simplifies webpack's config. Build time was reduced from `01:36` to `01:25`.


levithomason:

Issue #3329, opened 2018-12-08
Add 'aria-pressed' attribute for toggle button


shufflerAbhi:
## Feature Request
Add `aria-pressed` attribute for toggle button. 
### Problem description
Accessibility guidelines suggest that toggle buttons have an additional `aria-pressed` attribute. This ensures that screen readers understand that the button supports being pressed and not-pressed.
[MDN guide on accessible toggle buttons ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Toggle_buttons)

### Proposed solution
The `aria-pressed` attribute can be set based on the value of `toggle` and `active` props. If `toggle` is true, the value of `active` can be used to set `aria-pressed` to the corresponding value. If `toggle` is false or undefined, then `aria-pressed` should be undefined.

layershifter:
Fixed in #3330

PR #3330, opened 2018-12-08
feat(Button) : Add aria-pressed attribute for toggle button

shufflerAbhi:
This adds the `aria-pressed` attribute for toggle button. #3329 

The value of `active` prop determines the value of `aria-pressed` when `toggle` is set. Otherwise `aria-pressed` is left undefined.

A few test cases have been added to verify functionality.


levithomason:
Awesome, thanks!

Issue #3331, opened 2018-12-10
Docs: "Props" toggle is not working


mac-s-g:
it looks like there were some changes to the docs website in recent history

https://react.semantic-ui.com/elements/input/

I can't enable or view props for the Input component.  this is one example of an issue that is pretty consistent for similar components.

I also don't know where to look to find information about controlled Input states and events.  The information all used to be displayed at the location linked above.

is there any way to access the older version of the docs website?  i really think the new site needs some  QA or user testing.  making documentation harder to find could have a serious impact on adoption.

gotta say though, this is a really amazing project.  so thanks for all the hard work.

and thanks for reading my support request.

Mac

layershifter:
> I can't enable or view props for the Input component. this is one example of an issue that is pretty consistent for similar components.

PR with fix #3332, docs are also deployed recently.

> is there any way to access the older version of the docs website?

Yes, you can pull from Git the matching tag and build it locally, [CONTRIBUTING.md#commands](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/.github/CONTRIBUTING.md#commands).

```sh
git clone https://github.com/Semantic-Org/Semantic-UI-React.git
git checkout tags/v0.83.0
yarn
yarn start
```

>  really think the new site needs some QA or user testing

Fully agree, feel free to create issues for non-working or incorrectly working things as bug reports. We will try to fix them.

It's an opensource project, so we really need your help and contributions 🐱 You can also create examples for component usages 👍

PR #3332, opened 2018-12-10
docs(Props): fix "Props" checkbox

layershifter:
Refs #3331.

Fixes issue with uncheckable props checkbox.


levithomason:

PR #3334, opened 2018-12-11
Add support for data-testid attribute to Input

ikovic:
We rely on `data-testid` attributes for our automated e2e and integration tests. We can't set the attribute value to the correct `input` element since setting that as a prop on an `Input` will set the attribute value on the wrapping `div` instead of the child `input`.

I've browsed the codebase and I've also come across this issue:
https://github.com/Semantic-Org/Semantic-UI-React/issues/2979

Not sure if this is the best way to solve it, so please comment if there's anything you want me to change. Thanks!

### DIFF
--- a/src/lib/htmlPropsUtils.js
+++ b/src/lib/htmlPropsUtils.js
@@ -78,7 +78,9 @@ export const htmlInputEvents = [
   'onTouchStart',
 ]
 
-export const htmlInputProps = [...htmlInputAttrs, ...htmlInputEvents]
+export const testAttributes = ['data-testid']
+
+export const htmlInputProps = [...htmlInputAttrs, ...htmlInputEvents, ...testAttributes]
 
 export const htmlImageProps = ['alt', 'height', 'src', 'srcSet', 'width']
 


layershifter:
Is there any reason why this will not work for you?

```jsx
<Input data-testid='foo' />
```
```js
document.querySelector('[data-testid="foo"] input') // will point to input
```

ikovic:
@layershifter thanks for answering! At the moment we are doing exactly that. I'd still prefer if there was some kind of option to set the attribute directly on the `input`. Maybe another prop type on Input - eg `innerInputAttributes`?

Main reason is to be consistent - we have a convention of using `data-testid` on all interactive elements, this just makes us go a bit sideways. Not too big of a problem anyway :)

brianespinosa:
You can technically do this which I think will work for you in most cases.
```
<Input ...props>
  <input data-testid='foo' />
</Input>

ikovic:
@brianespinosa thank you, I'll give it a try! I'd use that approach if there are no negative side effects.

layershifter:
We have a working solution there, so I'm going to close this.

PR #3336, opened 2018-12-12
BREAKING (Sidebar): remove `duration` prop

### LINKED ISSUE #3328: Sidebar: "duration" prop does not appear to do anything
### Steps
1. Visit https://react.semantic-ui.com/modules/sidebar/#types-sidebar
2. Add the prop `duration={5000}` in the `<Sidebar/>` component of the first example on the page
3. Open the sidebar in that example

### Expected Result
The Sidebar animation is slower

### Actual Result
Nothing is changed

### Version
0.83.0

It doesn't look like this prop works with any `<Sidebar/>` configuration unless I'm not understanding how `duration` is supposed to work.

Fabianopb:
`duration` prop has been removed from the `Sidebar` component. However I've maintained a static variable in the `Sidebar` class to still provide the default duration of 500 ms that is used to manage the callback cycle when showing and hiding the sidebar.

@layershifter, as `animationDuration` is now a static prop it can actually be assigned externally (like I've done in the modified tests for this PR). If that's ok, should we add that to the type definitions? It could be done as easily as modifying the `Sidebar` type declaration in the end of `Sidebar.d.ts` to something like:

```ts
declare const Sidebar: SidebarComponent & { animationDuration: number }
```
Cheers!
(Closes #3328)

### DIFF
--- a/src/modules/Sidebar/Sidebar.d.ts
+++ b/src/modules/Sidebar/Sidebar.d.ts
@@ -27,9 +27,6 @@ export interface StrictSidebarProps {
   /** Direction the sidebar should appear on. */
   direction?: 'top' | 'right' | 'bottom' | 'left'
 
-  /** Duration of sidebar animation. */
-  duration?: number | string
-
   /**
    * Called before a sidebar begins to animate out.
    *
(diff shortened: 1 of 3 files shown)


Fabianopb:
Friendly ping @layershifter and @levithomason, I'd be happy to get some feedback whenever you guys get some time! :)

layershifter (approved):
@Fabianopb sorry, I'm a little busy now. I don't want to expose this in typings now, we can do this if we will have a request for it.

LGTM, thanks :+1:

levithomason:

PR #3340, opened 2018-12-14
docs(Placeholder): remove extra HTML and classes

levithomason:


levithomason:

PR #3343, opened 2018-12-18
fix(Textarea): fix broken ctrl-z on firefox

ericls:
In TextArea ctrl-z is currently broken on firefox if autoHeight is enabled.

I feel like it might be a firefox bug but this workaround will make it work.


layershifter:
The issue is really strange 🤔 
@ericls thanks for the fix!

levithomason:

PR #3351, opened 2018-12-24
fix(Checkbox): prevent `onClick` from being called twice

### LINKED ISSUE #3348: Checkbox: double call of `onClick` when event fired by mouse
e.g. use the example from the docs and change the Checkbox onChange to onClick:

```
import { Button, Checkbox } from 'semantic-ui-react'

export default class CheckboxExampleRemoteControl extends Component {
  state = { checked: false }
  toggle = () => this.setState({ checked: !this.state.checked })

  render() {
    return (
      <div>
        <Button onClick={this.toggle}>Toggle it</Button>
        <Checkbox label='Check this box' onClick={this.toggle} checked={this.state.checked} />
      </div>
    )
  }
}
```

Fabianopb:
Alright, I have to say that this was not trivial at all, but now the handlers seem to work for all cases. After plenty of investigation and trying different approaches, the solution was to leave the call of `onClick` to the click handler.

Here are gifs showing the component working for three cases (in all of them I'm checking and unchecking with clicks and then with the spacebar):

* Simple uncontrolled component
normal-checkbox

* Controlled component with `setState` as an **object**
controlled-checkbox

* Controlled component with `setState` as a **function** (maintaining the functionality implemented for #2587 fixed in #2748)
controller-checkbox-functional-setstate

Cheers and Merry Xmas! 🎄 

(Closes #3348)


layershifter:
Please check the `DOM Comparison` example.

### Case 1 `mouse click`: FAILED
`SUIR checkbox`
```
Click      { checked: false }
Change     { checked: true }
```
`DOM checkbox`
```
DOM Change { checked: true }
DOM Click  { checked: true }
```

### Case 2 `space key click`: PASSED
`SUIR checkbox`
```
Change     { checked: true }
Click      { checked: true }
```
`DOM checkbox`
```
DOM Change { checked: true }
DOM Click  { checked: true }
```

### Case 3 `ID: mouse click on checkbox`: FAILED
`SUIR checkbox with ID`
```
Change     { checked: true }
```
`DOM checkbox with ID`
```
DOM Change { checked: true }
DOM Click  { checked: true }
```

### Case 4 `ID: mouse click on label`: FAILED
`SUIR checkbox with ID`
```
Change     { checked: true }
```
`DOM checkbox with ID`
```
DOM Change { checked: true }
DOM Click  { checked: true }
```

---

@Fabianopb I briefly checked the changes, they shown that we need more strict tests there and changes to match DOM checkbox.

Something like this:

```js
wrapper.find('label').simulate('click')

onClick.should.have.been.calledOnce()
onClick.should.have.been.calledWith(e, { checked: true })
onChange.should.have.been.calledOnce()
onChange.should.have.been.calledWith(e, { checked: true })

onChange.should.have.been.calledAfter(onClick)
```

https://github.com/domenic/sinon-chai

Fabianopb:
Sorry @layershifter where can I find these `DOM Comparison` examples?

layershifter:
They are in the bottom in Checkbox docs:
https://react.semantic-ui.com/modules/checkbox/#usage-dom-comparison
http://localhost:3000/modules/checkbox/#usage-dom-comparison

Fabianopb:
Alright thanks, I'll take a look!

Fabianopb:
@layershifter I've found a quite curious solution, which was to remove the `onClick` call completely from `handleClick` (in other words, completely removing the click handler from `Checkbox`).

By doing that, the interactions still work as I've shown in all those gifs and the `DOM Comparisons` are all valid (you can already check from my latest commit)!

However the test `handles events transparently` fails as it can't find `onClick` being fired from `handleClick`. As everything works as expected, should we treat this case as an exception?

Fabianopb:
@layershifter I've created four tests for DOM Comparisons, simulating the change event on:
* `input` without `id`
* `label` without `id`
* `input` with `id`
* `label` with `id`

The `change` event is used because `onClick` is invoked within `handleChange` (as already done previously). By doing that we also cover spacebar key "click" as it fires the same events as a click in this component.

However, I had to disable the common test `isConformant` due to the empty `handleClick` handler.

My proposition is that we **merge this PR** because the bug is fixed and the DOM behavior is conformant, however, we should **open a new issue** to discuss and tackle the empty handler and disabled test (I could open that and start the discussion). What do you say? :)

Fabianopb:
Thanks @layershifter and Happy New Year! 🎉

layershifter:
@Fabianopb I updated tests to work with DOM deeply. It will be awesome if you will be able to take a look and check them 👍

Fabianopb:
It looks good to me @layershifter, nice set up with the  `assertMatrix` and `domEvent.fire`!

I've added two more tests for `onChange` and `onClick` on controlled components with `setState` as function, which cover the functionality of this bug fix and it's also a kind of confirmation test for #2748.    From me that should be it, I think this is ready to merge! :)

src/modules/Checkbox/Checkbox.js
```diff
@@ -163,10 +163,28 @@ export default class Checkbox extends Component {
...
+    // https://github.com/Semantic-Org/Semantic-UI-React/pull/3351
+    //
+    // We need to avoid duplicate calls to onClick.
+    // We need to handle clicks on the root and bubble clicks from the label and input.
+    // The order of the events needs to match that of a vanilla DOM checkbox.
+    //
+    // The browser calls onClick AFTER onChange since onClick is on the root element.
+    // Because of this, we call onClick in handleChange to preserve the correct order.
+    // In that case, we skip calls to onClick here to avoid duplicate calls.
+    if (onChange) return
```
Fabianopb: Wouldn't this cause issues in case the user has both `onClick` and `onChange` defined?

Fabianopb: Digging a bit deeper, this change actually breaks the use of `onClick` like described in #3348 , I've updated the tests on controlled components to warn us accordingly. You will notice that if you revert this commit, the tests will pass.

levithomason: Thanks for the double checking.  I have a couple more updates I'll push, then I think we can get a final on this and merge it.

Fabianopb: Thinking again, it's a bit misleading to have an empty handler like I had done before, I'll push a new commit removing it completely and adding the explanation close to the onClick invocation.

Fabianopb:
Some 10 days since the last discussions, just checking, what's the status? :)

layershifter:
Merged, @Fabianopb thanks for working on it 👍

Fabianopb:
Awesome @layershifter thanks!

test/specs/commonTests/isConformant.js
```diff
@@ -13,6 +13,7 @@ import hasValidTypings from './hasValidTypings'
  * Assert Component conforms to guidelines that are applicable to all components.
  * @param {React.Component|Function} Component A component that should conform.
  * @param {Object} [options={}]
+ * @param {String[]} [options.disabledHandlers=[]] An array of listeners that are disabled.
```
levithomason, line 16: Heads up @layershifter we'll need to undo this sometime soon.  Ignoring events means we aren't testing that all events work correctly for the end user.  I don't think it is high priority, but something to keep on our tech debt list.

levithomason:

PR #3354, opened 2018-12-27
docs(Responsive): update ResponsiveExampleMaxWidth.js (added two words to clarify)

janderudder:
Added "and lower" to the 2569px width example, because that's what it does and the wording was confusing on first sight.


levithomason:

PR #3358, opened 2018-12-31
feat(Icon): `disabled` prop disables `onClick`

### LINKED ISSUE #3341: Icon: `disabled` prop should disable `onClick` callback
## Bug Report

Based on past and open issues, it sounds like `onClick` callback execution has been updated to not fire when `disabled === true` for some but not all elements. I think `Icon`s should not execute `onClick` callbacks when `disabled === true` to maintain parity with other elements (eg `Button`) which have been updated to maintain parity with HTML.

### Steps
A clear and concise description of steps to reproduce the problem.

```jsx
import React from 'react'
import { Icon } from 'semantic-ui-react'

const IconExampleDisabled = () => <Icon disabled name='users' onClick={() => console.log('hi') }/>

export default IconExampleDisabled
```

Click the icon.

### Expected Result

When the icon is clicked, the `onClick` callback should not be called and "hi" should not be logged.

### Actual Result

When the icon is clicked, "hi" is logged.

### Version

0.82.3

### Testcase

https://codesandbox.io/s/2o4qo44vrn

danielsyang:
Added behaviour of not running `onClick` function when Icon is `disabled`.
Very similar with what is already available in the `Button` component.

Fixes #3341.

### DIFF
--- a/src/elements/Icon/Icon.js
+++ b/src/elements/Icon/Icon.js
@@ -99,6 +99,17 @@ class Icon extends PureComponent {
     return ariaOptions
   }
 
+  handleClick = (e) => {
+    const { disabled } = this.props
+
+    if (disabled) {
+      e.preventDefault()
+      return
+    }
+
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
   render() {
     const {
       bordered,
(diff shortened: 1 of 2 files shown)


spencer-brown:
Awesome. Thanks, @shuminyang !

levithomason:

PR #3359, opened 2019-01-01
Fix: Icon onClick callback event disbled for 'disabled' props

harish-a:
As requested in issue #3341, I have added fix for that issue.


spencer-brown:
Awesome. Thanks, @harish-a !

Issue #3194, opened 2018-10-03
Dropdown: closing Dropdown inside a Modal by clicking Escape also closes Modal


vkramskikh:
## Bug Report

There is no way to prevent an event triggered by a Dropdown to affect a Modal containing it. Reproduces with Escape, Enter and other keys/events.

### Steps
1. Create a search+selection Dropdown inside a Modal
2. Try start searching some value in the Dropdown and hit Escape to close it

### Expected Result
Dropdown is closed, Modal is open

### Actual Result
Modal is closed

### Version
0.82.5

### Testcase
https://codesandbox.io/s/9y8vzn4nvr

yuritoledo:
Hey @vkramskikh , try to add this on modal component: `closeOnEscape={false}`

It should work :)

vkramskikh:
Sure it will work. But I want it to close on escape, just not when Dropdown is open and users want the Dropdown to close, not the Modal. Same applies to Enter key - users want to choose item in the Dropdown, not to submit the form in Modal

blnk-space:
Hi guys!
I have currently used just one string ('dropdown') to implement the check and stop propagation.

However, ideally I'd want semantic-ui-react to have a centrally defined array of classes, that if found inside a Portal, should take over the keyboard event. Let me know which other components except for dropdown need to have this behavior.

jongsue:
I think you should handle escape event by props onClose
something like this
https://codesandbox.io/s/mjrlq64w8x

Modal could have any Component, So It can't decide to ignore which escape event come from

fiskpatte:
@vkramskikh , I had a similar problem. If a certain input field inside my modal was visible, I wanted to close it with escape. Otherwise I wanted the modal to close as normal with escape.
My solution was to use the modal as controlled component, then set `closeOnEscape={false}` on the modal, then inside of it I listened for the escape click event and if the input field was visible I just closed that one, otherwise I called the function I passed in that closed the modal.

```
<Modal 
    open={this.state.newGroupModalOpen}
    closeOnEscape={false}>
         <Modal.Content>
               <NewModifierGroup 
                      cancel={this.closeModifierModal.bind(this)}
               />
         </Modal.Content>                    
</Modal>
```
NewModifierGroup:
```
    
    componentDidMount(){
        document.addEventListener("keydown", this.catchEscape);
    }

    componentWillUnmount(){
        document.removeEventListener("keydown", this.catchEscape); 
    }

[...]

PR #3362, opened 2019-01-05
Fix(Portal): Modal handles keyboard escape event instead of child dropdown

aviral-public:
fixes issue: #3194 Dropdown: closing Dropdown inside a Modal by clicking Escape also closes Modal


aviral-public:
closing this as it's a bad solution, doesn't work when dropwdown is defaultOpen

Issue #3339, opened 2018-12-13
Popup: should contain `disabled` prop


spirkovskii:
### Problem description
I have an issue when using Popup to display validation error on some input.

If the input is invalid I render it in the popup via trigger prop, if it's not invalid I render just the input.
The problem with this is when the state changes from invalid to valid and vice versa, my input loses state.

![example](https://user-images.githubusercontent.com/14183595/49969110-ff7d2880-ff27-11e8-9440-3af236ccdf1d.PNG)

### Proposed solution
I would like the popup to have prop disabled, which would only render the component passed via trigger.
If you think this makes sense, I am willing to implement it

levithomason:
Tips:

1. You can use the [`context` Popup prop](https://react.semantic-ui.com/modules/popup/#usage-context) to pass a node reference.
2. You can use the [Ref component](https://react.semantic-ui.com/addons/ref/) to ensure you always get a DOM node reference.

This will always render the same `Input` element, even when the `invalid` state changes.  The Popup will only be added to the page when the input is invalid.  Since the Input is not a trigger of the Popup, it should not lose state or focus.  It will always exist in the DOM:

```jsx
import { Input, Popup, Ref } from 'semantic-ui-react'

// ...

<>
  {invalid && <Popup context={this.inputRef} />}

  <Ref innerRef={node => (this.inputRef = node)}>
    <Input />
  </Ref>
</>
```

spirkovskii:
Thank you for your answer Levi,

I think it would be good idea to add example with using Ref in the documentation, since Ref is buried in the bottom of the components under addons, and no clue that it could be used like this was provided in the docs.

This solves my problem but I still think 'disabled' prop should be added, it may solve some other specific scenarios for the consumer and after all it doesn't bring any breaking changes and would save few lines of unnecessary implementation.

levithomason:
Agreed on both counts.  We have an update coming for the Popup as well, #3377.  Would you like to add an example for the `Ref`?

PR #3377, opened 2019-01-12
feat(Popup): add `disabled` prop

husseyexplores:
Fixes: https://github.com/Semantic-Org/Semantic-UI-React/issues/3339

Let me know if something is missing from this.


src/modules/Popup/Popup.d.ts
```diff
@@ -109,6 +109,9 @@ export interface StrictPopupProps extends StrictPortalProps {
 
   /** Element to be rendered within the confines of the viewport whenever possible. */
   keepInViewPort?: boolean
+
+  /** Disables the popup and only renders the element passed via trigger prop */
```
levithomason: Suggestion:

```suggestion
  /** A disabled popup only renders its trigger. */
```

This will also be the title of the doc site example.

The `disabled` prop is a new popup state, so we should add a doc site example to `docs/src/examples/modules/Popup/States/PopupExampleDisabled.js`.

Our docs are powered by the structure of the examples directory.  You can reference `/State` examples in other components for a guide as to how to structure this one for the Popup.

levithomason (dismissed):
This looks great, let's just add a doc site example.  See my comment.

src/modules/Popup/Popup.d.ts
husseyexplores: Thank you for the feedback!

I've renamed the description as per the suggestion and also added the State example.
Let me know what else can be improved.

layershifter (approved):

levithomason:

PR #3378, opened 2019-01-14
[WIP]: chore(docs): update React Static

layershifter:
# WIP


layershifter:
Actually, there is a blocker:
https://github.com/nozzle/react-static/pull/983
https://github.com/nozzle/react-static/issues/974

levithomason:
Static is great for SEO, but it has complicated things quite a bit and degraded perf.  What about using a prerender service instead?

layershifter:
I want to resolve performance issues before this merge :rocket:

PR #3380, opened 2019-01-15
fix(Input): add `select` to type definitions

skeate:
`select` was added in 0.82 but the type definition wasn't updated 😞


levithomason:

PR #3382, opened 2019-01-16
chore(tests): make reporter more silent

layershifter:
Current tests output is too verbose, this PR makes reporter more silent.


levithomason:

PR #3383, opened 2019-01-16
chore(package): update karma and webpack

layershifter:
This PR breaks partial test runs because uses `karma-webpack` instead of `karma-webpack-with-fast-source-maps`. However, `karma-webpack-with-fast-source-maps` is outdated and doesn't support Webpack 4.

We should switch to Jest if they will improve the performance in Jest 24/25.


karma.conf.babel.js
```diff
@@ -91,12 +91,8 @@ export default (karmaConfig) => {
       'test/tests.bundle.js': ['webpack'],
     },
     webpack: {
+      ...webpackConfig,
```
levithomason, line 94: This is now spreading keys that aren't applicable to karma's webpack config (`name`, `mode`, `target`, `output`, etc).  I think we should stick with explicitly passing only the keys we want as it was before.

levithomason (approved):
Merge after reviewing comments 👍

levithomason:
Yes, we need a faster runner.  Test optimization is needed in this project.

karma.conf.babel.js
layershifter: Now we have only `webpack.config.karma.js`, so I cleaned up it more and removed `name`, `output`, `target` from there.

`mode` is required setting for Webpack 4, otherwise it will run in production mode.

levithomason:

PR #3390, opened 2019-01-18
chore(package): upgrade keyboard-key

### LINKED ISSUE #3251: Docs: make the filter keyboard shortcut other than "/" (for international keyboards)
I search for components a lot.

With my German keyboard I cannot use the shortcut "/".

levithomason:
Fixes #3251

The latest release of `keyboard-key` now respects event objects that already have a `key` property.  This should resolve many bugs like what was noted in the linked issue.

The long term fix there is to add locale support to keyboard-key.  Namely, we'd need to handle the proper <kbd>shift</kbd> values for every keyboard locale.  See the new [Locale Caveat](https://github.com/levithomason/keyboard-key/#locale-caveat) section over at `keyboard-key`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "@babel/runtime": "^7.1.2",
     "@semantic-ui-react/event-stack": "^3.0.1",
     "classnames": "^2.2.6",
-    "keyboard-key": "^1.0.2",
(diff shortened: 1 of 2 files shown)


levithomason:

PR #3391, opened 2019-01-18
fix(Dropdown): prevent calling onChange unless value changed

zarend:
For the Dropdown component, ensures that onChange is called only if the user selectes a value that is different from the current value of the Dropdown. This is so we can have parity with the regular HTML select.


levithomason:
Excellent work, thank you much!

levithomason:

PR #3393, opened 2019-01-20
fix(Icon): update typings for `corner` prop

areading314:
See docs: https://react.semantic-ui.com/elements/icon/


layershifter:
Thanks much, missed this in #3315 👍

levithomason:

PR #3395, opened 2019-01-21
fix(docs): fix a polyfill loading for IE11

### LINKED ISSUE #3369: Cannot browse documentation with IE11
### Steps
* Open IE11
* Browse to https://react.semantic-ui.com/modules/modal/

### Expected Result
The documentation is displayed.

### Actual Result
The dev tools show an error:
`Object doesn't support property or method 'assign'`.

Error is in the `babel.min.js` file.

### Version
Documentation live as of January 9th, 2019.

oolleegg55:
As @levithomason have said here https://github.com/Semantic-Org/Semantic-UI-React/issues/3369#issuecomment-455391652 I moved a babel polyfill to the top of the page to load it first.
Also I removed import from index file.

fixes #3369

### DIFF
--- a/config.js
+++ b/config.js
@@ -43,7 +43,7 @@ const config = {
   // ----------------------------------
   // Compiler Configuration
   // ----------------------------------
-  compiler_devtool: (__DEV__ || __TEST__) && 'cheap-eval-source-map',
+  compiler_devtool: (__DEV__ || __TEST__) && 'cheap-source-map',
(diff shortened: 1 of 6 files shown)


layershifter:
I will add some more change because we have locally the similar issue, https://github.com/halfzebra/create-elm-app/issues/217

levithomason:

PR #3396, opened 2019-01-21
docs(elements/Icon/Variations/IconExampleColored): add white

JorgeMadson:
Adding white color to the icon colors.
Why we don't have a white color for icons?


layershifter:
SUI core doesn't have definitions for `white` color, https://github.com/Semantic-Org/UI-Icon/blob/master/icon.css

You can use `inverted`:

```jsx
const Example = () => (
  <div style={{ background: 'red' }}>
    <Icon inverted name='user' />
  </div>
)
```

PR #3400, opened 2019-01-24
fix(Visibility): handle context scroll instead of just window scroll

johnkeane475:
Currently adding a context to the Visibility component will add event listeners for that context, however the inner workings of the Visibility component still use the window scroll position for calculations.

This means that we are unable to use the Visibility component on any context other than the window.
This PR seeks to resolve this issue.

The code changes I have made are working for my use case, however I require some assistance in getting the tests to work as well as ensuring that the changes I have made are suitable for all use cases of the Visibility component.

Thanks :)


src/behaviors/Visibility/Visibility.js
```diff
@@ -363,7 +364,8 @@ export default class Visibility extends Component {
     const { bottom, height, top, width } = this.ref.getBoundingClientRect()
     const [topOffset, bottomOffset] = normalizeOffset(offset)
 
-    const direction = window.pageYOffset > this.pageYOffset ? 'down' : 'up'
+    const newOffset = (this.props.context ? this.props.context.scrollTop : window.pageYOffset)
```
levithomason: Since there are 4 instances of this, we should probably consider a method:

```jsx
getPageYOffset() {
  const { context } = this.props
  return context ? context.scrollTop : window.pageYOffset
}
```

test/specs/behaviors/Visibility/Visibility-test.js
```diff
@@ -254,16 +254,16 @@ describe('Visibility', () => {
...
+      // it('returns up when scrolling down', () => {
+      //   const onUpdate = sandbox.spy()
+      //   mount(<Visibility onUpdate={onUpdate} context={window} />)
+      //
+      //   window.pageYOffset = 5
+      //   domEvent.scroll(window)
+      //   onUpdate.should.have.been.calledWithMatch(null, {
+      //     calculations: { direction: 'down' },
+      //   })
+      // })
```
levithomason: ~~Does this test fail?~~ Per the PR description, I've marked this one as needing some help! 👍

levithomason:
Looking good so far! As noted, we just need some more hands on the test case.

johnkeane475:
Thanks for the feedback @levithomason 
I've pushed a new commit with your suggested tidy-up.
I've also uncommented the failing test for clarity.

johnkeane475:
@layershifter Thank you very much for your help with this PR.
Is there a schedule for when the next version will be released?

layershifter:
Only Levi has release rights and he is on a vacation right now. I have asked him for a release, but I don't have any ETA 🙂

johnkeane475:
No problem, thanks :)

PR #3403, opened 2019-01-27
Use string enums rather than string unions to define types

nessup:
Specifying the value of `color` on a `Progress` component yields a TypeScript error, even if passing in a defined string constant (such as `green`). Is there any reason the union types such as `SemanticCOLORS` and `SemanticDIRECTIONALTRANSITIONS` aren't defined as string enums?

I've included a proposed fix for the `SemanticCOLORS` type. By turning it into an enum, you get an inferred type definition as well as string values that can be used at runtime. Happy to submit a complete PR including a bunch of enums if this seems to be interesting to the maintainers. I'm not sure where such enums should live, and whether they should be made accessible to regular JavaScript.

### DIFF
--- a/src/generic.d.ts
+++ b/src/generic.d.ts
@@ -73,20 +73,21 @@ export type SemanticShorthandItem<TProps> =
 // Styling
 // ======================================================
 
-export type SemanticCOLORS =
-  | 'red'
-  | 'orange'
-  | 'yellow'
-  | 'olive'
-  | 'green'
-  | 'teal'
-  | 'blue'
-  | 'violet'
-  | 'purple'
-  | 'pink'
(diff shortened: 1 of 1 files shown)


layershifter:
Use these exported types, please. This will satisfy TS compiler.

1. ` const enum` will fail compiling with `--isolatedModules`, https://github.com/DefinitelyTyped/DefinitelyTyped/issues/23002#issuecomment-359051201
2. Usual `enum` will point to an unexisting export entry, https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31248#pullrequestreview-183468400

layershifter:
Closing for housekeeping.

PR #3404, opened 2019-01-28
docs(ComponentExample): allow to disable HTML preview

layershifter:
Extracted from #3029.


levithomason:

PR #3407, opened 2019-01-31
fix(Portal): fix onBlur handling in IE11

### LINKED ISSUE #3365: Popup: onClick handler inside a popup doesn't work in Internet Explorer 11
## Bug Report

In Internet Explorer 11 the following is happening. When element placed inside a popup its `onClick` handler doesn't fire if `Popup` has `on='focus'`. If `Popup`'s `on` property set to any other value `onClick` handler works as intended.

In Chrome and Firefox `onClick` handler works in all cases.

### Steps
Open https://arfedulov.github.io/test/SUIRTest/. Click on buttons inside each popup.

### Expected Result
You should see alert message in all three cases.

### Actual Result
In Chrome and Firefox you see alert message after clicking on any of three buttons. But in Internet Explorer 11 you see the message only in two cases.

### Version
Internet Explorer 11.0.9600.17631
semantic-ui-react@0.84.0

### Testcase
Example here https://arfedulov.github.io/test/SUIRTest/.
Exactly the same code here https://codesandbox.io/s/1rnom7m5oj. But I couldn't open this link in IE.

layershifter:
Fixes #3365.

### Before in IE11

portal-before

### After

portal-after


levithomason:

PR #3417, opened 2019-02-06
feat(Accordion): add `icon` shorthand for AccordionTitle

sako9:
Added an example on how to add a custom icon via the standard Shorthand


sako9 (commented):
Looks good!

levithomason:

PR #3420, opened 2019-02-08
docs(Portal): fix controlled Portal usage

### LINKED ISSUE #3408: Portal: controlled portal is not closed on button click
### Steps
https://react.semantic-ui.com/addons/portal/#types-controlled

1.  Click "Open portal"
2. Click  "Close portal"

### Expected Result
Portal closed

### Actual Result
Portal is not closed

### Version
0.84.0

### Testcase
Docs: https://react.semantic-ui.com/addons/portal/#types-controlled

Fabianopb:
As discussed in #3408 the usage of `Portal` as a controlled component was wrong in the docs creating a false positive bug. The proper way puts the control inside the `Portal`'s `trigger`:

image

(Closes #3408)

### DIFF
--- a/docs/src/examples/addons/Portal/Types/PortalExampleControlled.js
+++ b/docs/src/examples/addons/Portal/Types/PortalExampleControlled.js
@@ -4,9 +4,8 @@ import { Button, Grid, Header, Segment, Portal } from 'semantic-ui-react'
 export default class PortalExampleControlled extends Component {
   state = { open: false }
 
-  handleClick = () => this.setState({ open: !this.state.open })
-
   handleClose = () => this.setState({ open: false })
+  handleOpen = () => this.setState({ open: true })
 
   render() {
     const { open } = this.state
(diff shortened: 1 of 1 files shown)


layershifter:
Actually, the issue comes from EventStack that incorrectly works with browser frames (click on `button` should not be fired actually).

There is an experiment with new approach: https://codesandbox.io/s/9yw3vq6774 I need to spend more time on it to test other scenarios and perform cleanup.

I will update the example later today.

Fabianopb:
So would that mean that the fix should be done in EventStack in the end?

layershifter:
> So would that mean that the fix should be done in EventStack in the end?

@Fabianopb Actually, yes. But, I'm not ready to ship a _correct_ fix for right now. I need to resolve the issue correctly and forever 👍

layershifter (approved):

layershifter:
@Fabianopb I've updated an example, now it works more correctly. Please let me know if it is fine 🚶

Fabianopb:
Now the external button is solely used to open the `Portal`, which can be closed via the internal button or by clicking outside. Sounds fair too!

levithomason:

PR #3422, opened 2019-02-08
fix(Dropdown): retain focus on the input if the Dropdown receives a click

### LINKED ISSUE #3346: Multiple select search dropdown input field is too small
### Steps
1. In a multiple select search dropdown with no values, click inside the dropdown once to open it
2. Click a second time somewhere on the very right inside the dropdown. Notice how your cursor still indicates as if you're hovering over an input field.
3. Try to type a search query

### Expected Result
The focus should be on the input, so you can type your search query

### Actual Result
The focus is not on the input, meaning you can't type. Instead you have to click the most left part to get focus.

### Version
0.84.0

### Testcase
https://codesandbox.io/s/mjnyx49vj9

The input field should probably fill the whole width of the dropdown or you should be able to click anywhere in the width to regain focus on the input field.

Say your dropdown looks like this:
Image of test case

now if you click anywhere right of the small input field, you will lose focus off the input and won't be able to type. This situation is very easy to get with such fluid dropdowns.  If you know about this issue, you know to simply click within the input field again, but everyone else will not understand this.

This issue is not really present in the non-react version. There the input field is still small, but you can click along the entire width to regain focus to type.

jongsue:
fix #3346

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -691,7 +691,10 @@ export default class Dropdown extends Component {
     e.stopPropagation()
 
     if (!search) return this.toggle(e)
-    if (open) return
+    if (open) {
+      if (this.searchRef) this.searchRef.focus()
+      return
+    }
     if (searchQuery.length >= minCharacters || minCharacters === 1) {
       this.open(e)
       return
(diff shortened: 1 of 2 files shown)


levithomason:

PR #3425, opened 2019-02-10
fix(Checkbox): disable click on mouse right/middle button

### LINKED ISSUE #3419: Checkbox: become checked even when clicked using the right/middle mouse button
### Steps
Click any checkbox on https://react.semantic-ui.com/modules/checkbox/ with the right or middle mouse button

### Expected Result
Nothing (except the browser's default action for that mouse button, i.e. opening the context menu / autoscrolling)

### Actual Result
The checkbox was toggled

---

This is quite annoying as someone who uses both middle-button autoscrolling and right-button-based mouse gestures a lot: I do not expect either of these actions to trigger actions on websites, and for the standard checkboxes in both the browser and the operating system (I think it's consistent across windows, linux, probably macos as well) only the left mouse button toggles the checkbox.

harish-a:
Fixes #3419.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -181,6 +181,10 @@ export default class Checkbox extends Component {
 
   handleMouseUp = (e) => {
     debug('handleMouseUp()')
+
+    // Right and middle mouse button check disabled.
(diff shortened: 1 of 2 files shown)


levithomason:

PR #3428, opened 2019-02-12
chore(Dropdown): use EventStack component

layershifter:
This PR continues work to use `EventStack` as a component.
The goal: to have declarative API there an simplify class methods.


layershifter:
@Fabianopb as you spent some effort on `Dropdown` previously, can you please check these changes? 🙏

levithomason:

PR #3430, opened 2019-02-12
fix(Dropdown): sets focus to the search input after selection

### LINKED ISSUE #3349: Dropdown: focus is lost after selection
Drop-down looses focus when I select an item, can anyone please tell me how should I maintain focus.

jongsue:
Addresses #3349.

focus-after-selection


Fabianopb:
Next time just take a look if someone else has already committed to start working on the issue...

layershifter:
I need to merge #3428 before.
Tried the similar fix today, but there were issues with broken key handlers.

layershifter:
I added some small changes. Actually, this PR don't fix #3349 because the focus will be still lost on usual `<Dropdown selection />`. Going to merge this PR because it fixes the issue for `<Dropdown search selection />`.

levithomason:

Issue #3003, opened 2018-07-12
Dropdown Text Prop


ecnaidar:
## Bug Report
Dropdown allows passing `text` prop that will be later rendered inside of your dropdown component.
In reality any react node is working because it is rendered as `{text}` internally but PropType checking errors since in expects `string`. 
It might be reasonable to relax PropTypes constraints since it working anyways but it will be slightly stupid because `text` name implies passing a string as opposed to `content`.

## Why am I even doing that if there is a `label` prop?
`label` prop overrides dropdown arrow icon and I want to have a clear indication that there is dropdown available and another icon at the same time:

<img width="140" alt="screen shot 2018-07-12 at 13 01 52" src="https://user-images.githubusercontent.com/10078339/42627338-64b94110-85d5-11e8-8672-7ddfce7a4fc3.png"> 

### Version
0.81.1

### Testcase
https://codesandbox.io/s/1v80o4ykr3

levithomason:
https://react.semantic-ui.com/modules/dropdown#usage-trigger

![image](https://user-images.githubusercontent.com/5067638/42655348-b43e81c8-85d0-11e8-8d90-50b632ab6067.png)

ecnaidar:
@levithomason Hoped I won't be that person who can't read. Thanks for link and sorry for disturbance.

levithomason:
No worries, there are far too many examples on that page.  This isn't the first issue asking about this feature for `text` either :)

We need an improved discovery process in the docs.   Especially for large components, like Dropdown.  Users should be able to easily find what they need and right now, it is tough.

ecnaidar:
@levithomason Depending on how documentation is generated potentially cheap improvement might be having a link next to prop that directs to example showcasing prop usage and the other way around.

PR #3431, opened 2019-02-12
refactor(Dropdown): allow "object" as a valid PropTypes for "text"

ghost:
Hello,
I'm using [react-intl](https://github.com/yahoo/react-intl) and it would be great to have the possibility to use it as an object instead of as a string. Actually, it's working but i have an error in my console saying me : 
> prop `text` of type `object` supplied to `Dropdown`, expected `string`

Here is an example :
```
<Dropdown button text={<FormattedMessage {...text} />}>
  <Dropdown.Menu>
    [...]
  </Dropdown.Menu>
</Dropdown>
```

### DIFF
--- a/src/modules/Dropdown/Dropdown.d.ts
+++ b/src/modules/Dropdown/Dropdown.d.ts
@@ -270,7 +270,7 @@ export interface StrictDropdownProps {
   tabIndex?: number | string
 
   /** The text displayed in the dropdown, usually for the active item. */
-  text?: string
+  text?: string | object
 
   /** Custom element to trigger the menu to become visible. Takes place of 'text'. */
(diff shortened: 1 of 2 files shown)


layershifter:
https://github.com/Semantic-Org/Semantic-UI-React/issues/3003#issuecomment-404647591
https://github.com/Semantic-Org/Semantic-UI-React/issues/3003#issuecomment-404627408

You're looking for a `trigger` prop 😺

layershifter:
Currently I'm working on some cleanups in `Dropdown`, need to check `text`/`trigger`, may be `text` can be safely removed to avoid confusions.

ghost:
Thanks, it works like a charm with `trigger`. I missed it

PR #3435, opened 2019-02-13
fix(Checkbox): prevent click propagation from the input element

### LINKED ISSUE #3433: Checkbox: onClick fires parent onClick twice
## Bug Report
I just upgraded to v0.85 (from 0.82.5) seeing this release is supposed to fix this issue, https://github.com/Semantic-Org/Semantic-UI-React/pull/3351. My issue might be slightly different though.

I have the following subset of code:
```
         <li
            onClick={this.liClickHandler}
          >
            <div className="border--left" />
            <div className="infoIconContainer">
              <Popup trigger={<Icon name="info circle" />} content={node.data.tooltipText} inverted />
            </div>
            <div className="checkboxContainer">
              <Checkbox
                id={node.id}
                checked={node.data.active}
                indeterminate={node.data.indeterminate}
                radio={treeType === 'Radio'}
                onClick={e => e.stopPropagation()}
              />...
```

When the user clicks the checkbox, which is a radio button, it runs the `li`'s `onClick` event twice. If the user clicks the `li` it runs once. 

It appears to only do this when I specify the `id` property. This can be validated in the test case below
 
### Steps
Use the checkbox with an ID specified and a parent click event and note it fires twice. It fires twice for radio and checkboxes.

Put onclick on the checkbox and it only fires once.

### Expected Result
it fires once

### Actual Result
it fires twice. In debugging the `this.liClickHandler`, I notice that the `event` object is slightly different the 2nd time it calls.

First time, `event.target` is `label`, the second time it's the `input`. Appears both controls are pushing the click.

### Version
    "semantic-ui-css": "^2.4.1",
    "semantic-ui-react": "^0.85.0",

### Testcase
https://codesandbox.io/s/88zzv64rn8

### LINKED ISSUE #3440: Checkbox: no longer prevents default after updating to version 0.85.0
## Bug Report
Hi, we are using a Card component with a toggle Checkbox in its header. The Card component is also used as a link (from react-router-dom), so we need to prevent the link from firing when we click the toggle. Everything was working fine until we updated to version 0.85.0 from 0.83.0.

### Steps
Use a card component (as a link) with a checkbox in it and try preventing default when you click the checkbox to prevent the link from firing.

### Expected Result
Checkbox is checked without being redirected to the link.

### Actual Result
Prevent default doesn't work anymore so you are redirected to the link.

### Version
"semantic-ui": "^2.4.2"
"semantic-ui-react": "^0.85.0"

### Testcase
I created an example without a link which uses stopPropagation instead but the described behavior can still be observed if switched between the two versions. I wonder if this could be caused by this PR: https://github.com/Semantic-Org/Semantic-UI-React/pull/3351

https://codesandbox.io/s/0oxky24lyp

EDIT: Here is an example with a link: https://codesandbox.io/s/2zxzr1jw5p

Thanks for any help!

Fabianopb:
Initially the click was being fired from both `label` and `input`
checkbox-antes

Now it only comes from `label`
checkbox-depois

(Closes #3433)
(Closes #3440)

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -1,7 +1,7 @@
 import cx from 'classnames'
 import _ from 'lodash'
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { createRef } from 'react'
 
 import {
   AutoControlledComponent as Component,
@@ -117,6 +117,9 @@ export default class Checkbox extends Component {
 
   static autoControlledProps = ['checked', 'indeterminate']
 
+  inputRef = createRef()
+  labelRef = createRef()
+
   componentDidMount() {
     this.setIndeterminate()
   }
@@ -139,29 +142,50 @@ export default class Checkbox extends Component {
     return disabled ? -1 : 0
   }
 
-  handleInputRef = c => (this.inputRef = c)
-
-  handleChange = (e, fromMouseUp) => {
-    debug('handleChange()')
+  handleClick = (e) => {
+    debug('handleClick()', _.get(e, 'target.tagName'))
     const { id } = this.props
     const { checked, indeterminate } = this.state
 
-    if (!this.canToggle()) return
-    if (fromMouseUp && !_.isNil(id)) return
+    const hasId = !_.isNil(id)
+    const isLabelClick = e.target === this.labelRef.current
+    const isLabelClickAndForwardedToInput = isLabelClick && hasId
 
-    // We don't have a separate click handler as it's already called in here,
-    // and also to avoid duplicate calls, matching all DOM Checkbox comparisons.
-    _.invoke(this.props, 'onClick', e, {
-      ...this.props,
-      checked: !checked,
-      indeterminate: !!indeterminate,
-    })
+    // https://github.com/Semantic-Org/Semantic-UI-React/pull/3351
+    if (!isLabelClickAndForwardedToInput) {
+      _.invoke(this.props, 'onClick', e, {
+        ...this.props,
+        checked: !checked,
+        indeterminate: !!indeterminate,
+      })
+    }
+
+    if (this.isClickFromMouse) {
+      this.isClickFromMouse = false
+
+      if (isLabelClick && !hasId) {
+        this.handleChange(e)
+      }
+
+      if (hasId) {
+        // To prevent two clicks from being fired from the component we have to stop the propagation
+        // from the "input" click: https://github.com/Semantic-Org/Semantic-UI-React/issues/3433
+        e.stopPropagation()
+      }
+    }
+  }
+
+  handleChange = (e) => {
+    const { checked } = this.state
+
+    if (!this.canToggle()) return
+    debug('handleChange()', _.get(e, 'target.tagName'))
+
     _.invoke(this.props, 'onChange', e, {
       ...this.props,
       checked: !checked,
       indeterminate: false,
     })
-
     this.trySetState({ checked: !checked, indeterminate: false })
   }
 
@@ -174,8 +198,10 @@ export default class Checkbox extends Component {
       checked: !!checked,
       indeterminate: !!indeterminate,
     })
-    _.invoke(this.inputRef, 'focus')
 
+    _.invoke(this.inputRef.current, 'focus')
+    // Heads up!
+    // We need to call "preventDefault" to keep element focused.
     e.preventDefault()
   }
 
@@ -183,15 +209,12 @@ export default class Checkbox extends Component {
     debug('handleMouseUp()')
     const { checked, indeterminate } = this.state
 
+    this.isClickFromMouse = true
     _.invoke(this.props, 'onMouseUp', e, {
       ...this.props,
       checked: !!checked,
       indeterminate: !!indeterminate,
     })
-
-    // Handle mouseUp only on the left mouse button.
-    // https://github.com/Semantic-Org/Semantic-UI-React/issues/3419
-    if (e.button === 0) this.handleChange(e, true)
   }
 
   // Note: You can't directly set the indeterminate prop on the input, so we
@@ -200,13 +223,7 @@ export default class Checkbox extends Component {
   setIndeterminate = () => {
     const { indeterminate } = this.state
 
-    if (this.inputRef) this.inputRef.indeterminate = !!indeterminate
-  }
-
-  handleInputClick = (e) => {
-    e.stopPropagation()
-    // To prevent two clicks from being fired from the component we have to stop the propagation
-    // from the "input" click: https://github.com/Semantic-Org/Semantic-UI-React/issues/3433
+    if (this.inputRef.current) this.inputRef.current.indeterminate = !!indeterminate
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


layershifter:
Actually we should check:
- only one `click` was propagated with `id` on `label`
- only one `click` was propagated with `id` on `input`
- only one `click` was propagated without `id` on `input`
- only one `click` was propagated without `id` on `label`

Fabianopb:
I can check that later, but just to confirm, the `id` goes directly on the `Checkbox`, right? And then it's passed to the `input` and `label` within the component.

So if we're testing
```jsx
mount(
  <div role='presentation' onClick={onClick}>
    <Checkbox id='foo' />
  </div>,
)
```

We could also test
```jsx
mount(
  <div role='presentation' onClick={onClick}>
    <Checkbox />
  </div>,
)
```

Then it's all or nothing with the id, right?

layershifter:
> the `id` goes directly on the `Checkbox`, right?

Absolutely

```jsx
const wrapperWithId = mount(
  <div role='presentation' onClick={onClick}>
    <Checkbox id='foo' />
  </div>,
)
const wrapper = mount(
  <div role='presentation' onClick={onClick}>
    <Checkbox />
  </div>,
)

wrapperWithId.find('input').click() // 1
wrapperWithId.find('label').click() // 2
wrapper.find('input').click() // 3
wrapper.find('label').click() // 4
```

Fabianopb:
Oh that's right, we set the `id` in the `Checkbox`, but then we can simulate the click both on the `input` and on the `label`. I'll check that later!

Fabianopb:
Tests updated @layershifter !
`inputs` don't propagate click because there's *always* a `label` on top of them, so the click event is always propagated from `label`.

test/specs/modules/Checkbox/Checkbox-test.js
```diff
@@ -201,6 +201,50 @@ describe('Checkbox', () => {
...
+    assertions.forEach((assertion) => {
+      it(assertion.description, () => {
+        const onClick = sandbox.spy()
+        wrapperMount(
+          <div role='presentation' onClick={onClick}>
+            <Checkbox id={assertion.id} />
+          </div>,
+        )
+        const target = document.querySelector(`.ui.checkbox ${assertion.target}`)
+        domEvent.click(target)
```
layershifter:
```diff
-const target = document.querySelector(`.ui.checkbox ${assertion.target}`)
-domEvent.click(target)
+domEvent.click(`.ui.checkbox ${assertion.target}`)

This should be a valid change 😺

File: test/specs/modules/Checkbox/Checkbox-test.js [hunk 2]
```diff
@@ -201,6 +201,50 @@ describe('Checkbox', () => {
...
+        if (assertion.propagates) onClick.should.have.been.calledOnce()
+        else onClick.should.not.have.been.called()
```
layershifter:
```diff
-if (assertion.propagates) onClick.should.have.been.calledOnce()
-else onClick.should.not.have.been.called()
+onClick.should.have.callCount(Number(assertion.propagates))
```

We can test the call count directly, https://github.com/domenic/sinon-chai

test/specs/modules/Checkbox/Checkbox-test.js
```diff
@@ -201,6 +201,50 @@ describe('Checkbox', () => {
     })
   })
 
+  describe('click propagation', () => {
+    const assertions = [
```
layershifter:
```suggestion
    const assertMatrix= [
```

src/modules/Checkbox/Checkbox.js
```diff
@@ -258,6 +258,7 @@ export default class Checkbox extends Component {
           tabIndex={this.computeTabIndex()}
           type={type}
           value={value}
+          onClick={e => e.stopPropagation()}
```
layershifter: Can we create a `handleInputClick` method and make a comment there with a link to the matching issue?

layershifter: And sort props in alphabetic order ❤️

layershifter:
@Fabianopb I made some comments, can you please address them? Nice job 👍

Fabianopb:
Awesome @layershifter thanks for the comments, I'll try address them later today!

src/modules/Checkbox/Checkbox.js
Fabianopb: Done!

test/specs/modules/Checkbox/Checkbox-test.js
Fabianopb: Sure, I'll change that

test/specs/modules/Checkbox/Checkbox-test.js
Fabianopb: Good to know!

test/specs/modules/Checkbox/Checkbox-test.js
Fabianopb: Nice idea!

Fabianopb (commented):
@layershifter I just pushed the changes!

layershifter:
@Fabianopb I added changes that I picked from @levithomason branch 😸 Can you please help to test these changes? Did I miss something? I.e. double clicks, event propagation, right mouse click?

test/specs/modules/Checkbox/Checkbox-test.js
```diff
@@ -424,40 +387,77 @@ describe('Checkbox', () => {
...
+        },
+        target: 'label',
+      },
+      {
+        description: 'click on input: fires on mouse click',
+        events: {
+          label: ['mouseup', 'click'],
+          input: ['click'],
+        },
         target: 'label',
```
Fabianopb: This `target` property doesn't seem to be used anymore, is that so?

Fabianopb (commented):
@layershifter that looks good to me, just one detail I got from re-testing the behavior manually:

* without "id" the parent listens for a click coming necessarily from the `label`
* with "id" the parent listens for a click coming necessarily from the `input`

Is that ok/intended?

test/specs/modules/Checkbox/Checkbox-test.js
layershifter: Nice catch 👍

layershifter:
> Is that ok/intended?

For me it looks fine, I don't see a big issue. I've tested it again, works like a charm 💎 

Thank you!

levithomason:
@Fabianopb You've contributed here before and you've shown care and initiative for ensuring SUIR becomes better. We appreciate this and want to enable you to do more. I've added you as a collaborator. You're now able to manage issues and PRs as well as create branches here in the main repo. You no longer need to work from a fork.  Merging to master is the only reserved permission.

We are very select about who we add and when, so know that we trust your abilities and your judgment. Please feel free to weigh in on any issues and PRs that you have an opinion about. You have a share of ownership and say here now.

Once you accept the invite (see your email), you might want to show the Semantic Org badge publicly on your profile. To do that, navigate to the "people" page and change the "private" dropdown to "public" for your user in the table. Here's a [direct link](https://github.com/orgs/Semantic-Org/people?utf8=%E2%9C%93&query=Fabianopb) to your user in the table.

/cc
@Semantic-Org/react-collaborators
@Semantic-Org/react-maintainers

Please welcome @Fabianopb to our ranks 😄 🎉

levithomason:

Issue #3418, opened 2019-02-07
TextArea with autoHeight doesn't shrink when lines deleted


mashabow:
### Step

1. Input multiple lines in a TextArea with `autoHeight`  // → TextArea's height grows to fit the lines
2. Delete some lines.

### Expected Result

TextArea's height shrinks to fit the remaining lines.

### Actual Result

TextArea's height doesn't change on deleting.

![feb-07-2019 18-40-45](https://user-images.githubusercontent.com/6268183/52402673-3e82c100-2b08-11e9-9ce6-5fda376fbcc8.gif)

### Version
0.85.0
Does not occcur in 0.84.0.

### Testcase
**https://codesandbox.io/s/qzn3nzwr1j 0.85.0**
https://codesandbox.io/s/104r4oy993 0.84.0 (correct)

layershifter:
The issue comes from #3343.

PR #3436, opened 2019-02-13
fix(TextArea): Make TextArea's height shrink (#3418)

codrex:
This fixes the issue reported [here](https://github.com/Semantic-Org/Semantic-UI-React/issues/3418)

## Before
bug

## After
fix

### DIFF
--- a/src/addons/TextArea/TextArea.js
+++ b/src/addons/TextArea/TextArea.js
@@ -45,6 +45,8 @@ class TextArea extends Component {
     rows: 3,
   }
 
+  initialHeight = null
+
   componentDidMount() {
     this.updateHeight()
   }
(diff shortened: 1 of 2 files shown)


layershifter:
@codrex thank you for the PR, checked in Chrome.

### Zoom 100%

Works fine.

zoom-100

### Zoom 90%

Appears a scroll bar.

zoom-90

codrex:
@layershifter You welcome. I will check this out.

codrex:
fix2

@layershifter updates have been made to the PR. A scroll bar still appears and disappears at a given size. Going through the current implementation I discovered the same behavior.

layershifter:
@codrex thank you for your effort there, I really appreciate it 👍

However, I opened #3465 with the detailed description why I want to do that changes. Sorry for the late response, I wanted to hear the opinion from the team that uses SUIR very long time.

PR #3438, opened 2019-02-14
feat(Transition): allow custom animations

### LINKED ISSUE #3323: Transition: allow custom animations
## Feature Request
Semantic originally allows custom transitions to be created., but the React `Transition` component only allow the original transitions (especially for directional animations).

### Problem description
There is no way to define a custom transition animation on my CSS and use it as a directional transition on my React components.

### Proposed solution
Add `SUI.pushAnimation("jump")` and `SUI.pushDirectionalAnimation("reveal left")` allowing them to be used while defining a Transition.

The only problem with this solution might be the PropTypes check, which relies on a static array, which would have to change to a dynamic function to get the updated value.

### MVP
[Fork, update, and replace this pen if you can show the proposed solution]:
https://codesandbox.io/s/2l3n74j9y

jongsue:
Fixes #3323.

### DIFF
--- a/src/modules/Transition/Transition.d.ts
+++ b/src/modules/Transition/Transition.d.ts
@@ -16,7 +16,7 @@ export interface StrictTransitionProps {
   /** Primary content. */
   children?: React.ReactNode
 
-  /** Whether it is directional animation event or not. */
+  /** Whether it is directional animation event or not. Use it only for custom transitions. */
   directional?: boolean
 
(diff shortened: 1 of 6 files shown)


layershifter:
Added small styling changes, LGTM! Good job 👍

jongsue:
Thank you for review @layershifter 👍

levithomason:

PR #3444, opened 2019-02-17
fix(Input): update supported values for the `size` prop

### LINKED ISSUE #3439: Input: with size "tiny" is same as size "medium"
## Bug Report
The support for the "tiny" size for the Input element seems to be missing.

### Steps
<Input size="tiny" placeholder="foo" />

### Expected Result
I would expect the size "tiny" to be bigger than "mini" and smaller than "medium".

### Actual Result
Instead "tiny" is the same as "medium". I believe the css is missing for this size.

### Version
0.85.0

### Testcase
Even your own test case in the documentation is missing the "tiny" size:
https://react.semantic-ui.com/elements/input/#variations-size

layershifter:
Fixes #3439.

This PR updates supported values for the `size` prop, see [`input.css`](https://github.com/Semantic-Org/UI-Input/blob/master/input.css#L487).

### DIFF
--- a/src/elements/Input/Input.d.ts
+++ b/src/elements/Input/Input.d.ts
@@ -65,7 +65,7 @@ export interface StrictInputProps {
   onChange?: (event: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => void
 
   /** An Input can vary in size. */
-  size?: SemanticSIZES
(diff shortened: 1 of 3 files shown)


levithomason:

woozyking:
Would this be a breaking change, since previously there was `medium` size supported?

This came up only when we ran some snapshot testing on our data table component, as seen https://travis-ci.com/EQWorks/extended-semantic-components/builds/105847120

PR #3445, opened 2019-02-17
fix(Dropdown): correctly handle `focus` value in state

layershifter:
Fixes a regression introduced in #3428.

---

### Regression

The value of `focus` in a state wasn't handled correctly previously.

regress

### After 

rec1
rec2


levithomason:

PR #3448, opened 2019-02-20
feat(Sticky): accept refs as value for `context` and `scrollContext` props

layershifter:
This PR makes following changes:
- `Sticky` now uses `createRef()` API internally ⭐️ 
- `Sticky` now accepts a refs created with `React.createRef()` into `context` and `scrollContext`, it's really great simplification 👍


levithomason:

PR #3449, opened 2019-02-22
feat(MountNode): support refs as value for `node` prop

layershifter:
Changes similar to #3448. So, `MountNode` now accepts a refs created with `React.createRef()` into `node` prop, it's really great simplification 👍


levithomason:

PR #3450, opened 2019-02-22
docs(Modal): fix Modal.Description description

hollg:
Change description of Modal.Description from "A modal can have a header" to "A modal can contain a description with one or more paragraphs", which is consistent with Card.Description's docs


levithomason:

PR #3451, opened 2019-02-22
feat(Sidebar): support object refs as value for `target` prop

layershifter:
Changes similar to #3448.

So, `Sidebar` now accepts a refs created with `React.createRef()` into `target` prop, it's really great simplification 👍


levithomason:

PR #3452, opened 2019-02-22
fix(Dropdown): retains focus after selection

### LINKED ISSUE #3349: Dropdown: focus is lost after selection

jongsue:
Fixes #3349.

@layershifter Could you please review this one?


src/lib/AutoControlledComponent.js
```diff
@@ -183,8 +183,9 @@ export default class AutoControlledComponent extends Component {
    * Second argument is a state object that is always passed to setState.
    * @param {object} maybeState State that corresponds to controlled props.
    * @param {object} [state] Actual state, useful when you also need to setState.
+   * @param {function} [callback] a setState callback
    */
-  trySetState = (maybeState, state) => {
+  trySetState = (maybeState, state, callback = () => {}) => {
```
levithomason: Thanks for this, I'd been wanting to update this to match React's API.  There are only a few cases where we use the `state` argument as well.  I'm thinking we should also remove it to match the React API 1:1.  Otherwise, to use the callback you need to pass some null `state`:
```jsx
this.trySetState(
  { count: 1 },             // maybeState
  null,                     // state, I don't need...
  () => console.log('done') // complete callback
)
```

The current `trySetState` implementation throws warnings about including state keys that are not listed in `autoControlledProps`, however, it could instead ignore them and include them as-is in the actual `setState` call.  This way we can just combine the maybe state and the actual state.

src/modules/Dropdown/Dropdown.js
```diff
@@ -470,7 +470,7 @@ export default class Dropdown extends Component {
       this.scrollSelectedItemIntoView()
     } else if (prevState.open && !this.state.open) {
       debug('dropdown closed')
-      this.handleClose()
+      // this.handleClose()
```
levithomason: cruft?

src/modules/Dropdown/Dropdown.js
```diff
@@ -489,7 +489,7 @@ export default class Dropdown extends Component {
     const { closeOnChange, multiple } = this.props
     const shouldClose = _.isUndefined(closeOnChange) ? !multiple : closeOnChange
 
-    if (shouldClose) this.close(e)
+    if (shouldClose) this.close(e, () => {})
```
levithomason: cruft?

src/modules/Dropdown/Dropdown.js
```diff
@@ -1127,13 +1131,13 @@ export default class Dropdown extends Component {
...
 
-  close = (e) => {
+  close = (e, callback = this.handleClose) => {
     const { open } = this.state
     debug('close()', { open })
 
     if (open) {
       _.invoke(this.props, 'onClose', e, this.props)
-      this.trySetState({ open: false })
+      this.trySetState({ open: false }, undefined, callback)
```
levithomason: Case-in-point 😉

levithomason:
After making a selection in a `search` selection, the input DOM node retains focus but the Dropdown loses its blue border.  I'm pretty sure this means `this.state.focus` is not true.

**After making a selection:**
image

Note the missing blue border, as there is when first focusing the Dropdown:
image

levithomason:
⚠️ Heads Up!

I also merged master and fixed conflicts here.  You'll need to pull before resuming work.

jongsue:
> After making a selection in a `search` selection, the input DOM node retains focus but the Dropdown loses its blue border. I'm pretty sure this means `this.state.focus` is not true.
> 
> **After making a selection:**
> image
> 
> Note the missing blue border, as there is when first focusing the Dropdown:
> image

semantic.css ` .ui.selection.dropdown:focus`  seems to make the blue border.

스크린샷 2019-03-14 오전 1 00 31

As you mentioned After making a selection in a search selection, the input DOM node retains focus **not Dropdown**.
That's why there is no blue border.
I don't know how to solve this problem. do you have any idea?

src/lib/AutoControlledComponent.js
jongsue: I tried to remove the `state` arguments out of  `trySetState`.
I'm not sure I did understand what you said.
please review this code again

src/modules/Dropdown/Dropdown.js
jongsue: oops.
I deleted it

src/modules/Dropdown/Dropdown.js
jongsue: This is on purpose.
`callback` the 2nd argument in `close` method has default value.
But not using default value in this case.

```
close = (e, callback = this.handleClose) => {
    const { open } = this.state
    debug('close()', { open })

    if (open) {
      _.invoke(this.props, 'onClose', e, this.props)
      this.trySetState({ open: false }, callback)
    }
  }
```

jongsue:
I tried adding ‘active’ class on Dropdown when ‘focus’ state is true.
Then we could get the blue border but there is ‘border-bottom-left-radius’ to make it look like not right

amrita-syn:
Any update on this? Would save a lot of effort...

jongsue:
@levithomason could you review this again please ?
The blue border is css issues. 
If Semantic UI would fix the problem then It'll fix ours too.

layershifter:
Yep, that blue border is a CSS issue, want to check this additionally. For the late feedback 🐑

layershifter (approved):
Tested it manually, works perfectly 👍

PR #3453, opened 2019-02-22
chore(Input|TextArea): use createRef() API

layershifter:
This PR continues adoption of `createRef()`, no user-facing changes.


levithomason:

PR #3454, opened 2019-02-22
BREAKING(Portal): use createRef() API internally

layershifter:
This PR continues adoption of `createRef()`.

----

# BREAKING CHANGES

Will affect you **only** if you used the `PortalInner` component directly before.

`PortalInner` returned `node` previously in `onMount`/`onUnmount` callbacks. Now you should use `innerRef`.

#### Before

```js
<PortalInner onMount={(e, data) => console.log(data.node)} />
```

#### After

```js
const innerRef = React.createRef()

<PortalInner innerRef={innerRef} />
console.log(innerRef.current)
```


levithomason:

PR #3456, opened 2019-02-22
fix(Search): prevent blur event when SearchResult is clicked

### LINKED ISSUE #3298: Search: does not trigger `onResultSelect` if custom open prop is supplied
### Steps
Add a search component with a managed state for "open", and custom functions for onFocus, onBlur and onResultSelect. 

### Expected Result
Results appear when ```open = true```. ```value``` gets set to ```result.value``` when a result is either clicked or selected with arrow-keys and enter.

### Actual Result
clicking a result does not trigger "onResultSelect", hence value is not updated. 

When commenting out "open" it works as expected.

### Version
0.83.0

### Testcase
https://codepen.io/j_sve/pen/Rqybdd

jongsue:
fixes #3298 

I couldn't write the test code.

If you know how to do it, please let me know
thanks in advance


src/modules/Search/Search.js
```diff
@@ -552,6 +552,7 @@ export default class Search extends Component {
         key={childKey || result.title}
         active={selectedIndex === offsetIndex}
         onClick={this.handleItemClick}
+        onMouseDown={e => e.preventDefault()}
```
layershifter: 1. Please move this code to a separate class method and place it under `handleItemClick`:

```jsx
  handleItemMouseDown = (e) => {
    debug('handleItemMouseDown()')
    // We should prevent default to prevent blur events.
    // https://github.com/Semantic-Org/Semantic-UI-React/issues/3298
    e.preventDefault()
  }
```

2. As we there, let's use `_.invoke()`:

```jsx
  handleFocus = (e) => {
    debug('handleFocus()')

    _.invoke(this.props, 'onFocus', e, this.props)
    this.setState({ focus: true })
  }

  handleBlur = (e) => {
    debug('handleBlur()')

    _.invoke(this.props, 'onBlur', e, this.props)
    this.setState({ focus: false })
  }
```

```diff
- onMouseDown={e => e.preventDefault()}
+ onMouseDown={this.handleItemMouseDown}
```

3. About a new unit test, we want to assert that `onBlur` will not happen on the item selection, we can test this:

```jsx
    it('is not called on an item click', () => {
      const onBlur = sandbox.spy()
      wrapperMount(<Search results={options} onBlur={onBlur} />)

      openSearchResults()
      wrapper.find('SearchResult').at('0').simulate('click', nativeEvent)
      onBlur.should.have.not.been.called()
    })
```

layershifter (changes requested):
Needs some changes, thanks for taking this 👍

layershifter (approved):

layershifter:
Applied changes to get this fix 👍

levithomason:

PR #3459, opened 2019-02-23
feat(Popup): support an object from `createRef()` as context value

layershifter:
Changes similar to #3448.

So, `Popup` now accepts a refs created with `React.createRef()` into `context` prop, it's really great simplification 👍


levithomason:

Issue #1925, opened 2017-08-03
IE: TextArea: wrong autoHeight counting


slunecniborecek:
Based on https://github.com/Semantic-Org/Semantic-UI-React/issues/1843 which was closed. Unfortunately the very last commit contains an issue in IE browsers

**steps 1:**
copy a long text
paste using context menu

**steps 2:**
make text area to grow some height
select all text in it
using context menu, click 'delete' or 'cut'

**Expected:**
textarea is resized properly in both steps1 and steps2

**Actual:**
the height is not changed

slunecniborecek:
@levithomason ^

levithomason:
Thanks for the follow-up.  Looks like we'll need to use the `onInput` method for IE.  When this is done, we'll need to also update the callback signature to follow our conventions.  See the `onChange` propType description and handler for that pattern.

nickpedersen:
Could I help out with this one? I can prepare a PR right now

Issue #1478, opened 2017-03-19
TextArea autoHeight initialState with SSR


lokhmakov:
With server side render (SSR) TextArea does not set init styles and showed full sized before scripts started.

[http://imgur.com/a/IgjYt](url)

I use some hack:

```js
<TextArea
  style={ description === `` ? {  minHeight: 0, resize: `none`, height: `41px` } : {} }
  autoHeight
/>
```

But think we can do it inside `TextArea`.

levithomason:
Thanks @lokhmakov.

We have to wait for one render in order to measure the textarea and size it.  We should likely apply the minimum height style on the first render to prevent this issue.  We may need to remove the style when making the measurement to ensure the content can size the textarea.

lokhmakov:
If we cannot determine initial height then we can set initial rows, like this:

```js
  rows={ value === `` ? 1 : null }
```

levithomason:
Closing for housekeeping as there isn't much activity over here.  We're open to a PR on this issue.

PR #3465, opened 2019-02-26
breaking(TextArea): remove `autoHeight` prop

### LINKED ISSUE #3418: TextArea with autoHeight doesn't shrink when lines deleted

layershifter:
Fixes #3418.

# BREAKING CHANGES

This PR removes the `autoHeight` prop from the `TextArea` component.

### Motivation

We had numerous issues with `TextArea` component and `autoHeight` prop:
- #3418
- #3343
- #2173
- #1925
- #1843
- #1478
- #1405
- https://github.com/Semantic-Org/Semantic-UI-React/issues/982#issuecomment-266915712

I had a conversation with team from Ukraine that continuously uses SUIR and they said that `autoHeight` is buggy and they are using [`react-textarea-autosize`](https://www.npmjs.com/package/react-textarea-autosize), too 😼 

So, as @levithomason originally proposed I will suggest to use [`react-textarea-autosize`](https://www.npmjs.com/package/react-textarea-autosize):
>I would encourage users to use that module if it suits their needs. SUI requires no special classNames nor markup to style the `textarea` when used in a `Form`, so it will work OOTB.

### Before

```jsx
import { Form, TextArea } from 'semantic-ui-react'

//

<Form.TextArea autoHeight />
<TextArea autoHeight />
```

### After

```jsx
import { Form, TextArea } from 'semantic-ui-react'
import TextareaAutosize from 'react-textarea-autosize'

//

<Form.TextArea control={TextareaAutosize} />
<TextareaAutosize  />
```

I also created a Codesandbox example with how-tos, https://codesandbox.io/s/1v67906ll4. It will be available on the [Prototypes](https://react.semantic-ui.com/prototypes) page after the next release, but you can start to migrate right now.


levithomason:

TristanPouliquen:
@layershifter  Hi, I'm trying to migrate to your new implementation with TextAreaAutosize to work with forms in a Semantic Accordion.

However, I can't seem to get the textarea value in the onChange function. I looked at your Sandbox and I also get `undefined` when I log `data.value`in your example (how the value is kept in memory I do not know as you also do `value={this.state.value}` and the value is also undefined in the state I get from my console.log.

Do you have any idea as to why ? Thanks in advance !

layershifter:
> However, I can't seem to get the textarea value in the onChange function. 

You should use:

```jsx
onChange={e => this.setState({ value: e.target.value })}
```

I've update CodeSandbox to reflect this 👍

TristanPouliquen:
Thanks for your swift answer ! It works as a charm :+1:

PR #3471, opened 2019-03-01
fix(mixed): handle properly refs on third-party components

### LINKED ISSUE #1321: Ensure DOM nodes can be found for `trigger` elements and `as` components

layershifter:
Fixes #1321.

See #1190 & #1321 for context.


levithomason:

PR #3474, opened 2019-03-04
fix(Ref): update `node` if it was changed

### LINKED ISSUE #3443: Modal: unexpectedly closes when using inline `as` prop
## Bug Report
if you have the `as` prop for a Modal as an inline function, once the component re-renders at least once, it loses its ability to track clicks that are inside the modal, and the triggers it to close.

### Steps
* Create a modal component.
* use an inline function for `as`
* do something to make the component re-render (click the button)
* click anywhere inside the modal

### Expected Result
Modal stays open

### Actual Result
Modal closes

### Version
0.85.0

### Testcase
https://codesandbox.io/s/918jowm18w

layershifter:
Fixes #3443.


levithomason:

PR #3482, opened 2019-03-08
fix(Dropdown): not calls `onChange` when value is not changed on item click

### LINKED ISSUE #3437: Dropdown: calls `onChange` even when value is not changed
### Steps
1. Open https://codesandbox.io/s/jpnzzkl7pv
2. Open the first dropdown and select the "text4" option
3. Open the same dropdown again and select the "text4"

### Expected result
`onChange` was called once as value wan't changed. The native `<select />` has the same behavior.

### Actual result
`onChange` was called twice.

### Version
0.85.0

### Testcase

https://codesandbox.io/s/jpnzzkl7pv

jongsue:
fixes #3437


levithomason:

PR #3485, opened 2019-03-10
fix(lib): fully remove debug in production

layershifter:
Fixes ##3484. This PR:
- fixes issue with not fully remove debug part in CJS builds, see [lib/index.js](https://cdn.jsdelivr.net/npm/semantic-ui-react@0.85.0/dist/commonjs/lib/index.js), `var _debug = require("./debug");` was present there
- removes `debug()` function as it was never used


levithomason:

Issue #3479, opened 2019-03-07
Dropdown code sandbox not working


mattnedrich:
### Steps
1. Visit https://react.semantic-ui.com/modules/dropdown/
2. Select "Code Sandbox" for the "Selection" or "Search Selection" variants

### Expected Result
It allows you to open a code sandbox for the dropdown component

### Actual Result
Progress indicator displayed and never finishes

Fabianopb:
Just checked at the current master, bug confirmed, I'll take a look soon!

layershifter:
It's caused by some internal imports like this:
```js
import { friendOptions } from '../common'
```

This code is not uploaded to CodeSandbox 🤔

layershifter:
@Fabianopb I've updated some components in #3486, feel free to update `Search` component's docs 👍

Fabianopb:
Sounds good, I'll work on that!

Fabianopb:
#3486 and #3494 merged, should we close the issue?

layershifter:
Removed some imports in #3511 (`ComponentExampleRenderSource.js`), @Fabianopb thanks for your work there 👍

PR #3486, opened 2019-03-10
docs(Dropdown|Select|Visibility): remove imports, use faker and components

layershifter:
This PR partially fixes the old issue: all examples should be copy-pastable, so they should not contain internal imports.

Refs #3479.


levithomason:

PR #3487, opened 2019-03-10
docs(CodeSandbox): fix broken icon fonts

layershifter:
### Before

image

### After

image


docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox.js
```diff
@@ -95,8 +96,6 @@ class ComponentControlsShowCode extends React.Component {
         examplePath='/'
         example={exampleCode}
         dependencies={dependencies}
-        /* Magic trick to reload sources on passed code update */
-        key={exampleCode}
```
layershifter, line 99: We don't need this more, was fixed in `react-codesandboxer@3.1.2`

levithomason:

PR #3493, opened 2019-03-12
fix(Pagination): update handling of `onClick` for `ellipsisItem`

### LINKED ISSUE #3490: Pagination: `onClick` don't work for `ellipsisItem`
### Problem description
`Pagination.Item type="ellipsisItem"`s do not support the `onClick` handler.  This prevents me from easily using the ellipsis button as a "jump several pages" button.  An `onClick` handler can be attached to the `content` of the Pagination.Item, but this requires CSS overrides to have the entire button be clickable.  

### Proposed solutions
1. Pagination.Items fire their `onClick` handlers even if they are ellipsisItems.  
Upsides: Maximum consistency and simplicity - `onClick` simply works as normal on everything inside a Pagination.
Downsides: This is a breaking change, in that anyone depending on the previous behavior (for instance, using a common props shorthand for custom ellipsisItem and  next/prevItem) will have to go change their code to not supply that handler for ellipsisItems.  
2. Pagination.Items offer a second `onEllipsisClick` prop that is onClick for ellipsisItems only.  
Upsides: Maintains backwards compatibility, nobody has to change any code.  
Downsides: Less consistent, requires you to look at the docs to understand how to handle a click on an ellipsisItem. 

### MVP
For solution 1, replace `handleClick` in addons/Pagination/PaginationItem.js with this code:
```js
handleClick = (e) => {
  _.invoke(this.props, 'onClick', e, this.props)
}
```
For solution 2, replace it with this code:
```js
handleClick = (e) => {
  const { type } = this.props
  if (type === "ellipsisItem") {
    _.invoke(this.props, 'onEllipsisClick', e, this.props)
  } else {
    _.invoke(this.props, 'onClick', e, this.props)
  }
}
```

...update the propTypes to this:

```js
static propTypes = {
  /** A pagination item can be active. */
  active: PropTypes.bool,

  /** A pagination item can be disabled. */
  disabled: PropTypes.bool,

  /**
   * Called on click for non-ellipsisItem items.
   *
   */
  onClick: PropTypes.func,

  /**
   * Called on on click if the type is "ellipsisItem".
   *
   */
  onEllipsisClick: PropTypes.func,

  /**
   * Called on key down.
   *
   */
  onKeyDown: PropTypes.func,

  /** A pagination should have a type. */
  type: PropTypes.oneOf([
    'ellipsisItem',
    'firstItem',
    'prevItem',
    'pageItem',
    'nextItem',
    'lastItem',
  ]),
}
```
...and replicate that new proptype into the type definition for PaginationItem.  

I'll open a PR for whichever of these solutions the maintainers prefer, if they agree this is a worthwhile change :smile:

layershifter:
Fixes #3490.

Allows to correctly handle `onClick` on `ellipsisItem`:
```jsx
<Pagination
  defaultActivePage={5}
  ellipsisItem={{
    content: '...',
    onClick: (e, props) => console.log('ellipsisItem click', props),
  }}
  totalPages={10}
/>
```

### DIFF
--- a/src/addons/Pagination/Pagination.js
+++ b/src/addons/Pagination/Pagination.js
@@ -108,7 +108,7 @@ export default class Pagination extends Component {
     key: `${type}-${value}`,
     onClick: (e, itemProps) => {
       _.invoke(predefinedProps, 'onClick', e, itemProps)
-      this.handleItemClick(e, itemProps)
+      if (itemProps.type !== 'ellipsisItem') this.handleItemClick(e, itemProps)
     },
   })
 
--- a/src/addons/Pagination/PaginationItem.js
+++ b/src/addons/Pagination/PaginationItem.js
@@ -45,9 +45,7 @@ class PaginationItem extends Component {
   }
 
   handleClick = (e) => {
-    const { type } = this.props
-
-    if (type !== 'ellipsisItem') _.invoke(this.props, 'onClick', e, this.props)
+    _.invoke(this.props, 'onClick', e, this.props)
   }
 
   handleKeyDown = (e) => {
--- a/test/specs/addons/Pagination/Pagination-test.js
+++ b/test/specs/addons/Pagination/Pagination-test.js
@@ -64,5 +64,24 @@ describe('Pagination', () => {
         .simulate('click')
       onPageChange.should.have.not.been.called()
     })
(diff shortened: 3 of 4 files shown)


levithomason:

PR #3494, opened 2019-03-12
docs(Search): remove imports of other examples

Fabianopb:
As discussed in #3479 , this PR remove imports in Search examples of other examples. This creates a bit of repetition, which doesn't feel very nice, but all sandboxes work now.

I wonder if it would make sense to add some kind of general test to assure that this won't repeat in other examples... 🤔

### DIFF
--- a/docs/src/examples/modules/Search/Variations/index.js
+++ b/docs/src/examples/modules/Search/Variations/index.js
@@ -18,7 +18,7 @@ const SearchVariationsExamples = () => (
     />
     <ShorthandExample
       title='Input'
-      description='A search can be passed an input via shorthand props.'
+      description='A search can receive an input via shorthand props.'
       examplePath='modules/Search/Variations/SearchExampleInput'
     />
   </ExampleSection>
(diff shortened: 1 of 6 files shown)


levithomason:
>I wonder if it would make sense to add some kind of general test to assure that this won't repeat in other examples... 🤔

Perhaps.  We have `examples-test.js` which ensures all examples render without errors.  Wonder if we could extend that.

This may also be something we can do with lint rules.  There is an eslint import plugin.  We might be able to restrict imports via an .eslintrc in the doc examples directory.  That would be easier to maintain.

We can merge this for now 👍

docs/src/examples/modules/Search/Variations/index.js
```diff
@@ -18,7 +18,7 @@ const SearchVariationsExamples = () => (
     />
     <ShorthandExample
       title='Input'
-      description='A search can be passed an input via shorthand props.'
+      description='A search can receive an input via shorthand props.'
```
levithomason, line 21: 👍

levithomason (approved):

levithomason:

Fabianopb:
> 
> This may also be something we can do with lint rules. There is an eslint import plugin. We might be able to restrict imports via an .eslintrc in the doc examples directory. That would be easier to maintain.
> 

@levithomason I've been fiddling with this idea for a while, I can submit a PR soon!

Issue #3495, opened 2019-03-12
Typescript type definition for Tab panes is incorrect


mbakiev:
## Bug Report
The Typescript type definition for 'panes' property of Tab appears to be incorrect. Typescript definition has `content`, but JS property is `pane`. This results in a compilation error when trying to set `pane` in typescript or no content shown for tabs when using `content`.

Type definition from Tab.d.ts:
```typescript
  panes?: {
    content?: SemanticShorthandItem<TabPaneProps>;
    menuItem?: any;
    render?: () => React.ReactNode;
  }[]
```

Javascript definition from Tab.js
```javascript
panes: _propTypes.default.arrayOf(_propTypes.default.shape({
    menuItem: _lib.customPropTypes.itemShorthand,
    pane: _lib.customPropTypes.itemShorthand,
    render: _propTypes.default.func
  }))
```

### Steps
Use typescript to create a Tab using `content` in the panes object:
```typescript
const panes = [
  { menuItem: 'Tab 1', content: <Tab.Pane>Tab 1 Content</Tab.Pane> }
]
<Tab panes={panes} />
```

### Expected Result
Display the desired content under the tabs.

### Actual Result
No content displayed

### Version
v0.85.0

layershifter:
Fixed in #3496.

PR #3496, opened 2019-03-12
fix(Tab): panes type definition

mbakiev:
Type definition for Tab prop pane is incorrect.

In reference to #3495


mbakiev:
Not clear on why this change is causing coverage problems in search.

levithomason:
Thanks. The project coverage is odd but the patch coverage is correct and changes look good.

levithomason:

PR #3497, opened 2019-03-13
perf(docs): improve doc examples performance

levithomason:
Doc example pages were getting a bit sluggish.  This improves performance a bit by removing some listeners and moving styles to the CSS sheet and implementing more shouldComponentUpdate.


levithomason:

PR #3500, opened 2019-03-13
fix(Sticky): preserve item height

### LINKED ISSUE #3499: Sticky component not preserves the height of the container
## Bug Report
When the sticky is above some other static content, when the sticky item becomes sticky, the rest of the content jump with the height of the sticky element.

### Steps
```javascript
<div className="App" ref={ref}>
      <Sticky context={ref.current}>
        <Header as="h2" icon textAlign="center">
        <Icon name="users" circular />
        <Header.Content>Friends</Header.Content>
      </Header>
      </Sticky>
      {Array(10)
        .fill(null)
        .map(() => (
          <img
            alt=""
            src="https://react.semantic-ui.com/images/wireframe/paragraph.png"
          />
        ))}
    </div>
```

### Expected Result
The reset of the content won't be affected.

### Actual Result
Rest of the content should not jump at all.

### Version
0.86.0

### Testcase
Here you can see that the images below jump when the sticky element become sticky.
https://codesandbox.io/s/8xnm9j0ql2

I will PR a fix for that issue.

felixmosh:
This PR fixes #3499


felixmosh:
Any update regarding this PR?

layershifter:
@felixmosh sorry for delay, thank you ❤️

PR #3503, opened 2019-03-14
docs(examples): prevent internal imports via eslint

Fabianopb:
Following up #3494 , @levithomason had the idea to prevent local imports in the docs examples via linter. Here's one solution: we have a local `.eslintrc` in the examples preventing undesired import patterns. However the rule is not valid for the `index.js` files in the same folder, so I've disabled the rule for those files.

Now the eslint will complain if there's a local import in the examples, like this:
Screen Shot 2019-03-14 at 15 29 47
Screen Shot 2019-03-14 at 15 40 27

Related to this PR, it seems that the prettier has done some additional changes to some of the `index.js` files. The docs seem to still work fine, but I wonder why that happened...

I'd be glad to get some feedback!
Cheers!


layershifter:
Instead of editing all files, I suggest to add `overrides`:
```
"overrides": [
 {
  "files": ["index.js"],
     "rules": {
        "no-restricted-imports": "off"   
     }
 }
```

Fabianopb:
Well pointed @layershifter, I didn't know you could override rules like this, thanks!

layershifter (approved):
👍

PR #3523, opened 2019-03-20
docs(FormExampleSubcomponentControl): gender neutral option

tanohzana:
I added a gender neutral option in the example as I felt some communities might be impacted by the binary choice here.


tanohzana:
There are other parts to edit, I'll make another PR.

PR #3524, opened 2019-03-20
docs(mixed): add gender other to gender selectors

tanohzana:
I added a gender neutral option in the selectors in examples as I felt some communities might be impacted by the binary choice here.

### DIFF
--- a/src/lib/SUI.js
+++ b/src/lib/SUI.js
@@ -609,6 +609,7 @@ export const GENDERS = [
   'mars stroke vertical',
   'mercury',
   'neuter',
+  'other',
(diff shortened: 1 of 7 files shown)


src/lib/SUI.js
```diff
@@ -609,6 +609,7 @@ export const GENDERS = [
   'mars stroke vertical',
   'mercury',
   'neuter',
+  'other',
```
layershifter: This array contains definitions for existing icons, please revert this change.

tanohzana: ff96fdc should fix it

PR #3528, opened 2019-03-20
fix(docs): remove component page hash replacement

### LINKED ISSUE #3527: Docs: Component documentation page replaces some URL hashes by 'null' string
## Bug Report
Component documentation page replaces some URL hashes by 'null' string for "old" hash without example path.

## First case
1. Go to https://react.semantic-ui.com
1. Select 'Elements' → 'Button' in the main (left) menu.
1. Select 'Groups' → 'Group' in the section's (right) menu.
1. Refresh the page.

## Second case
1. Go to https://react.semantic-ui.com/elements/button/#groups-group

### Expected Result
The hash is preserved by the page.

### Actual Result
The hash is replaced by the string `null`.

### Version
Current production version of the documentation.

GreenTeaCake:
Closes #3527 and affects #2732 (removes `componentWillMount` from `ComponentDoc`).

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentDoc.js
+++ b/docs/src/components/ComponentDoc/ComponentDoc.js
@@ -5,8 +5,7 @@ import { withRouteData } from 'react-static'
 import { Grid, Header, Icon } from 'semantic-ui-react'
 
 import DocsLayout from 'docs/src/components/DocsLayout'
-import { docTypes, examplePathToHash, getFormattedHash, scrollToAnchor } from 'docs/src/utils'
-import { isBrowser } from 'src/lib'
+import { docTypes, examplePathToHash, scrollToAnchor } from 'docs/src/utils'
 import ComponentDocLinks from './ComponentDocLinks'
 import ComponentDocSee from './ComponentDocSee'
 import ComponentExamples from './ComponentExamples'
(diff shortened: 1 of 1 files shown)


layershifter:
@grumblerchester thanks for pushing this. It's weird, I removed this part of code and it still works 🦀 Can you please confirm?

GreenTeaCake:
@layershifter Yeah! Deletion resolves the issue too and everything works absolutely fine for old hashes.
But I'm not sure what "hash replacement for new hashes" is supposed to do. 🤢

layershifter:
Can you please clarify what is "old" and "new" hashes?

GreenTeaCake:
@layershifter Please see [isOldHash](https://github.com/Semantic-Org/Semantic-UI-React/blob/3db04daa15eb6f8af6c1570ce1a48e753f3d25da/docs/src/utils/getFormattedHash.js#L5) that is used by [getFormattedHash](https://github.com/Semantic-Org/Semantic-UI-React/blob/3db04daa15eb6f8af6c1570ce1a48e753f3d25da/docs/src/utils/getFormattedHash.js#L23).
I have an assumption that these terms were introduced during some migration but not sure if the migration is completed.

layershifter:
I see, yep it was introduced in #2760 to fix #2737.

Actually, it was year ago and I don't see there a big issue that some of extra old links can be broken as they will still point to the correct page.

PR #3532, opened 2019-03-24
breaking(Popup): switch to Popper.js

### LINKED ISSUE #1126: Popup in Fixed Menu positioning
### Steps
1. Have a fixed menu
2. Have a popup for a menu item inside the fixed menu
3. Have a long page with scrollbar
4. Open the popup for the menu item
5. Scroll the page

### Expected Result
Popup should not move as the page scrolls

### Actual Result
Popup move as the page scrolls

### Version
0.63.5

### Testcase
http://codepen.io/anon/pen/ygyrze?editors=0010#0

### LINKED ISSUE #3226: Popup: position on window or container resize is not updated
### Steps
When window or the container element is resized, the position is not updated
resize

### Expected Result
The position of popup should always be where its supposed to be

### Actual Result
It changes the position from the actual element

### Version
latest

### Testcase
GIF file

### LINKED ISSUE #3478: Popup: doesn't reposition when changing its context
## Bug Report
I have a shared Popup that is meant to be used as a tooltip. I explicitly don't want to have a Popup associated with all items on my page that can have a tooltip because this creates a LOT of popups and hurts performance.
So my solution is to use global state via https://github.com/diegohaz/constate to share the ref of the DOM node where the Popup should be opened.
Sadly, positioning the popup via the `context` property doesn't work correctly in my use case.

### Steps
- Open https://codesandbox.io/s/vmm391vpv7
- Open the main popup
- Hover over any item
- A tooltip opens
- Hover over another item next to the first one
- The tooltip doesn't move
- Move the mouse to the border of the main popup so that the main popup stays open but the tooltip closes 
- Hover over another item in a different row or column
- The popup opens in a new position

### Expected Result
- When closing a popup and reassigning its `context`, the popup should be positioned at the updated context when it's opened the next time.
- The popup should point to the correct corner of its `context`

### Actual Result
- The popup doesn't move when being reopened. It only moves if it completely disappeared first.
- The popup doesn't point to the correct corner of its `context` DOM element

I suspect I could make it work if I could somehow wait until the popup is completely gone until reopening it again. I've tried several combinations of state logic but none has worked for me so far.
Even then it will probably still have issues with pointing at the correct corner.

### Version
0.85.0

### Testcase
https://codesandbox.io/s/vmm391vpv7

### LINKED ISSUE #3505: Popup: position not kept if trigger element changes
### Steps
1 - Render a popup with a button (better if inside a custom component)
2 - Set the `onOpen` handler to simulate the button onClick or directly onClick to add a label to that button (managing this.setState for example, hence the custom component)
3 - Click the button in order to add a label to it
4 - Close the popup (by making any other action)
5 - Open the popup again, it should be located in the upper left corner of the screen (top: 0, left: 0)

### Expected Result
The expected result was to see the Popup under or around the Trigger element

### Actual Result
Actual result: 2nd render of popup is on coordinates 0,0

### Version
v0.83.0

### Testcase
https://codesandbox.io/s/62m4v7j13w

### Guesses
I've taken a peek at the `<Portal>` element controlling the popup behaviour, as well as the `<RefXXX`> elements. Sadly I don't have the time to investigate further, but I'm guessing that since the trigger element is re-rendered somewhere along the line the reference is lost, and thus the calculated positions are (0,0)

layershifter:
# BREAKING CHANGES

This PR removes our positioning logic and uses [`react-popper`](https://www.npmjs.com/package/react-popper) instead.

## Removal of `keepInViewport`

This prop was removed and will be introduced again later.

## Removal of `horizontalOffset` & `verticalOffset`, use `offset` now

Popper.js uses more flexible `offset`, please follow their [docs](https://popper.js.org/popper-documentation.html#modifiers..offset).

```diff
<Popup
-  horizontalOffset={50}
-  verticalOffset={50}
+ offset='50, 50'
/>
```

(!) Offset in Popper works differently than `horizontalOffset` & `verticalOffset`:

> For length is intended the main axis relative to the placement of the popper.
> This means that if the placement is top or bottom, the length will be the width. In case of left or right, it will be the height.

## Changes in behavior

`Popup` changes its position more accurate now.

#### Before

popup-before

#### After

popup-after

----

### Fixes #1126.

<details>
  <summary>Before</summary>

  popup-fixed-bug
</details>

<details>
  <summary>After</summary>

  popup-fixed-fix
</details>

### Fixes #3226

<details>
  <summary>After</summary>

popup-resize-fix
</details>

### Fixes #3478

<details>
  <summary>Before</summary>

popup-reposition-before
</details>

<details>
  <summary>After</summary>

popup-reposition-after
</details>


semireg:
Is there a guide anywhere that shows how to upgrade from something like 0.84.0 with regards to Popper? The CHANGELOG says "Breaking" but offers no hints on how to fix the expected breaks. Thank you for the hard work, everyone!

HardenIan:
> Is there a guide anywhere that shows how to upgrade from something like 0.84.0 with regards to Popper? The CHANGELOG says "Breaking" but offers no hints on how to fix the expected breaks. Thank you for the hard work, everyone!

I have a feeling the "Breaking" change is the offset props being removed and the new combined offset being added.

PR #3533, opened 2019-03-25
docs(ComponentExample): replace deprecated lifecycle methods

GreenTeaCake:
Relates to #2732: Removes `componentWillMount` and `componentWillReceiveProps` from `docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js`.


layershifter:
Wow, that's great ❤️

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -64,36 +62,44 @@ class ComponentExample extends PureComponent {
...
+  static getDerivedStateFromProps(nextProps, prevState = {}) {
+    const {
+      exampleKeys,
+      examplePath,
+      exampleSources,
+      location: { hash },
+    } = nextProps
+
+    const anchorName = examplePathToHash(examplePath)
+    const isActiveHash = anchorName === getFormattedHash(exampleKeys, hash)
```
layershifter: As we don't support old hashes more, we can safely remove `getFormattedHash()` fully:

```suggestion
    const isActiveHash = anchorName === location.hash
```

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -64,36 +62,44 @@ class ComponentExample extends PureComponent {
...
-    const nextSourceCode = exampleSources[examplePath]
+  static getDerivedStateFromProps(nextProps, prevState = {}) {
+    const {
+      exampleKeys,
+      examplePath,
+      exampleSources,
+      location: { hash },
+    } = nextProps
+
+    const anchorName = examplePathToHash(examplePath)
```
layershifter: We can move `anchorName` to `constructor()` as we need it only once

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -64,36 +62,44 @@ class ComponentExample extends PureComponent {
...
-    this.setState({
-      showCode: this.isActiveHash(),
-      sourceCode: this.getOriginalSourceCode(),
-    })
-  }
-
-  componentWillReceiveProps(nextProps) {
-    const { examplePath, exampleSources, location } = nextProps
-    const nextSourceCode = exampleSources[examplePath]
+  static getDerivedStateFromProps(nextProps, prevState = {}) {
```
layershifter:
```suggestion
  static getDerivedStateFromProps(props, state) {
```

Let's use original naming proposed by React team

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -64,36 +62,44 @@ class ComponentExample extends PureComponent {
...
 
-  clearActiveState = () => {
-    this.setState({
-      showCode: false,
-      showHTML: false,
-    })
+  constructor(props) {
+    super(props)
+
+    const state = ComponentExample.getDerivedStateFromProps(props)
```
layershifter: `getDerivedStateFromProps()` will be dispatcher right after `constructor()`, we don't need this

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
layershifter: It also seems that it's a single place where we are using `exampleKeys`, so we can remove them, too 💯

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: We use computation for state here to be able to enable source code displaying and set initial value for source code (see an assignment below).

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: We can do that only in case if `examplePath` property is never changed. Current usage of `ComponentExample` implies such usage. But it is a little bit confusing from general perspective.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: Done

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: Done

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
layershifter: Based on nature of `ComponentExample` it's impossible in production and almost impossible in development, so actually we can do this :)

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
```diff
@@ -64,36 +61,43 @@ class ComponentExample extends PureComponent {
...
-    const nextSourceCode = exampleSources[examplePath]
+  static getDerivedStateFromProps(props, state = {}) {
+    const {
+      examplePath,
+      exampleSources,
+      location: { hash },
+    } = props
+
+    const anchorName = examplePathToHash(examplePath)
+    const isActiveHash = anchorName === hash
```
layershifter:
```suggestion
    const isActiveHash = anchorName === location.hash
```

Please use `location.hash`, without it's unclear what is it. It also will save us from nested destruction

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
layershifter: I understand what you mean, but I prefer separation of concerns there: if we will change something in `getDerivedStateFromProps()` it will break constructor.

Other confusion is placed around lifecycle: https://code.likeagirl.io/understanding-react-component-life-cycle-49bf4b8674de (we call `constructor()`, `getDerivedStateFromProps()` and then again `getDerivedStateFromProps()`)

```js
const originalSourceCode = props.exampleSources[props.examplePath]

this.state = {
  anchorName: examplePathToHash(props.examplePath),
  originalSourceCode,
  showCode: anchorName === props.location.hash,
  sourceCode: originalSourceCode,
}
```

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: Ok. I got it. Another reason to have `anchorName` in state is the ability to use it in static context during derived state computation. If the point for having it as a class field is to perform a computation just once I can do a conditional assignment to the non-empty anchorName.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: Done

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
GreenTeaCake: I can just remove [line n93](https://github.com/Semantic-Org/Semantic-UI-React/pull/3533/files#diff-488a7257d83543e18082c7216d0aecd9R93) (`...state,`).

layershifter:
@grumblerchester I've pushed some changes and merged `master`, can you please check them? 🙄

GreenTeaCake:
@layershifter Hey! I've checked the code locally. Everything works fine

PR #3535, opened 2019-03-26
docs(ComponentDoc): replace deprecated lifecycle methods

GreenTeaCake:
Relates to #2732  Replace `componentWillReceiveProps` by constructor and `getDerivedStateFromProps`.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentDoc.js
+++ b/docs/src/components/ComponentDoc/ComponentDoc.js
@@ -34,15 +34,18 @@ class ComponentDoc extends Component {
   state = {}
   examplesRef = createRef()
 
-  getChildContext() {
+  static getDerivedStateFromProps(props, state) {
+    const { displayName } = props
+
     return {
-      onPassed: this.handleExamplePassed,
+      displayName,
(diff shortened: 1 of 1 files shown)


docs/src/components/ComponentDoc/ComponentDoc.js
```diff
@@ -31,18 +31,31 @@ class ComponentDoc extends Component {
...
+  static getDerivedStateFromProps(props, state) {
+    const { displayName } = props
+
+    const derivedState = {}
+    if (displayName !== state.displayName) {
+      derivedState.displayName = displayName
+      derivedState.activePath = undefined
+    }
+    return derivedState
+  }
```
layershifter, line 42: Actually, we this can be simplified 🚀 

#### 1. We want always set `displayName` to state:
```js
return { displayName: props.displayName }
```
#### 2. We want reset only if `displayName`s are not equal:
```js
return { 
  activePath: props.displayName === state.displayName ? state.activePath : undefined,
  displayName: props.displayName,
}
```
#### 3. And for this code we can remove `constructor()` at all

At first render after `getDerivedStateFromProps()` will be called, we will reset `activePath` to undefined, but it's undefined by default 👍

PR #3540, opened 2019-03-29
fix(build): fix exports in UMD build

### LINKED ISSUE #3536: Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
### Steps
Using semantic-ui-react.min.js in a script tag doesn't work, e.g. when trying to externalise the dependency in webpack and use a CDN-build instead.

```html
<!DOCTYPE html>
<html>
    <body>
        This does't work!
        <script src="https://cdnjs.cloudflare.com/ajax/libs/react/16.8.5/umd/react.production.min.js"></script>
        <script src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.8.5/umd/react-dom.production.min.js"></script>
        <script src="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui-react/0.86.0/semantic-ui-react.min.js"></script>
    </body>
</html>
```

### Expected Result
Semantic-ui-react.min.js is available in the webpage

### Actual Result
The following error occurs in the console:

```js
semantic-ui-react.min.js:1342 Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
    at Module.<anonymous> (semantic-ui-react.min.js:1342)
    at Module.<anonymous> (semantic-ui-react.min.js:1342)
    at n (semantic-ui-react.min.js:1)
    at semantic-ui-react.min.js:1
    at semantic-ui-react.min.js:1
    at semantic-ui-react.min.js:1
    at semantic-ui-react.min.js:1
```

### Version
0.86.0 (doesn't work with 0.85.0 either).

Also tested with different versions of react/react-dom (16.3.0 / 16.6.0 / 16.8.5).

layershifter:
Fixes #3536.

---

This PR also adds `test:umd` step to ensure that UMD build works correctly.


src/umd.js
```diff
@@ -3,8 +3,4 @@
 // Do not replace this with named exports.
 // We need to export an object here for browser builds.
 // Otherwise, we end up with every component on the window.
-import * as semanticUIReact from './index'
-
-module.exports = {
-  ...semanticUIReact,
-}
+export * from './index'
```
layershifter, line 6: It's antipattern to combine CJS and ES modules in one file...

PR #3553, opened 2019-04-04
fix(elements): allow 'right' on iconPosition for Label

matt-koevort:
Using 'right' as iconPosition currently produces a warning for failed prop type.


shane-matuszek-takealot:
I'm having this same issue right now

layershifter:
Duplicate of #1802.

Issue #3565, opened 2019-04-11
using umd in script tag is broken


soederpop:
### Steps

Try including the semantic-ui-react umd build via script tag, and it is unavailable on window

### Expected Result

Expected to be able to use the library via a script tag and access via window.semanticUIReact

### Actual Result

```
typeof window.semanticUIReact === 'undefined'
```
### Version

0.86.0

### Testcase

You can see in the codesandbox that semanticUIReact is no longer put into window

https://codesandbox.io/s/jp7qrp41vy?fontsize=14

soederpop:
I was able to fix this in a fork by using

```
  entry: {
    [pkg.name]: `expose-loader?semanticUIReact!${paths.src('umd.js')}`,
  },
```

in your webpack umd config.  

tbh I thought webpack with libraryTarget `umd` and library `semanticUIReact` would have already done this for you (probably as it did previously).  

It might not since you don't have a default export?  I wish I could be more help

layershifter:
Duplicate of #3536 as was fixed in #3540, we also have a test that asserts for working UMD build:

![image](https://user-images.githubusercontent.com/14183168/56021527-577b3e80-5d12-11e9-9367-959ee61c644f.png)

soederpop:
Here's some simple HTML

```html
<html>
  <head>
      <script src="https://unpkg.com/react@16.8.4/umd/react.production.min.js"></script>
      <script src="https://unpkg.com/react-dom@16.8.4/umd/react-dom.production.min.js"></script>
      <script src="https://unpkg.com/semantic-ui-react@0.86.0/dist/umd/semantic-ui-react.min.js"></script>
      <script>
        alert(typeof window.semanticUIReact === 'undefined' ? 'Oh Crap' : 'Hmm')
      </script>
  </head>
  <body>
      <h1>hi</h1>
  </body>
</html>
```

here's the result:

![Screen Shot 2019-04-12 at 10 06 29 AM](https://user-images.githubusercontent.com/5178345/56047358-dc625a00-5d0a-11e9-913c-a13af8dbe9ba.png)

soederpop:
Here's that same HTML with the last working UMD build of semantic-ui, 0.84

```html
<html>
  <head>
      <script src="https://unpkg.com/react@16.8.4/umd/react.production.min.js"></script>
      <script src="https://unpkg.com/react-dom@16.8.4/umd/react-dom.production.min.js"></script>
[...]

PR #3566, opened 2019-04-12
chore(package): add `browser` field that points to UMD build

soederpop:
- uses expose-loader to do this
- added the umd build path as the browser property on package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "jsnext:main": "dist/es/index.js",
   "main": "dist/commonjs/index.js",
   "module": "dist/es/index.js",
+  "browser": "dist/umd/semantic-ui-react.min.js",
   "types": "index.d.ts",
   "files": [
     "src",


layershifter:
https://github.com/Semantic-Org/Semantic-UI-React/issues/3565#issuecomment-482478984

UMD build was already fixed, however thank you for your effort there ❤️ 
I also want to avoid add `browser` field, for me it's not clear how Webpack and Rollup will handle them...

soederpop:
Please reconsider.

For webpack you have to opt in with config.  

```
{ resolve: { aliasFields: ['browser'] } }
```

here are the docs on how it is used https://github.com/defunctzombie/package-browser-field-spec

Rollup also lets you opt into this:

https://github.com/rollup/rollup-plugin-node-resolve

So making this change won't effect anyone who uses these bundlers, except in the case that the developers make this very common config enhancement (or their boilerplate does) and then semantic-ui-react isn't able to help them optimize

but the biggest benefit would be for tools like unpkg.com and other cdn's, who would know to link to the umd build by default.  instead they redirect to the commonjs build which is obviously not usable in the browser.

so it doesn't hurt anyone, and only helps the people who want it.

Also, the UMD build is actually broken, despite the test saying otherwise

```html
<html>
  <head>
      <script src="https://unpkg.com/react@16.8.4/umd/react.production.min.js"></script>
      <script src="https://unpkg.com/react-dom@16.8.4/umd/react-dom.production.min.js"></script>
      <script src="https://unpkg.com/semantic-ui-react@0.86.0/dist/umd/semantic-ui-react.min.js"></script>
      <script>
        alert(typeof window.semanticUIReact === 'undefined' ? 'Oh Crap' : 'Hmm')
      </script>
  </head>
  <body>
      <h1>hi</h1>
  </body>
</html>
```

see the result:

Screen Shot 2019-04-12 at 10 06 29 AM

tested on all browsers.

used the file from the npm package, from unpkg.com, and built locally -- same error

also, take a look at the UMD build's contents:

```
/*! no static exports found */
/*! all exports used */
/*! ModuleConcatenation bailout: Module is not an ECMAScript module */function(t,n){t.exports=e},
/*!***************************************************************!*\
  !*** ./node_modules/@babel/runtime/helpers/defineProperty.js ***!
  \***************************************************************/
```

30% or more of the bundle is this kind of stuff, even though TerserPlugin is configured with comments set to false in the output.

maybe the test isn't testing the same build you're actually distributing.

soederpop:
Also I confirmed this bug did not exist until 0.85.0.  80-84 are all fine.

I took the liberty of modifying your UMD test to include the UMD builds that have been published and the tests do fail starting at 0.85 and 0.86, but pass at 0.84

I do think there may be an issue between what your tests are actually testing, vs what you're distributing.

layershifter:
Fix was not published yet, that's why it it's still broken. You can manually build master and get working UMD. I got your point, let's revert changes with `expose-loader` and merge.

soederpop:
Will do! 

~I don't think the version I built from master without expose-loader was working but maybe I rushed through it, will test it out again.~

Sorry to bombard you I didn't realize that the master branch was ahead of the npm release even though the version numbers are the same.

soederpop:
thanks a lot for reconsidering. removed the expose-loader.  confirmed it works fine. apologies for the false alarm.

PR #3574, opened 2019-04-18
feat(Table): add `headerRows` to allow to render multiple headers

abaschen:
Same behavior as renderBodyRow with tableData but for the Headers. Necessary when having more than one line in headers.
If headerRow is set then renderHeaderRow and headerData are ignored.

If accepted, maybe do the same for footer?


layershifter:
https://codesandbox.io/s/wym08vp8zl?module=/example.js

Can you check this CodeSandbox. Does it satisfiy all of your needs? 🐱

abaschen:
Indeed!! I was currently checking and about to commit tests and changes etc... but with this yes it's perfect!

abaschen:
Oops no I've commented too fast. It doesn't satisfy it. how can you have more than one row with rowSpan and so on?

layershifter:
Can you please provide example markup in HTML for it?

layershifter:
Do you mean something like this?
https://codesandbox.io/s/v0344yp7n7

abaschen:
Yes like this but with shorthands

layershifter:
@TecHunter please check changes, going to merge after your confirmation 👍

abaschen:
Okay for me. 

Maybe do the same for footer?

layershifter:
@TecHunter let's do 👍 We can add it in a separate PR

Issue #3554, opened 2019-04-04
Popup: closed while "mouse click inside but release outside" of the Popup block


jiayimu:
### Steps
I am creating a login popup component in React by using React Semantic UI Popup. I am using "Events triggering the popup on click". I was trying to copy the content in the Popup block, but the popup was closed when my mouse released outside of the popup box.  

```
import React, { Component } from 'react';
import { Button, Message, Popup, Icon, Menu } from 'semantic-ui-react';

// code inside of render return function
<Popup trigger={<Button icon>LOGIN</Button>}
       on='click'
       className='login-popup'
>
       <div className='popup-main'>
              <Message attached='bottom'>
                        Log In
              </Message>
              <LoginForm
                  {...this.props}
              />
        </div>

        <Message attached='bottom'>
              Don't have account? Signup instead.
        </Message>
</Popup>
```

### Expected Result
I would like to keep the Popup open unless the user clicks the mouse outside of the Popup

### Actual Result
The Popup closed when the mouse clicked inside and released outside of the popup box.

### Version
0.85.0

### Testcase
I tested it on the offical website and still get the same result.

layershifter:
https://react.semantic-ui.com/modules/popup/#usage-click

I tried this example and everything works as expected. Please provide a minimal CodeSandbox example and detailed steps.

jiayimu:
@layershifter 
Thank you for your checking. I think I might not explain the issue clearly. I asked this in stack overflow ([here](https://stackoverflow.com/questions/55502216/react-semantic-ui-popup-was-hidden-while-mouse-click-inside-but-release-outside)). I updated my expected result as below. Hope the update makes more sense.

**Expected Result**:  
The Popup box should not be closed if the mousedown event happens inside of the Popup, no matter where the mouseup event triggered.

layershifter:
Fixed in #3575.

mstarcevic:
@jiayimu 

I came across your post while trying to resolve a similar issue. I eventually resolved it by setting one of the input fields in the popup to autoFocus. Prior to that, clicking on an icon or somewhere not specifically an input, closed the popup straight away.

I see the issue has been closed but thought the comment might be helpful in any case.

PR #3575, opened 2019-04-18
fix(Popup|Portal): do not close when mouse click was occurred inside

mihai-dinculescu:
This PR is my proposed fix for #3554.

Afaik the `onClick` event provides no information whatsoever about the position where the mouse button was pressed down. For this reason I've added a listener for `mouseDown` and I'm storing the latest occurence of this event in the component.

~~I would like to also write a unit test but I'm not really sure how to go about mocking a click that originates in one place and ends in another.~~


mihai-dinculescu:
~~I've updated `domEvent.click` to also fire `mousedown` and `mouseup` so that it mimicks reality.~~

layershifter:
@joergbaier @mihai-dinculescu Thank you for addressing this 👍 

We have a similar stuff in `Dropdown` with `isMouseDown` and I prefer to keep existing listeners with `click` instead of `mouseup`. So I am wondered, do we really need to use `document.mouseup` instead of `document.click`?

mihai-dinculescu:
I don't see any way around using `mousedown`. `click` provides no information whatsoever about it.

However, I do see why we need to be more thoughtful before merging this. Touch gestures do not provide `mousedown` and `mouseup` events.

After further consideration, I now think that we should use `mousedown` in tandem with `click`, instead of `mousedown` and `mouseup`.

I should have a rework of this branch ready soon.

mihai-dinculescu:
Ok @layershifter, it should look much better now!

PR #3582, opened 2019-04-24
fix(Modal): do not close when mouse click was occurred inside

joergbaier:
The default modal closes when:

1. When a click starts in the modal but ends in the dimmer
2. When a click starts in the dimmer but ends in the modal

This can be reproduced in the first example
https://react.semantic-ui.com/modules/modal/

This change will only close the Modal if the click starts and ends in the dimmer.

### DIFF
--- a/src/modules/Modal/Modal.js
+++ b/src/modules/Modal/Modal.js
@@ -186,13 +186,12 @@ class Modal extends Component {
   handleDocumentClick = (e) => {
     debug('handleDocumentClick()')
     const { closeOnDimmerClick } = this.props
-
-    const previousDocumentMouseDownEvent = this.latestDocumentMouseDownEvent
-    delete this.latestDocumentMouseDownEvent
+    const currentDocumentMouseDownEvent = this.latestDocumentMouseDownEvent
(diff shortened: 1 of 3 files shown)


joergbaier:
It looks like https://github.com/Semantic-Org/Semantic-UI-React/pull/3575 could use the same change as well.

mihai-dinculescu:
I'm not sure if changing the tests from `domEvent.click` to `domEvent.mouseDown` and `domEvent.mouseUp` is the right aproach. This will not hit the `onClick` handlers and will act differently from reality. I believe that my aproach from https://github.com/Semantic-Org/Semantic-UI-React/pull/3575 to update `domEvent.click` to also fire `mousedown` and `mouseup` is more resilient.

joergbaier:
@mihai-dinculescu I will update this PR to match yours tomorrow.

fraynn:
Thanks for fixing this issue. Do you have any idea on when we can expect this fix to be pushed to npm ?

layershifter:
I think that we will finally make a release on this week 🎉

PR #3586, opened 2019-04-27
fix(Dropdown): use `placeholder` as default text

### LINKED ISSUE #3581: Dropdown: with button and search: search text isn't visible
### Steps
Create a Dropdown component using button and search options e.g.
```
<Dropdown
  className="icon"
  floating
  labeled
  icon="filter"
  button
  search
  options={myOptions}
  text="Search..." />
```
Type into the search box to filter the available options. The list will filter correctly but the search text is not visible in the search box.

### Expected Result
The search text is visible in the search box as well as filtering the available options. This is the behaviour in semantic-ui.

### Actual Result
The search text is not visible.

### Version
0.82.0

### Testcase
https://codesandbox.io/s/j2o9r3771w

This issue is also present on the semantic-ui-react Dropdown documentation page at https://react.semantic-ui.com/modules/dropdown/#types-search-dropdown

eyas2014:
Fixes #3581.

---

We should always display some `text` in `div.text` because SUI does the same thing: https://semantic-ui.com/modules/dropdown.html Actually, this is also an answer why we don't have any unit tests for this behavior.

SUI has styles that will prevent visibility off `div.text`:

image


layershifter:
Sorry for the late feedback. I want to check this change additionally because it was not covered by unit tests before 😮

mattgd:
Hi @layershifter! Any word on this PR? Is there anything I might be able to do to help with testing to expedite this?

layershifter:
@mattgd @eyas2014 thank you for contributions and follow ups 👍

I checked this issue deeply and `value` was invisible because we haven't render any `text`, so @eyas2014 provided a correct fix ⭐️ But, we should render a `placeholder` instead of `searchQuery` by default as SUI does:

```jsx
  <div class="ui floating dropdown labeled search button icon active visible">
    <input class="search" autocomplete="off" tabindex="0" value="VALUE" />
    <span class="text filtered">PLACEHOLDER</span>
  </div>
```

https://codesandbox.io/s/semantic-ui-react-example-55i18

src/modules/Dropdown/Dropdown.js
```diff
@@ -1193,9 +1193,8 @@ export default class Dropdown extends Component {
       search && searchQuery && 'filtered',
     )
     let _text = placeholder
-    if (searchQuery) {
-      _text = null
```
layershifter, line 1197: SUI renders `text` always and has styles to make it invisible 🐙

PR #3600, opened 2019-05-09
docs(RatingExampleRating): your description

zaghadon:
just some minor changes which can help an extremely curious beginner like me <3


layershifter:
Examples below contain valid code for `icon='heart'` and `icon='star'`: 

https://react.semantic-ui.com/modules/rating/#types-star
https://react.semantic-ui.com/modules/rating/#types-heart

image

Sorry, but I don't see any positive effect from these changes.

PR #3603, opened 2019-05-09
feat(Form): make `error` prop shorthand in Form.Field

### LINKED ISSUE #3602: Add support for error messages in <Form.Field/>
### Problem description
Some errors might require the developer to give the user some context on what went wrong. And this will most likely happen by showing an error message. But the Form.Field component lacks this functionality. 

### Proposed solution
To extend the error prop to not only accepts bools, but also strings to be shown as error messages. There is currently no css to support this feature. 

### MVP
Something as simple as this
Screenshot 2019-05-09 at 19 26 13
Screenshot 2019-05-09 at 19 16 54

bubunyo:
Fixes #3602. 

This PR makes `error` prop in `Form.Field` shorthand that allows to define custom error message inside it.

```jsx
<Form.Input  error='Please enter your last name' label='Last name' />
<Form.Checkbox
  label='I agree to the Terms and Conditions' 
  error='You must agree to the terms and conditions'
/>
```

image


layershifter:
https://semantic-ui.com/behaviors/form.html#validating-on-blur-and-other-events

SUI has similar thing actually, will share more thoughts tomorrow.

bubunyo:
Sure, As much as the SUI way attempts to solve the same problem, I think their way can be easily replicated, using already existing components. I don't see this PR being merged unless there is an earlier PR on SUI to deals with the formatting. CSS is not my strong suit, but I am attempting to submit a PR for that. I will welcome any help on that.

minkasu-kuppusamy (approved):

layershifter:
@bubunyo I pushed some updates to this PR, can you please check? 🙏

layershifter (approved):

PR #3607, opened 2019-05-13
fix(Popup): fix positioning in scrollable container

### LINKED ISSUE #3569: Popup: pass scroll target as a prop
Popup probably listening to the scroll of a window only. If popup in a div it stays on one position when scrolling the div. Would be great to have scroll target as a prop
I'm using 0.79.1 but the issue exists in the latest build.

### Testcase
https://codesandbox.io/s/74mkmk5600

layershifter:
Fixes #3569.

`Popper.JS` relies on [`getScrollParent()`](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getScrollParent.js) function which recursively uses [`getParentNode()`](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getParentNode.js) under hood. `getParentNode()` uses `parentNode` field and our `ReferenceProxy` doesn't have it before 😢 

### Before

popup-bug

### After

popup-update

### DIFF
--- a/src/modules/Popup/lib/createReferenceProxy.js
+++ b/src/modules/Popup/lib/createReferenceProxy.js
@@ -19,7 +19,7 @@ class ReferenceProxy {
   }
 
   get parentNode() {
-    return this.ref.current.parentNode
+    return this.ref.current ? this.ref.current.parentNode : undefined
   }
 }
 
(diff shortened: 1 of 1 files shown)


rijk:
Awesome! :) Can we test this somewhere? + any idea when the next update will be released?

layershifter:
I expect to have a release on next week 📅

PR #3654, opened 2019-06-10
feat(Popup): add pinned and popperModifiers prop

### LINKED ISSUE #3615: Popup: positioning glitch when using negative vertical offset
## Bug Report
Implementing the new Popper `offset` parameter (#3532), something weird happens:

(first is without offset, then with offset):

tooltip offset issue short

### Steps
A Popup configured with

```
position="bottom center"
offset="0, -4"
```

(near the top of the page, that could have something to do with it).

### Expected Result
Tooltip appears shifted up, but still in the same orientation (bottom center).

### Actual Result
Orientation changes to top, making it very glitchy.

### Version
0.87.1

### Testcase
https://codesandbox.io/s/semantic-ui-react-2m6rb

As a workaround I tried to use the `positionFixed` parameter, but it doesn't appear to be passed to Popper.

layershifter:
Fixes #3615.

## `pinned`

Will stop flipping of opened Popup.

popup-pinned

## `popperModifiers`

We can't map all PopperJS modifiers to component's props, so it will be better to just export them to allow customize them.

### Restore rending `children` in `ComponentExample`

I just noticed that we're are missing helpful descriptions in docs:

image

Fixed 💥

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
@@ -214,7 +214,10 @@ const Wrapper = (props) => {
       {...rest}
       /* eslint-disable-next-line react/prop-types */
       sourceCode={sourceCode}
-    />
+    >
+      {/* eslint-disable-next-line react/prop-types */}
+      {props.children}
+    </ComponentExample>
   )
 }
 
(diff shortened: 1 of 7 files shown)


RRSR:
Even after adding `pinned` the popup move up if squeezed

`
<StyledPopup
    basic
    content={<SomeKindaList />}
    flowing
    hoverable
    onOpen={props.p1}
    onClose={props.p2}
    open={props.p3}
    position='bottom right'
    trigger={(
      <StyledLinkMenuItem isOpenPopup={props.p1}>
        ..........
      </StyledLinkMenuItem>
    )}
    offset='0, -10'
    pinned
  />
`

So if I decrease the window size the popup moves up. How to fix the same.

PR #3657, opened 2019-06-11
feat(Popup): add `popperDependencies` prop

### LINKED ISSUE #3639: Popup: expose reference for react-popper instance
### Problem description
I want to recalculate `Popup` position when the size of content is changing.
However, it seems like there's no easy way to achieve this without having popper instance.

https://github.com/FezVrasta/react-popper/issues/274

### Proposed solution
- simple way: Just add reference prop for `<Popper />` element to the `PopupProps`
- complex way: Add feature to recalculate popup position to the `Popup` element.

layershifter:
Fixes #3639.

This PR adds `popperDependencies` that behaves like dependencies in React Hooks and allows to schedule a position update for `Popup`.  Should be used in cases when content is changing.


src/modules/Popup/Popup.js
```diff
@@ -171,6 +177,12 @@ export default class Popup extends Component {
     return { contentRestProps, portalRestProps }
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.open && shallowEqual(this.props.popperDependencies, prevProps.popperDependencies)) {
```
rokoroku: ~`shallowEqual` for dependency prop doesn't fit to hook-like dependency input.~

~React hook uses internal `areHookInputsEqual(nextDeps, prevDeps)` methods which iteratively check both input items using  [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) in order.~

~See: https://github.com/facebook/react/blob/45acbdc0ba05c29f849011eaed5e5b221b219d06/packages/react-reconciler/src/ReactFiberHooks.js#L301-L348~

~And there is a clone of above method in ~npm~ GitHub:
https://github.com/Andarist/are-hook-inputs-equal/blob/master/src/index.js~


Sorry, I found that the `shallowequal` library already do that! 😅

rokoroku (changes requested):

rokoroku (approved):

src/modules/Popup/Popup.js
```diff
@@ -230,15 +242,24 @@ export default class Popup extends Component {
...
     debug('handlePortalMount()')
+
+    this.open = true
     _.invoke(this.props, 'onMount', e, this.props)
   }
 
   handlePortalUnmount = (e) => {
     debug('handlePortalUnmount()')
+
+    this.open = true
```
rokoroku:
```suggestion
    this.open = false
```
maybe?

src/modules/Popup/Popup.js
layershifter: I will cover this with UTs to be sure 🐱

src/modules/Popup/Popup.js
layershifter: Great catch 👍

Issue #3630, opened 2019-05-29
Search: Ability to pass a custom SearchCategory renderer to the Search component


PrincessMadMath:
### Problem description

Currently the Search component supports custom renderer for the category name and result items but not for the category layout.

For example, instead of having a two columns layout (with category on the left and results on the right), I would like to be able have a simple list with a header by category:

![image](https://user-images.githubusercontent.com/8277266/58582715-61a0c000-8220-11e9-99b4-06807c116b1b.png)

### Proposed solution

Search can receive a new prop "categoryLayoutRenderer", which is passed down to SearchCategory. The default value of this props is the current category layout.

The prop categoryLayoutRenderer, is a function that receive the rendered category name and rendered results items.

Exemple: 

![image](https://user-images.githubusercontent.com/8277266/58583504-2606f580-8222-11e9-90af-9af8d967e365.png)

### MVP

We opened a PR to show the POC of how it can be done: https://github.com/Semantic-Org/Semantic-UI-React/pull/3672

PrincessMadMath:
Updated the description with a PR for the MVP: https://github.com/Semantic-Org/Semantic-UI-React/pull/3672

levithomason:
Fixed in #3672

PR #3672, opened 2019-06-18
feat(Search): custom category layout renderer

PrincessMadMath:
POC for: #3630


patricklafrance:
Hey @layershifter!

We would love to hear your feedback on this PR. This is a customization we have been looking for for a while!

Thank you,

Patrick

docs/src/examples/modules/Search/Types/index.js
```diff
@@ -21,7 +21,7 @@ const SearchTypesExamples = () => (
       examplePath='modules/Search/Types/SearchExampleCategory'
     />
     <ComponentExample
-      title='Category (Custom Render)'
+      title='Category (Custom Render) v2'
```
levithomason: Let's not version the example titles in this way.  It will be confusing with the semver already in place.

levithomason:
This is really fantastic, thanks for the great work.  I'll get this merged and shipped.

PR #3685, opened 2019-07-01
WIP fix(Transition): hiding automatically when setState of parent called on mount

### LINKED ISSUE #3641: Transition: hiding automatically when setState of parent called on mount
### Steps
1. When the child of Transition is a stateful component and is not enclosed by any div in the parent component, hide of transition is not hiding the child but the visible prop is actually false at the moment.
2. When the child of Transition is a stateful component and enclosed by div in parent component, if any function gets triggered from child's componentDidMount which do a setState in parent, transition is hiding the child but the visible prop is still true.
If unmountOnHide is set true, this auto hide happens after every visible trigger.

### Expected Result
1. Shall hide the child component.
2. Shall not auto hide on open.

### Actual Result
1. Hide not happening.
2. Auto closing after making it visible.

### Version
0.87.1

### Testcase
1. https://codesandbox.io/s/transitionhideissue-upze1
2. https://codesandbox.io/s/transitionautocloseissue-lj3km

DzmitryRamaniuk:
Fixes [#3641](https://github.com/Semantic-Org/Semantic-UI-React/issues/3641)


DzmitryRamaniuk:
I fixed, review plz and now I'm trying to write appropriate test

layershifter:
This issue was fixed in `semantic-ui-react@1.0.0` by #3982. Closing for housekeeping.

PR #3692, opened 2019-07-07
fix(Checkbox): handle click properly on label

### LINKED ISSUE #3683: Checkbox: label only clickable when simple element
## Bug Report
Checkbox label only clickable when simple element

Sandbox-CodeSandbox

### Steps
Nest a span in label, and clicking no longer works

### Expected Result
Toggle the checkbox

### Actual Result
Nothing

### Version
0.87.2

### Testcase
https://codesandbox.io/s/semantic-ui-react-78ovl

I haven't verified this, but my guess is this from #3435 which requires an exact match for the click event on the label element, and doesn't account for clicks on nested elements.

layershifter:
Fixes #3683.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -148,8 +148,9 @@ export default class Checkbox extends Component {
     const { id } = this.props
     const { checked, indeterminate } = this.state
 
+    const isLabelClick = _.invoke(this.labelRef.current, 'contains', e.target)
+
     const hasId = !_.isNil(id)
-    const isLabelClick = e.target === this.labelRef.current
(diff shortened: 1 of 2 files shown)


src/modules/Checkbox/Checkbox.js
```diff
@@ -148,8 +148,9 @@ export default class Checkbox extends Component {
     const { id } = this.props
     const { checked, indeterminate } = this.state
 
+    const isLabelClick = _.invoke(this.labelRef.current, 'contains', e.target)
```
layershifter, line 151: Previously there was an equality comparison, but in this case `contains()` should be used.

PR #3693, opened 2019-07-07
fix(Checkbox): handle change on root element

### LINKED ISSUE #3502: Checkbox: `slider` is not clickable in the entire area
## Bug Report
If you click in the space between the nub & the line it won't trigger the handler
### Steps
click in the are of the arrow 
image

### Expected Result
It should allow click on the checkbox in that area.

### Actual Result
It not clicks there, therefore it is hard to click on the toggle.

### Version
0.86.0

### Testcase
It should work as the jQuery version that is clickable in that area.
Checkout how it works https://semantic-ui.com/modules/checkbox.html#fitted

layershifter:
Fixes #3502.

In SUI the whole area of `slider` is clickable, previously we had an behavior mismatch. This PR fixes it.

## Before
before

## After 
after

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -148,7 +148,9 @@ export default class Checkbox extends Component {
     const { id } = this.props
     const { checked, indeterminate } = this.state
 
+    const isInputClick = _.invoke(this.inputRef.current, 'contains', e.target)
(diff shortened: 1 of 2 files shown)


felixmosh:
I'm not sure that the solution is good, I think that what can be fixed using css is better, but it is your decision.
Thanx for the fix!

layershifter:
> I think that what can be fixed using css is better, but it is your decision.

I can agree with it, but we should match SUI behavior 😺

PR #3702, opened 2019-07-12
feat(Dropdown): allow selecting an item when pressing the spacebar

nelsonleite:
Due to accessibility matters (and similarity to the native <select>), this adds the possibility to select an item via Spacebar and not only with Enter


layershifter:
### native `<select />`

https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select
Value is not selectable by Spacebar.

### `listbox`, `combobox`, `menubutton`
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html

No `Spacebar`, too.

> Due to accessibility matters (and similarity to the native )

Can you please clarify to which accessibility pattern you're referring?

nelsonleite:
The native `select` selects the value with the spacebar at least in Chrome and Safari. There is no specific a11y pattern that specifies this spacebar usage, but I was trying to mimic the behaviour in some browsers 🙂

layershifter:
@nelsonleite I have discussed this change also with Stardust UI team as there is also `Dropdown` component. I will follow up the discussion again, but I think that your PR will be included to next SUIR release 👍 🎉 🚀

layershifter (approved):

PR #3704, opened 2019-07-13
fix(Dropdown): allowing a ReactNode for the text prop

laurenz-glueck:
This PR is updating the `PropTypes` of the `Dropdown` module to allow a `ReactNode` for the `text` prop.


layershifter:
> https://github.com/Semantic-Org/Semantic-UI-React/issues/3003#issuecomment-404647591
> https://github.com/Semantic-Org/Semantic-UI-React/issues/3003#issuecomment-404627408

> You're looking for a `trigger` prop 😺 

https://github.com/Semantic-Org/Semantic-UI-React/pull/3431#issuecomment-462870180

Thanks for contributing, but we already have a `trigger` prop. Probably we should remove `text` prop at all to avoid confusions.

PR #3723, opened 2019-07-28
Add support for generic data sources to `Select`

yhnavein:
The problem I had with the `Select` is that I had to re-map my data sources manually to feed them as options.

The approach here is similar to what was done in the `ng2-semantic-ui`, i.e. let's allow more flexible `options` property + add possibility to specify which fields should be used as a `label` and `value`.

Updated docs to reflect this new approach.
I tried to make it as non-breaking as possible :)


layershifter:
Thank you for contributing!

We are not going to introduce such functionality now as it's very different from our ideas of shorthand APIs. To be honest, you can create a wrapper component on your side to do it 🚀

yhnavein:
That's a bummer!
And yeah, I know I can create a wrapper. But it's so basic functionality that I was even surprised it's not even available by default.
As a developer I expect to just attach a data source to a component and be ready to go. Not manually map object properties every single time. 

I understand your concern but this is just just too much

PR #3760, opened 2019-08-24
feat(Popup): add `positionFixed` to support fixed mode in `Popper.JS`

### LINKED ISSUE #3718: Popup: wrong positioning when `mountNode` uses `position: fixed`
### Steps
-add Popup control
-set mountNode property

### Expected Result
Popup should be renderer in a set node and visually be placed there

### Actual Result
Popup is rendered in a set node but visually is NOT placed there
It is placed far away outside
Wrong calculation of `transform` css value

### Version
0.87.3

In the version 0.86.0 all is ok

fxOne:
Fixes #3718.

Add support for popper.js to position the popup correctly if used with a mountNode inside a `position: fixed` container

### DIFF
--- a/src/modules/Popup/Popup.d.ts
+++ b/src/modules/Popup/Popup.d.ts
@@ -107,7 +107,7 @@ export interface StrictPopupProps extends StrictPortalProps {
     | 'top center'
     | 'bottom center'
 
-  /** Enables the popper position 'fixed' mode */
+  /** Tells `Popper.js` to use the `position: fixed` strategy to position the popover. */
   positionFixed?: boolean
 
   /** An object containing custom settings for the Popper.js modifiers. */
--- a/src/modules/Popup/Popup.d.ts
+++ b/src/modules/Popup/Popup.d.ts
@@ -107,6 +107,9 @@ export interface StrictPopupProps extends StrictPortalProps {
     | 'top center'
     | 'bottom center'
 
+  /** Enables the popper position 'fixed' mode */
+  positionFixed?: boolean
+
   /** An object containing custom settings for the Popper.js modifiers. */
   popperModifiers?: object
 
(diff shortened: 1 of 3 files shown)


src/modules/Popup/Popup.js
```diff
@@ -153,6 +156,7 @@ export default class Popup extends Component {
     on: ['click', 'hover'],
     pinned: false,
     position: 'top left',
+    positionFixed: false,
```
layershifter:
```suggestion
```

It's not required to set defaults to `false`

test/specs/modules/Popup/Popup-test.js
```diff
@@ -289,6 +289,20 @@ describe('Popup', () => {
     })
   })
 
+  describe('positionFixed', () => {
+    it(`is "false" by default`, () => {
+      wrapperMount(<Popup open />)
+
+      wrapper.should.have.prop('positionFixed', false)
```
layershifter:
```suggestion
      wrapper.should.not.have.prop('positionFixed')
```

https://github.com/Semantic-Org/Semantic-UI-React/pull/3760/files#r317650640

layershifter (approved):
Left minor comments, looks good to me ❤️

fxOne:
Thanks for the feedback, I removed the unnecessary code.

src/modules/Popup/Popup.d.ts
```diff
@@ -107,6 +107,9 @@ export interface StrictPopupProps extends StrictPortalProps {
     | 'top center'
     | 'bottom center'
 
+  /** Enables the popper position 'fixed' mode */
```
layershifter:
```suggestion
  /** Tells `Popper.js` to use the `position: fixed` strategy to position the popover. */
```

src/modules/Popup/Popup.js
```diff
@@ -127,6 +127,9 @@ export default class Popup extends Component {
     /** Position for the popover. */
     position: PropTypes.oneOf(positions),
 
+    /** Enables the popper position 'fixed' mode */
```
layershifter:
```suggestion
    /** Tells `Popper.js` to use the `position: fixed` strategy to position the popover. */
```

Issue #3764, opened 2019-08-28
Dropdown: search string with space causes select the first option


honzajerabek:
### Steps
1. Have a Dropdown (both `selection` and `search` true) and options that are multiple words starting same first word
2. Open the dropdown and start typing the first word, that is common for some number of options
3. Press space to continue writing
4. the first available option is immediately selected

note: I've tried number of combinations of `selectOnNavigation`, `selectOnBlur` props as well as stopping propagation on `onSearchChange` and other events

### Expected Result
Dropdown lets me to search any string I want

### Actual Result
Dropdown cuts typing the search word on the spacebar press and selects the first available option from the search results

### Version
0.88.0

### Testcase
https://codesandbox.io/s/semantic-ui-react-9undx

ambroselittle:
This issue undoubtedly is caused by this: https://github.com/Semantic-Org/Semantic-UI-React/pull/3702 

That PR should be named "force" selection not "allow". :D  

Suggestion that if search is enabled, this behavior is not enabled. And/or add an option to configure. But really, it shouldn't happen during search.

evanstern:
Yes, at the very least this needs to be an option, not the default.

ambroselittle:
By the way, temporary workaround that tested OK for us - downgrade to 0.87.3. Unless you need something in 0.88, of course.

honzajerabek:
> Suggestion that if search is enabled, this behavior is not enabled. And/or add an option to configure. But really, it shouldn't happen during search.

@ambroselittle Agree, it does behave like that in non-react semantic-ui - https://jsfiddle.net/Lqanh61p/1/

i guess I'll have some time to make a pr

evanstern:
> i guess I'll have some time to make a pr

👍 -- ~I was about to say the same thing but if you want to do this that's fine with me!~ (just saw that you already submitted it)

> By the way, temporary workaround that tested OK for us - downgrade to 0.87.3. Unless you need something in 0.88, of course.

Yeah, we want the fixes in 0.88 that stop message spam about `UNSAFE_` lifecycle methods so I'd rather not downgrade. Luckily, we are at a place in development where we can live with a buggy UI element until it gets fixed.

layershifter:
Thanks for reporting and special thanks to @honzajerabek 🍾 
Will be released during this week 👍

layershifter:
Release in `0.88.1` 🚀

Issue #3130, opened 2018-09-05
Dropdown: automatically highlights the first element


Brqqq:
### Steps
1. Create a dropdown with static data inside of it, with 2 or more items to illustrate it better
2. Open the dropdown by clicking on it. 

### Expected Result
None of the options should be highlighted, until you hover over it with your mouse or use your "up" or "down" arrow keys.

Currently the non-React version of Semantic UI already behaves like this.

### Actual Result
The first item is highlighted. So it acts as if you have your mouse hovered over this first item.

### Version
0.82.3

### Testcase
This issue can be seen on the doc page demos. A specific example of a simple dropdown where this issue can be seen:
https://codesandbox.io/s/218np7y82j

Compare this to the non-React version where this issue is NOT present:
https://jsfiddle.net/7adbsm6r/5/

I have assumed it's a bug, because of this behaviour is different than the non-React version. The non-React version behaves in the way that I described in the "expected result".

Here's an image showing the difference. This is an uncontrolled dropdown and when you first open it, **without hovering your mouse over any option**

![Comparison between the react and non-react version of semantic](https://imgur.com/YLey5sp.png)

fifthsegment:
I think this is the author's intended behaviour, because there's a test case in the Dropdown-test.js file which specifically tests if the first item is selected in the case that no value or default value is provided. But you're right, the first value shouldn't be selected by default. 

```javascript
describe('selected item', () => {
    it('defaults to the first item', () => {
      wrapperShallow(<Dropdown options={options} selection />)
        .find('DropdownItem')
        .first()
        .should.have.prop('selected', true)
```

yuritoledo:
hey @Brqqq , try to add this props on Dropdown, setting it to `false`

![image](https://user-images.githubusercontent.com/18474538/46445546-cec18b80-c74d-11e8-8ee9-1c76a22ff7b9.png)

yuritoledo:
Nop. I guess I saw a behaviour when have the `multiple` props, but its to use only when is multiple :/

willbryk720:
Try selectOnBlur={false}
Found in #811 
It works with Form.Select at least

Brqqq:
@willbryk720 It was suggested in the replies here too. It does not work.

[...]

PR #3766, opened 2019-08-29
fix(Dropdown): closing the search menu on spacebar press

honzajerabek:
Fixes the #3764 .

 In tests, I've only focused on open/closed state of the Dropdown components, because `DropdownItem` gets automatically set `active` and `selected`, which is probably a different issue (#3130 ), so I decided not to dig into it.


evanstern:
👍 Thanks for getting this done so fast!

ambroselittle (commented):
Seems like the right quick fix for this.

layershifter (approved):
Tested in manually, works like a charm 💎 

Thank you for fixing this ❤️

PR #3774, opened 2019-09-06
chore(Ref): use component from Stardust UI

layershifter:
This PR is a part of convergence effort between Semantic UI React & Stardust UI.

### 🔮 `Ref` component is reexported from Stardust UI

Stardust UI has the same `Ref` component as Semantic UI. To simplify maintenance we will support `Ref` component only under Stardust.

Docs for this component will be also removed from Semantic UI, but still will be available under Stardust: https://stardust-ui.github.io/react/components/ref/definition

These changes should not affect you as SUIR user, _however_.

### 😨 Possible breaking changes

Things below were never advertised as public API, but should be mentioned:
- TypeScript interfaces `StrictRefProps`, `RefFindNodeProps`, `StrictRefFindNodeProps`, `RefForwardProps`,  `StrictRefForwardProps` are no longer exported.
- `RefFindNode` & `RefForward` components are no longer exported.

*** 

Also bumps `react-popper` to latest.


montogeek:
@layershifter Sorry for mention you here, What does it mean "convergence effort between Semantic UI React & Stardust UI."? StardustUI provides the JS/component part and SemanticUI the style/look and feel part? It is to make the core lightweight?

layershifter:
@montogeek 

Stardust UI was forked from SUIR, in the current state components there have a bit different API, design and ideas. To clarify, Stardust is more accessible, uses CSS-in-JS for styles and has some API changes in shorthands that were implemented based on consumers feedback and our experience. However, there are also caveats, it's less stable, supports less features and don't offer other themes than Teams.

However, there is a lot of things that can and should be shared. At least to simplify maintenance, first it's a `Ref` component that is totally same. Next things will be probably shorthand factories and some React hooks for internal implementation that will allow to get accessibility improvements and simplify components like `Dropdown`.

> StardustUI provides the JS/component part and SemanticUI the style/look and feel part?

In the end we will probably end with this, but it's a long ride 🚗 Currently, just converge utils.

montogeek:
Sounds really good, specially the CSSinJS use, overriding SUI specificity is hard.

PR #3785, opened 2019-09-13
fix(Search): merge nested shorthand props for the <input> element

### LINKED ISSUE #3750: Search: Issues when getting input element ref
It's unclear what the expected way to get a ref to a `Search` component's `<input>` element is. The methods I've been able to make work are a little less than ideal for various reasons.

### Use Case
I have a `Search` component that stays mounted permanently at the top of my page. In the `onResultSelect` callback I trigger route changes (using react-router), however the search input retains focus. That behaviour might be desirable in some cases, but in this case I would like to unfocus the input when this happens, which means getting a handle on the DOM element so I can call `.blur()` on it.

### Actual Result
These are the methods I've been able to make work.

1. **Nested shorthand props:**

    ```jsx
    // In class body
    inputRef = React.createRef()

    // In render
    <Search
      ...
      input={{
        input: { ref: this.inputRef }
      }}
    />
    ```
    Problem: The `<input>` element loses the props that `Search` sets, the ones found here:

    https://github.com/Semantic-Org/Semantic-UI-React/blob/dd66e920a845c199ad2cc05ddec38293a235ec20/src/modules/Search/Search.js#L531

    This is because the linked code doesn't account for the possibility of having to merge the nested shorthand props, instead the object I provide just overwrites them. Of course I could manually add those props myself, but that seems hacky/brittle. Should this be considered a bug? 

2. **Undocumented property on `Input` component:**

    ```jsx
    // In class body
    inputcomponentRef = React.createRef()

    handleResultSelect = (e, { result }) => {
      ...
      this.inputComponentRef.current.inputRef.current.blur()
    }

    // In render()
    <Search
      ...
      input={{ ref: this.inputComponentRef }}
    />
    ```

    Problem: Since it's undocumented, it shouldn't be relied on. See #1980.

    Actually the `Input` component provides a (documented) `.focus()` method directly, presumably to make this sort of thing easier. Since that's the case, maybe its `.blur()` counterpart deserves to be added too? But I feel like that's a slippery slope that just skirts around the issue.

3. **Function shorthand + `Ref` component:**

    ```jsx
    // In class body
    wrapperDivRef = React.createRef()

    handleResultSelect = (e, { result }) => {
      ...
      this.wrapperDivRef.current.querySelector('input.prompt').blur()
    }

    // In render()
    <Search
      ...
      input={(Input, inputProps) => (
        <Ref innerRef={this.wrapperDivRef}>
          <Input {...inputProps} />
        </Ref>
      )}
    />
    ```

    This would be nicer if the `.querySelector()` call could be avoided, but the markup requires the wrapper div. Also this uses `ReactDOM.findDOMNode()` under the hood, which is deprecated. It also seems like this might have slight performance implications? I realize I can extract the function to stop a new one being created on every render and maintain referential consistency, but there's still the overhead of the HOC if nothing else.

    Probably still the best option currently.

### Expected Result
I get a reference to the `<input>` element in a 'clean'/user-friendly manner.

### Version
0.87.3

noinkling:
...instead of overriding the hardcoded ones.

Fixes #3750.


layershifter:
Thanks for contributing, I will review these changes and will let you know soon.

src/modules/Search/Search.js
```diff
@@ -528,11 +532,12 @@ export default class Search extends Component {
       defaultProps: {
         ...rest,
         icon,
-        input: { className: 'prompt', tabIndex: '0', autoComplete: 'off' },
         onChange: this.handleSearchChange,
```
layershifter, line 548: Let's move out `autoComplete` & `tabIndex` to a higher level:

```tsx
      defaultProps: {
        ...rest,
        autoComplete: 'off',
        icon,
        input: { className: 'prompt' },
        onChange: this.handleSearchChange,
        onClick: this.handleInputClick,
        tabIndex: '0',
        value,
      },
```

***

Shorthand value will still win in all cases:
```tsx
<Input placeholder='Search...' tabIndex={1} input={{ tabIndex: -1 }} />
// <div class="ui input"><input type="text" placeholder="Search..." tabindex="-1" /></div>
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,10 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
+const overrideSearchInputProps = (inputProps) => ({
+  input: { className: 'prompt', tabIndex: '0', autoComplete: 'off', ...inputProps.input },
+})
```
layershifter: Shame on us, but we don't have deep handling of shorthands. This an edge case, actually.
It should be a bit more smart as we accept more things to shorthands that just an object:

```jsx
const overrideSearchInputProps = (predefinedProps) => ({
  const { input } = predefinedProps

  if (_.isUndefined(input)) return { className: 'prompt' }
  if (_.isPlainObject(input)) return { ...input, className: cx(input.className, 'prompt') }

  return input
}
```

- `isUndefined` will handle default case
- `isPlainObject` will merge `className` properly as in our implementation you will loose `prompt`

***

Let's also move it to class method:

```js
handleSearchInputOverrides = (predefinedProps) => ({
```

layershifter (changes requested):
Great job ⭐️ I left some comments, please address them when you will have time.

src/modules/Search/Search.js
noinkling: You're right. Since the functional form of `overrideProps` always provides an object, I accidentally conflated it by assuming that the nested `.input` property would be an object.

However with the suggested changes it now feels too hacky; factories.js already has logic for merging classnames (as well as the `style` prop) and determining which type of shorthand it is. Therefore I think it probably needs another approach.

noinkling:
As per my comment on the feedback, I think my attempt to delegate handling of the nested shorthand to the `Input` component using existing mechanisms failed.

Another approach I initially considered was to have `Search` call the `createHTMLInput()` factory itself to simulate the recursive behaviour desired. However it feels wrong not to delegate to `Input` for this, since:

1. It would be duplicating logic making it more brittle in case `Input` were to change.

2. It would be doing extra, unnecessary work during the render since it would run through `createHTMLInput()` twice.

So yeah, for these reasons I think nested shorthands probably need a more generic solution inside `createShorthand()` itself, e.g. by looking at the component's prop metadata, or the caller explicitly specifying the props for the nested shorthands that should be deep merged. Or maybe you could get away with an explicit utility function like `mergedPropsForShorthand()`.

Personally I don't think I'm up for such an involved feature, so feel free to close this PR if it's no longer useful.

src/modules/Search/Search.js
layershifter:
```suggestion
const overrideSearchInputProps = (predefinedProps) => ({
  const { input } = predefinedProps

  if (_.isUndefined(input)) {
    return { className: 'prompt' }
  }
  if (_.isPlainObject(input)) {
    return { ...input, className: cx(input.className, 'prompt') }
  }

  return input
}
```

test/specs/modules/Search/Search-test.js
```diff
@@ -774,6 +774,16 @@ describe('Search', () => {
...
   })
 
+  describe('input', () => {
+    it(`merges nested shorthand props for the <input>`, () => {
+      wrapperMount(<Search input={{ input: { tabIndex: '-1' } }} />)
+      const inputElement = wrapper.find('input')
+
+      inputElement.should.have.prop('tabIndex', '-1')
+      inputElement.should.have.className('prompt')
+    })
```
layershifter:
```suggestion
    it(`merges nested shorthand props for the <input>`, () => {
      wrapperMount(<Search input={{ input: { className: 'foo', tabIndex: '-1' } }} />)
      const input = wrapper.find('input')

      input.should.have.prop('tabIndex', '-1')
      input.should.have.className('foo')
      input.should.have.className('prompt')
    })
```

layershifter: I also added a case for class merge

src/modules/Search/Search.js
```diff
@@ -531,11 +544,12 @@ export default class Search extends Component {
       defaultProps: {
         ...rest,
         icon,
-        input: { className: 'prompt', tabIndex: '0', autoComplete: 'off' },
         onChange: this.handleSearchChange,
         onClick: this.handleInputClick,
```
layershifter, line 549:
```suggestion
        onClick: this.handleInputClick,
        tabIndex: '0',
```

src/modules/Search/Search.js
```diff
@@ -531,11 +544,12 @@ export default class Search extends Component {
       defaultProps: {
         ...rest,
         icon,
```
layershifter, line 547:
```suggestion
        autoComplete: 'off'
        icon,
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
+const overrideSearchInputProps = (predefinedProps) => ({
```
layershifter:
```suggestion
const overrideSearchInputProps = (predefinedProps) => {
```

src/modules/Search/Search.js
```diff
@@ -530,12 +543,15 @@ export default class Search extends Component {
       autoGenerateKey: false,
       defaultProps: {
         ...rest,
+        autoComplete: 'off'
```
layershifter:
```suggestion
        autoComplete: 'off',
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
+const overrideSearchInputProps = (predefinedProps) => {
+  const { input } = predefinedProps
+
+  if (_.isUndefined(input)) {
+    return { className: 'prompt' }
```
layershifter:
```suggestion
    return { predefinedProps: input: { className: 'prompt' } }
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
...
+  }
+  if (_.isPlainObject(input)) {
+    return { ...input, className: cx(input.className, 'prompt') }
```
layershifter:
```suggestion
    return { ...predefinedProps, input: { ...input, className: cx(input.className, 'prompt') } }
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
...
+  const { input } = predefinedProps
+
+  if (_.isUndefined(input)) {
+    return { predefinedProps: input: { className: 'prompt' } }
+  }
+  if (_.isPlainObject(input)) {
+    return { ...predefinedProps, input: { ...input, className: cx(input.className, 'prompt') } }
+  }
+
+  return input
```
layershifter:
```suggestion
  return predefinedProps
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
+const overrideSearchInputProps = (predefinedProps) => {
+  const { input } = predefinedProps
+
+  if (_.isUndefined(input)) {
+    return { predefinedProps: input: { className: 'prompt' } }
```
layershifter:
```suggestion
    return { ...predefinedProps: input: { className: 'prompt' } }
```

src/modules/Search/Search.js
```diff
@@ -27,6 +27,19 @@ import SearchResults from './SearchResults'
 
 const debug = makeDebugger('search')
 
+const overrideSearchInputProps = (predefinedProps) => {
+  const { input } = predefinedProps
+
+  if (_.isUndefined(input)) {
+    return { ...predefinedProps: input: { className: 'prompt' } }
```
layershifter:
```suggestion
    return { ...predefinedProps, input: { className: 'prompt' } }
```

test/specs/modules/Search/Search-test.js
```diff
@@ -774,6 +774,17 @@ describe('Search', () => {
...
 
+  describe('input', () => {
+    it(`merges nested shorthand props for the <input>`, () => {
+      wrapperMount(<Search input={{ input: { className: 'foo', tabIndex: '-1' } }} />)
+      const input = wrapper.find('input')
+
+      input.should.have.prop('tabIndex', '-1')
+      input.should.have.className('foo')
+      input.should.have.className('prompt')
+    })
```
layershifter, line 785:
```suggestion
    it(`will not merge for a function`, () => {
      wrapperMount(<Search input={{ input: (Component, props) => <Component {...props} /> }} />)
      const input = wrapper.find('input')

      input.should.have.prop('autoComplete', 'off')
      input.should.have.not.className('prompt')
    })
```

layershifter (approved):
I had some time to think about it 🤔

I don't think that we need a solution for all components as this use case is rare (actually a single one). I updated a PR and tests in it to cover more scenarios.

Thanks for raising this 👍

PR #3789, opened 2019-09-22
docs(DropdownExampleSearchSelection): updated countries

gisturiz:
countryOptions was expanded to include 250 main countries from A to Z.


layershifter:
Thank you for contributing, but all examples contain a limited set of counties to be representative and avoid long lists.

PR #3793, opened 2019-09-26
Move @types/react to dependencies

wolfd:
The TypeScript maintainers [recommend](https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338) putting `@types/*` packages in `dependencies` instead of `devDependencies`.

When using pnpm (a really awesome package manager that attempts to solve the issues with phantom dependencies), I have to manually add `@types/react` to the list of dependencies in order to get this package to work.

For more on phantom dependencies: https://rushjs.io/pages/advanced/phantom_deps/

Thanks!


layershifter:
Thank you for contributing, but I will not merge this PR.

First reason, typings are definitely not production dependency. Second, I haven't seen ever that other libraries do the same. Can you please refer to someone?

wolfd:
So the reason that it's required for TypeScript users is that Semantic-UI's type definitions have imports in them, which TypeScript has to resolve. Importing "react" from a type file, when I'm using the library, requires me to have 
`@types/react` installed, which it won't be if I'm running in NODE_ENV=production mode.

I shouldn't need any `devDependencies` of any package when I'm using it in my application, but in this case, I need to trick my package manager into changing the requirements of this library when it's installing packages to be able to use it.

Without it, I get cryptic errors like this:
```
TypeScript error in somefile.tsx(116,8):
JSX element type 'Search' does not have any construct or call signatures.  TS2604
    114 |     }
    115 |     return (
  > 116 |       <Search
        |        ^
...
```

For reference, if anyone is using pnpm + semantic-ui-react + TypeScript, and sees cryptic errors like above:

`pnpmfile.js`:
```
"use strict";
module.exports = {
  hooks: { readPackage }
};

function readPackage(pkg) {
  if (pkg.name === "semantic-ui-react") {
    // TODO(danny): remove when PR resolved
    // https://github.com/Semantic-Org/Semantic-UI-React/pull/3793
    pkg.dependencies["@types/react"] = "^16.4.14";
  }
  return pkg;
}
```

For more discussion on this:
https://stackoverflow.com/a/46011417
https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338

Angular seems to put types in deps: https://github.com/angular/angular/blob/fbad4ff65ca73cb7eec16ec5a3e8980014df6d74/package.json#L48-L65

However, in doing a bit more reading, it looks like it might be a good idea to put `@types/react` in `peerDependencies`. https://github.com/microsoft/types-publisher/issues/81#issuecomment-272250338 Seems to suggest that it should mirror, but there's no official comment on it. I just asked in the thread if that's a good solution.

wolfd:
@layershifter did you get a chance to consider my additional comments?

nihalanisumit:
@wolfd I am having this issue 
TS2604: JSX element type 'Header' does not have any construct or call signatures.

https://stackoverflow.com/questions/61569712/ts2604-jsx-element-type-header-does-not-have-any-construct-or-call-signature
Any help would be appreciated 
cc: @layershifter

PR #3794, opened 2019-09-26
docs(SearchExampleStandard): use React hooks

rpatelcode:
Converted into Functional functional component


layershifter (approved):
Thanks for contributing, I did few changes. It will be great if we can do update remaining examples...

PR #3822, opened 2019-10-28
feat(FormField): make form field error accessible

### LINKED ISSUE #3821: Form field errors are not accessible for screen readers
### Problem description
Currently form field errors are by default not accessible for screen readers. The error message is not pronounced and the input is not marked as invalid. 

### Proposed solution
Use the `aria-describedby` attribute to link the error message to the input. Additionally, use `aria-invalid` attribute to indicate that the input is invalid.

aoelen:
Fixes #3821 

Form fields should have an id attribute in order for it to work (since aria-describedby needs an id).

Relates to: https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=331#error-identification. More specifically, it relates to techniques: [ARIA19](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA19.html) and [ARIA21](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21.html)

### DIFF
--- a/docs/src/examples/collections/Form/Shorthand/FormExampleFieldControlId.js
+++ b/docs/src/examples/collections/Form/Shorthand/FormExampleFieldControlId.js
@@ -38,6 +38,16 @@ const FormExampleFieldControlId = () => (
       placeholder='Opinion'
     />
     <Form.Field
+      id='form-input-control-error-email'
+      control={Input}
+      label='Email'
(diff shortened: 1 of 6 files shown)


snitin315 (commented):
Nice Work @aoelen , but it seems your checks were not successful , check it once.

levithomason:
Thanks much and happy holidays 🎄 ☃️

PR #3834, opened 2019-11-08
refactor(Checkbox): remove hardcoded hidden classname

wgeorgecook:
The checkbox has a nonsensical 'hidden' classname which sets a
z-index of -1 for the browser. I don't see any reason for this
documented and it can cause grief in your styles. I removed it
in this commit.

### DIFF
--- a/src/modules/Checkbox/Checkbox.js
+++ b/src/modules/Checkbox/Checkbox.js
@@ -293,7 +293,6 @@ export default class Checkbox extends Component {
           <input
             {...htmlInputProps}
             checked={checked}
-            className='hidden'
             disabled={disabled}
             id={id}
             name={name}


wgeorgecook:
I did go through the history to see why this was included, but couldn't find a specific commit since others were masking it. This className was not in the original commit, so I suspect it leaked in and just hadn't caused any harm.

levithomason:
There is a reason documented in the LESS stylesheet:

https://github.com/Semantic-Org/Semantic-UI-LESS/blob/master/definitions/modules/checkbox.less#L255

```less
/* Initialized checkbox moves input below element
 to prevent manually triggering */
.ui.checkbox input.hidden {
  z-index: -1;
}
```

The styles are shared between multiple framework implementations, so changes need to work in all cases.

Can you elaborate on what issues it is causing?

wgeorgecook:
@levithomason this specifically was causing issues with Selenium testing. The style was preventing the driver from clicking elements. I still don't understand why we're wanting to prevent manual triggers here.

PR #3836, opened 2019-11-13
fix(FeedSummary): add missing space around content

ridvankaradag:


levithomason:
Thanks for noticing and contributing!  It seems this is because there are missing required spaces before and after the `content`.  I've added this logic to the `FeedSummary` component so it is always correct.

PR #3848, opened 2019-11-26
fix(Search): use result.id for SearchResult key

unbug:
In many cases, the title value is not unique, which leads to the SearchResult never got updated. The solution is taking a new `key` property or the `id` property as the component key.


levithomason:
Thank you and happy holidays 🎄🎅!

PR #3881, opened 2020-01-09
docs(LabelExampleImageColored): update example to match image names

isnifer:
Replace Veronika with jenny due to image name


MonkeyAlex (approved):
Looks correct.
If you're keen, you may also correct the next label from Helen to Christian (see line#18)

isnifer:
@MonkeyAlex did it)

PR #3901, opened 2020-02-24
fix(FormFiled): pass `id` when contains child elements or has `content` prop

### LINKED ISSUE #3900: Form.Field - Cannot set ID when contains child elements or content prop
## Bug Report
When creating a `<Form.Field>` with child elements or content prop, ID is ignored and cannot be set on the div that is created from the `<Form.Field>`.

This bug is caused by the following PR: https://github.com/Semantic-Org/Semantic-UI-React/pull/3822
Since the PR added `id` prop as one of the handled props for this component if it is provided by the user it will be filtered out on the `getUnhandledProps` function.

When no control is provided the created element props will be taken from the `rest` variable.
Because the `id` was filtered out from the `rest` variable it will not be passed to the child elements or the content that was set.

Let me know if you agree that this is a bug and I can open a PR for the fix and review

### Steps
Create a form and add <Form.Field>:
```
<Form>
    <Form.Field id="testId" content="Some content"/>
 </Form>
```

### Expected Result
```
<form class="ui form">
    <div id="testId" class="field">Some content</div>
</form>
```

### Actual Result
```
<form class="ui form">
    <div class="field">Some content</div>
</form>
```

### Version
0.88.2

### Testcase
https://codesandbox.io/s/semantic-ui-react-9gnsc

srotbart:
Fixes https://github.com/Semantic-Org/Semantic-UI-React/issues/3900

Fixed issue where ID cannot be set when contains child elements or content props for FormField


srotbart:
@levithomason 
Hi Levi, could you please address this PR?

src/collections/Form/FormField.js
```diff
@@ -75,17 +75,18 @@ function FormField(props) {
   // No Control
   // ----------------------------------------
 
+  const noControlProps = { ...rest, id }
   if (_.isNil(control)) {
     if (_.isNil(label)) {
       return (
-        <ElementType {...rest} className={classes}>
+        <ElementType {...noControlProps} className={classes}>
```
layershifter:
```suggestion
        <ElementType {...rest} className={classes} id={id}>
```

src/collections/Form/FormField.js
```diff
@@ -75,17 +75,18 @@ function FormField(props) {
...
           {childrenUtils.isNil(children) ? content : children}
         </ElementType>
       )
     }
 
     return (
-      <ElementType {...rest} className={classes}>
+      <ElementType {...noControlProps} className={classes}>
```
layershifter:
```suggestion
      <ElementType {...rest} className={classes} id={id}>
```

src/collections/Form/FormField.js
```diff
@@ -75,17 +75,18 @@ function FormField(props) {
   // No Control
   // ----------------------------------------
 
+  const noControlProps = { ...rest, id }
```
layershifter:
```suggestion
```

layershifter:
I was a bit confused with a use case for this as according to spec input elements should have `id` (https://www.w3.org/TR/WCAG20-TECHS/H44.html) to be linked with labels. However, at the same time it's frustrating that `id` is passed nowhere.

Thanks for addressing this 👍

PR #3902, opened 2020-02-27
Replaced image urls

akhilaravind:
Images was not rendering in the sandbox fiddles, so replaced image URL so that it will render the image in sandbox fiddles. Additional url was added to image src that was preventing the image render.


rogemus (approved):

layershifter:
@akhilaravind thanks for contributing!

However, as you may mention in this case it will not be enough to fix a single example. So I did a fix #3758 for all examples some time ago.

Please let us know if the issue still exists. I verified it on this example and it works like a charm.

PR #3916, opened 2020-03-25
fix(Input): add 'inputMode' attribute to input

RoyalHunt:
The inputmode attribute provides a hint to browsers for devices with onscreen keyboards to help them decide which keyboard to display.


layershifter:
@RoyalHunt good catch 👍

layershifter (approved):

Issue #3917, opened 2020-03-25
Wrong position of Popup component


saramorillon:
### Steps
- Open the documentation page of the Popup component, section "Position"
- Or reproduce the example given in the document in your own project

### Expected Result
The position of the different popups should be correct.

### Actual Result
The position of the **top left** and **top right** one hand, and on the other hand **bottom left** and **bottom right** popups are reverted.

### Version
0.88.2

### Testcase
https://react.semantic-ui.com/modules/popup/#variations-position

### Screenshots

![image](https://user-images.githubusercontent.com/4224278/77584734-9da9bd80-6ee3-11ea-96bb-33f42f31a616.png)

----

![image](https://user-images.githubusercontent.com/4224278/77584761-a7332580-6ee3-11ea-8bcf-c5102bdfd17d.png)

brianespinosa:
@saramorillon taking a look at the way the position class names are implemented and [how they correspond to Semantic UI](https://semantic-ui.com/modules/popup.html#positioning), the positioning is indeed consistent. We rely on the parent SUI project as our source of truth on functionality.

PR #3921, opened 2020-03-29
fix(Popup): fixing the popup position

Yuval1231:
Regarding to issue [#3917](https://github.com/Semantic-Org/Semantic-UI-React/issues/3917)

* Making a new hash that provides the right positions of the placements;
* Changing the position of the popup;

**Before**:
popup - before

**After:**
popup - after

### DIFF
--- a/src/modules/Popup/lib/positions.js
+++ b/src/modules/Popup/lib/positions.js
@@ -2,12 +2,12 @@ import _ from 'lodash'
 
 export const positionsMapping = {
   'top center': 'top',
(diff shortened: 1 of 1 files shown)


brianespinosa:
@Yuval1231 please see my comment and link in the related issue report. It does appear that we are consistent with Semantic UI in terms of how these popups are implemented. This PR would introduce a breaking change and change existing functionality for all users of this library.

I am going to go ahead and close this PR based on the fact that current implementation is consistent with the parent library.

PR #3922, opened 2020-03-29
fix(TransitionablePortal): fixed deprecated lifecycle method #2732

shachar731945:
* removed deprecated componentWillReceiveProps lifecycle

* changed it to componentDidUpdate

* changed test description to componentDidUpdate

it is worth mentioning that it is not advised to use getDerivedStateFromProps in this case due to the different time this hook is getting called in contrast to componentDidUpdate which also does not require change from any of the test

### DIFF
--- a/src/addons/TransitionablePortal/TransitionablePortal.js
+++ b/src/addons/TransitionablePortal/TransitionablePortal.js
@@ -76,11 +76,10 @@ export default class TransitionablePortal extends Component {
   // Lifecycle
   // ----------------------------------------
 
-  // eslint-disable-next-line camelcase
-  UNSAFE_componentWillReceiveProps({ open }) {
-    debug('componentWillReceiveProps()', { open })
-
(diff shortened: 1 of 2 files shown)


shachar731945:
my circle ci checks are not working because i have setState in componentDidUpdate. I would like to know if this check is missused in the circleci checks or is there an actual reason for it. thanks from ahead.

NicholasG04:
See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md maybe as to why?

layershifter:
Oops, folks 😿 I missed that this PR exists...

I did an update in #3966 and you can check it if you're interested. In general when there is `UNSAFE_componentWillReceiveProps` that performs `setState()` it should be replaced with `getDerivedStateFromProps()` as otherwise `componentDidUpdate()` will cause double render.

PR #3931, opened 2020-04-09
Added a product built using Semantic UI React

yomeshgupta:
Added https://code.devtools.tech in the `Built With` section. The product using the latest version of Semantic UI React bundled using Webpack 4.


layershifter:
@yomeshgupta thanks for contributing, however I removed this section recently as there were many outdated resources that creates more confusion than trust to the project. Sorry about that.

Issue #3933, opened 2020-04-10
When supplying a label to a FormField children do not render


reduxdj:
## Bug Report
Noticed this, and it baffled me a bit

### Steps
```jsx
        <Form.Field tabIndex={4} label="some label" error={formit.errors.dob}>
          <input />
        </Form.Field>
```
### Expected Result
The result that you expected.

Some Label
Input Box

### Actual Result
Some Label

### Version
.88.2

### Testcase
[Fork, update, and replace this pen to show the bug]:
https://codesandbox.io/s/2l3n74j9y

calebdeji:
Would like to work on this, is this open?

brianespinosa:
Please create a reproducible test case using the supplied Codesandbox link at the bottom of the bug report template.

Closing this report until a reproducible case is provided.

layershifter:
This is expected behavior you should use Children or Shorthand API. I suggest to use `Form.Input` in this case, all scenarios should be possible with it.

I created CodeSandbox (https://codesandbox.io/s/kind-sun-7zfgn?file=/example.js), but it has weird errors, however in local delevopment the warning is fully visible:

```
Warning: Failed prop type: Prop `label` in `FormField` conflicts with props: `children`. They cannot be defined together, choose one or the other.
```

This also is mentioned in our docs:

```
  // Heads Up!
  // Do not disallow children with `label` shorthand
  // The `control` might accept a `label` prop and `children`
  /** Mutually exclusive with children. */
```

PR #3937, opened 2020-04-21
Fixed : FormField cannot render children if control is null. Prior to…

calebdeji:
Prior to this commit, when supplying a label without control to a FormField, children do not render. Since Formfield should render children irrespective of the absence of control, FormField will render the children if present.

Relate to https://github.com/Semantic-Org/Semantic-UI-React/issues/3933


layershifter:
Thanks for contributing ❤️ 

However, it's an expected behavior, please check my comment https://github.com/Semantic-Org/Semantic-UI-React/issues/3933#issuecomment-665700324.

PR #3938, opened 2020-04-27
fix(Search): fix an error when `showNoResults=false` and arrow down is pressed

ysugimoto:
This PR fixes an unexpected error on `Search` module.

When the Search module built with `showNoResults={false}` props, the `<div class="result">` element won't present in DOM. And if a user types cursor-down (keycode: 40) in search input,  then it raises an error:

```
Search.js:260 Uncaught TypeError: Cannot read property 'querySelector' of null
    at Search.eval [as scrollSelectedItemIntoView] (Search.js:260)
    at Search.eval [as moveSelectionBy] (Search.js:252)
    at Array.eval (Search.js:113)
    at EventSet.dispatchEvent (event-stack.development.js:164)
    at EventPool.dispatchEvent (event-stack.development.js:275)
    at eval (event-stack.development.js:341)
    at Map.forEach (<anonymous>)
    at HTMLDocument.eval (event-stack.development.js:340)
```

This PR can avoid that error by checking `menu` element existence.


layershifter (approved):
Verified, the issue is fixed 👍

Issue #3771, opened 2019-09-02
Popup: Incorrect position of arrow


dmitriirudakov:
### Steps
1. Create an element with a popup
2. Move that element to the right side of the page

### Expected Result
I don't know an exact position of a trigger element on a page in advance but I expect popup and popup arrow will be displayed correctly everywhere as it was in 0.86.0.
![old](https://user-images.githubusercontent.com/17123205/64114274-a39c4d80-cd9d-11e9-8a9d-f70409ca4961.gif)

### Actual Result
Popup arrow position is incorrect.
The package doesn't calculate a popup position based on an actual popup trigger position.
It passes down a position from props to popper and applies css styles based on that instead of calculating a correct value. 
In the previous version a position was changed to 'top right' but in current version it's 'top left' from defaultProps always.
![new](https://user-images.githubusercontent.com/17123205/64114293-adbe4c00-cd9d-11e9-9af0-c68fbb076bdb.gif)

### Version
0.88.0

### Testcase
https://codesandbox.io/s/semantic-ui-react-o7mfd

twens:
similar issue
left - before semantic update(without popper js), rigth - after
![image](https://user-images.githubusercontent.com/15820414/69047493-e4eae380-0a03-11ea-8862-98e47acef489.png)

twens:
looks like it is related to this
Issue when use sidebar- it does not overlay panel (outside)
![image](https://user-images.githubusercontent.com/15820414/69048010-2334d280-0a05-11ea-88a6-75dd865cbf63.png)

Issue #3819, opened 2019-10-27
Use forward refs in Components to access ref instances


mxschmitt:
### Problem description

Currently, not all of components properly forward the passed in `ref` prop to the underlying component. This leads to end-users not being able to access the underlying DOM node and execute their browsers native functions.

For example when you try to execute `checkValidity` on the form element itself.

### Proposed solution

Upgrade the components to forward refs: https://reactjs.org/docs/forwarding-refs.html

### MVP

https://codesandbox.io/s/naughty-solomon-fqk4f?fontsize=14

Here you see who examples, the current `<Form>` component and the native browsers one. The ref instances are totally different.

pasticks:
Here's a workaround that takes advantage of semantics 'as' prop and prop forwarding to give you access to the rendered DOM element. It'll work with any semantic components, you just need to pass in the render type, which in most cases will probably be the default.

[https://codesandbox.io/s/semantic-ui-react-refs-workaround-oq4zk](https://codesandbox.io/s/semantic-ui-react-refs-workaround-oq4zk)

patrickp-dev:
Here's a workaround that takes advantage of semantics 'as' prop and prop forwarding to give you access to the rendered DOM element. It'll work with any semantic components, you just need to pass in the render type, which in most cases will probably be the default.

https://codesandbox.io/s/semantic-ui-react-refs-workaround-ibnnv

jacobweber:
@patrickp-dev Thanks for the workaround, but it seems to be causing some bad side-effects for me. I added an controlled Input and a useState hook to your example:

https://codesandbox.io/s/semantic-ui-react-refs-workaround-2hr0e

But with this workaround, the input loses focus after typing.

Not sure what's happening, but I'm wondering if there's any other way to work around this.

(Use case: I'm trying to capture keyboard events in a Modal, which means I need to focus() the modal's div when it's rendered. So I need a ref to focus. I saw some mention of a Ref component in other tickets, but it seems to be removed.)

patrickp-dev:
@jacobweber Hey jacob, had a look at your example. The lose of input focus was caused by the fact the 'as' prop was being reassigned a new instance of the higher order component that was returned by the forwardRefAs on each rerender, (my bad). All that's required is pulling the HOC instantiation out of the 'App' component so that it doesn't remount the entire form on each state change. I've updated your example to get it working and have update my original example. https://codesandbox.io/s/semantic-ui-react-refs-workaround-h0f64

[...]

PR #3947, opened 2020-05-21
chore: update `react-popper` to v2

ayasakov:
Popper.js v1 is out of date, v2 was released in Dec 2019, time to upgrade 🚀

---

# ⚠️ BREAKING CHANGES

`Popup` component will use `@popperjs/core` v2. As result, there are two breaking changes in `Popup` component itself.

## `offset` can't be specified as `string` anymore

Previously it was possible to pass different units to the `offset` prop as units. This behavior was changed and `offset` accepts a tuple or a function. Examples in our docs were also updated.

### Before

```jsx
<>
  <Popup offset="50px" />
  <Popup offset="-100%p" />
</>
```

### After

```jsx
<>
  <Popup offset={[50, 0]} />
  <Popup offset={({ popper }) => [-popper.width, 0]} />
</>
```

### `popperModifiers` are defined as arrays now

The `popperModifiers` prop should be present as an `array` with changed format (see [Popper docs](https://popper.js.org/docs/v2/migration-guide/#10-update-modifiers)).

### Before

```jsx
<Popup popperModifiers={{ preventOverflow: { padding: 0 } }} />
```

### After

```jsx
<Popup popperModifiers={[{ name: "preventOverflow", options: { padding: 0 } }]} />
```

### a wrapping element around `.popup`

To avoid warnings from Popper.js about `margin`s (https://popper.js.org/docs/v2/migration-guide/#4-remove-all-css-margins) we added a wrapping element around `.popup`:

```html
<div>
  <div class="ui popup">A popup content</div>
</div>
```

In #4094 the automatic transfer of `zIndex` was added.


layershifter:
@ayasakov this is great, love it ❤️ 

_But_, as you mentioned this is a breaking change. I will think how we can stage it, probably this will be for a next release as in a current I would like to remove warnings related to `UNSAFE_` methods.

ayasakov:
I rebased my PR to the last master code. There is a console warning from the Popper.js regarding `margin` position in styles. The following lines can fix it:
```
.ui.top.popup {
  margin: 0;
}
.ui.left.center.popup {
  margin: 0;
}
.ui.right.center.popup {
  margin: 0;
}
.ui.bottom.popup {
  margin: 0;
}
```
Looks need to add an additional change in another Semantic repo.

@layershifter are there any updates? That would be great to merge it in the next release :)

layershifter:
@ayasakov just to clarify what is blocking there 🤔 

1. I like Popper v2, I like changes, great job 👍 
2. We released v1 recently and started to follow Semver as this PR contains breaking changes it can be included only to a next major release i.e. v2.
3. There are changes that should be done to be fully compatible with `StrictMode` (#3819), they will also require major release... However, I don't think that they can be done in next two weeks, it will be too optimistic. 

To conclude: I don't want to block the progress there as Popper v2 is really awesome. There will be also another small release with few deprecations and then we can go to v2. To give you ETA I expect that this can be released in two-three weeks.

Please let me know that do you think 🙏

ayasakov:
@layershifter sounds great! I'm waiting.

src/modules/Popup/Popup.js
```diff
@@ -375,8 +372,7 @@ Popup.propTypes = {
 
 Popup.defaultProps = {
   disabled: false,
-  eventsEnabled: true,
```
layershifter, line 378: As `eventListeners` is available via `modifier` I decided to keep prop to avoid an additional breaking change

src/modules/Popup/Popup.js
```diff
@@ -298,14 +293,16 @@ Popup.propTypes = {
...
+   * array with two numbers in the form [skidding, distance].
+   *
+   * The first number, skidding, displaces the popper along the reference element.
+   *
+   * The second number, distance, displaces the popper away from, or toward, the
+   * reference element in the direction of its placement. A positive number displaces
+   * it further away, while a negative number lets it overlap the reference.
    */
-  offset: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  offset: PropTypes.arrayOf(PropTypes.number),
```
layershifter: `offset` can be also a function, I added an example for this

src/modules/Popup/Popup.js
```diff
@@ -214,17 +213,17 @@ export default class Popup extends Component {
...
-      popperModifiers,
-    )
+    const defaultModifiers = [
+      { name: 'arrow', enabled: false },
+      { name: 'flip', enabled: !pinned },
+      { name: 'preventOverflow', enabled: !!offset },
+      { name: 'offset', options: { offset } },
+    ]
+    const modifiers = popperModifiers
+      ? _.unionBy(popperModifiers, defaultModifiers, 'name')
```
layershifter: Merge there is not required as Popper will do it for us 😺 

https://github.com/popperjs/popper-core/blob/f1f9d1ab75b6b0e962f90a5b2a50f6cfd307d794/src/createPopper.js#L97-L99

ayasakov:
I like it! Great work!

layershifter:
> I rebased my PR to the last master code. There is a console warning from the Popper.js regarding `margin` position in styles. The following lines can fix it:

_One drop of poison infects the whole tun of wine_ 🤦 

@ayasakov Definitely this is the worst part there as we don't have control over SUI/FUI CSS. I don't have any better ideas that 1236630 as we definitely cannot go with warnings.

@ayasakov It will also not fix #3771 as Popper does not take `arrow` into account as it's defined in SUI CSS as preudo-element.

layershifter (approved):
@ayasakov great work there 👍 I pushed few minor changes & update PR's description.

If I missed something please let me know 🙏

layershifter:
Released in `semantic-ui-react@2.0.0` 🚢

PR #3964, opened 2020-06-16
docs(TableExampleSortable): your description

eddy2048:
Updated for react hooks


docs/src/examples/collections/Table/Variations/TableExampleSortable.js
```diff
@@ -72,6 +61,8 @@ export default class TableExampleSortable extends Component {
           ))}
         </Table.Body>
       </Table>
-    )
-  }
+    </div>
+  );
 }
+
+export default Table;
```
layershifter:
```suggestion
```

docs/src/examples/collections/Table/Variations/TableExampleSortable.js
```diff
@@ -1,62 +1,51 @@
...
-    data: tableData,
-    direction: null,
-  }
-
-  handleSort = (clickedColumn) => () => {
-    const { column, data, direction } = this.state
+function Table() {
+  const [column, updateColumn] = useState(null);
+  const [data, updateData] = useState(tableData);
+  const [direction, updateDirection] = useState(null);
```
layershifter: It seems that `useRedurer` will be a better solution in this case. Can we use it?

docs/src/examples/collections/Table/Variations/TableExampleSortable.js
```diff
@@ -1,62 +1,51 @@
...
-export default class TableExampleSortable extends Component {
-  state = {
-    column: null,
-    data: tableData,
-    direction: null,
-  }
-
-  handleSort = (clickedColumn) => () => {
-    const { column, data, direction } = this.state
+function Table() {
```
layershifter:
```suggestion
export default TableExampleSortable() {
```

- default export works for functions 😸 
- let's keep existing example name

docs/src/examples/collections/Table/Variations/TableExampleSortable.js
```diff
@@ -1,62 +1,51 @@
...
+import React, { useState } from "react";
+import { Header, Table } from "semantic-ui-react";
 
 const tableData = [
-  { name: 'John', age: 15, gender: 'Male' },
-  { name: 'Amber', age: 40, gender: 'Female' },
-  { name: 'Leslie', age: 25, gender: 'Other' },
-  { name: 'Ben', age: 70, gender: 'Male' },
-]
+  { name: "John", age: 15, gender: "Male" },
```
layershifter: It looks that you have a different formatting there. Please run `yarn prettier:fix` & `yarn lint:fix` it should fix your issues and pass build.

docs/src/examples/collections/Table/Variations/TableExampleSortable.js
```diff
@@ -1,62 +1,51 @@
 import _ from 'lodash'
-import React, { Component } from 'react'
-import { Table } from 'semantic-ui-react'
+import React, { useState } from "react";
```
layershifter:
```suggestion
import React from "react";
```

I suggest to use `React.useState`/`React.useReducer` as it gives more context from where hooks came.

PR #3985, opened 2020-07-14
chore: use `react-intersection-observer` in docs to improve perf

### LINKED ISSUE #3978: The docs (https://react.semantic-ui.com) cause excessive CPU usage in Chrome
## Bug Report

NOTE: Issue #3784 was accidentally closed (the PR to close was a README fix, so no way it solved the issue). This is a direct copy/paste from #3784 as this issue still persists (Sitting on the docs for semantic-ui easily causes 50% cpu usage)

### Steps
Use Chrome to open the documentation pages for any component which contains examples

### Expected Result
Chrome should not use more CPU than with other similar sites

### Actual Result
According to Chrome's dev tools, the Rendering task "Recalculate Styles" runs continuously.
After a few moments my laptops fan kicks in and the laptop turns into a heater

### Version
0.88.2

### Screenshot
Screenshot

knight-coder-llc:
Fixes #3978.

Added react intersection observer to docs component examples to improve cpu usage over getBoundingRect function call. Replaced react visibility sensor.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js
@@ -3,7 +3,8 @@ import cx from 'classnames'
 import copyToClipboard from 'copy-to-clipboard'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import VisibilitySensor from 'react-visibility-sensor'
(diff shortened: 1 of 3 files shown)


layershifter (approved):
Pushed few more changes, LGTM. Also verified SSR scenario for docs, works as expected

Thanks ❤️

PR #4004, opened 2020-07-29
fix(Dropdown): handle keyboard events for arrows, space and enter keys locally

### LINKED ISSUE #3994: Dropdown: `allowAdditions` with Space key does not work
Hi,

Why does not the dropdown component allow for space character from version 0.88.0?
It works nice in version 0.87.3.
error

layershifter:
Fixes #3994.

---

This PR is focused on the single thing to move event handling from `document` (i.e. `EventStack`) to `Dropdown`s scope as it creates various issues when multiple `Dropdown` are opened. And there are no reasons to have these listeners globally.

### DIFF
--- a/docs/src/examples/modules/Dropdown/States/DropdownExampleDisabledItem.js
+++ b/docs/src/examples/modules/Dropdown/States/DropdownExampleDisabledItem.js
@@ -8,7 +8,7 @@ const options = [
 ]
 
 const DropdownExampleDisabledItem = () => (
-  <Dropdown text='Dropdown' options={options} open />
+  <Dropdown text='Dropdown' options={options} />
 )
 
 export default DropdownExampleDisabledItem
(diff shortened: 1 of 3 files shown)


docs/src/examples/modules/Dropdown/States/DropdownExampleDisabledItem.js
```diff
@@ -8,7 +8,7 @@ const options = [
 ]
 
 const DropdownExampleDisabledItem = () => (
-  <Dropdown text='Dropdown' options={options} open />
+  <Dropdown text='Dropdown' options={options} />
```
layershifter, line 11: This `Dropdown` should not be be opened by default,

src/modules/Dropdown/Dropdown.js
```diff
@@ -52,28 +52,17 @@ export default class Dropdown extends Component {
 
   static getAutoControlledStateFromProps(nextProps, computedState, prevState) {
     // These values are stored only for a comparison on next getAutoControlledStateFromProps()
-    const derivedState = { __value: nextProps.value, __options: nextProps.options }
+    const derivedState = { __options: nextProps.options, __value: computedState.value }
```
layershifter, line 55: This should use `computedState.value` as we want to compute new `selectedIndex` based the component's state

Issue #3948, opened 2020-05-21
Responsive: HomepageLayout not rendering properly with SSR on mobile mode.


ysfBenali:
Check the issue on this link : https://react.semantic-ui.com/layouts/homepage/ 

![menu](https://user-images.githubusercontent.com/26151852/82568199-fefdae00-9b6d-11ea-9f56-7e78c8775424.PNG)

geraldoramos:
**Plus one for this issue**

I spent hours thinking this was an SSR issue related to next.js but after testing the [original](https://react.semantic-ui.com/layouts/homepage/) I confirmed this is not working when loading on mobile.

In order to reproduce on the desktop using the developer tools, make sure to load the site starting as a mobile device (ie: Iphone X), If you start as a desktop and then resize, it works perfectly. The issue is when it's loaded initially as a mobile device.

Sometimes my nextjs cli spits out this error, but I'm not sure if it's related:

```
Warning: Failed prop type: Element is not defined
    in Sidebar (at pages/index.jsx:162)
    in MobileContainer (at pages/index.jsx:212)
```

Briggsdf:
Any updates on this? We're having this issue on our website as well https://gmgt.live .

Briggsdf:
Just for any future people who encounter this. Install NoSSR and then do this,
```
 <NoSSR>
    <MobileContainer>{children}</MobileContainer>
    <DesktopContainer >{children}</DesktopContainer>
  </NoSSR>
```

deniapps:
This is the issue due to SSR. As you see in the code:
```
// Heads up!
// We using React Static to prerender our docs with server side rendering, this is a quite simple solution.
// For more advanced usage please check Responsive docs under the "Usage" section.
const getWidth = () => {
  const isSSR = typeof window === 'undefined'

  return isSSR ? Responsive.onlyTablet.minWidth : window.innerWidth
}
```
The getWidth() does not check SSR from mobile. In which, isSSR is false, so then the width equals to Tablet.minWidth, and then the layout messed up. 

The solution is using "mobile-detect" npm to detect the mobile device from the server side by checking the request header's user-agent. The code like this: (copied from **Responsive docs under the "Usage" section**)
```
//NOTE: I add this in _app.js (NextJS), then save it in the context, 
//so I don't need to call this on every page
deniApp.getInitialProps = async (appContext) => {
  const agent = appContext.ctx.req
    ? appContext.ctx.req.headers["user-agent"]
[...]

Issue #3398, opened 2019-01-23
Homepage layout problem


username7171:
I think there is a bug in the official demonstration in semantic-ui-react

https://react.semantic-ui.com/layouts/homepage/

It is Okay when you open it the first time, however, when you refresh the page, it becomes messy. Where is the bug? 

A clear and concise description of steps to reproduce the problem.

### Expected Result
![image](https://user-images.githubusercontent.com/36737465/51613889-76620400-1f5f-11e9-9cf8-c63b018f4f59.png)

### Actual Result
![image](https://user-images.githubusercontent.com/36737465/51613901-7eba3f00-1f5f-11e9-8e52-e3b88c83a7d5.png)

layershifter:
Was fixed in #3373.

username7171:
> Was fixed in #3373.
Sorry but I think the bug remains,
I have downloaded the code and run on local to test it. 
You can also try to visit 
https://react.semantic-ui.com/layouts/homepage/
and refresh the page to reproduce the bug.

layershifter:
It was fixed, but was not released yet 🏷  Served locally (`yarn build:docs:staging`) production build and everything is fine there:

![image](https://user-images.githubusercontent.com/14183168/51618372-dee8bb00-1f36-11e9-84a6-64c285f26d9e.png)

username7171:
Thanks so much for your reply. May I ask one silly question?
![image](https://user-images.githubusercontent.com/36737465/51619227-d27e5580-1f6a-11e9-9f9a-af3edac70ac1.png)
There are 3 files changed corresponding to this issue.
I am already using the updated version of the middle file , which is HomepageLayout.js.
But the bug remains. Is it because HomepageLayout.js. file rely on the first and third file?

layershifter:
> But the bug remains.

Sorry, what do you mean? I don't understand. You have issues with `Responsive` and SSR?
Please check https://github.com/Semantic-Org/Semantic-UI-React/issues/3361#issuecomment-453202401 and https://codesandbox.io/s/ly3zpw7yzm

username7171:
Oh.Thank you so much. All 3 solutions provided by you are great. The NoSSR component works perfectly.

Issue #3110, opened 2018-08-28
Responsive: Not rendering correctly with SSR


teoraunio:
## Bug Report
I am trying to create a top bar which uses two Responsive elements to determine if it should render content for over or under 1000 pixel resolution. The app is using server side rendering.

The problem _(code below to understand the props and classes)_:
When the window size is over 1000 pixels, it will render a Responsive element that has `minWidth={1000}` as expected, but the first element inside the Responsive is actually from the other Responsive element with `maxWidth={1000}`

So what happens is:
1. The server renders the element with `maxWidth=1000` as `window` doesn't yet exist and the width is set to 0 inside the Responsive component.
2. Client re-renders it according to acquired window width and now chooses to use the Responsive element with `minWidth=1000`
3. The `<div>` element that should have class `desktop-content-1` has falsely class `mobile-content` that is defined inside the other `Responsive` element.

I am aware that this can be fixed by wrapping both Responsive elements in separate containers, but it means using unnecessary elements and it should at least be specified in the documentation.

### Steps
You need an app that uses Server Side Rendering with the following piece of code rendered to reproduce the problem:

```
<div className="top-bar">
  <Responsive minWidth={1000}>
    <div className="desktop-content-1" />
    <div className="desktop-content-2" />
  </Responsive>
  <Responsive maxWidth={1000}>
    <div className="mobile-content" />
  </Responsive>
</div>
```

### Expected Result
The width of the window is over 1000 pixels.

```
<div class="top-bar">
  <div>
    <div class="desktop-content-1" />
    <div class="desktop-content-2" />
  </div>
</div>
```

### Actual Result
The width of the window is over 1000 pixels.

```
<div class="top-bar">
  <div>
    <div class="mobile-content" />
    <div class="desktop-content-2" />
  </div>
</div>
```

### Version
0.77.1

layershifter:
@teoraunio thanks for the detailed report!

1. Please update to latest version :cat: 
2. The latest version contains the [`getWidth`](https://github.com/Semantic-Org/Semantic-UI-React/blob/v0.82.3/src/addons/Responsive/Responsive.js#L50) prop, by default width will be always `0` for SSR. You can override it with this prop.
3. Use `fireOnMount` prop, it should solve your issue, because listeners are fired on `resize` & `scroll` by default.

```jsx
const getWidth = () => {
  if (isBrowser) return window.innerWidth
  return 1000
}

<div className="top-bar">
[...]

PR #4008, opened 2020-07-30
docs(Responsive): deprecate component

### LINKED ISSUE #4002: Responsive: does not initialize correctly when used with Gatsby
## Bug Report

Site generated with Gatsby produces a page that doesn't properly set the DOM on page load from `Responsive`. The page works fine when used with vanilla react and with Gatsby develop.

A test repo https://github.com/sveinse/gatsby-semantic-ui-bug has been created. Please see its README.md. The issue is related to `DesktopContainer` and `MobileContainer` in https://github.com/sveinse/gatsby-semantic-ui-bug/blob/master/src/components/layout.js

I'm not sure at this stage if this is a Gatsby issue or a semantic-ui-react issue. The issue has been filed with Gatsby at https://github.com/gatsbyjs/gatsby/issues/26060 

As described in https://github.com/gatsbyjs/gatsby/issues/26060#issuecomment-664979848 this seems to be related to the default value of `Responsive.getWidth` when building the pages. When `gatsby build` is run, the page is rendered. The default `getWidth` will in this case return 0. The only combination that seems to work is when `getWidth` returns a value that puts both `DesktopContainer` and `MobileContainer` out of view. However, that results in a static page that doesn't contain anything and everything has to be loaded dynamically on page load.

If the page on build contains either `DesktopContainer` or the `MobileContainer`, the oposite view on page load does not work. I.e. if the static page contains the `DesktopContainer`, the mobile page load is incorrect. If the static page contains the `MobileContainer`, the desktop page load is incorrect.

### Steps

1. Clone https://github.com/sveinse/gatsby-semantic-ui-bug and `npm install`
2. Run `gatsby build && gatsby serve`

### Expected Result

In mobile view (narrow browser):

develop_mobile

### Actual Result

serve_mobile

### Version
1.0.0

layershifter:
# ⚠️ Deprecation

This PR marks `Responsive` as deprecated, we will remove it in the next major release, i.e. `2.0.0`.

Fixes #4002.

# ❓ Why?

We did with this component two design mistakes 🚒

### 💅 Not CSS origin

This component does not have connection between Semantic UI CSS and our implementation. This causes mismatch our defined breakpoints (https://semantic-ui.com/elements/container.html) in CSS and JS part.

### ⚙️ SSR support

With previous item it's possible to survive, but this item is really annoying as SSR and SSG are popular and provide great experience (Next.js, Gatsby). Numerous issues without simple solution is a sign of this failure (#3110, #3398, #3948, #4002).

# ➡️ How to migrate?

In our example ([HomepageLayout](https://react.semantic-ui.com/layouts/homepage)) I made a switch to [@artsy/fresnel](https://github.com/artsy/fresnel). These changes are part of this PR.

This package was chosen as a suggested replacement as it works great with Next.js & Gatsby (https://github.com/artsy/fresnel#usage-with-gatsby-or-next) and it does not use conditional rendering (https://github.com/artsy/fresnel#why-not-conditionally-render).

### Before

```jsx
import React from "react";
import { Responsive, Segment } from "semantic-ui-react";

const App = () => (
  <Segment.Group>
    <Responsive as={Segment} {...Responsive.onlyMobile}>
      Mobile
    </Responsive>
    <Responsive as={Segment} {...Responsive.onlyTablet}>
      Tablet
    </Responsive>
    <Responsive as={Segment} {...Responsive.onlyComputer}>
      Computer
    </Responsive>
    <Responsive as={Segment} {...Responsive.onlyLargeScreen}>
      Large Screen
    </Responsive>
    <Responsive as={Segment} {...Responsive.onlyWidescreen}>
      Widescreen
    </Responsive>
  </Segment.Group>
);
```

### After

```jsx
import { createMedia } from "@artsy/fresnel";
import React from "react";
import { Segment } from "semantic-ui-react";

const AppMedia = createMedia({
  breakpoints: {
    mobile: 320,
    tablet: 768,
    computer: 992,
    largeScreen: 1200,
    widescreen: 1920
  }
});

const mediaStyles = AppMedia.createMediaStyle();
const { Media, MediaContextProvider } = AppMedia;

const App = () => (
  <>
    <style>{mediaStyles}</style>
    <MediaContextProvider>
      <Segment.Group>
        <Segment as={Media} at="mobile">
          Mobile
        </Segment>
        <Segment as={Media} at="tablet">
          Tablet
        </Segment>
        <Segment as={Media} greaterThanOrEqual="computer">
          Computer
        </Segment>
        <Segment as={Media} at="largeScreen">
          Large Screen
        </Segment>
        <Segment as={Media} at="widescreen">
          Widescreen
        </Segment>
      </Segment.Group>
    </MediaContextProvider>
  </>
);
```

# Examples

Preconfigured examples on CodeSandbox:

- Create React App - https://codesandbox.io/s/semantic-ui-react-usage-of-artsyfresnel-f2wo9
- Gatsby - https://codesandbox.io/s/semantic-ui-react-gatsby-artsyfresnel-ws5g1
- Next.js - https://codesandbox.io/s/semantic-ui-react-and-artsyfresnel-26qe5

All of them can be [deployed](https://codesandbox.io/docs/deployment) to [Vercel](https://vercel.com/) for verification.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentDoc.js
+++ b/docs/src/components/ComponentDoc/ComponentDoc.js
@@ -2,7 +2,7 @@ import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React, { Component, createRef } from 'react'
 import { withRouter, withRouteData } from 'react-static'
-import { Grid, Header, Icon } from 'semantic-ui-react'
+import { Grid, Header, Icon, Label, Popup } from 'semantic-ui-react'
 
 import DocsLayout from 'docs/src/components/DocsLayout'
 import { docTypes, examplePathToHash } from 'docs/src/utils'
@@ -47,7 +47,7 @@ class ComponentDoc extends Component {
   }
 
   render() {
-    const { componentsInfo, displayName, seeTags, sidebarSections } = this.props
+    const { componentsInfo, displayName, deprecated, seeTags, sidebarSections } = this.props
     const activePath = _.findKey(this.state.exampleStates)
     const componentInfo = componentsInfo[displayName]
     const contextValue = { ...this.props, onVisibilityChange: this.handleExampleVisibility }
@@ -60,7 +60,16 @@ class ComponentDoc extends Component {
             <Grid.Column>
               <Header
                 as='h1'
-                content={displayName}
+                content={
+                  <>
+                    <span>{displayName}</span>
+                    {deprecated && (
+                      <Popup trigger={<Label color='black'>Deprecated</Label>}>
+                        This component is deprecated and will be removed in the next major release.
+                      </Popup>
+                    )}
+                  </>
+                }
                 subheader={_.join(componentInfo.docblock.description, ' ')}
               />
               <ComponentDocSee seeTags={seeTags} />
@@ -107,6 +116,7 @@ class ComponentDoc extends Component {
 ComponentDoc.propTypes = {
   componentsInfo: PropTypes.objectOf(docTypes.componentInfoShape).isRequired,
   displayName: PropTypes.string.isRequired,
+  deprecated: PropTypes.bool.isRequired,
   history: PropTypes.object.isRequired,
   location: PropTypes.object.isRequired,
   seeTags: docTypes.seeTags.isRequired,
--- a/docs/src/examples/addons/Responsive/index.js
+++ b/docs/src/examples/addons/Responsive/index.js
@@ -1,10 +1,26 @@
 import React from 'react'
+import { Icon, Message } from 'semantic-ui-react'
 
 import Types from './Types'
 import Usage from './Usage'
 
 const ResponsiveExamples = () => (
   <div>
+    <Message icon warning>
+      <Icon name='warning sign' />
+
+      <Message.Content>
+        <Message.Header>Deprecation notice</Message.Header>
+        <p>
+          <code>Responsive</code> component is deprecated and will be removed in
+          the next major release. Please follow our{' '}
+          <a href='https://github.com/Semantic-Org/Semantic-UI-React/pull/4008'>
+            upgrade guide
+          </a>
+          .
+        </p>
+      </Message.Content>
+    </Message>
     <Types />
     <Usage />
   </div>
--- a/docs/src/layouts/HomepageLayout.js
+++ b/docs/src/layouts/HomepageLayout.js
@@ -1,5 +1,7 @@
 /* eslint-disable max-classes-per-file */
+/* eslint-disable react/no-multi-comp */
 
+import { createMedia } from '@artsy/fresnel'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 import {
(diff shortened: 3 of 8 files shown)


sveinse:
I'm not sure its important: The code in _after_ is incorrect: It still use `Responsive` while I believe it should be using `Media`?

I think also the docs should have a gentle reminder that it might not be the best solution to put the entire page under conditional render like the home page example does because it might end up duplicating data when using SSR.

layershifter:
@sveinse I update PR description (final version) & add a comment to `HomepageLayout`. Please check one more time 🐱

sveinse:
@layershifter looks good ✔️. I've tested the _after_ example successfully and I've glanced at the PR commit. Looks good from my humble eyes.

charliematters:
This might be obvious to other people, but I struggled with using the above examples with the `<Button>` component. Looking at the `<Media>` component, it always renders a div and only uses the `className` prop. As I was using the `href` prop it was producing buttons which didn't work.

To fix this I used the render-props option they document here: https://github.com/artsy/fresnel#media.

Even then it's not perfect as my responsive buttons don't neatly sit in button groups any more (as the :first and :last css selectors don't consider elements that are present but hidden), but it does at least work

layershifter:
> This might be obvious to other people, but I struggled with using the above examples with the <Button> component.

@charliematters There is no obvious things, thanks for rising this 👍

### If you are not using SSR/SSG

```jsx
<Button.Group>
  <Media at="mobile">
    {(mediaClassNames, renderChildren) => {
      return renderChildren ? <Button>Hello mobile!</Button> : null;
    }}
  </Media>
  <Button>Always</Button>
  <Media greaterThan="mobile">
    {(mediaClassNames, renderChildren) => {
      return renderChildren ? <Button>Hello desktop!</Button> : null;
    }}
  </Media>
</Button.Group>
```

```html
<div class="ui buttons">
  <button class="ui button">Hello mobile!</button
  ><button class="ui button">Always</button>
</div>
```

This code will do exactly the same thing as `Responsive` component previously: borders will look properly due conditional rendering.

### If you're using SSR/SSG

If you're using Next.js/Gatsby/_other tool_ and will try to use the code above you will get this warning:

```
Warning: Text content did not match. Server: "Hello mobile!" Client: "Always"
```

> Even then it's not perfect as my responsive buttons don't neatly sit in button groups any more (as the :first and :last css selectors don't consider elements that are present but hidden), but it does at least work

And this is correct, styling that is built with `:last`/`:first` does not play well with SSR and responsiveness. To avoid this warning I propose to use `NoSSR` and the responsive block in this case will be rendered only on client:

image

```jsx
<NoSSR>
  <Button.Group>
    <Media at="mobile">
      {(mediaClassNames, renderChildren) => {
        return renderChildren ? <Button>Hello mobile!</Button> : null;
      }}
    </Media>
    <Button>Always</Button>
    <Media greaterThan="mobile">
      {(mediaClassNames, renderChildren) => {
        return renderChildren ? <Button>Hello desktop!</Button> : null;
      }}
    </Media>
  </Button.Group>
</NoSSR>
```

https://codesandbox.io/s/semantic-ui-react-and-artsyfresnel-forked-uluj1?file=/pages/index.js

gb9box:
Hi @layershifter ,
How can I implement responsive design for the iPad Pro Portrait/Landscape?
I'd like to split design by device orientation (portrait/landscape) like as following:
```
/* Portrait */
@media only screen 
  and (min-width: 1024px) 
  and (max-height: 1366px) 
  and (orientation: portrait) 
  and (-webkit-min-device-pixel-ratio: 1.5) {
}

/* Landscape */
@media only screen 
  and (min-width: 1024px) 
  and (max-height: 1366px) 
  and (orientation: landscape) 
  and (-webkit-min-device-pixel-ratio: 1.5) {

}
```

Could you guide me on how can I implement it?

layershifter:
@gb9box It's not related to this change actually... You can use `interactions` option in [`createMedia()`](https://github.com/artsy/fresnel#createmedia) in [@artsy/fresnel](https://github.com/artsy/fresnel#createmedia).

PR #4012, opened 2020-07-31
docs(Modal): improve examples, use React hooks

layershifter:
This PR:
- improves examples for `Modal` component
- refactors them use React hooks instead of class components
- adds bundle size fixture


docs/src/examples/modules/Modal/Usage/ModalExampleCallbacks.js
```diff
@@ -0,0 +1,122 @@
...
+import { Button, Grid, Label, Modal, Segment } from 'semantic-ui-react'
+
+function exampleReducer(state, action) {
+  switch (action.type) {
+    case 'CLEAR_LOG':
+      return { ...state, log: [] }
+    case 'OPEN_MODAL':
+      return {
+        log: [
+          {
```
assuncaocharles, line 11: This will add the newest logs in the beginning of the array, is it intentional?

layershifter: Yep, as it creates an order for entries output:

image

PR #4014, opened 2020-07-31
fix(Dropdown): update search prop typing

### LINKED ISSUE #4013: Dropdown: `search` prop has invalid typings
### Steps
Use a dropdown with a default search prop

### Expected Result
That it uses the default search function

### Actual Result
```
Type 'true | ((options: DropdownItemProps[], value: string) => boolean | DropdownItemProps[])' is not assignable to type '(options: DropdownItemProps[], value: string) => boolean | DropdownItemProps[]'.
  Type 'true' is not assignable to type '(options: DropdownItemProps[], value: string) => boolean | DropdownItemProps[]'.ts(2322)
```

### Version
1.1.0

### Testcase
[Before](https://github.com/Semantic-Org/Semantic-UI-React/blob/33b5d5a860d3e27008d498841c282bb8880f8008/src/modules/Dropdown/Dropdown.d.ts):
```
search?: boolean | ((options: DropdownItemProps[], value: string) => DropdownItemProps[])
```

[After](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.d.ts):
```
search?: (options: DropdownItemProps[], value: string) => DropdownItemProps[] | boolean
```

The new type means that you can provide a function, but that function has to return an array of options, or a boolean. Previously, you could provide a boolean (e.g. `<Dropdown search />`) or alternatively a function

charliematters:
Fixes #4013. Reverted the Dropdown.search prop typing to the previous version as per #4013

### DIFF
--- a/src/modules/Dropdown/Dropdown.d.ts
+++ b/src/modules/Dropdown/Dropdown.d.ts
@@ -246,7 +246,7 @@ export interface StrictDropdownProps {
    * A selection dropdown can allow a user to search through a large list of choices.
    * Pass a function here to replace the default search.
    */
-  search?: (options: DropdownItemProps[], value: string) => DropdownItemProps[] | boolean
+  search?: boolean | ((options: DropdownItemProps[], value: string) => DropdownItemProps[])
 
   /** A shorthand for a search input. */
   searchInput?: any


layershifter (approved):
I updated PR title to auto close the issue, https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

Thanks 👍

charliematters:
I didn't know that was a thing - thanks!

felixmosh:
Thank you for the fix 🙏🏼
@layershifter when this fix will be published?

layershifter:
> @layershifter when this fix will be published?

I will ask @levithomason to do it today.

layershifter:
@felixmosh published in `1.1.1`: https://github.com/Semantic-Org/Semantic-UI-React/releases/tag/v1.1.1

PR #4019, opened 2020-08-03
fix(Popup|Visibility): update types definitions for props

### LINKED ISSUE #4016: Popup: invalid typings for `on` property
### Steps
Use Popup component and try to use the `on` property as array
` <Popup on={['hover', 'click']}`

It throws TS error: `Type '"hover"' is not assignable to type '"focus"'`

Issue may be with prop declaration:  
```
/** Events triggering the popup. */
on?: 'hover' | 'click' | 'focus' | 'hover' | 'click' | 'focus'[]
```

### Expected Result
Expect to use array form of prop with all permitted values

### Actual Result
Only can use an array of ['focus'] and not any other

### Version
1.1.1

daigof:
Fixes #4016.

### DIFF
--- a/src/behaviors/Visibility/Visibility.d.ts
+++ b/src/behaviors/Visibility/Visibility.d.ts
@@ -59,7 +59,7 @@ export interface StrictVisibilityProps {
    * Value that context should be adjusted in pixels. Useful for making content appear below content fixed to the
    * page.
    */
-  offset?: number | string | number | string[]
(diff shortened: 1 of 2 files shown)


layershifter (approved):
I updated PR's description to auto close the issue, https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

Thanks 👍

Issue #3915, opened 2020-03-23
Refs are not passed to components


neutraali:
This is super confusing and feels like the `ref` -prop isn't being consistently supported through SUIR.

For example:

1. **Input** (others?) supports `ref` [Link](https://react.semantic-ui.com/elements/input/#usage-ref-focus)
2. **Image**  (others?) does _not_ support `ref` [Link](https://react.semantic-ui.com/elements/image/)
3. **Ref** is being referenced in past issues [Link](https://github.com/Semantic-Org/Semantic-UI-React/issues/3717)
4. **Ref** is removed from docs; moved to _Stardust UI_ || _Fluent UI_ || _Whatever UI_ [Link](https://github.com/Semantic-Org/Semantic-UI-React/pull/3774)

Let's say I want to:
```javascript
import Image from 'semantic-ui-react';

let ref = React.useRef(null);
let image = () => <Image ref={ref} />
```

What is the current style of passing `ref`? Would be awesome if this was stated somewhere in the SUIR docs in large, bolded letters.

traverse:
Actually ran into the same problem when trying to implement drag and drop on a table, table rows don't pass refs either.

layershifter:
@neutraali you're totally right. This is extremely confusing 🤕 

Convergence effort is dead so I will not even try to move it forward, I will revert #3774 to move back docs for `Ref`s 🔙 

Current big goal is to get rid of deprecated lifecycle methods (#3919), the next big goal will be solve `ref`s issues and use `React.forwardRef`.

brianespinosa:
@layershifter I think this would also have to be a breaking major version change now that we officially published at 1.0.0 version. @levithomason should we update the docs to mention that since we have a full major version release that we want to stick to semantic versioning while in maintenance?

Also @layershifter I have seen you taking some time to go in and clean up some of the open issues here while also working refactoring for all of the `UNSAFE_*` methods. Thank you! I am going to make an effort to go through and respond to at least a few open issues every day that are still in the `triage` bucket.

brianespinosa:
Duplicate of #3819

layershifter:
FYI: this #4039 PR restores docs for `Ref` component.

PR #4023, opened 2020-08-05
chore: update Ref dependency to latest version

jluxenberg:
The `@stardust-ui/{react-component-event-listener, react-component-ref}` components have been moved to `@fluentui/{react-component-event-listener, react-component-ref}` (https://github.com/stardust-ui/react now redirects to https://github.com/microsoft/fluent-ui-react)

Change dependency so that the new packages are used, and upgrade to their latest versions.

One wrinkle: `toRefObject()` was removed from `react-component-ref` (see https://github.com/microsoft/fluent-ui-react/pull/2287). I've rewritten the code that depended on this function.


layershifter:
@jluxenberg that's great 🎉 

However, can we also partially revert #3774 to include `Ref*`, examples and its functions in our codebase? Per #3915 for consumers it is unclear that `Ref` component exists at all and what should be used.

src/addons/MountNode/MountNode.js
```diff
@@ -4,36 +4,37 @@ import { Component } from 'react'
...
+    }
+
+    handleClassNamesChange(this.ref, [this])
   }
 
   componentWillUnmount() {
-    const nodeRef = getNodeRefFromProps(this.props)
-
-    nodeRegistry.del(nodeRef, this)
-    nodeRegistry.emit(nodeRef, handleClassNamesChange)
```
layershifter: I don't think that it's valid to remove this part as it solves duplicated classnames issue, for example:

```tsx
<MountNode className="foo" node={document.body} />
<MountNode className="foo" node={document.body} />
<MountNode className="bar" node={document.body} />
```

Should add only `foo bar`.

This also handles unmount scenario: when one of `MountNode` components with `className="foo"` will be removed - `foo` should still be defined on `document.body` as there is another component that defines it.

jluxenberg: Got it; thanks for this additional test case. I will take a look and see if my change can be reworked.

layershifter: @jluxenberg do not spend cycles on `MountNode`, I am working on a hook that will replace it...

layershifter: @jluxenberg FYI: #4027

package.json
```diff
@@ -73,9 +73,9 @@
   "homepage": "https://github.com/Semantic-Org/Semantic-UI-React#readme",
   "dependencies": {
     "@babel/runtime": "^7.10.5",
+    "@fluentui/react-component-event-listener": "^0.51.0",
+    "@fluentui/react-component-ref": "^0.51.0",
```
layershifter:
```suggestion
    "@fluentui/react-component-event-listener": "~0.51.0",
    "@fluentui/react-component-ref": "~0.51.0",
```

These packages do not follow Semver, let's be more accurate 😺 Please also run `yarn` after, it will fix version requirements in `yarn.lock`

src/addons/MountNode/MountNode.js
layershifter: I merged the original PR and solved merge conflicts in this. Can you please check? It also seems that we have a single comment left 👍

layershifter:
> However, can we also partially revert #3774 to include `Ref*`, examples and its functions in our codebase? Per #3915 for consumers it is unclear that `Ref` component exists at all and what should be used.

We can address this one is a separate PR if you're interested 🤔

PR #4026, opened 2020-08-09
fix(Popup): do not throw if `context` & `trigger` are missing

### LINKED ISSUE #3896: Popup: Context menu Uncaught (in promise) TypeError
### Steps
We migrated from to version 0.85.0 to 0.88.2 and our context menu stopped working.
Here is how the context menu is implemented: https://codesandbox.io/s/semantic-ui-react-f1k0i

In order to reproduce, please switch the semantic-ui version from 0.85.0 to 0.88.2

### Expected Result
Open a pop-up window when doing right-click on the table header

### Actual Result
JavaScript error
`Uncaught (in promise) TypeError: Cannot read property 'left' of undefined`

### Version
0.88.2

### Testcase
https://codesandbox.io/s/semantic-ui-react-f1k0i

layershifter:
Fixes #3896.

This PR adds an example with context menu usage and fixes an issue when `trigger` and `context` are not defined to prevent throwing errors.

### DIFF
--- a/src/modules/Popup/lib/createReferenceProxy.js
+++ b/src/modules/Popup/lib/createReferenceProxy.js
@@ -7,7 +7,7 @@ class ReferenceProxy {
   }
 
   getBoundingClientRect() {
-    return _.invoke(this.ref.current, 'getBoundingClientRect', {})
+    return _.invoke(this.ref.current, 'getBoundingClientRect') || {}
   }
 
   get clientWidth() {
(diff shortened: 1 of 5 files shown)


src/modules/Popup/lib/createReferenceProxy.js
```diff
@@ -7,7 +7,7 @@ class ReferenceProxy {
   }
 
   getBoundingClientRect() {
-    return _.invoke(this.ref.current, 'getBoundingClientRect', {})
+    return _.invoke(this.ref.current, 'getBoundingClientRect') || {}
```
layershifter, line 10: This is my the most common mistake with Lodash: `_.invoke()` as 3rd parameter has a call argument, not a default value 🤦

PR #4027, opened 2020-08-10
chore(MountNode): deprecate component

layershifter:
# ⚠️ Deprecation

This PR marks `MountNode` as deprecated, we will remove it in the next major release, i.e. `2.0.0`.

---

The `MountNode` component was originally introduced in #2407. The main issue that is component should not be a part of the public API as it solves our specific issue with the `Modal` component.

This PR internally replaces usages of `MountNode` with `useClassNamesOnNode()` hook. The refactored version of utils does not use either `lodash` or `lodash/fp`, this allows to cut almost **30KB** from bundled code 🎉 

---

**This component will not be longer exported in V2.**

_If you are using this functionality, please make a follow up in this PR and we will export `useClassNamesOnNode()`._


src/addons/MountNode/MountNode.js
```diff
@@ -1,44 +1,21 @@
...
-
-    nodeRegistry.add(nodeRef, this)
-    nodeRegistry.emit(nodeRef, handleClassNamesChange)
-  }
+function MountNode(props) {
+  useClassNamesOnNode(props.node, props.className)
 
-  componentDidUpdate() {
-    nodeRegistry.emit(getNodeRefFromProps(this.props), handleClassNamesChange)
+  // A workaround for `react-docgen`
```
layershifter:
```suggestion
  // A workaround for `react-docgen`: https://github.com/reactjs/react-docgen/issues/336
```

docs/src/examples/addons/MountNode/index.js
```diff
@@ -1,8 +1,26 @@
...
 
 const MountNodeExamples = () => (
   <div>
+    <Message icon warning>
+      <Icon name='warning sign' />
+
+      <Message.Content>
+        <Message.Header>Deprecation notice</Message.Header>
+        <p>
+          <code>Responsive</code> component is deprecated and will be removed in
```
layershifter:
```suggestion
          <code>MountNode</code> component is deprecated and will be removed in
```

PR #4029, opened 2020-08-11
feat: introduce `children` function to shorthands

layershifter:
This PR adds support for `children` callback usage in any existing shorthands. This change was originally made for Stardust UI / Fluent UI Northstar (https://github.com/microsoft/fluent-ui-react/pull/1951) which had the more complicated API.

In case of Semantic UI React this will a soft migration to avoid any breaking changes: new API is added and the previous API becomes deprecated. It's expected to remove the old one with functions in V3.

## 💡 Motivation for changes

#1724 is discussing to add more types for options in `Dropdown`, the same issue can be met in `Menu`. In Fluent UI Northstar these issues were solved with `kind` prop:

```jsx
<Menu items={[{ kind: 'divider', key: 'divider' }]} />
```

However, the current implantation with having shorthand as a function blocks it as there is no way to combine `{ kind: 'smth' }` and `() => {}` 🤔 

## 🔧 Actual changes

The API was introduced and docs were added, now if you would like to customize rendered elements please use `children` function on any shorthand:

```tsx
<Button
  content='Like'
  icon={{
    children: (Component, componentProps) => <Component {...componentProps} color='red' />,
    name: 'question',
  }}
/>
```

To avoid breaking changes in future releases please modify your existing usages:

```diff
<Button
  content='Like'
-  icon={(Component, componentProps) => <Component {...componentProps} />}
+  icon={{ children: (Component, componentProps) => <Component {...componentProps} /> }}
/>
```

The warning was added to inform users about them.

---

Please share feedback and concerns 👍


src/generic.d.ts
```diff
@@ -56,18 +56,31 @@ export interface StrictHtmlSpanProps {
 // Types
 // ======================================================
 
+/**
+ * @deprecated Will be removed in v2
```
layershifter:
```suggestion
 * @deprecated Will be removed in v3
```

layershifter:

PR #4031, opened 2020-08-11
chore: support React 17

### LINKED ISSUE #4028: React 17 Compatibility
I think it would be good for us to potentially track any issues here surrounding React 17. The good news is this release is intended to move things forward progressively.

It seems like the only thing that might be worrying to us is [no more event pooling optimization](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#no-event-pooling). I know that we did at some point add in some functionality to help with this using `eventStack` but I am not sure if this change in 17 will impact us due to this.

@layershifter since you are the author of that, you are probably best to answer if this has any impact for us in SUIR. If it is irrelevant, I think we could probably close this whole issue based on the rest of the release notes for this RC.

layershifter:
Fixes #4028.

---

### On public side

This PR updates `peerDependencies` to avoid warnings related to React versions mismatch.

### On our side

Updates React & dev dependencies to run our doc site and tests with React 17, with this we can ensure that nothing is broken. Currently shipped with unofficial adapter for Enzyme and there is no support for React 17 (https://github.com/enzymejs/enzyme/issues/2429).

### DIFF
--- a/docs/src/components/Document.js
+++ b/docs/src/components/Document.js
@@ -40,11 +40,9 @@ const Document = ({ Body, children, Head, Html, siteData: { dev, versions } }) =
         }.js`}
       />
       <script
-        src={
-          dev
-            ? ` https://cdn.jsdelivr.net/npm/@hot-loader/react-dom@${versions.react}/umd/react-dom.development.js`
-            : `https://cdn.jsdelivr.net/npm/react-dom@${versions.react}/umd/react-dom.production.min.js`
-        }
+        src={`https://cdn.jsdelivr.net/npm/react-dom@${versions.react}/umd/react-dom${
+          dev ? '.development' : '.production.min'
+        }.js`}
       />
       <script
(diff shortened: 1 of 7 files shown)


wojtekmaj:
Note: For now, you can use 3rd party adapter I built based on `enzyme-adapter-react-16`: https://www.npmjs.com/package/@wojtekmaj/enzyme-adapter-react-17

docs/src/components/Document.js
```diff
@@ -40,11 +40,9 @@ const Document = ({ Body, children, Head, Html, siteData: { dev, versions } }) =
         }.js`}
       />
       <script
-        src={
-          dev
-            ? ` https://cdn.jsdelivr.net/npm/@hot-loader/react-dom@${versions.react}/umd/react-dom.development.js`
```
layershifter, line 45: There is no `@hot-loader/react-dom` for 17.0.1. It also creates weird issues with HMR...

PR #4047, opened 2020-08-26
fix(Dropdown): handle `text` as a content in `renderLabel`

### LINKED ISSUE #4046: Dropdown text no longer has a `<div class="text">` wrapper if text is an element
### Steps
1. Create a `Dropdown` and let a item's text be a `<span>`.
2. Select this item.

### Expected Result
The `<span>` should be wrapped by a `<div class="text">`. Just like the previous versions.

### Actual Result
The `<span>` no longer has the wrapper.

This cause my CSS selectors stop working.

### Version
v1.2.1 (Maybe since https://github.com/Semantic-Org/Semantic-UI-React/commit/a2de6cd0e2a68c47ce90b2bf6383572179957c99).

### Testcase
https://codesandbox.io/s/semantic-ui-react-forked-9cxcn?file=/index.js

layershifter:
Fixes #4046.

---

This PR fixes a regression from #4003 ([v1.1.0](https://github.com/Semantic-Org/Semantic-UI-React/releases/tag/v1.1.0)).

That change affected `renderLabel` prop in `Dropdown` (will be used to render labels with `multiple`). In #4003 I decided to avoid breaking changes and keep the existing behavior via this condition:

```js
if (React.isValidElement(text) || _.isFunction(text)) {
```

However, I can't consider the previous behavior as valid as it will break the layout once `text` will be a JSX element (see behaviors below).

## test snippet

```jsx
<Grid columns={2}>
  <Grid.Column>
    <Dropdown
      placeholder="Select Friend"
      fluid
      selection
      multiple
      options={[
        {
          key: "Jenny Hess",
          text: <span>Jenny Hess</span>,
          value: "Jenny Hess"
        },
        {
          key: "Elliot Fu",
          text: "Elliot Fu",
          value: "Elliot Fu"
        }
      ]}
    />
  </Grid.Column>
  <Grid.Column>
    <Dropdown
      placeholder="Select Friend"
      fluid
      selection
      options={[
        {
          key: "Jenny Hess",
          text: <span>Jenny Hess</span>,
          value: "Jenny Hess"
        },
        {
          key: "Elliot Fu",
          text: "Elliot Fu",
          value: "Elliot Fu"
        }
      ]}
    />
  </Grid.Column>
</Grid>
```

Mention that in the first case `text` is a JSX element.

## pre v1.1.0 behavior

image

https://codesandbox.io/s/busy-hooks-5p5pq

## v1.1.0 behavior

image
https://codesandbox.io/s/distracted-dawn-xg28c

## fix in this PR

image

---

To keep previous behavior use:

```jsx
<Dropdown renderLabel={({ text }) => text} />
```

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -41,9 +41,9 @@ const getKeyAndValues = (options) =>
 function renderItemContent(item) {
   const { flag, image, text } = item
 
-  // TODO: remove this in v2
+  // TODO: remove this in v3
   // This maintains compatibility with Shorthand API in v1 as this might be called in "Label.create()"
-  if (React.isValidElement(text) || _.isFunction(text)) {
+  if (_.isFunction(text)) {
     return text
   }
 


layershifter:

PR #4060, opened 2020-09-08
fix(Transition): schedule changes only on `status` change

### LINKED ISSUE #4059: Transition: rerender of a component resets `setTimeout()`
### Steps

1. Click "toggle visibility" button and open your console
2. Wait 5 seconds till onComplete() callback will be executed
3. ✅ There will be a similar log in your console:
```
visibility onStart() +0ms
visibility onComplete() +5s
```

1. Click "toggle visibility" button and open your console
2. Wait around 1 second and press multiple times rerender, wait till onComplete() callback will be executed
3. ❌ There will be a similar log in your console:
```
visibility onStart() +0ms
visibility onComplete() +8s
```

### Expected Result
In both cases there should be a same behavior as visibility of `Transition` was not changed.

### Actual Result
Timer will be reset on each render 💣 

### Version
1.2.1

### Testcase
https://codesandbox.io/s/intelligent-jepsen-5xupd

layershifter:
Fixes #4059.

---

See the linked issue for more details. The check that we previously had was false positive on each component's render that may cause unexpected race conditions. It also wrongly cleared a `setTimeout()` timer which caused desynchronization issues between CSS and JS.

### DIFF
--- a/src/modules/Transition/Transition.js
+++ b/src/modules/Transition/Transition.js
@@ -87,16 +87,17 @@ export default class Transition extends Component {
     const durationType = TRANSITION_CALLBACK_TYPE[nextStatus]
     const durationValue = normalizeTransitionDuration(duration, durationType)
 
-    clearTimeout(this.timeoutId)
-    this.timeoutId = setTimeout(
-      () => this.setState((state) => ({ status: state.nextStatus })),
-      durationValue,
-    )
+    this.timeoutId = setTimeout(() => this.setState({ status: nextStatus }), durationValue)
   }
 
   updateStatus = (prevState) => {
-    if (this.state.status !== this.state.nextStatus && this.state.nextStatus) {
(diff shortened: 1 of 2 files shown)


src/modules/Transition/Transition.js
```diff
@@ -87,16 +87,17 @@ export default class Transition extends Component {
...
-    clearTimeout(this.timeoutId)
-    this.timeoutId = setTimeout(
-      () => this.setState((state) => ({ status: state.nextStatus })),
-      durationValue,
-    )
+    this.timeoutId = setTimeout(() => this.setState({ status: nextStatus }), durationValue)
   }
 
   updateStatus = (prevState) => {
-    if (this.state.status !== this.state.nextStatus && this.state.nextStatus) {
```
layershifter, line 98: The previous check does not have sense as we should trigger a timer on `status` change

layershifter:

PR #4072, opened 2020-09-28
chore: use lodash-es to improve bundle size

layershifter:
This PR reduces bundle size for projects that are using ES modules i.e. Webpack, Rollup, Next.js, Gatsby, etc. As SUIR depends on Lodash heavily the usage of `lodash-es` allows to improve tree-shaking as [ModuleConcatenationPlugin](https://webpack.js.org/plugins/module-concatenation-plugin/) understands currently only ES modules.

_The solution was inspired by [TSDx](https://github.com/formium/tsdx#using-lodash)._

---

This may have negative impact on projects that are using `lodash` dependency, feel free to leave feedback in this PR.


noinkling:
Just thought you might want to be aware of lodash/lodash#4879, I don't know what's going on there.

layershifter:
@noinkling good call 🙁 Let's see how it will go, overall I would like to use less Lodash on our components.

PR #4084, opened 2020-10-04
Popup: Add z-index to popper container

### LINKED ISSUE #4083: Popup: zIndex is not applied in v2
### Steps
Originally we have a Popup inside a Modal. On hover the Popup will not show up.

### Expected Result
Should behave as in 1.3.1.

### Actual Result
Popup is behind the Modal.

### Version
2.0.0

### Testcase => Reason
The <div> created as a popup container no longer has ANY classes, esp. not .ui.popup. Especially the z-index is not set.

popup1
popup2

The class is set one level below the outer div and the z-Index (1900) is no longer able to move the Popop contents above the modal.

felixmosh:
fixes #4083

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -179,7 +179,7 @@ export default class Popup extends Component {
       // Heads up!
       // A wrapping `div` there is a pure magic, it's required as Popper warns on margins that are
       // defined by SUI CSS. It also means that this `div` will be positioned instead of `content`.
-      <div ref={popperRef} style={popperStyle}>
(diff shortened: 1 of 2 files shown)


siarheipashkevich:
please merge this PR

layershifter:
@felixmosh thanks for contributing, however the solution with syncing `zIndex` from `.popup` is more flexible for users than a hardcoded value. I implemented it #4094 and we plan to release it on Monday. Will appreciate your eyes there 🙏

felixmosh:
Closing since #4094 fixes the same bug

PR #4112, opened 2020-11-16
fix(Feed|CardGroup|ItemGroup|Search): change key checks from falsy to nullish semantics

### LINKED ISSUE #4111: `0` or other falsy but non-nullish values being ignored for certain `key` properties
## Bug Report

Using falsy checks for keys in certain places means we're unable to use `0` (and possibly `false`, `NaN`, etc.) as a key. In #1729 the resolution seemed to be to use nullish semantics, but at least a few components didn't get the memo:

https://github.com/Semantic-Org/Semantic-UI-React/blob/694e3993a5f711e1bd513ff823619c1c25209e7b/src/views/Feed/Feed.js#L37

https://github.com/Semantic-Org/Semantic-UI-React/blob/694e3993a5f711e1bd513ff823619c1c25209e7b/src/views/Item/ItemGroup.js#L52

https://github.com/Semantic-Org/Semantic-UI-React/blob/694e3993a5f711e1bd513ff823619c1c25209e7b/src/modules/Search/Search.js#L414

https://github.com/Semantic-Org/Semantic-UI-React/blob/c82836802975f99eddfdd78b6b72ab6110bafa4a/src/views/Card/CardGroup.js#L63

I was going to open a PR fixing these cases to use Lodash's `isNil()` as elsewhere, however I was thinking now that the nullish coalescing operator (`??`) is part of JS, maybe you guys would like to add the Babel plugin for it so we could use that instead? I notice that the optional chaining plugin is already there.

Edit: I just realized that nullish coalescing is probably already available from preset-env.

### Steps
e.g:
```jsx
const items = [
  { childKey: 0, content: 'Foo' },
  { childKey: 1, content: 'Bar' }
];
<Item.Group items={items} />
```

### Expected Result
The generated `Item`s should have a `key` of `"0"` and `"1"` respectively.

### Actual Result
The first item has a generated key based on its other props: `"Foo---"`

### Version
2.0.1

### Testcase
Directly from one of the doc examples: https://codesandbox.io/s/tooji?module=/example.js

noinkling:
Fixes #4111.

The nullish coalescing operator has been available in @babel/preset-env since v7.8, so it should be safe to use (build/tests passed).

### DIFF
--- a/src/modules/Search/Search.js
+++ b/src/modules/Search/Search.js
@@ -411,7 +411,7 @@ export default class Search extends Component {
 
     return (
       <SearchResult
-        key={childKey || result.id || result.title}
+        key={childKey ?? (result.id || result.title)}
         active={selectedIndex === offsetIndex}
         onClick={this.handleItemClick}
         onMouseDown={this.handleItemMouseDown}
@@ -436,7 +436,7 @@ export default class Search extends Component {
 
     return _.map(categories, ({ childKey, ...category }) => {
       const categoryProps = {
-        key: childKey || category.name,
+        key: childKey ?? category.name,
         active: _.inRange(selectedIndex, count, count + category.results.length),
         layoutRenderer: categoryLayoutRenderer,
         renderer: categoryRenderer,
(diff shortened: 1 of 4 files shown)


layershifter:
## Before

image

## After

image

---

Thanks for fixing this 👍

PR #4116, opened 2020-11-20
fix(Image): pass `loading` attribute to an `img` element

noinkling:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading (it's part of the WHATWG spec now, not just experimental)

This allows use of the attribute when the image uses a wrapping element, e.g: `<Image loading="lazy" wrapped />`  
Currently it gets set on the wrapper `div` which is useless.


layershifter:
Thanks for fixing this ❤️

PR #4118, opened 2020-11-23
fix(Item): actually make `content` a shorthand prop for `ItemContent`

noinkling:
It *says* it is in the docs, but actually `Item` currently just forwards the `content` prop to `ItemContent` like a typical content prop. This PR will allow use of `as`, `className` and `verticalAlign` via shorthand, and is especially useful when providing the `items` array shorthand to `ItemGroup`. 

With this PR nothing changes for people who were using a primitive value, but it might be a breaking change if people were supplying an element or array, they would need to change it to `content={ content: <element /> }` or similar to match previous behaviour.

Also with this change `ItemContent` will no longer unconditionally render, at least one of the shorthands has to be present.

### DIFF
--- a/src/views/Item/Item.js
+++ b/src/views/Item/Item.js
@@ -1,4 +1,5 @@
 import cx from 'clsx'
+import _ from 'lodash'
 import PropTypes from 'prop-types'
 import React from 'react'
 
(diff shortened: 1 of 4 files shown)


src/views/Item/Item.js
```diff
@@ -29,17 +30,22 @@ function Item(props) {
     )
   }
 
+  let contentShorthandValue = content
+  if (
+    contentShorthandValue === undefined &&
+    [description, extra, header, meta].some((prop) => !_.isNil(prop))
+  ) {
+    contentShorthandValue = {}
```
noinkling, line 38: Only setting this when `content` is strictly `undefined` here and letting `ItemContent.create` handle everything else, because it allows the user to still use `null` (or a boolean) to prevent rendering when one of the other shorthands is set. Maybe `true` should also be included though?

src/views/Item/Item.js
```diff
@@ -29,17 +30,22 @@ function Item(props) {
...
-      <ItemContent
-        content={content}
-        description={description}
-        extra={extra}
-        header={header}
-        meta={meta}
-      />
+      {ItemContent.create(contentShorthandValue, {
+        autoGenerateKey: false,
+        overrideProps: { description, extra, header, meta },
```
noinkling, line 47: Alternatively you could use `defaultProps` here, but I think it makes sense for them to override because they're more 'specific'.

Issue #4152, opened 2021-02-11
Allow passing props to <input /> component


aminpaks:
Hi guys,

There are cases that we wanna pass some props to the `input` DOM element for example adding some `data` attributes.
Would you be open to a PR that we implement a simple prop such as `inputProps` and pass that to the real input element?

Right now this is not possible as all the props are passed down to the container and not the `input` element.

brianespinosa:
@aminpaks this is already possible... it is just not as straightforward as passing props directly. You can place the `<input />` DOM element inside of the component as a child and then place any of your attributes you want on that element directly. I threw a codepen example of this for you. https://codesandbox.io/s/semantic-ui-react-forked-k8c7o?file=/index.js

There is also an example of this in the docs for placing the element as a child, but I do not know if it specifically shows that you can do this to also attach params.

aminpaks:
@brianespinosa Is this documented somewhere?

brianespinosa:
@aminpaks yes https://react.semantic-ui.com/elements/input/#variations-icon-child

Maybe you missed it because it looks like it is labeled incorrectly as passing an icon child instead of passing an input child. 🤦🏽

aminpaks:
> Maybe you missed it because it looks like it is labeled incorrectly as passing an icon child instead of passing an input child. 🤦🏽

Yeah it's not obvious at first. Maybe we should mention it specifically.

brianespinosa:
I am sure we would accept a PR updating the anchor link and text in that section.

aminpaks:
@brianespinosa Would you help me identify the issue in the pipeline? #4153

PR #4153, opened 2021-02-13
docs(Input): Include an example for passing props to input element

aminpaks:
Addresses #4152.


brianespinosa:
@aminpaks when I look at the failed test above, the output says it is missing a PR label. There are no Github labels applied to this PR.

aminpaks:
Of course, I'm not familiar with the process of the pipeline.

Thanks for the help 🙏

layershifter:
@brianespinosa FYI: Labels are assigned manually by maintainers, it allows to power `lerna-changelog` to generate fancy changelogs 📜

PR #4169, opened 2021-03-12
feat: Introduce `name` attribute to `MenuItem` component

### LINKED ISSUE #4168: Expose 'name' attribute for MenuItem
### Problem description
When using Semantic UI React with [Elastic APM](https://github.com/elastic/apm-agent-rum-js) (for real user monitoring), we get unhelpful log messages such as "Click - div" or "Click - a", as the monitoring library relies on the `name` attribute to give it a concise definition. As an example, I want to have `Click - a["authorTab"]` logged when clicking on a Tab's `MenuItem` so Elastic RUM can provide network activity (API requests) when changing tabs, etc.

#### Tab:

Tab Code Example

#### Menu:

Menu Code Example

### Proposed solution

The solution here is to expose the `name` attribute for the `MenuItem` component which will reflect for both Tab MenuItems and regular navigation MenuItem's too. Code examples can be found in the MVP below.

### MVP

[https://codesandbox.io/s/semantic-ui-react-forked-gnxi3](https://codesandbox.io/s/semantic-ui-react-forked-gnxi3?file=/index.js)

reecebenson:
Fixes #4168.

### DIFF
--- a/src/collections/Menu/MenuItem.js
+++ b/src/collections/Menu/MenuItem.js
@@ -61,14 +61,14 @@ export default class MenuItem extends Component {
 
     if (!childrenUtils.isNil(children)) {
       return (
-        <ElementType {...rest} className={classes} onClick={this.handleClick}>
+        <ElementType {...rest} name={name} className={classes} onClick={this.handleClick}>
           {children}
         </ElementType>
       )
     }
 
     return (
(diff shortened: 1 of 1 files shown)


layershifter:
Unfortunately, we cannot merge this change as it is, please check https://github.com/Semantic-Org/Semantic-UI-React/issues/4168#issuecomment-1018401512.

PR #4174, opened 2021-03-16
docs(theming) update theme.config override example

### LINKED ISSUE #4173: Improve documentation for theming
Currently under the Theming with Create React App section of https://react.semantic-ui.com/theming in the step `This file is mostly prepared, but you have to change some details in the bottom of it:` It gives an example of what it should be.

I was confused why after copying and pasting the Folders section of this that it was still not working. It is because I also need to change and update the `Import Theme` section inside the theme.confg.

I propose to update the documentation to show that.
```
/*******************************
            Folders
*******************************/

@themesFolder : 'themes';
@siteFolder  : '../../src/semantic-ui/site';

/*******************************
         Import Theme
*******************************/

@import (multiple) 'semantic-ui-less/theme.less';
@fontPath : '../../../themes/@{theme}/assets/fonts';
```

the133448:
This matches the format inside node_modules/semantic-ui-less/theme.config.example
fixes #4173


docs/src/pages/Theming.mdx
```diff
@@ -153,7 +153,11 @@ This file is mostly prepared, but you have to change some details in the bottom
 @themesFolder : 'themes';
 @siteFolder  : '../../src/semantic-ui/site';
 
-@import (multiple) "~semantic-ui-less/theme.less";
+/*******************************
+         Import Theme
+*******************************/
+
+@import (multiple) 'semantic-ui-less/theme.less';
```
layershifter:
```suggestion
@import (multiple) '~semantic-ui-less/theme.less';
```

As I remember a tilde there is required. Can you please check?

the133448:
Thanks @layershifter , your right! Don't know why I missed that.

PR #4210, opened 2021-05-16
fixe error events typings  4165

### LINKED ISSUE #4165: Dropdown: Event Typings are Incorrect
## Bug Report
The typing for the Dropdown `onClick` event is a `KeyboardEvent` rather than a `MouseEvent`.

https://github.com/Semantic-Org/Semantic-UI-React/blob/ff703557a3090ea281cb7a263bc486a978fbabdd/src/modules/Dropdown/Dropdown.d.ts#L154

It also looks as though `onBlur` and `onFocus` should be `FocusEvent`s?

https://github.com/Semantic-Org/Semantic-UI-React/blob/ff703557a3090ea281cb7a263bc486a978fbabdd/src/modules/Dropdown/Dropdown.d.ts#L138

https://github.com/Semantic-Org/Semantic-UI-React/blob/ff703557a3090ea281cb7a263bc486a978fbabdd/src/modules/Dropdown/Dropdown.d.ts#L170

### Steps
Use any Dropdown and add a typed `onClick` handler.

### Expected Result
The `onClick` event is typed as a `MouseEvent`.

### Actual Result
The `onClick` event is typed as a `KeyboardEvent`.

### Version
2.0.3

### Testcase
https://codesandbox.io/s/semantic-ui-react-forked-pmkh6

ivankenou:
fixed #4165 
error event typings


layershifter:
Unfortunately it's a duplicate of #4181.

PR #4259, opened 2021-07-28
fix(Modal|Portal): fix an error with unsafe setState()

### LINKED ISSUE #4200: Warning when closing Modal "Can't perform a React state update on an unmounted component."
## Bug Report
My first attempt at using a Modal was to have a state variable indicating if the modal was open, and only render the Modal when open was true.  This worked fine for closing the modal with the closeIcon, but gave me a warning when I clicked outside the modal to close it. 

After some testing I changed my code to always render the Modal and set the open property to my open state variable. 
I suggest explaining this in the documentation (or fixing the cause of the warning). 

### Steps
Render modal depending on state.
Click outside the open modal, or use escape to close it.

### Expected Result
Modal should close without errors in console.

### Actual Result
Modal is closed with this error in the console:
"Warning: Can't perform a React state update on an unmounted component."
Two errors when using escape to close.

### Version
2.0.3

### Testcase
https://codesandbox.io/s/semantic-ui-example-forked-yt0xt

layershifter:
Fixes #4200.

---

This PR fixes a reported issue in #4200. It happens when a component gets unmounted before `.setState()` call.
Test that covers this scenario is also added.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -196,8 +196,8 @@ class Portal extends Component {
   close = (e) => {
     debug('close()')
 
-    _.invoke(this.props, 'onClose', e, { ...this.props, open: false })
     this.setState({ open: false })
+    _.invoke(this.props, 'onClose', e, { ...this.props, open: false })
   }
 
   closeWithTimeout = (e, delay) => {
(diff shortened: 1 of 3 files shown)


laukaichung:
Thank you for the fix. When would you expect it to be released?

layershifter:
@laukaichung it was released yesterday 🎉

PR #4282, opened 2021-09-02
feat(Search): add "placeholder" & documentation

### LINKED ISSUE #3806: search placeholder is implemented but not documented
hi.
in "react semantic ui" document
search placeholder API is not documented while is implemented for component
please add
tnx

rohanrav:
Fixes #3806.


rohanrav:
Here is the issue this PR addresses: #3806

gbhasha (approved):

docs/src/examples/modules/Search/Types/SearchExampleStandard.js
```diff
@@ -67,6 +67,7 @@ function SearchExampleStandard() {
       <Grid.Column width={6}>
         <Search
           loading={loading}
+          placeholder='Search'
```
layershifter:
```suggestion
          placeholder='Search...'
```

_nit_ 🐱

layershifter:
@rohanrav great job, I also added a small unit test 🐱

noinkling:
Hmm, this change causes the search's `<input>` element to have a superfluous (no value) `placeholder` attribute when it isn't provided. I would rather avoid that.

I think it might be because of assigning it an empty string in `defaultProps`? That doesn't seem necessary to me anyway.

layershifter:
> Hmm, this change causes the search's `<input>` element to have a superfluous (no value) `placeholder` attribute when it isn't provided. I would rather avoid that.
> 
> I think it might be because of assigning it an empty string in `defaultProps`? That doesn't seem necessary to me anyway.

@noinkling good call, yeah it should be removed from `defaultProps`. I will make a PR.

layershifter:
@noinkling #4327

PR #4291, opened 2021-10-06
docs(Modal): update props' description

bonellia:
I suppose Modal is just a specialized Portal (or vice-versa?) but I suppose props docs for modal could be less ambiguous with these small changes.


src/modules/Modal/Modal.js
```diff
@@ -373,12 +373,12 @@ Modal.propTypes = {
...
 
-  /** Element to be rendered in-place where the portal is defined. */
+  /** Element to be rendered in-place where the modal is defined. */
   trigger: PropTypes.node,
 
   /**
-   * NOTE: Any unhandled props that are defined in Portal are passed-through
-   * to the wrapping Portal.
+   * NOTE: Any unhandled props that are defined in Modal are passed-through
+   * to the wrapping Modal.
```
layershifter:
```suggestion
   * to the inner Portal.
```

This is more accurate 😉

bonellia: Of course :)

PR #4309, opened 2021-12-20
fix(Search): type definition for layout renderer

### LINKED ISSUE #4308: Type definition for Search Category Layout Renderer is incorrect
## Bug Report

https://github.com/Semantic-Org/Semantic-UI-React/blob/61a08733941fb60d6ff09f30c3e2fc13e40448da/src/modules/Search/SearchCategory.js#L25

↓Actually, the argument of this function is a single object, but it is written as two objects in types.
https://github.com/Semantic-Org/Semantic-UI-React/blob/61a08733941fb60d6ff09f30c3e2fc13e40448da/src/modules/Search/SearchCategory.d.ts#L29-L39

↓ Actually, The argument of this function is a single object, but it is written as two objects in JSDoc.
https://github.com/Semantic-Org/Semantic-UI-React/blob/61a08733941fb60d6ff09f30c3e2fc13e40448da/src/modules/Search/Search.d.ts#L61-L68

### Expected

Both should be
```ts
/** 
 * Renders the SearchCategory layout. 
 * 
 * @param {object} props - The SearchCategoryLayout props object.
 * @returns {*} - Renderable SearchCategory layout. 
 */ 
(props: SearchCategoryLayoutProps) => React.ReactElement<any>
```

### Version
2.0.4

bhbs:
Fixes #4308 

- Corrected Types.
- `SearchCategoryLayoutProps` is defined and should be used.

### DIFF
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -5,6 +5,7 @@ import { InputProps } from '../../elements/Input'
 import SearchCategory, { SearchCategoryProps } from './SearchCategory'
 import SearchResult, { SearchResultProps } from './SearchResult'
 import SearchResults from './SearchResults'
+import { SearchCategoryLayoutProps } from './SearchCategoryLayout'
 
 export interface SearchProps extends StrictSearchProps {
   [key: string]: any
@@ -61,11 +62,10 @@ export interface StrictSearchProps {
   /**
    * Renders the SearchCategory layout.
    *
-   * @param {object} categoryContent - The Renderable SearchCategory contents.
-   * @param {object} resultsContent - The Renderable SearchResult contents.
+   * @param {object} props - The SearchCategoryLayout props object.
    * @returns {*} - Renderable SearchCategory layout.
    */
-  categoryLayoutRenderer?: (props: SearchCategoryProps) => React.ReactElement<any>
+  categoryLayoutRenderer?: (props: SearchCategoryLayoutProps) => React.ReactElement<any>
 
   /**
    * Renders the SearchCategory contents.
--- a/src/modules/Search/Search.d.ts
+++ b/src/modules/Search/Search.d.ts
@@ -65,7 +65,7 @@ export interface StrictSearchProps {
    * @param {object} props - The SearchCategoryLayout props object.
    * @returns {*} - Renderable SearchCategory layout.
    */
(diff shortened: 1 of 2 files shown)


bhbs:
@layershifter 
Hi, could you review this?
Thank you for nice library.

src/modules/Search/SearchCategory.d.ts
```diff
@@ -29,14 +30,10 @@ export interface StrictSearchCategoryProps {
...
-   * @param {object} categoryContent - The Renderable SearchCategory contents.
-   * @param {object} resultsContent - The Renderable SearchResult contents.
+   * @param {object} props - The SearchCategoryLayout props object.
    * @returns {*} - Renderable SearchCategory layout.
    */
-  layoutRenderer?: (
-    categoryContent: React.ReactElement<any>,
-    resultsContent: React.ReactElement<any>,
-  ) => React.ReactElement<any>
+  layoutRenderer?: (props: SearchCategoryLayoutProps) => React.ReactElement<any>
```
layershifter:
```suggestion
  layoutRenderer?: (props: Pick<SearchCategoryLayoutProps, 'categoryContent' | 'resultsContent'>) => React.ReactElement<any>
```

We don't pass all props there only `categoryContent` & `resultsContent`, so let's make it more accurate 😉

src/modules/Search/Search.d.ts
```diff
@@ -61,11 +62,10 @@ export interface StrictSearchProps {
   /**
    * Renders the SearchCategory layout.
    *
-   * @param {object} categoryContent - The Renderable SearchCategory contents.
-   * @param {object} resultsContent - The Renderable SearchResult contents.
+   * @param {object} props - The SearchCategoryLayout props object.
    * @returns {*} - Renderable SearchCategory layout.
    */
-  categoryLayoutRenderer?: (props: SearchCategoryProps) => React.ReactElement<any>
+  categoryLayoutRenderer?: (props: SearchCategoryLayoutProps) => React.ReactElement<any>
```
layershifter:
```suggestion
  categoryLayoutRenderer?: (props: Pick<SearchCategoryLayoutProps, 'categoryContent' | 'resultsContent'>) => React.ReactElement<any>
```

Same as in other comment :)

src/modules/Search/SearchCategory.d.ts
bhbs: You're right! Thank you! 😀

PR #4330, opened 2022-02-01
chore(Search): use React.forwardRef()

layershifter:
Similarly to #4234, adds native ref forwarding to `Search`.

The difference compared to other components is that `Search` remains class component. Ref forwarding is done via `wrapper` component around it. All `Search` tests were updated to use `mount()` renderer as some other tests, too.

The test that was testing internal implementation was refactored.

### DIFF
--- a/src/lib/getUnhandledProps.js
+++ b/src/lib/getUnhandledProps.js
@@ -10,6 +10,8 @@ const getUnhandledProps = (Component, props) => {
   const { handledProps = [] } = Component
 
   return Object.keys(props).reduce((acc, prop) => {
+    // "childKey" and "innerRef" are internal props of Semantic UI React
+    // "innerRef" can be removed when "Search" & "Dropdown components will be removed to be functional
     if (prop === 'childKey' || prop === 'innerRef') return acc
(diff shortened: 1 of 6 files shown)


src/lib/getUnhandledProps.js
```diff
@@ -10,7 +10,7 @@ const getUnhandledProps = (Component, props) => {
   const { handledProps = [] } = Component
 
   return Object.keys(props).reduce((acc, prop) => {
-    if (prop === 'childKey') return acc
+    if (prop === 'childKey' || prop === 'innerRef') return acc
```
layershifter, line 15: TODO: Add comment

src/modules/Search/Search.js
```diff
@@ -681,9 +687,16 @@ Search.defaultProps = {
   showNoResults: true,
 }
 
-Search.autoControlledProps = ['open', 'value']
+SearchInner.autoControlledProps = ['open', 'value']
+
+if (process.env.NODE_ENV !== 'production') {
```
layershifter, line 692: TODO: Add comment

PR #4332, opened 2022-02-03
fix(Progress): clamp bar width and improve propTypes check for progress prop

kangaechigai:
Fix some inconsistencies in `Progress` bar width and invalid prop combos.

1. bar width: `progress='value'` and `value > total`
e.g. `<Progress progress='value' value={51} total={50} />` 
 
**Expected Result**: progress bar with text `51` and width clamped to 100% to match behavior with any other value for `progress`
 
**Actual Result**: progress bar with text `51` and width 102%

1. bar width: `value` without `total`
e.g. `<Progress value={50} />`
 
**Expected Result**: progress bar with width 50% to match behavior with `progress='value'`
 
**Actual Result**: progress bar with min width

1. bad prop combo: `progress='ratio` without `value` or `total`
e.g. `<Progress progress="ratio" />`
 
**Expected Result**: `propTypes` error
 
**Actual Result**: no `propTypes` error, progress bar with text `undefined/undefined`

1. bad prop combo: `progress='value'` without `value`
e.g. `<Progress progress="value" />`
 
**Expected Result**: `propTypes` error
 
**Actual Result**: no `propTypes` error

### Version
2.1.1

### Testcase
https://codesandbox.io/s/progress-problem-with-progress-value-odipb?file=/index.js

### DIFF
--- a/src/modules/Progress/Progress.js
+++ b/src/modules/Progress/Progress.js
@@ -161,7 +161,12 @@ Progress.propTypes = {
   precision: PropTypes.number,
 
   /** A progress bar can contain a text value indicating current progress. */
-  progress: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['percent', 'ratio', 'value'])]),
+  progress: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.oneOf(['percent']),
+    customPropTypes.every([PropTypes.oneOf(['value']), customPropTypes.demand(['value'])]),
+    customPropTypes.every([PropTypes.oneOf(['ratio']), customPropTypes.demand(['value', 'total'])]),
+  ]),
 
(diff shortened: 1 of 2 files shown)


brianespinosa:
I thought for sure that the progress bar would allow to go past 100 in the parent SUI project, but I wanted to verify. Sure enough, it does not let you go past 100%. [I put a Codesandbox together to check just to make sure, as that parent project is our source of truth](https://codesandbox.io/s/semantic-ui-progress-clamp-100-weteg). We get an error if the percentage exceeds 100%. Looks like this was something that was missed when the Progress component was built. But maybe this was intentional. 

This could be considered a breaking change for this component as some people might be using it to show values exceeding 100%. I know there are scenarios in some of the projects I use the React bindings where we have used the Progress component in this way.

PR #4337, opened 2022-02-16
fix(Popup): do not hide Popup with "hideOnScroll" when scroll comes from inside the Popup

### LINKED ISSUE #4305: Popup with hideOnScroll option hides on a scroll from inside the Popup
### Steps
Create a Popup with `hideOnScroll` option and custom content with an inner scroll inside. Try to scroll popup content, popup will hide.

### Expected Result
Popup should remain visible when the scroll comes from inside the Popup. I think there is no point to hide the popup in this case.

### Actual Result
Popup hides when scroll comes from inside the popup

### Version
2.0.4

### Testcase
https://codesandbox.io/s/semantic-ui-react-popup-hides-on-inner-scroll-jdtmw

https://user-images.githubusercontent.com/1057056/145740145-54366c2e-51dc-4083-bf9e-776d5c8c1427.mp4

oskarwrobel:
Fixes an issue when Popup with `hideOnScroll` property hides on a scroll that comes from inside the Popup.
Closes https://github.com/Semantic-Org/Semantic-UI-React/issues/4305.

### DIFF
--- a/src/modules/Popup/Popup.js
+++ b/src/modules/Popup/Popup.js
@@ -36,6 +36,7 @@ export default class Popup extends Component {
   zIndexWasSynced = false
 
   triggerRef = React.createRef()
+  elementRef = React.createRef()
 
   static getDerivedStateFromProps(props, state) {
     if (state.closed || state.disabled) return {}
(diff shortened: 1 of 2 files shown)


layershifter:
~~@oskarwrobel thanks for contributing ❤️  I tested on deployed doc site:~~
~~It still closes on scrolling inside... Can you please check~~

EDIT: NVM, it's a different case.

layershifter (approved):

PR #4342, opened 2022-02-18
chore: refactor tests for shorthands to use mount()

layershifter:
This PR refactors tests for shorthands use `mount()` from Enzyme.

This is mainly a requirement for #4338. _And extracted from there to better understand coverage issues happened in that PR._

### DIFF
--- a/src/elements/Flag/Flag.js
+++ b/src/elements/Flag/Flag.js
@@ -530,14 +530,13 @@ Flag.propTypes = {
   name: customPropTypes.suggest(names),
 }
 
-Flag.defaultProps = {
-  as: 'i',
-}
-
 // Heads up!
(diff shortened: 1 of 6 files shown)


src/elements/Flag/Flag.js
```diff
@@ -530,14 +530,13 @@ Flag.propTypes = {
...
-}
-
 // Heads up!
 // .create() factories should be defined on exported component to be visible as static properties
 const MemoFlag = React.memo(Flag)
 
 MemoFlag.create = createShorthandFactory(MemoFlag, (value) => ({ name: value }))
+MemoFlag.defaultProps = {
+  as: 'i',
+}
```
layershifter, line 540: This is really strange change that is required for "_how_ Enzyme works". It does not affect functionality, so it's okay.

PR #4356, opened 2022-04-18
fix: make typings compatible with @types/react@18

### LINKED ISSUE #4352: Type check with @types/react 18 failing
### Steps
Just install @types/react, @types/react-dom v18 and typescript and update react and react-dom to v18 in any project using semantic-ui-react

### Expected Result
No error

### Actual Result
Type checking throws this errors:
```ts
JSX element type 'Header' does not have any construct or call signatures
```
The same error happens with Segment, Grid, List and Image

### Version
2.1.2

### Testcase

[Sandbox example](https://codesandbox.io/s/semantic-ui-react-forked-oyzfb2?file=/index.tsx)

image

I think this error is happening because many elements extend from the type `StatelessComponent`.
This has breen deprecated.
image

Lurk:
fixes #4352

## what changed
react typing package is updated to the latest version (18.0.5) (also added to package.resolutions because of react-helm)
Most of the work in typings was React.StatelessComponent to React.FC replacement

### DIFF
--- a/package.json
+++ b/package.json
@@ -197,8 +197,7 @@
     "react-router": "^5.0.0",
     "react-router-dom": "^5.0.0",
     "react-test-renderer": "^17.0.0",
-    "react-universal-component": "^3.0.3",
-    "@types/react": "18.0.5"
+    "react-universal-component": "^3.0.3"
   },
   "changelog": {
     "repo": "Semantic-Org/Semantic-UI-React",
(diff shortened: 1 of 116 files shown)


layershifter (approved):
@Lurk great job, and good for us that we declare [`children` explicitly](https://github.com/eps1lon/types-react-codemod#implicit-children).

Lurk:
@layershifter Yeah, if there were no explicit children already, the pull request would be much bigger 🙃

ExeAy:
Is this planned to be released any time soon?

layershifter:
Released in `semantic-ui-react@2.1.3`.

PR #4364, opened 2022-05-05
Add forward ref to Form

bolivier:
Similarly to #4234, adds native ref forwarding to Form.

### DIFF
--- a/test/specs/collections/Form/Form-test.js
+++ b/test/specs/collections/Form/Form-test.js
@@ -34,6 +34,24 @@ describe('Form', () => {
     rendersContent: false,
   })
 
+  common.forwardsRef(Form, {
+    tagName: 'form',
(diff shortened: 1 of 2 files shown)


test/specs/collections/Form/Form-test.js
```diff
@@ -34,6 +34,24 @@ describe('Form', () => {
...
+  })
+
+  common.forwardsRef(Form, {
+    tagName: 'div',
+    requiredProps: {
+      children: <input />,
+      as: 'div',
+    },
+  })
+
```
layershifter:
```suggestion
```

This is redundant as there are no conditions in `Form` around `ElementType`

layershifter (approved):
We can simply remove redundant tests. Other than that LGTM ❤️

bolivier:
:+1: Cool

PR #4383, opened 2022-09-27
docs: updates to contribution.md

carkod:
Corrected some words, added link to Cypress. In the case of `hand written` this should be spelled as `hand-written` or `handwritten`. [See](https://ell.stackexchange.com/questions/128289/how-to-spell-correctly-hand-written-or-handwritten#:~:text=Handwritten%20and%20hand%2Dwritten%20are,but%20lost%20that%20in%20years.)


layershifter (approved):
👍

PR #4408, opened 2022-12-13
chore(Select,FormSelect,FormDropdown): use React.forwardRef()

felixmosh:
For some reason `common.forwardsRef` fails on num of spy calls


layershifter:
@felixmosh I fixed unit tests and `Dropdown` component that was the source of the problem.

PR #4428, opened 2023-05-13
replace loadsh invoke with optional chaining

thing18:


layershifter:
@thing18 you need to remove some of Lodash imports as in some files it was a single usage. Run `yarn lint` locally to detect failures.

PR #4430, opened 2023-06-29
add mobile styling

highsmithcodes:
create mobile styling for the general layout, currently this site is not mobile friendly.

### DIFF
--- a/docs/src/Style.js
+++ b/docs/src/Style.js
@@ -1,6 +1,11 @@
 const style = {}
 const sidebarWidth = 250
 
+style.container = {
+  display: 'flex',
+  flexDirection: 'row', // Default value for non-mobile devices
+}
+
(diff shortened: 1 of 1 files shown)


docs/src/Style.js
```diff
@@ -25,4 +28,13 @@ style.main = {
...
 }
 
+// Manually check the screen size and apply appropriate styles
+const screenWidth = window.innerWidth
+if (screenWidth <= 768) {
+  style.container.flexDirection = 'column'
+  style.menu.position = 'relative'
+  style.menu.width = '100%'
+  style.sidebarMain.marginLeft = '0%'
+}
```
layershifter, line 38: @highsmithcodes we can't really on `window` there as it's unsafe for SSR 😢

PR #4446, opened 2023-11-29
fix(Popup): fix positioning in scrollable container

### LINKED ISSUE #4444: Popup moves as the page scrolls
## Bug Report

This is related to these couple of issues: https://github.com/Semantic-Org/Semantic-UI-React/issues/1126 and https://github.com/Semantic-Org/Semantic-UI-React/issues/3569. Popup probably listening to the scroll of a window only. If popup in a div it stays on one position when scrolling the div.

This was fixed with this PR: https://github.com/Semantic-Org/Semantic-UI-React/pull/3607, and added to version `0.87.3`. I can confirm that it works on that version, but testing in newer versions this doesn't work anymore, as commented in the same issue thread: https://github.com/Semantic-Org/Semantic-UI-React/issues/3569

I'm not sure if there's an open issue for this, but if not, I wanted to give it more visibility to this regression issue. Tested on version `2.1.4`, which is the current latest version.

### Steps

1. Have a trigger for the popup (like a button or input)
2. Have a long page with scrollbar
3. Open the popup
4. Scroll the page

### Expected Result
Popup should not move as the page scrolls.

### Actual Result
Popup moves as the page scrolls.

### Screenshots
**Current behavior**
57651284-1e621280-75cd-11e9-81dc-bf482fa92e03

**Expected behavior**
57651198-f4105500-75cc-11e9-86ee-380f76a6a336

### Version
`2.1.4`

### Testcase
https://codesandbox.io/p/sandbox/semantic-ui-react-example-forked-v8rk2z

CoryDuncan:
Fixes https://github.com/Semantic-Org/Semantic-UI-React/issues/4444

This PR addresses a regression for the `Popup` component where a `Popup` inside a non-document scrollable container would not update its position on scroll.

**Context:**
`Popper.JS` attempts to find scroll ancestors based on the reference element here: https://github.com/floating-ui/floating-ui/blob/92195083958c9496988ed4526b82b75b7256b5a4/src/createPopper.js#L80-L83.

The problem is that the [reference element provided via `Popup`](https://github.com/Semantic-Org/Semantic-UI-React/blob/ba4b72031d202388768779e4f908b9cc7b0271d0/src/modules/Popup/Popup.js#L321) fails the `isElement()` check and the fallback, a `contextElement` property does not exist.

This PR updates `ReferenceProxy` to add the `contextElement` property.

Original fix: https://github.com/Semantic-Org/Semantic-UI-React/pull/3607

### DIFF
--- a/src/modules/Popup/lib/createReferenceProxy.js
+++ b/src/modules/Popup/lib/createReferenceProxy.js
@@ -21,6 +21,10 @@ class ReferenceProxy {
   get parentNode() {
     return this.ref.current ? this.ref.current.parentNode : undefined
   }
+
+  get contextElement() {
+    return this.ref.current;
+  }
 }
 
 /**


c3-juanca:
@CoryDuncan really appreciate your help!

@layershifter what's necessary to get this merge to version `2.1.4`?

layershifter:
@CoryDuncan thanks ❤️ I will check some PRs/issues this week and trigger a release for next beta in v3.

CoryDuncan:
> I will check some PRs/issues this week and trigger a release for next beta in v3.

@layershifter would it be possible to release this as a patch in v2? The project I'm working on is using version `2.1.4` and doesn't have any near term plans to migrate to `3` (at least while it's in beta).

c3-juanca:
@layershifter I'm in the same position as Cory. Is it possible to release this as a patch in v2?

adithya1310:
@layershifter  would it be possible to release this as a patch in v2 itself as we would need this feature to be fixed for our requirement

layershifter:
@CoryDuncan @c3-juanca @adithya1310 as an exception I released the fix in `semantic-ui-react@2.1.5`.

c3-juanca:
Really appreciate it, @layershifter 🙌

adithya1310:
[heart]         Adithya Ganapathy reacted to your message:
________________________________

PR #4449, opened 2023-12-11
chore: remove usage of deprecated `.defaultProps`

### LINKED ISSUE #4426: FormInput: Support for defaultProps will be removed [v3]
Hello, I am using the library in version "semantic-ui-css": "^2.5.0", and "semantic-ui-react": "^2.1.4", along with react 18 and next 13 , and I am having the following warning or error, I searched but found no solution, would not be compatible with the version I use?

image

tourman:
This PR:
- removes usage of deprecated `.defaultProps` from all components
- replaces `getElementType()` with `getComponentType()` as `getElementType()` relied on presence of `.defaultProps`
- updates unit tests to not use `.defaultProps`

Fixes #4426.

-----

@layershifter please take a look at the basic changes on https://github.com/Semantic-Org/Semantic-UI-React/issues/4426. If you are fine with them, I'll proceed with the tests. You can review it with the hidden whitespace.

- The entries of `.defaultProps` are replaced with `function getDefaultProps()` to ensure minimum code changes. As the linter rule `no-use-before-define` forbids using functions before their declarations, I've taken the rule using `eslint --print-config` and modified it to pass the validation.
- Instead of using the default arguments, I've suggested the solution with `_.defaults` as the `props` may be used in children functions/components.


tourman:
@layershifter oh I see now, that you'd prefer `_.defaults` instead of just `defaults` style, using Lodash.

layershifter:
@tourman thanks for contributing ❤️ It's one of the issue that prevents me from shipping the next beta. However, I am more leaning towards changes proposed in https://github.com/Semantic-Org/Semantic-UI-React/issues/4426#issuecomment-1695396294. To me `getDefaultProps()` and `_.defaults()` looks like additional complexity and code debt. WDYT?

therealgilles:
I updated to v3.0.0-beta.2 but I am still seeing this:
```JS
Warning: Image: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
    at Image (https://localhost:3000/static/js/client.js:94066:22)
    at ItemImage (https://localhost:3000/static/js/client.js:102192:20)
    at div
```

Is the npm release broken?

PR #4465, opened 2024-02-20
Pass props and other data separately into all component callbacks

### LINKED ISSUE #4464: StrictTextAreaProps has a type that includes number and undefined for the "value" field
### Steps
```
<Form.TextArea onChange={(_: unknown, data: TextAreaProps) => {
  const s: string =  data.value;
}}/>
```

### Expected Result
No TypeScript errors.

### Actual Result
`Type 'string | number | undefined' is not assignable to type 'string'.`

### Version
2.1.5

### Testcase
https://codesandbox.io/p/sandbox/semantic-ui-react-forked-5tf5yh

### Notes
It doesn't appear to be possible to actually set the value of the textarea to a number or undefined, so it's not clear why I'd have to cast, and it's very inconvenient to cast undefined to a String since `String(undefined)` is `"undefined"` instead of `""`.

skondrashov:
Resolves #4464

All callbacks to components which need to process some internal value of the component have had to pull the data out of the object intended to mirrors the props that were passed into the component back to the parent. This PR separates those concepts, so that the callbacks receive internal values of components as separate parameters from the props object.

These are clearly breaking changes but they make the API cleaner and make it possible to see what values (including their types) the callbacks.

To make them easier to review, the changes are broken up into commits by each component which needed to be updated. All tests pass locally after every commit.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -426,7 +426,6 @@ export {
 export {
   default as Dropdown,
   DropdownProps,
-  DropdownOnSearchChangeData,
   StrictDropdownProps,
 } from './dist/commonjs/modules/Dropdown'
 export {
--- a/index.d.ts
+++ b/index.d.ts
@@ -254,12 +254,7 @@ export {
   StrictImageGroupProps,
 } from './dist/commonjs/elements/Image/ImageGroup'
 
-export {
-  default as Input,
-  InputProps,
-  InputOnChangeData,
-  StrictInputProps,
-} from './dist/commonjs/elements/Input'
+export { default as Input, InputProps, StrictInputProps } from './dist/commonjs/elements/Input'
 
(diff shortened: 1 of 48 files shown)


skondrashov:
Hm, based on the failing test maybe I misunderstood the purpose of this type? I was using it to determine what value I was getting out of the component, but maybe it's correct if it's the value that can be passed into the component?

skondrashov:
Updated based on what I've learned. These are breaking changes, and honestly I would like to completely remove the `props` parameter from all components as part of this, which would be even more breaking, but it doesn't make sense to pass the props back up to the parent, the parent has access to the props already kind of by definition. Would simplify and standardize the API in my opinion, might be good to fit into 3.0.0.

layershifter:
ACK on this, I won't be able to check this in upcoming two weeks, but I will follow up later. Thx

PR #4482, opened 2024-05-21
chore: replace lodash invoke with optional chaining

CrazyStevenz:
This PR takes the original work of https://github.com/Semantic-Org/Semantic-UI-React/pull/4428 and addresses the import comments, as well as some other linting errors. I used their commit directly, so they appear as the author, thus properly crediting them in my opinion.

Semantic UI React is the only package we're using that requires `lodash`, so I'd like to help get closer to the point where you can stop depending on it.


CrazyStevenz:
I'm not sure what the type of `onMouseLeave` in the `Rating.js` component should be to satisfy the failing test.

PR #4492, opened 2024-08-29
feat(Dropdown): Adds support for closeOnChange when removing items

### LINKED ISSUE #4491: Implement closeOnChange for removeItemOnBackspace and handleLabelRemove
### Problem description
I noticed that #1252 implemented an optional prop `closeOnChange` for Multiple Selection Dropdown that only works when selecting an item to add but does not apply when removing an item.

### Proposed solution
Add `this.closeOnChange(e)` to the end of the following functions: `removeItemOnBackspace` and `handleLabelRemove` which are both in `src/modules/Dropdown/Dropdown.js`

### MVP
Proof of concept (showing the problem): https://codesandbox.io/p/sandbox/semantic-ui-react-closeonchange-poc-69t37k
Add items to selection, then remove items via label or backspace while dropdown is focused/open and element will remain open

Forked, Cloned, Installed, and Tested in PR #4492

stevenrick:
Fixes #4491 

This PR adds support for closeOnChange for item removal via label and backspace in a Multiple Selection Dropdown

### DIFF
--- a/src/modules/Dropdown/Dropdown.js
+++ b/src/modules/Dropdown/Dropdown.js
@@ -430,6 +430,7 @@ class DropdownInner extends Component {
 
     this.setState({ value: newValue })
     this.handleChange(e, newValue)
+    this.closeOnChange(e)
   }
 
   closeOnDocumentClick = (e) => {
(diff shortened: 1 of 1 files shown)


stevenrick:
one note - i called this a feature, but it could be argued it is actually a bugfix - it's not obvious what the intended purpose of closeOnChange is but I'd argue it should be to close on ANY change (including removal) not just additions, hence this PR

PR #4499, opened 2024-11-13
chore(Popup): replace event-stack

dominikdosoudil:
I think that removing event-stack might be better done incrementally as each the components are independent.

I've replicated that Popup only listens for scroll when the Popup is open and enabled.
I would prefer the whole useEffect hidden under the condition however that might require much more refactoring. This way it's cleaner to see what happened in the diff. Let me know the other way would be preferred.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -152,26 +152,23 @@ function Portal(props) {
     if (!hideOnScroll) {
       return
     }
-    const abortController = new AbortController()
 
-    window.addEventListener(
-      'scroll',
-      (e) => {
-        debug('handleHideOnScroll()')
+    const handleScroll = (e) => {
+      debug('handleHideOnScroll()')
 
-        // Do not hide the popup when scroll comes from inside the popup
-        // https://github.com/Semantic-Org/Semantic-UI-React/issues/4305
-        if (_.isElement(e.target) && contentRef.current.contains(e.target)) {
-          return
-        }
+      // Do not hide the popup when scroll comes from inside the popup
+      // https://github.com/Semantic-Org/Semantic-UI-React/issues/4305
+      if (_.isElement(e.target) && contentRef.current.contains(e.target)) {
+        return
+      }
 
-        closePortal(e)
-      },
-      { signal: abortController.signal, passive: true },
-    )
+      closePortal(e)
+    }
+
+    window.addEventListener('scroll', handleScroll, { passive: true })
 
     return () => {
-      abortController.abort()
+      window.removeEventListener('scroll', handleScroll)
     }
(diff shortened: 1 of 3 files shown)


dominikdosoudil:
Because of codecov's complaining about the uncovered setTimeout, I've decided to try to get rid of it as it seems to be hacky and not much clear what it does. 

As I understood, it causes the Portal component to remount for a moment (50ms) which causes it's inner state reset which causes visually closing the popup even if Popup's state "closed" is false. There were 2 cleaner ways how to do that IMO: fully control open state of Portal (which seemed to be not so easy) or let Portal handle the scroll event and close it's inner state and even propagate it via the onClose handler (because it's all there). It appears to me that this approach works well, no weird timeout, and even the functionality can be reused more.

src/addons/Portal/Portal.js
```diff
@@ -146,6 +147,33 @@ function Portal(props) {
...
+
+        closePortal(e)
+      },
+      { signal: abortController.signal, passive: true },
+    )
+
+    return () => {
+      abortController.abort()
+    }
+  }, [closePortal, hideOnScroll])
```
layershifter, line 173: As now we need to care about `closePortal` being stable, let's wrap it with `useEventCallback()` i.e.

```diff
- const closePortal = (e) => {
+ const closePortal = useEventCallback((e) => { 
    debug('close()')

    setOpen(false)
    _.invoke(props, 'onClose', e, { ...props, open: false })
- }
+ }) 
```

src/addons/Portal/Portal.js
```diff
@@ -146,6 +147,33 @@ function Portal(props) {
...
+
+        // Do not hide the popup when scroll comes from inside the popup
+        // https://github.com/Semantic-Org/Semantic-UI-React/issues/4305
+        if (_.isElement(e.target) && contentRef.current.contains(e.target)) {
+          return
+        }
+
+        closePortal(e)
+      },
+      { signal: abortController.signal, passive: true },
```
layershifter: `signal` is nice, but does not really work with old browsers. SUIR & SUI still work in IE11, so let's break people another time :)

For now, let's change it to `.removeEventListener()` 🐱

layershifter (commented):
some nits, LGTM overall, great progress ❤️

src/addons/Portal/Portal.js
dominikdosoudil: @layershifter what about the `passive` option?

Apparently it's not supported in IE as well however IE might hopefully just ignore the option. What do you think?

https://caniuse.com/?search=passive

layershifter: Re `passive`: that's fine. It will be just ignored.

Issue #4409, opened 2022-12-13
v3: roadmap to release


layershifter:
This PR tracks work for Semantic UI React v3 🎉 

## Pre beta-0 ✅ 

- [x] chore: replace `ta-scripts` with `release-it`

## For beta-0 ✅ 

- [x] All components support native [ref forwarding](https://reactjs.org/docs/forwarding-refs.html) #4233
- [x] **BREAKING CHANGE** remove `Ref` component #4286
- [x] **BREAKING CHANGE** remove `Visibility` component #4257
- [x] **BREAKING CHANGE** `Input`, `Button` & `TextArea` are forwarding refs to native elements
- [x] **BREAKING CHANGE** `Transition` static props #4266

## Post beta-0 ✅ 

- [x] Document breaking changes #4431

## For beta-2 ✅ 

- [x] fix: remove usages of `.defaultProps` #4426, see https://github.com/Semantic-Org/Semantic-UI-React/issues/4426#issuecomment-1627680271
- [x] chore: update typings, ensure that they support `ref` forwarding #4432, see https://github.com/Semantic-Org/Semantic-UI-React/issues/4432#issuecomment-1627741193
- [x] #4433

## For beta-3

- [x] #4457
- [x] #4490 
- [ ] **BREAKING CHANGE** Event handling: remove `eventStack`
- [ ] test: `validateTrigger` should handle `forwardRef`
- [ ] chore: remove `.displayName` in production
- [ ] docs: update contributing guide
- [ ] docs: update Github templates
- [ ] docs: update introduction to remove AutoControlledComponent mentions
- [ ] docs: rename classname generators (currently they violate rules of hooks as their names start with `use`)
- [x] chore: enable ESLint plugin for rules of hooks #4498

## For RC

- [ ] chore: integrate monosize
- [ ] chore: add bundle size fixtures for all components
- [ ] chore: remove dependency on lodash
- [ ] chore: deprecate "dot-syntax"
  - [ ] add warnings
  - [ ] add a codemod

## After release
- [ ] docs: update CodeSandbox templates

vinodhrajamuralidharan:
@layershifter what is the status user.React.forwardRef PR status? I see it is still in open

layershifter:
@vinodhrajamuralidharan I had a conversation with @levithomason about that PR. The agreement was that he will review it and then we can proceed. I am also waiting for that PR to be merged 😉

sunech:
[...]

PR #4504, opened 2024-11-21
chore(Portal): Listen for Escape keypress using addEventListener instead of event-stack

dominikdosoudil:
It is only one removed EventStack instance from Portal. I wanna go step by step with it just to know that I am not going some wrong direction.

There are few decisions I would like to explain:
- I've put the event handling into the `PortalInner`. It seems to me like the right place as PortalInner is only mounted when Portal is open (-> listeners only added when portal is open as well as with EventStack component).
- I've made `onClose` and `closeOnEscape` required in `PortalInner` because it seems to me that PortalInner is and should only be used inside the Portal itself. Therefor there should not be option for SUIR developer to forget prop drill it.

Overall, I looked into it a bit and I think that most of the listeners should be in the `PortalInner` as it has access to the ref which will be needed for other handlers. I should even make `Portal` itself a bit lightweight when closed.

I hope that it makes sense and does not disrespect some SUIR architecture.

### DIFF
--- a/src/addons/Portal/Portal.js
+++ b/src/addons/Portal/Portal.js
@@ -1,5 +1,4 @@
 import EventStack from '@semantic-ui-react/event-stack'
-import keyboardKey from 'keyboard-key'
 import _ from 'lodash'
 import PropTypes from 'prop-types'
 import * as React from 'react'
@@ -130,18 +129,6 @@ function Portal(props) {
       debug('handleDocumentClick()')
       closePortal(e)
     }
-  }
-
-  const handleEscape = (e) => {
-    if (!closeOnEscape) {
-      return
-    }
-    if (keyboardKey.getCode(e) !== keyboardKey.Escape) {
-      return
-    }
-
-    debug('handleEscape()')
-    closePortal(e)
   }
 
   // ----------------------------------------
@@ -277,6 +264,8 @@ function Portal(props) {
             onMount={() => _.invoke(props, 'onMount', null, props)}
             onUnmount={() => _.invoke(props, 'onUnmount', null, props)}
             ref={contentRef}
+            onClose={closePortal}
+            closeOnEscape={closeOnEscape}
           >
             {children}
           </PortalInner>
@@ -295,7 +284,6 @@ function Portal(props) {
           />
           <EventStack name='mousedown' on={handleDocumentMouseDown} pool={eventPool} />
           <EventStack name='click' on={handleDocumentClick} pool={eventPool} />
-          <EventStack name='keydown' on={handleEscape} pool={eventPool} />
         </>
       )}
       {trigger &&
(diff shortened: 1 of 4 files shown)


dominikdosoudil:
Btw there is some behaviour change in the "Multiple Modals" example. 

Prod: when I click "Open first Modal", then "Proceed", then hit `Escape` both modals close. When I open the first one again, the second one cannot be opened. The dimmer is also darker so it seems that something opens twice.

This PR: Still both modals close with 1 Escape hit, however second opening of second modal works.

I believe that the production behaviour is a bug tho.

dominikdosoudil:
Test error is `done() called multiple times in hook <"after each" hook> (of root suite);` however that does not happen to me locally. Is that really some problem in the code or just test environment having a bad mood?

layershifter:
> Btw there is some behaviour change in the "Multiple Modals" example.
> 
> Prod: when I click "Open first Modal", then "Proceed", then hit `Escape` both modals close. When I open the first one again, the second one cannot be opened. The dimmer is also darker so it seems that something opens twice.
> 
> This PR: Still both modals close with 1 Escape hit, however second opening of second modal works.
> 
> I believe that the production behaviour is a bug tho.

Yeah, that's something that is ugly about `EventStack.` Can you please check https://github.com/Semantic-Org/Semantic-UI-React/issues/4409#issuecomment-2493216774?

dominikdosoudil:
> > Btw there is some behaviour change in the "Multiple Modals" example.
> > Prod: when I click "Open first Modal", then "Proceed", then hit `Escape` both modals close. When I open the first one again, the second one cannot be opened. The dimmer is also darker so it seems that something opens twice.
> > This PR: Still both modals close with 1 Escape hit, however second opening of second modal works.
> > I believe that the production behaviour is a bug tho.
> 
> Yeah, that's something that is ugly about `EventStack.` Can you please check [#4409 (comment)](https://github.com/Semantic-Org/Semantic-UI-React/issues/4409#issuecomment-2493216774)?



@layershifter I've been exploring behaviour of the fluentui Popover and I've noticed that the nested popover closes on Escape only when something inside the Popover has focus because there is the [contains condition](https://github.com/microsoft/fluentui/blob/592b73f62e471a2e4668a0949583790a5ef17ef8/packages/react-components/react-popover/library/src/components/PopoverSurface/usePopoverSurface.ts#L86C5-L86C90)

Is that intentional? At first I thought that the escape closing one by one somehow starts working with the virtual parents but as I see it now, these are 2 different things.

If it is intentional and I am right that escape has nothing to do with Virtual Parents (there is nothing about escape in the [VP PR](https://github.com/microsoft/fluentui/pull/18148/files)) implementation of VP and escape handling should be 2 separate PRs IMO.

PR #4513, opened 2025-02-14
update react and react-dom peerdeps to include v 19

GGAlanSmithee:
I have been using semantic-ui-react (v3 beta) with react 19 for a good while without any issues. Can we please include that version in the react peer deps range so I can avoid having to do --legacy-peer-deps?

I do not update the package version in this PR, but if you can do that and release a new version, that would be amazing.

many thanks 🙏🏼

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,8 +181,8 @@
     "webpack-dev-middleware": "^3.7.2"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
-    "react-dom": "^16.8.0 || ^17.0.0 || ^18.0.0"
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0" || ^19.0.0",
+    "react-dom": "^16.8.0 || ^17.0.0 || ^18.0.0" || ^19.0.0"
   },
   "resolutions": {
     "babel-plugin-universal-import": "^2.0.2",
--- a/package.json
+++ b/package.json
@@ -181,8 +181,8 @@
     "webpack-dev-middleware": "^3.7.2"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0" || ^19.0.0",
-    "react-dom": "^16.8.0 || ^17.0.0 || ^18.0.0" || ^19.0.0"
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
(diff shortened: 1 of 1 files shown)


GGAlanSmithee:
seems like the Vercel deployment fails because of a too low nodejs version, not sure what you want to do about that.

kylewatkins1202:
I'm confused on what the issue is here? Does this project need to be updated to use Node 22 or does workflow need to be updated to use Node 22 for Vercel?

GGAlanSmithee:
@kylewatkins1202

The actual issue that this PR solves is that the specified peer-deps does not allow for react 19.

The issue with the deployment of this PR on Vercel _seems_ to be about node.js versions, but I don't know, and I don't think it has anything to do with my PR.

kylewatkins1202:
@GGAlanSmithee 
This is an odd one to say the least. From lookin at the other PRs they seem to pass the Vercel workflow even though they give the same error, checkout #4512 . It seems this one passed but still got the same error.

After doing some research, I believe that the issue is is that dependencies can cause the Node mismatch to be treated more strictly with Vercel. In this case, React 19 is supported by 18+. This is causing Vercel to say hey 16 is [deprecated](https://vercel.com/changelog/node-js-14-and-16-are-being-deprecated) & the dependencies aren't supported by the Node version.

The issue is most likely with there work flow. However you can override the workflow with 
```
"engines": { "node": "20.x" }
``  
in package.json

GGAlanSmithee:
> The issue is most likely with there work flow. However you can override the workflow with ...

ok, understood. Do you want me to do that in this PR, given that it's not really related to what I want to change?

> React 19 is supported by 18+

Not really sure what this means. As I understand semver and node `^18.0.0` will match all version, so long as the major version is `18`. Sorry, I might be misunderstanding you - I just want to be able to run `npm i` without `--legacy-peer-deps` in my project 😂

GGAlanSmithee:
@kylewatkins1202

do you know if there's anything actionable for me to do here to move this forward? Would be amazing to be able to use semantic-ui-react with react 19 out-of-the-box.

kylewatkins1202:
@GGAlanSmithee Sadly no, they need to update their Vercel workflow to use a higher version of Node. This is the only way this PR will be approvable. They are using too old of a version of Node on Vercel

GGAlanSmithee:
Alright, thanks @kylewatkins1202 🙏🏼

Issue #4488, opened 2024-07-24
bug: [Popup] doesn't take into account popper's attributes, only styles


vkramskikh:
## Bug Report
`state.attributes.popper` aren't applied to the popup wrapper div, only `state.styles.popper` are.

### Steps
1. Use `Popup` component with default props and modifiers (which include `hide` by default) and `on="click"`.
2. Open the popup by clicking the trigger.
3. Scroll the trigger out of the viewport.

### Expected Result
The popup wrapper div gets the attributes `data-popper-escaped` and `data-popper-reference-hidden` from the default `hide` modifier.

### Actual Result
No attributes are added to the div.

### Version
2.1.5

yamunajayan:
Hi I would like to look into this ticket

abhishek-kumar-91:
I want to work on this issue, Can you assign me?

MANASA-NUKALA:
"Hi, I'd love to work on this issue! Can you assign it to me?"

AlishaMahek03:
Hi! I can fix this issue. Can you assign it to me?

Navidreza80:
I really think that I can fix this so it'll be great if you assign this to me!

Exudev:
I would love to work on this issue and return react some of the great things it has done for me, can somone assign me this? @vkramskikh @layershifter @yamunajayan

PR #4525, opened 2025-07-02
fix(Popup): apply Popper attributes to wrapper div

Aditya-githubbb:
This pull request fixes issue #4488. It spreads Popper.js's positioning attributes (`attributes.popper`) into the Popup wrapper div, allowing for correct accessibility and layout behavior as expected by Popper.js consumers.


Aditya-githubbb:
Hi! First-time contributor here. The checks passed except Vercel (which seems unrelated). Let me know if any changes are needed. Thanks! 🙌

PR #4541, opened 2026-08-04
test(Dropdown): add regression coverage for Checkbox nested in Dropdown.Item/Header

### LINKED ISSUE #2121: Checkbox: onClick doesn't fire when inside a Dropdown
### Steps
Add a checkbox inside a dropdown, either as header or item.

### Expected Result
- `onClick` fires

### Actual Result
- `onClick` doesn't fire
- also note the weird font

### Version
0.74.1

### Testcase
https://codepen.io/anon/pen/zENBPx

AmanUllah687:
Closes #2121

## What
Adds regression tests confirming that a `Checkbox`'s `onClick` handler
fires correctly when nested inside a `Dropdown.Item` and inside a
`Dropdown.Header`.

## Why
#2121 reports that `onClick` doesn't fire on a Checkbox nested inside
a Dropdown (as an item or a header). That issue is currently locked
with reason "resolved," so I wasn't able to comment there directly,
hence explaining here instead.

I couldn't reproduce the bug against current `master` or the published
`2.1.5`. I rendered a Checkbox inside both a Dropdown.Item and a
Dropdown.Header and simulated a real mousedown → mouseup → click
sequence; onClick fired correctly in both cases.

Digging through git history, Checkbox's click-handling logic was
rewritten for unrelated reasons in the fixes referenced inline as
#3351 and #3433, and that rewrite looks to have fixed this issue as a
side effect. No test ever verified it though, so this PR adds that
coverage and closes the loop on the original report.

## Testing
- Added two tests under `describe('Checkbox child', ...)` in
  `Dropdown-test.js`
- `yarn test` passes locally (11033 tests completed, 0 failed)


aphenine:
@AmanUllah687  I've forked the library at https://github.com/Fomantic-UI-React/react-fomantic-ui 

Please resubmit your PR there.

PR #4542, opened 2026-08-06
fix(TextArea): correct value type in onChange/onInput callback data

### LINKED ISSUE #4464: StrictTextAreaProps has a type that includes number and undefined for the "value" field

Khushighosh:
Fixes #4464

   The onChange/onInput callbacks for TextArea were typed with `data: TextAreaProps`, which allows `value` to be `number | string | undefined` — reflecting valid *input* values. But at runtime, the callback always returns the actual typed string, so this made consumers cast/narrow `data.value` unnecessarily.

   This PR introduces a `TextAreaOnChangeData` type that narrows `value` to `string` specifically for the callback's returned data, without changing `StrictTextAreaProps.value` itself.


aphenine:
@Khushighosh I've forked the library at https://github.com/Fomantic-UI-React/react-fomantic-ui 

Please resubmit your PR there.

